PR #18, opened 2023-11-15
Add concept of reserved channels (ie. channels updated by Pregel), Add a reserved channel for is_last_step

nfcampos:


permchain/pregel/reserved.py
```diff
@@ -0,0 +1,8 @@
+from enum import StrEnum
+
+
+class ReservedChannels(StrEnum):
+    """Channels managed by the framework."""
+
+    is_last_step = "is_last_step"
```
eyurtsev, line 7: Is step the correct way of calling this? Is step referring to an update step rather than agent step here?

permchain/pregel/__init__.py
```diff
@@ -456,12 +466,23 @@ def _apply_writes_and_prepare_next_tasks(
...
+        if chan in [c.value for c in ReservedChannels]:
+            raise ValueError(f"Can't write to reserved channel {chan}")
         pending_writes_by_channel[chan].append(val)
 
+    # Update reserved channels
+    pending_writes_by_channel[ReservedChannels.is_last_step] = [
+        for_step + 1 == config["recursion_limit"]
+    ]
+
+    print(for_step, config["recursion_limit"])
```
eyurtsev, line 484: print statement

eyurtsev (approved):

permchain/pregel/reserved.py
nfcampos: A Pregel algorithm step

permchain/pregel/__init__.py
nfcampos: Oops

PR #24, opened 2024-01-01
start langgraph

hwchase17:


permchain/langgraph/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+    def connect(self, start: Actor, end: Union[Actor, DecisionPoint]):
+        if start.name not in self.nodes:
+            raise ValueError(f"Need to register `{start.name}` first")
+        if end.name not in self.nodes:
+            raise ValueError(f"Need to register `{end.name}` first")
+        if start.name in self.connections:
+            raise ValueError(f"Already found path for {start.name}")
+        self.connections[start.name] = end.name
+
+    def branch(self, start: DecisionPoint, end: Optional[Actor], condition: str):
```
jakerachleff, line 48: Would it make sense for `branch` to instead look like:

```
class Branch:
  def __init__(self, condition: str, action: Runnable):
    ...
... 

def branch(self, name: str, *branches: Branch):
  if len(branches) == 0:
    assert("ahhhhhh")
  self.branches = {}
  for branch in branches:
    self.branches[name][branch.condition] = branch
```

That way when defining your workflow you can do something like:

```
workflow.branch(
  "somebranch",
  Branch("path1", Path1Runnable),
  Branch("path2", Path2Runnable),
  Branch("stop", StopRunnable)
)
```

permchain/langgraph/__init__.py
```diff
@@ -0,0 +1,110 @@
+from langchain_core.runnables import Runnable, RunnableMap
+from typing import Callable, Union, Optional
+from permchain import Channel, Pregel
+
+class Actor:
+
+    def __init__(self, name: str, runnable: Runnable):
```
jakerachleff, line 7: I wonder if there's an alternative setup here where actor looks like:

```
class Actor:
  def __init__(self, key: str, runnable: Runnable, input_keys: list[str], entry: bool):
    ....
```

That would mean you could skip all the `connect` calls and just set up your inputs inline. For decisionPoints you could have something similar, but maybe each branch has its own key (e.g. `{decisionPoint.key}.{conditionKey}`).

You'd be able to support circular dependencies too since you are doing this by key instead of by value.

permchain/langgraph/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+
+class End:
+    name = "end"
+
+
+def branch(data, condition, mapping):
+    result = condition(data)
+    return Channel.write_to(mapping[result])
+
+class Graph:
```
jakerachleff, line 26: One high level question - we now will have two graph languages in LangChain. 

We'll have LangGraph for permchain, and LCEL for langchain (which is basically a DAG language disguised with piping). 

Is this going to be confusing for users?

Personally, I prefer the graph like syntax everywhere

permchain/langgraph/__init__.py
jakerachleff: If matching the proposed syntax above on `Actors`, then you'd have something more like:

```
my_decision = DecisionPoint(
  "somebranch",
   lambda x: some_fn_on_x(x),
   Branch("path1", actor1.key),
   Branch("path2", actor2.key),
   Branch("stop", actor3.key)
)
```

Then you don't even need the branch calls?

hinthornw (commented):
Ah this is the syntax I kept reaching for! love it

permchain/langgraph/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+    def __init__(self):
+        end_node = End()
+        self.nodes = {end_node.key: end_node}
+        self.edges = []
+        
+        # self.connections = {}
+        # self.branches = {}
+        self.entry_point: Optional[str] = None
+
+    def register_node(self, node: Actor):
```
eyurtsev, line 129: Could we rename `register` to `add`. It's semantically identical, and `add` is more friendly

PR #38, opened 2024-01-16
Nc/state

nfcampos:


langgraph/graph/graph.py
```diff
@@ -0,0 +1,132 @@
+from asyncio import iscoroutinefunction
```
eyurtsev, line 1: This namespace is very nested: langgraph.graph.graph.Graph

langgraph/graph/graph.py
```diff
@@ -0,0 +1,132 @@
...
+            raise ValueError(f"Need to add_node `{end_key}` first")
+
+        if not self.support_multiple_edges and start_key in set(
+            start for start, _ in self.edges
+        ):
+            raise ValueError(f"Already found path for {start_key}")
+
+        self.edges.add((start_key, end_key))
+
+    def add_conditional_edges(
```
eyurtsev, line 58: A better method name could be helpful

langgraph/prebuilt/tool_executor.py
```diff
@@ -0,0 +1,69 @@
...
+    """Interface for invoking a tool"""
+
+    tool: str
+    tool_input: Union[str, dict]
+
+
+class ToolInvocation(Serializable):
+    """Information about how to invoke a tool."""
+
+    tool: str
```
eyurtsev, line 23: Should this take a tool id for multiple tool invocation use case? (when invoking the same tool, but with different arguments in parallel)

langgraph/prebuilt/tool_executor.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class ToolExecutor(RunnableBinding):
+    tools: Sequence[BaseTool]
+    tool_map: dict
+    invalid_tool_msg_template: str
+
+    def __init__(
+        self,
+        tools: Sequence[BaseTool],
```
eyurtsev, line 36:
```suggestion
        tools: Sequence[BaseTool],
        *,
```

langgraph/prebuilt/tool_executor.py
```diff
@@ -0,0 +1,69 @@
...
+from langchain_core.runnables import RunnableBinding, RunnableLambda
+from langchain_core.tools import BaseTool
+
+INVALID_TOOL_MSG_TEMPLATE = (
+    "{requested_tool_name} is not a valid tool, "
+    "try one of [{available_tool_names_str}]."
+)
+
+
+class ToolInvocationInterface:
```
eyurtsev, line 13: Inherit from Protocol?

And extend interface to allow parallel tool calling?

langgraph/prebuilt/tool_executor.py
```diff
@@ -0,0 +1,69 @@
...
+class ToolInvocation(Serializable):
+    """Information about how to invoke a tool."""
+
+    tool: str
+    """The name of the Tool to execute."""
+    tool_input: Union[str, dict]
+    """The input to pass in to the Tool."""
+
+
+class ToolExecutor(RunnableBinding):
```
eyurtsev, line 29: I think we should wrap the response in a ToolResponse message that has an id propagated through, and a slot for an error message

langgraph/prebuilt/agent_executor.py
```diff
@@ -0,0 +1,123 @@
...
+            input: str
+            # The list of previous messages in the conversation
+            chat_history: Sequence[BaseMessage]
+            # The outcome of a given call to the agent
+            # Needs `None` as a valid type, since this is what this will start as
+            agent_outcome: Union[AgentAction, AgentFinish, None]
+            # List of actions and corresponding observations
+            # Here we annotate this with `operator.add` to indicate that operations to
+            # this state should be ADDED to the existing values (not overwrite it)
+            intermediate_steps: Annotated[list[tuple[AgentAction, str]], operator.add]
```
eyurtsev, line 26: Could we change the observation to `Any` type since the output from tools can be a dict etc. Would be helpful to have an option to stringify it as close to the prompt as possible?

langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -0,0 +1,128 @@
...
+    model = model.bind_functions(
+        [format_tool_to_openai_function(t) for t in tool_classes]
+    )
+
+    # Define the function that determines whether to continue or not
+    def should_continue(state):
+        messages = state["messages"]
+        last_message = messages[-1]
+        # If there is no function call, then we finish
+        if "function_call" not in last_message.additional_kwargs:
```
eyurtsev, line 30: note in passing: we should figure out how to expose an interface around this... `additional_kwargs` feels like should be hidden from end-user

hwchase17: yup agree

PR #72, opened 2024-01-29
Fix missing imports in the multi-agent-collaboration example

al1p:


al1p (commented):
Apparently format_tool_to_openai_function is depreacted. 
`format_tool_to_openai_function` should be replaced by `convert_to_openai_function`

`from langchain_core.utils.function_calling import convert_to_openai_function`

And 
   ` functions = [format_tool_to_openai_function(t) for t in tools]`
   
   should be 
    `functions = [convert_to_openai_function(t) for t in tools]`

hwchase17 (approved):
thank you!

PR #73, opened 2024-01-30
Update deps

nfcampos:


hinthornw:
Should we update the pyproject.toml bounds then?

nfcampos:
@hinthornw I thought about it but that can create some annoying issues for folks

PR #74, opened 2024-01-31
feat: stategraph parser

stugorf:
While developing in LangGraph it can be helpful for the developer to visualize the 
flowchart representing the logic flow of their StateGraph.This parser reads a python
file and creates an abstract syntax tree (AST) of the StateGraph. The AST is then
translated into Mermaid charting language and saved as a PNG.

NOTE: I have never committed to an open-source project

This PR is related to a conversation I had with @hwchase17


nfcampos:
hi @stugorf, thanks for this, we've recently added support for mermaid drawing in langchain-core, see examples here https://github.com/langchain-ai/langgraph/pull/311 so closing this one, thanks again

stugorf:
Thank you!

PR #89, opened 2024-02-07
Add test for astream_events

hinthornw:


nfcampos:
Fixed in https://github.com/langchain-ai/langchain/pull/17205

PR #107, opened 2024-02-12
Formatting

hinthornw:
Plus add trace links.


ampersand-five:
@hinthornw Noticed some bugs:

1) Task Fetching Unit, when it inserts the tool_messages, it's outside the loop where all the tasks are executed, so the args for each function are not being set correctly, this causes the replanner to loop and use the same function parameters and never corrects itself.
```python
  tool_messages = [                                                                                                                     vvv: Here
    FunctionMessage(name=name, content=str(obs), additional_kwargs={'idx': k, 'args':task['args']})
    for k, (name, obs) in new_observations.items()
  ]
```

2) Replan prompt is never actually inserted into the prompt sent to the LLM.
3) In the planner prompt, the 'num_tools' variable needs to have a +1 added to it because we're appending the join() function, without +1 it says there's one less available tools than there actually is because it only counts the passed in functions and not the join() function.  And add +1 for listing otherwise it starts at a 0 offset.
4) The task fetching unit, when it calls the tools it doesn't handle when there's not tasks or reaches it's end:
Current code: `tasks = itertools.chain([next(tasks)], tasks)`
A possible fix, wrap in try/except and set tasks to empty:
```python
    try:
        tasks = itertools.chain([next(tasks)], tasks)
    except StopIteration:
        # Handle the case where 'tasks' is empty or has reached its end. If 
        tasks = iter([])
```

PR #113, opened 2024-02-16
Format

hinthornw:


beyondExp:
We are also encountering the same Problem even with the about an Hour released langchain-fireworks==0.1.7 and the totally new Langchain  core langchain-core==0.2.26. donyoubknownmore about this @scottykaye?

PR #128, opened 2024-02-20
Add get_state() and update_state() methods to get and update checkpoint in between runs, Add debug arg to Graph/StateGraph

nfcampos:
- [x] Decide on what to do with interrupt_before arg


langgraph/channels/base.py
```diff
@@ -116,16 +116,16 @@ def create_checkpoint(
     checkpoint: Checkpoint, channels: Mapping[str, BaseChannel]
 ) -> Checkpoint:
     """Create a checkpoint for the given channels."""
-    checkpoint = Checkpoint(
+    values: dict[str, Any] = {}
```
nfcampos, line 119: before we were leaving old checkpoint values in there by mistake

PR #134, opened 2024-02-22
Fix `grade_documents` and `rewrite`

code-sum:
Hi, I’m so excited to be using LangGraph !

Yesterday, I found this bug in `/examples/langgraph_agentic_rag.ipynb`.

The problem is that when it comes to the case of `DECISION: DOCS NOT RELEVANT`
(e.g. "What does Lilian Weng say about the types of apple?")
some error occurs because of the data type of `score` and `msg`.

Developers at LangGraph can fix the bug by applying my solution.

Good luck and have a nice day !

스크린샷 2024-02-21 171452

스크린샷 2024-02-21 171520

스크린샷 2024-02-21 171729

스크린샷 2024-02-21 171812

스크린샷 2024-02-21 171929

### DIFF
--- a/examples/rag/langgraph_agentic_rag.ipynb
+++ b/examples/rag/langgraph_agentic_rag.ipynb
@@ -30,7 +30,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 1,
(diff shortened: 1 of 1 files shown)


hwchase17:
thanks! was resolved in a separate pr, but appreciate this :)

PR #149, opened 2024-02-25
feat: add dynamodb checkpoint saver

adosaiguas:
Checkpoint saver that stores history in AWS DynamoDB

PR #170, opened 2024-03-01
Add Amazon DynamoDB checkpointer and unit tests

donatoaz:
This commit adds a new checkpointer that persists data in a DynamoDB table.

It's a draft, since I see these points of improvement that I would appreciate feedback on:

- I never used Poetry, so I tried to have the boto3 and moto dependencies as optional, so that folks who don't want anything to do with AWS can be left undisturbed, but I'm not sure I did it the right way
- I didn't really understand the tests folder, I added a `unit` folder, but again, please feel free to indicate a more appropriate location for the unit test
- The code assumes the table to have a `"thread_id"` PK, but this could, maybe, be parameterizable, not sure, what do you think?

Overall, I'd appreciate any feedback.


donatoaz:
Oh dear, I had not seen https://github.com/langchain-ai/langgraph/pull/149 

I will close with this comment.

PR #198, opened 2024-03-12
Adding missing import in example code

Saidiibrahim:
There was a import of ChatPromptTemplate and MessagePlaceholder in the `agent_supervisor.ipynb` example. I've added the import and the code works fine.


pyproject.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://www.github.com/langchain-ai/langgraph"
 [tool.poetry.dependencies]
 python = ">=3.9.0,<4.0"
 langchain-core = "^0.1.25"
+langchain-openai = "0.0.2.post1"
```
hwchase17, line 13: this shouldnt change - this is dependencies for langgraph the package, not for examples

pyproject.toml
```diff
@@ -42,6 +43,7 @@ jupyter = "^1.0.0"
 langchain = "^0.1.0"
 langchainhub = "^0.1.14"
 langchain-openai = "^0.0.2"
+ipykernel = "^6.29.3"
```
hwchase17, line 46: we have jupyter, why is this needed?

Saidiibrahim:
Will attempt again later

PR #200, opened 2024-03-12
Fix human in the loop notebook

nfcampos:


hinthornw:
What was broken?

nfcampos:
yea, when I added the get/update state methods broken that

PR #209, opened 2024-03-15
Feat/Add @workflow.register_node decorator

Howe829:
Adding graph.register_node decorator for easily register a node to the workflow.

### DIFF
--- a/langgraph/graph/graph.py
+++ b/langgraph/graph/graph.py
@@ -89,17 +89,17 @@ class Graph:
             raise ValueError(f"Need to add_node `{end_key}` first")
 
         if not self.support_multiple_edges and start_key in set(
-                start for start, _ in self.edges
+            start for start, _ in self.edges
         ):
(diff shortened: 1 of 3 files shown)


hinthornw:
oh it's kinda convenient. Will defer to @nfcampos on the approval though

nfcampos:
I'm not opposed to this as long as we can do it inside the existing add_node method

Howe829:
Hi there, I've tried to make it inside the add_node method.PLTAL
It may looks  a little bit tricky but this is the best way for me  to make add_node as a decorator as well as thinking about the forward compatibility.

Glinte:
Any updates on this? Seems quite helpful

vbarda:
Thanks for your contribution! Going to close since this is already supported in the library https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/tests/test_pregel.py#L2189

PR #222, opened 2024-03-20
Utility to create tool invocation

hinthornw:
Context: our readme and examples have a lot of `message.additional_kwargs["function"]` stuff  that make me always scratch my head - it's a lot of superfluous code that make the tool invocation interface less friendly. Would be nice to be able to just go from message -> tool invocation.

TODO: Add docstring, etc. 

Open to other names, class methods, and other suggestions.

### DIFF
--- a/langgraph/prebuilt/tool_executor.py
+++ b/langgraph/prebuilt/tool_executor.py
@@ -1,8 +1,10 @@
-from typing import Any, Sequence, Union
+import json
+from typing import Any, List, Optional, Sequence, Union
 
 from langchain_core.load.serializable import Serializable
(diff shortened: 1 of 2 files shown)


langgraph/prebuilt/tool_executor.py
```diff
@@ -24,6 +26,31 @@ class ToolInvocation(Serializable):
     """The name of the Tool to execute."""
     tool_input: Union[str, dict]
     """The input to pass in to the Tool."""
+    id: Optional[str] = None
```
hinthornw, line 29: Simplifies the `zip()` when creating the tool message. Open to not including

nfcampos:
closing this in favor of ToolNode

PR #223, opened 2024-03-20
[RFC] Stepwise Agent Evaluator

hinthornw:
Teacher forcing example that computes the (macro) average step-wise success rate of an agent.

Code explicitly looks at message graph but i could look into cleaning up / generalizing.

a bit too much code tbh imo but i like the idea. Maybe would be better if the trajectory were more high level rather than full mocks. Open to lots of suggestions (e.g., simplifying the stepwise score )

https://smith.langchain.com/public/3b46ebd4-8cf9-41b8-8fbc-5df92528a148/d


hwchase17:
we are working on agent evals, will keep this in mind, but will close

PR #242, opened 2024-03-29
Added postgresql sync and async checkpoint classes

sharan9678:
I have added the postgresql sync and async checkpoints classes using psycopg3.
To run the test we need to have a postgresql installed locally and then change the connection string in the both sync and async pregal test files.

### DIFF
--- a/langgraph/checkpoint/aiopostgresql.py
+++ b/langgraph/checkpoint/aiopostgresql.py
@@ -1,0 +1,163 @@
+import pickle
+from contextlib import AbstractAsyncContextManager
+from types import TracebackType
+from typing import AsyncIterator, Optional
+
+import psycopg
+from langchain_core.pydantic_v1 import Field
+from langchain_core.runnables import RunnableConfig
+from typing_extensions import Self
+
+from langgraph.checkpoint.base import BaseCheckpointSaver, Checkpoint, CheckpointTuple
+
+
+class AsyncPostgresqlSaver(BaseCheckpointSaver, AbstractAsyncContextManager):
+    conn: psycopg.AsyncConnection
+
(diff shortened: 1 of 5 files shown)


ajjjj8:
@sharanrajt do we have any estimate on when this can be merged?

hinthornw:
Hi @sharan9678 , to avoid bloat in the langgraph package, we don't plan to add any additional integrations in the main langgraph package that add dependencies.

There is an existing (albeit outdated: would love help updating) postgres implementation in the langchain-postgres package (https://github.com/langchain-ai/langchain-postgres) here: https://github.com/langchain-ai/langchain-postgres/blob/main/langchain_postgres/checkpoint.py

ootkin:
I was reviewing the Langgraph documentation and came across this note:

"While this class does support asynchronous checkpointing, it is not recommended for production workloads due to limitations in SQLite's write performance. For production workloads, consider using a more robust database like PostgreSQL."

I agree with this recommendation, but the documentation does not provide guidance on using other production-ready databases (SQL, MongoDB, Redis, etc.).

I will check @sharan9678's code for further insights.

sharan9678:
Hi @ootkin,

I had some similar requirements for PostgreSQL hence I wrote this code, But as of now langchain-ai provides support for PostgreSQL Checkpointer. Check the documentation at [langchain-ai/langchain-postgres](https://github.com/langchain-ai/langchain-postgres).

flavioaltimari:
Could someone help on that? I need to build the checkpoint on postgresql, but lost here

FilippoFlorindiKAI:
I have the same problem and I need to store checkpointers in a postgreSQL db. They also removed the checkpointer from the [langchain-ai/langchain-postgres](https://github.com/langchain-ai/langchain-postgres) library.

PR #245, opened 2024-03-31
Add stream_mode=updates or values

nfcampos:
- stream(stream_mode="values") yields thre current channel values whenever a channel is updated
- stream(stream_mode="updates") yields the update sent to each channel by each node
- this removes the __end__ value present at the end of each call to Graph/StateGraph/MessageGraph.stream(). To access the __end__ value either call .invoke() or stream(stream_mode="values")


hinthornw:
Ah i may have some notebooks that check for `__end__` in the .stream() output and behave differently if so that I'll want to update

nfcampos:
@hinthornw what do the notebooks do with that? You can switch to `invoke()` or `stream(stream_mode="values")` depending on what you want to achieve

Issue #265, opened 2024-04-02
Support AgentExecutor features via create_agent_executor


ccurme:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

- [x] ~max_iterations~ handled by `recursion_limit` config?
- [X] ~max_execution_time~ OK with `step_timeout`? (via asyncio task timeouts)
- [X] return_intermediate_steps (always)
- [X] trim_intermediate_steps (via the message functioN)
- [ ] handle_parsing_errors
- [ X] early_stopping_method (via timeouts + recursion error handling)
- [ X] verbose (via deug)
- [x] [handle old style tools](https://github.com/langchain-ai/langchain/blob/3218463f6a3d841905648971735949a14a16e191/libs/langchain/langchain/agents/output_parsers/openai_tools.py#L45-L54) that do not expose a schema and expect a single string argument as an input
- [ ] ??

hinthornw:
Think this can be closed now!

PR #268, opened 2024-04-02
(WIP) Add support for early stopping in agent executor

ccurme:
Following https://github.com/langchain-ai/langgraph/issues/265, here we add support for early stopping (e.g., via `max_iterations` or `max_execution_time`) as in AgentExecutor.

WIP: looking for feedback on the general strategy for incorporating AgentExecutor's attributes, eventually including
- `return_intermediate_steps`
- `max_iterations`
- `max_execution_time`
- `early_stopping_method`
- `handle_parsing_errors`
- `trim_intermediate_steps`

Here I've added some to `AgentState`, but it's unclear to me how to set defaults when they are not passed in by the caller on invoke, stream, etc. For example, even if I set `iteration_count: int = 0` in AgentState, it gets set to 0 if not included in the input. Looking for suggestions on other ways to do this!

### DIFF
--- a/langgraph/prebuilt/agent_executor.py
+++ b/langgraph/prebuilt/agent_executor.py
@@ -1,4 +1,5 @@
 import operator
+import time
 from typing import Annotated, Optional, Sequence, TypedDict, Union
 
 from langchain_core.agents import AgentAction, AgentFinish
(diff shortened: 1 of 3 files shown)


nfcampos:
We support max_iterations natively in langgraph through the config key `recursion_limit`:
```
app.invoke(..., {"recursion_limit": 15})
```

nfcampos:
We support `step_timeout` arg natively in langgraph, see `step_timeout` constructor arg in Pregel/CompiledGraph. I don't think we need `total_timeout` when we have step timeout.

PR #281, opened 2024-04-04
Added CRAG With Gemini

tutu-sol:


hwchase17:
we are not going to add an example for every model provider, as that won't really scale. this is a great example and we're happy to highlight it separately, but i dont think it should live here!

PR #282, opened 2024-04-05
added packages to run multi-agent collaboration notebook

roninio:
Added packages to be able to run the notebook 
was missing langchain_core langgraph 
also changed that line would run as it was commented out


hinthornw (approved):

hinthornw:
Thank you!

PR #286, opened 2024-04-08
Docs

hinthornw:


langgraph/graph/state.py
```diff
@@ -135,10 +172,12 @@ def _get_state_key(input: dict, config: RunnableConfig, *, key: str) -> Any:
         state_keys = list(self.graph.channels)
         # state updaters
         state_write_entries = [
-            ChannelWriteEntry(key, None, skip_none=True)
-            if key == "__root__"
-            else ChannelWriteEntry(
-                key, RunnableCallable(_get_state_key, key=key, trace=False)
+            (
```
nfcampos, line 175: different version of formatter?

hinthornw: ah i think so ya

PR #292, opened 2024-04-10
Fix/call the correct variable `authoring_graph` during compilation

AryanMadhavVerma:
In the `hierarchical_agent_teams.ipynb` example, `research_graph` should be replaced by `authoring_graph` during compilation.


ruemura3 (changes requested):
I noticed that chain isn't used elsewhere. Could we replace authoring_graph.compile() with chain on line 646?

hinthornw (approved):

PR #293, opened 2024-04-10
Checkpointer: Support more types of serializable states using `dill`

nadav-okami:
This PR solves the `PicklingError` when using states with [types that `pickle` doesn't support but `dill` does](https://github.com/uqfoundation/dill?tab=readme-ov-file#major-features).


nfcampos:
hi, @nadav-okami thanks for this, I'm making the serializer used by checkpoints configurable to more easily support this https://github.com/langchain-ai/langgraph/pull/305

PR #296, opened 2024-04-11
Add prebuilt ToolNode, Make prebuilt create_tool_calling_executor compatible with any tool calling model

nfcampos:


langgraph/prebuilt/tool_node.py
```diff
@@ -0,0 +1,96 @@
...
+    else:
+        try:
+            return json.dumps(output)
+        except Exception:
+            return str(output)
+
+
+class ToolNode(RunnableCallable):
+    """
+    A node that runs the tols requested in the last AIMessage. It can be used
```
hinthornw, line 25:
```suggestion
    A node that runs the tools requested in the last AIMessage. It can be used
```

langgraph/prebuilt/tool_node.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def str_output(output: Any) -> str:
+    if isinstance(output, str):
+        return output
+    else:
+        try:
+            return json.dumps(output)
+        except Exception:
+            return str(output)
```
hinthornw, line 20: maybe debug log?

hinthornw:
nice

langgraph/prebuilt/tool_node.py
nfcampos: Not sure, lots of things in python/langchain (eg our good friend pydantic) aren't json serializable ?

PR #298, opened 2024-04-11
Pregel: Add support for falsy outputs in Pregel class

hmasdev:
## Description

This PR addresses a specific issue where a graph unexpectedly returns None for falsy values instead of the actual falsy output. This behavior was identified during operations where the expected behavior is to output falsy values, such as 0, False, or an empty collection, but instead, the graph outputs None.

The issue is illustrated with the following example:

```python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from langchain_core.runnables import RunnableLambda
>>> from langgraph.graph import Graph, END
>>> workflow = Graph()
>>> workflow.add_node('minus1', RunnableLambda(lambda n: n-1))
>>> workflow.set_entry_point('minus1')
>>> workflow.add_edge('minus1', END)
>>> runnable = workflow.compile()
>>> print(runnable.invoke(1))
None
>>> print(runnable.invoke(2))
1
>>> print(runnable.invoke(-1))
-2
```

For more details, refer to　https://github.com/langchain-ai/langchain/discussions/20201 .

The solution implemented in this PR introduces a new parameter allow_falsy_output to decide whether the graph should return None or the actual output when the result is falsy. This allows for more accurate output representation, as demonstrated below:

```python
>>> print(runnable.invoke(1, allow_falsy_output=True))
0
```

## Issues

https://github.com/langchain-ai/langchain/discussions/20201

## Dependencies

Nothing

## Question

The purpose of the following if statements in the langgraph/pregel/__init__.py file is unclear:

- https://github.com/langchain-ai/langgraph/blob/114137059350b95068db9e96f79924fdc379bbe2/langgraph/pregel/__init__.py#L701
- https://github.com/langchain-ai/langgraph/blob/114137059350b95068db9e96f79924fdc379bbe2/langgraph/pregel/__init__.py#L706
- https://github.com/langchain-ai/langgraph/blob/114137059350b95068db9e96f79924fdc379bbe2/langgraph/pregel/__init__.py#L915
- https://github.com/langchain-ai/langgraph/blob/114137059350b95068db9e96f79924fdc379bbe2/langgraph/pregel/__init__.py#L920

Could someone provide clarity on the intention behind these conditions?

### DIFF
--- a/langgraph/pregel/__init__.py
+++ b/langgraph/pregel/__init__.py
@@ -557,6 +557,7 @@ class Pregel(
         input_keys: Optional[Union[str, Sequence[str]]] = None,
         interrupt_before_nodes: Optional[Sequence[str]] = None,
         interrupt_after_nodes: Optional[Sequence[str]] = None,
+        allow_falsy_output: bool = False,
         debug: Optional[bool] = None,
     ) -> Iterator[Union[dict[str, Any], Any]]:
         config = ensure_config(config)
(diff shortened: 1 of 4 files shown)


nfcampos:
hi @hmasdev thanks for this, this is a good fix, but we don't need a new argument for this, we should always return falsey values, I'll update the PR to reflect this. Thanks again!

PR #299, opened 2024-04-11
In stream_mode=values yield all values, instead of only changed ones

nfcampos:


tests/test_pregel.py
```diff
@@ -2874,10 +2880,17 @@ def qa(data: State) -> State:
     ]
 
     assert [*app.stream({"query": "what is weather in sf"}, stream_mode="values")] == [
-        {"query": "what is weather in sf"},
-        {"query": "query: what is weather in sf"},
-        {"docs": ["doc1", "doc2", "doc3", "doc4"]},
-        {"answer": "doc1,doc2,doc3,doc4"},
+        {"query": "what is weather in sf", "docs": []},
```
hinthornw, line 2883: Ah so it's the full state now

hinthornw (approved):
Oh this is good

PR #305, opened 2024-04-12
Make ser/de configurable in checkpointer classes

nfcampos:
- Remove pydantic usage from base checkpointer class
- Make serialization configurable for all existing checkpointer classes, you can eg use dill or json instead of pickle

### DIFF
--- a/langgraph/checkpoint/base.py
+++ b/langgraph/checkpoint/base.py
@@ -1,13 +1,18 @@
-import asyncio
 from abc import ABC
 from collections import defaultdict
 from copy import deepcopy
 from datetime import datetime, timezone
-from typing import Any, AsyncIterator, Iterator, NamedTuple, Optional, TypedDict
+from typing import (
+    Any,
+    AsyncIterator,
+    Iterator,
+    NamedTuple,
+    Optional,
+    Protocol,
+    TypedDict,
+)
 
-from langchain_core.load.serializable import Serializable
-from langchain_core.runnables import RunnableConfig
-from langchain_core.runnables.utils import ConfigurableFieldSpec
+from langchain_core.runnables import ConfigurableFieldSpec, RunnableConfig
 
 from langgraph.utils import StrEnum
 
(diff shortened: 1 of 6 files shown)


langgraph/checkpoint/memory.py
```diff
@@ -1,47 +1,84 @@
...
-                            "thread_id": config["configurable"]["thread_id"],
-                            "thread_ts": thread_ts,
-                        }
-                    },
-                    checkpoint=checkpoints[thread_ts],
+                    config={"configurable": {"thread_id": thread_id, "thread_ts": ts}},
+                    checkpoint=self.serde.loads(checkpoints[ts]),
                 )
 
+    def list(self, config: RunnableConfig) -> Iterator[CheckpointTuple]:
```
nfcampos, line 46: this was missing

langgraph/checkpoint/base.py
```diff
@@ -93,8 +98,27 @@ class CheckpointTuple(NamedTuple):
...
+        ...
+
+    def loads(self, data: bytes) -> Any:
+        ...
+
+
+class BaseCheckpointSaver(ABC):
+    at: CheckpointAt = CheckpointAt.END_OF_STEP
+
+    serde: SerializerProtocol
```
hinthornw, line 112: so this is for specifying an inheritable default?

langgraph/checkpoint/memory.py
```diff
@@ -1,47 +1,84 @@
...
-        if value := self.get_tuple(config):
-            return value.checkpoint
+    storage: defaultdict[str, dict[str, Checkpoint]]
+
+    def __init__(
+        self,
+        *,
+        serde: Optional[SerializerProtocol] = None,
+        at: Optional[CheckpointAt] = None,
+    ) -> None:
```
hinthornw, line 27: Could we add docstrings :)

hinthornw (approved):

langgraph/checkpoint/memory.py
nfcampos: haha yea but only after settling on final checkpointer interface

langgraph/checkpoint/base.py
```diff
@@ -93,8 +98,27 @@ class CheckpointTuple(NamedTuple):
...
+class SerializerProtocol(Protocol):
+    def dumps(self, obj: Any) -> bytes:
+        ...
+
+    def loads(self, data: bytes) -> Any:
+        ...
+
+
+class BaseCheckpointSaver(ABC):
+    at: CheckpointAt = CheckpointAt.END_OF_STEP
```
andrewnguonly, line 110: More granular default check pointing.

langgraph/checkpoint/base.py
nfcampos: you mean the previous line?

andrewnguonly (approved):
High level, makes sense!

PR #315, opened 2024-04-17
Tutorial

hinthornw:


hwchase17 (commented):
1/ tools_called should be a prebuilt edge imo
2/ the second graph is messed up (has an edge from action to llm)
3/ Checkpointing - should we just call memory? at very least we should mention memory as a main motivator - we dont even mention it (despite the first thing being about memory)
4/ Check pointing should be split up in persistence/memory and then the state modification (state modification is advanced, should come after human in the loop, is a form of human in the loop but more advanced)

5/ imo in addition to the standard prebuilt node/edge, should also add a custom node/edge/state
6/ in that respect, should probably use StateGraph (with messages key) not messageGraph

PR #316, opened 2024-04-17
fix: examples plan-and-execute.ipynb

Brant-lzh:
Running plan-and-execute.ipynb reported an error:

```
File [~/anaconda3/envs/llm/lib/python3.11/site-packages/langgraph/pregel/__init__.py:1033](https://file+.vscode-resource.vscode-cdn.net/home/brant/python/langgraph/examples/plan-and-execute/~/anaconda3/envs/llm/lib/python3.11/site-packages/langgraph/pregel/__init__.py:1033), in _panic_or_proceed(done, inflight, step)
   1031             inflight.pop().cancel()
   1032         # raise the exception
-> 1033         raise exc
   1034         # TODO this is where retry of an entire step would happen
...
---> 90     destinations = [self.ends[r] for r in result]
     91 else:
     92     destinations = result

TypeError: 'bool' object is not iterable
```

I found that only by returning str will it be converted into an array

`langgraph/graph/graph.py`

```python
    async def _aroute(
        self,
        input: Any,
        config: RunnableConfig,
        *,
        reader: Optional[Callable[[], Any]],
        writer: Callable[[str], Optional[Runnable]],
    ) -> Runnable:
        result = await self.condition.ainvoke(
            reader(config) if reader else input, config
        )
        if isinstance(result, str):     #this 
            result = [result]
        if self.ends:
            destinations = [self.ends[r] for r in result]
        else:
            destinations = result
        return writer(destinations) or input
```

So I decided to edge func back str like PR did, and the result worked normally.
I wonder if I made the right modification?  thx.


hinthornw (approved):

Brant-lzh:
@hinthornw Merge into the main branch?

PR #319, opened 2024-04-17
chat agent executor: catch parsing errors in tool calls

ccurme:
```
                         +-----------+
                         | __start__ |
                         +-----------+
                                *
                                *
                                *
                           +-------+
                           | agent |*
                       ****+-------+ ****
                   ****          *       ****
               ****              *           ****
             **                   *              ****
+-----------------+               *                  **
| should_continue |**            *                    *
+-----------------+  ********    *                    *
           *       ***       ********                 *
            *         ****      *    ********         *
            *             **   *             ****     *
      +---------+         +-------+             +--------+
      | __end__ |         | retry |             | action |
      +---------+         +-------+             +--------+
```


tests/test_pregel.py
```diff
@@ -1891,7 +1891,6 @@ def search_api(query: str) -> str:
     assert app.get_input_schema().schema_json() == snapshot
     assert app.get_output_schema().schema_json() == snapshot
     assert json.dumps(app.get_graph().to_json(), indent=2) == snapshot
-    assert app.get_graph().draw_ascii() == snapshot
```
ccurme, line 1894: I'm finding that this test is flaky, lmk if it should be deterministic.

https://github.com/langchain-ai/langgraph/pull/319/commits/bac26213e949a78c10606a1b2c96faf6cb964f80

langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -192,12 +200,21 @@ async def acall_model(state: AgentState):
...
+            retry_message = [retry_message]
+        return {"messages": retry_message}
+
     # Define a new graph
     workflow = StateGraph(AgentState)
 
     # Define the two nodes we will cycle between
     workflow.add_node("agent", RunnableLambda(call_model, acall_model))
     workflow.add_node("action", ToolNode(tools))
+    workflow.add_node("retry", RunnableLambda(add_retry_message))
```
ccurme, line 217: alternative to adding a new node is to just process messages containing parsing issues inside the `call_model` functions. for instance, we could add an AIMessage functioning as a thought/observation, similar to AgentExecutor. but then in `should_continue` we would need to distinguish this case.

PR #330, opened 2024-04-19
Add managed values and IsLastStep

nfcampos:
- managed values are read-only state keys whose values are managed by langgraph
- this PR implements one: IsLastValue, a boolean which is True in the last iteration, eg to allow you to return a nice "ran out of iterations" message to user


langgraph/pregel/__init__.py
```diff
@@ -1254,7 +1352,12 @@ def _prepare_next_tasks(
                     val: Any = {
                         k: read_channel(channels, chan, catch=chan not in proc.triggers)
                         for k, chan in proc.channels.items()
+                        if isinstance(chan, str)
                     }
+                    for key, chan in proc.channels.items():
+                        for mv in managed:
+                            if isclass(chan) and isinstance(mv, chan):
+                                val[key] = mv(step, PregelTaskDescription(name, val))
```
nfcampos, line 1360: this mutates the value throughout so the input to an MV depends on the order in which we find it, which is silly

andrewnguonly: New implementation:
```python
val: dict = {  # <-- type annotation is changed to dict, I don't know why it was Any
    k: read_channel(channels, chan, catch=chan not in proc.triggers)
    for k, chan in proc.channels.items()
    if isinstance(chan, str)
}

managed_values = {}
for key, chan in proc.channels.items():
    for mv in managed:
        if isclass(chan) and isinstance(mv, chan):
            managed_values[key] = mv(step, PregelTaskDescription(name, val))

val.update(managed_values)
```

PR #361, opened 2024-04-29
migrate to new imports

eyurtsev:
Output after running:

```shell
langchain-cli migrate --include-ipynb .
```

from root directory.

Had to edit the tests by hand due to local scope imports -- which aren't
handeled currently by the migration script.


hinthornw:
Re-ran in a separate PR! Thanks for making the migration CLI

PR #362, opened 2024-04-29
For cond edges support specifying list of possible destinations as a list or typing annotation

nfcampos:
…list or typing annotation


hinthornw:
Nice!

PR #370, opened 2024-04-30
Implement stream_mode=debug

nfcampos:
- outputs 3 types of payloads
  - task: node about to be executed
  - task_result: result of node execution
  - checkpoint: state values at end of superstep

### DIFF
--- a/langgraph/pregel/debug.py
+++ b/langgraph/pregel/debug.py
@@ -1,11 +1,122 @@
+import json
 from collections import defaultdict
+from datetime import datetime, timezone
 from pprint import pformat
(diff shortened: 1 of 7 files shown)


langgraph/pregel/debug.py
```diff
@@ -1,13 +1,124 @@
...
+        "timestamp": ts,
+        "step": step,
+        "payload": {
+            "config": config,
+            "values": read_channels(channels, stream_channels),
+        },
+    }
+
+
 def print_step_tasks(step: int, next_tasks: list[PregelExecutableTask]) -> None:
```
nfcampos, line 122: will consolidate this with debug printing in a future pr

PR #391, opened 2024-05-03
Add checkpoint metadata fields

nfcampos:
- source: input, update or loop
- step: int
- make step counter continue from previous last step

### DIFF
--- a/langgraph/checkpoint/aiosqlite.py
+++ b/langgraph/checkpoint/aiosqlite.py
@@ -1,7 +1,7 @@
 import asyncio
 from contextlib import AbstractAsyncContextManager
 from types import TracebackType
-from typing import Any, AsyncIterator, Optional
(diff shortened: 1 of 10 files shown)


langgraph/checkpoint/aiosqlite.py
```diff
@@ -88,7 +89,7 @@ async def aget_tuple(self, config: RunnableConfig) -> Optional[CheckpointTuple]:
                     return CheckpointTuple(
                         config,
                         self.serde.loads(value[0]),
-                        self.serde.loads(value[2]) if value[2] is not None else None,
+                        self.serde.loads(value[2]) if value[2] is not None else {},
```
hinthornw, line 92: why is this needed?

nfcampos: olde checkpoints don't have metadata

PR #399, opened 2024-05-06
Checkpoint inputs before starting the first step for easier error recovery

nfcampos:
- this enables easier retrying, for any error just do .invoke(None, config) no matter which step the error happened on

### DIFF
--- a/langgraph/checkpoint/sqlite.py
+++ b/langgraph/checkpoint/sqlite.py
@@ -1,5 +1,6 @@
 import pickle
 import sqlite3
+import threading
 from contextlib import AbstractContextManager, contextmanager
 from types import TracebackType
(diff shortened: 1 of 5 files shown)


langgraph/checkpoint/sqlite.py
```diff
@@ -41,10 +42,17 @@ def __init__(
         super().__init__(serde=serde)
         self.conn = conn
         self.is_setup = False
+        self.lock = threading.Lock()
 
     @classmethod
     def from_conn_string(cls, conn_string: str) -> "SqliteSaver":
-        return SqliteSaver(conn=sqlite3.connect(conn_string))
+        return SqliteSaver(
+            conn=sqlite3.connect(
```
hinthornw, line 50: nice

PR #405, opened 2024-05-07
Update README

hinthornw:


README.md
```diff
@@ -77,9 +77,9 @@ So what did we do here? Let's break it down step by step:
 
 1. First, we initialize our model and a `MessageGraph`.
 2. Next, we add a single node to the graph, called `"oracle"`, which simply calls the model with the given input.
-3. We add an edge from this `"oracle"` node to the special string `END`. This means that execution will end after current node.
+3. We add an edge from this `"oracle"` node to the special string `END` (`"__end__"`). This means that execution will end after the current node.
 4. We set `"oracle"` as the entrypoint to the graph.
-5. We compile the graph, ensuring that no more modifications to it can be made.
+5. We compile the graph, ensuring that it can be run.
```
nfcampos, line 82: should we be more explicit here and say something like "we compile the graph, translating it to low-level pregel operations?"

README.md
```diff
@@ -10,27 +10,27 @@
...
-It extends the [LangChain Expression Language](https://python.langchain.com/docs/expression_language/) with the ability to coordinate multiple chains (or actors) across multiple steps of computation in a cyclic manner.
+It extends the [LangChain Expression Language](https://python.langchain.com/docs/expression_language/) with the ability to coordinate multiple chains (or actors) across multiple steps of computation in cycles.
 It is inspired by [Pregel](https://research.google/pubs/pub37252/) and [Apache Beam](https://beam.apache.org/).
-The current interface exposed is one inspired by [NetworkX](https://networkx.org/documentation/latest/).
+The current public interface draws inspiration from [NetworkX](https://networkx.org/documentation/latest/).
 
 The main use is for adding **cycles** to your LLM application.
-Crucially, LangGraph is **NOT optimized for acyclic**, or Directed Acyclic Graph (DAG), workflows.
-If you want to build a DAG, you should just use [LangChain Expression Language](https://python.langchain.com/docs/expression_language/).
+Crucially, LangGraph is not **optimized for acyclic**, or Directed Acyclic Graph (DAG), workflows.
```
nfcampos, line 18: is this actually true?

README.md
```diff
@@ -124,10 +124,10 @@ graph.add_node("oracle", chain)
 
 ## Conditional edges
 
-Now, let's move onto something a little bit less trivial. Because math can be difficult for LLMs, let's allow the LLM to conditionally call a `"multiply"` node using tool calling.
+Now, let's move onto something a little bit less trivial. LLMs struggle with math,so let's allow the LLM to conditionally call a `"multiply"` node using [tool calling](https://python.langchain.com/docs/modules/model_io/chat/function_calling/).
```
nfcampos, line 127: missing a space after math

README.md
```diff
@@ -141,47 +141,46 @@ def multiply(first_number: int, second_number: int):
...
+builder.set_entry_point("oracle")
 ```
 
 Now let's think - what do we want to have happened?
 
 - If the `"oracle"` node returns a message expecting a tool call, we want to execute the `"multiply"` node
 - If not, we can just end execution
 
-We can achieve this using **conditional edges**, which routes execution to a node based on the current state using a function.
+We can achieve this using **conditional edges**, which call a function on the current state and routes execution to a node the function's output.
```
nfcampos, line 161: I think might be missing a word here?

hinthornw: ha good catch

PR #424, opened 2024-05-09
Add conceptual guide

hinthornw:
Would benefit from some great images, but I tried to make a decent explainer of the core ideas behind langgraph


WesGBrooks:
Was just reading the docs to get up to speed on langgraph, and this has already been a helpful resource to help me understand the key ideas better than any prior docs! Thanks so much @hinthornw !

PR #442, opened 2024-05-14
Small fixes while migrating JS

### LINKED ISSUE #429: Incorrect Validation in Graph class
### Checked other resources

- [X] I searched the LangChain documentation with the integrated search.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

On github, in `langgraph/langgraph/graph/graph.py`:

```
97   class Graph:
...
241     def validate(self, interrupt: Optional[Sequence[str]] = None) -> None:
...
259         for source in all_sources:
260             if node not in self.nodes and node != START:
261                 raise ValueError(f"Found edge starting at unknown node '{source}'")
```

N/A

### Description

"node" in line 260 of `langgraph/langgraph/graph/graph.py` should instead be `source` to properly validate that all edges have a known source.

### System Info

N/A

nfcampos:
Resolves #429

### DIFF
--- a/langgraph/graph/state.py
+++ b/langgraph/graph/state.py
@@ -269,7 +269,7 @@ class CompiledStateGraph(CompiledGraph):
             elif end != END:
                 # subscribe to start channel
                 self.nodes[end].triggers.append(starts)
-        else:
+        elif end != END:
             channel_name = f"join:{'+'.join(starts)}:{end}"
             # register channel
             self.channels[channel_name] = NamedBarrierValue(str, set(starts))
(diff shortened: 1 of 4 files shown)


langgraph/graph/state.py
```diff
@@ -269,7 +269,7 @@ def attach_edge(self, starts: Union[str, Sequence[str]], end: str) -> None:
             elif end != END:
                 # subscribe to start channel
                 self.nodes[end].triggers.append(starts)
-        else:
+        elif end != END:
```
hinthornw, line 272: No Else condition now?

hinthornw (approved):

langgraph/graph/state.py
nfcampos: else would be if end is END, which we don't need to do anything for

PR #444, opened 2024-05-14
Add `FewShotExamples` managed value (redo)

andrewnguonly:
### Summary
This PR is a reimplementation of the [old "Add FewShotExamples managed value" PR](https://github.com/langchain-ai/langgraph/pull/332). A new PR is created because there are merge conflicts with the old PR. The changes in this PR are mostly derived from the old PR with a few noted exceptions.

### Implementation
1. Add `score` field to `CheckpointMetadata`. Clients can specify a score for a checkpoint to mark a thread as "good". This is different from the old PR.
1. Add `search()` and `asearch()` APIs to `BaseCheckpointSaver` for searching checkpoints by metadata without filtering on `thread_id`. Implement the APIs in `MemorySaver`, `SqliteSaver`, and `AsyncSqliteSaver`. The implementation is mostly copied from the `list()` method. The old PR implemented an API called `list_w_score()`.
1. Copy `FewShotExamples` class from the old PR.
1. Copy the tests from the old PR.
1. Copy `learning.ipynb` from the old PR.

### To Do
- [x] Run the `learning.ipynb` notebook.
- [x] Close PR https://github.com/langchain-ai/langgraph/pull/332.

PR #447, opened 2024-05-14
Fix memory checkpointer list/alist

nfcampos:
- add tests for memory checkpointer list and alist
- fix async step counting

### DIFF
--- a/langgraph/checkpoint/memory.py
+++ b/langgraph/checkpoint/memory.py
@@ -1,5 +1,6 @@
 import asyncio
 from collections import defaultdict
+from functools import partial
 from typing import AsyncIterator, Iterator, Optional
 
 from langchain_core.runnables import RunnableConfig
@@ -112,7 +113,8 @@ class MemorySaver(BaseCheckpointSaver):
                 continue
             if limit is not None and limit <= 0:
                 break
-            limit -= 1
+            elif limit is not None:
+                limit -= 1
             yield CheckpointTuple(
                 config={"configurable": {"thread_id": thread_id, "thread_ts": ts}},
                 checkpoint=self.serde.loads(checkpoint),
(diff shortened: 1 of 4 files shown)


langgraph/checkpoint/memory.py
```diff
@@ -181,12 +189,16 @@ async def alist(self, config: RunnableConfig) -> AsyncIterator[CheckpointTuple]:
...
+            None, partial(self.list, before=before, limit=limit), config
+        )
         while True:
-            try:
-                yield await loop.run_in_executor(None, next, iter)
-            except StopIteration:
-                return
+            # handling StopIteration exception inside coroutine won't work
+            # as expected, so using next() with default value to break the loop
+            if item := await loop.run_in_executor(None, next, iter, None):
```
hinthornw, line 198: sad

andrewnguonly (approved):
Additional context:
1. https://peps.python.org/pep-0479/
2. https://peps.python.org/pep-0492/#why-stopasynciteration

From reference 2:
```
Moreover, with semantics from [PEP 479](https://peps.python.org/pep-0479/), all StopIteration exceptions raised in coroutines are wrapped in RuntimeError.
```

langgraph/checkpoint/memory.py
```diff
@@ -181,12 +189,16 @@ async def alist(self, config: RunnableConfig) -> AsyncIterator[CheckpointTuple]:
             AsyncIterator[CheckpointTuple]: An asynchronous iterator of checkpoint tuples.
         """
         loop = asyncio.get_running_loop()
-        iter = loop.run_in_executor(None, self.list, config)
+        iter = await loop.run_in_executor(
+            None, partial(self.list, before=before, limit=limit), config
```
andrewnguonly, line 193: I'll adopt this change in https://github.com/langchain-ai/langgraph/pull/444.

PR #454, opened 2024-05-15
Add single arg overload for add_node, which takes function/runnable's…

nfcampos:
… name as the node name

### DIFF
--- a/langgraph/graph/graph.py
+++ b/langgraph/graph/graph.py
@@ -14,6 +14,7 @@ from typing import (
     get_args,
     get_origin,
     get_type_hints,
+    overload,
 )
 
 from langchain_core.runnables import Runnable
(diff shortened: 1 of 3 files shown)


langgraph/graph/graph.py
```diff
@@ -106,18 +107,31 @@ def __init__(self) -> None:
     def _all_edges(self) -> set[tuple[str, str]]:
         return self.edges
 
-    def add_node(self, key: str, action: RunnableLike) -> None:
+    @overload
+    def add_node(self, node: RunnableLike) -> None:
+        ...
+
+    @overload
+    def add_node(self, node: str, action: RunnableLike) -> None:
```
hinthornw, line 115: I guess the alternative of having the action be first and name second would be weird because of the current behavior

hinthornw (approved):

langgraph/graph/graph.py
nfcampos: yea I think that would be confusing to have both

PR #457, opened 2024-05-15
Handle Async Callable

hinthornw:
```
class MyAsyncCallable:
    async def __call__(self, state):
        ...
```

### DIFF
--- a/langgraph/utils.py
+++ b/langgraph/utils.py
@@ -149,17 +149,12 @@ class DrawableGraph(Graph):
 
 
 def _isgencheck(thing: RunnableLike) -> bool:
-    return (
-        inspect.isasyncgenfunction(thing)
-        or inspect.isgeneratorfunction(thing)
-    )
+    return inspect.isasyncgenfunction(thing) or inspect.isgeneratorfunction(thing)
 
 
 def _isgenerator(thing: RunnableLike) -> bool:
     return (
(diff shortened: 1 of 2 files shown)


langgraph/utils.py
```diff
@@ -148,6 +148,31 @@ def extend(
         )
 
 
+def _isasyncgencheck(thing: RunnableLike) -> bool:
+    return (
+        inspect.isasyncgenfunction(thing)
+        or inspect.isasyncgen(thing)
+        or hasattr(thing, "__call__")
```
nfcampos, line 155: This seems recursive, which I don't think is on purpose?

langgraph/utils.py
```diff
@@ -148,6 +148,31 @@ def extend(
         )
 
 
+def _isasyncgencheck(thing: RunnableLike) -> bool:
+    return (
+        inspect.isasyncgenfunction(thing)
+        or inspect.isasyncgen(thing)
```
nfcampos, line 154: I don't think we need this one?

langgraph/utils.py
hinthornw: tired brain. ya

langgraph/utils.py
hinthornw: good catch

PR #458, opened 2024-05-15
Add graph for hierarchical agent teams

Undertone0809:
Add graphical displays for specific scenes to better represent the current state.

image

image

image


hinthornw:
Great!

PR #484, opened 2024-05-16
Add retry_policy for graphs

nfcampos:
- default is to not retry, fully configurable
- configuration options follow temporal https://docs.temporal.io/retry-policies#properties

### DIFF
--- a/langgraph/pregel/__init__.py
+++ b/langgraph/pregel/__init__.py
@@ -21,6 +21,7 @@ from typing import (
     overload,
 )
 
+from langchain_core.callbacks.manager import AsyncParentRunManager, ParentRunManager
 from langchain_core.globals import get_debug
 from langchain_core.load.dump import dumpd
 from langchain_core.pydantic_v1 import BaseModel, Field, root_validator
@@ -92,6 +93,7 @@ from langgraph.pregel.io import (
 )
 from langgraph.pregel.log import logger
 from langgraph.pregel.read import PregelNode
+from langgraph.pregel.retry import RetryPolicy, arun_with_retry, run_with_retry
 from langgraph.pregel.types import (
     All,
     PregelExecutableTask,
@@ -211,6 +213,9 @@ class Pregel(
 
     checkpointer: Optional[BaseCheckpointSaver] = None
     """Checkpointer used to save and load graph state. Defaults to None."""
+
+    retry_policy: Optional[RetryPolicy] = RetryPolicy()
+    """Retry policy to use when running tasks. Set to None to disable."""
 
     config_type: Optional[Type[Any]] = None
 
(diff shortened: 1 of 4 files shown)


langgraph/pregel/__init__.py
```diff
@@ -487,7 +492,9 @@ def update_state(
         saved = self.checkpointer.get_tuple(config)
         checkpoint = copy_checkpoint(saved.checkpoint) if saved else empty_checkpoint()
         # find last node that updated the state, if not provided
-        if as_node is None and not saved:
+        if as_node is None and not any(
+            v for vv in checkpoint["versions_seen"].values() for v in vv.values()
```
hinthornw, line 496: is this related?

langgraph/pregel/retry.py
```diff
@@ -0,0 +1,142 @@
...
+    ):
+        return False
+    if isinstance(exc, httpx.HTTPStatusError):
+        return 500 <= exc.response.status_code < 600
+    if isinstance(exc, requests.HTTPError):
+        return 500 <= exc.response.status_code < 600 if exc.response else True
+    return True
+
+
+class RetryPolicy(NamedTuple):
```
hinthornw, line 37: ooc: why a named tuple vs a typeddict?

langgraph/pregel/retry.py
```diff
@@ -0,0 +1,142 @@
...
+    while True:
+        try:
+            # clear any writes from previous attempts
+            task.writes.clear()
+            # run the task
+            if stream:
+                async for _ in task.proc.astream(task.input, task.config):
+                    pass
+            else:
+                await task.proc.ainvoke(task.input, task.config)
```
hinthornw, line 114: Could be nice to log the attempt number as metadata so you could track e.g., frequency of retries in ls
Would also presumably make it easier in something like stream events to check if you need to overwrite something? not sure

langgraph/pregel/retry.py
```diff
@@ -0,0 +1,142 @@
...
+            # if successful, end
+            break
+        except Exception as exc:
```
hinthornw, line 117: why not catch teh broader `BaseException` here?

hinthornw (approved):

langgraph/pregel/__init__.py
nfcampos: No, just an edge case I found while migrating js

langgraph/pregel/retry.py
nfcampos: because it allows defaults

langgraph/pregel/retry.py
nfcampos: Isn't the only difference that it includes KeyboardInterrupt and friends? I don't think we'd want to ever retry on those

langgraph/pregel/retry.py
nfcampos: yea agree, but not sure today where to log it, do you have an idea?

PR #489, opened 2024-05-18
Add support for passing `Callable` metadata filter to `FewShotExamples` managed value

andrewnguonly:
### Summary
`Callable` metadata filters are used for filtering on checkpoint metadata that is only known at runtime. For example, in OpenGPTs, an `assistant_id` (an instance of a graph), is only known after it is created. In this scenario, a developer should define a function that retrieves the `assistant_id` from the passed `RunnableConfig`.

### Implementation
1. Update `FewShotExamples` to support passing a `Callable` metadata filter.
1. ???

### Next Steps
1. Update `langgraph` version in LangGraph API after this feature is merged and released.
1. Update OpenGPTs, pass function to `FewShotExamples` for filtering checkpoint metadata by `assistant_id`.

### DIFF
--- a/langgraph/pregel/__init__.py
+++ b/langgraph/pregel/__init__.py
@@ -760,10 +760,11 @@ class Pregel(
                             )
                         )
                         checkpoint_config = {
+                            **checkpoint_config,
                             "configurable": {
                                 **checkpoint_config["configurable"],
                                 "thread_ts": checkpoint["id"],
-                            }
+                            },
                         }
                     # increment start to 0
                     start += 1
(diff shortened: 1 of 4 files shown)


langgraph/pregel/__init__.py
```diff
@@ -760,10 +760,11 @@ def stream(
                             )
                         )
                         checkpoint_config = {
+                            **checkpoint_config,
```
andrewnguonly, line 763: Hey @nfcampos, I forgot why we're doing this. Can you explain again? Thank you!

nfcampos: This is to ensure all checkpoints after the first contain the same metadata as the first one

nfcampos (approved):

PR #506, opened 2024-05-20
Add compat with stream_events v2

nfcampos:


eyurtsev (approved):
need to bump min core to 0.2.0 or else do a conditional import based on version in astream_events

PR #510, opened 2024-05-21
Merge configurable fields with previous checkpoint config before each run

andrewnguonly:
### Summary
The following scenario occurs in OpenGPTs/LangChain API, which is what prompted this change.
1. Thread run is posted and graph execution is interrupted before tool execution.
2. Inspecting the thread's checkpoint metadata shows that `graph_id` and `assistant_id` are present in the checkpoint metadata. This is correct an expected.
3. Resume graph execution (i.e. post another thread run with null input).
4. Inspecting the thread's checkpoint metadata shows that `graph_id` and `assistant_id` are NOT present in the checkpoint metadata. This is unexpected.

The solution/feature to accommodate this scenario is to merge a run's configurable fields with the previous checkpoint config (if any). This ensures that when a graph resumes execution from an interrupt, it should contain the configurable fields from the current run. This assumes that a client continues to send the same configurable fields across each run.

### Implementation.
1. Update `Pregel` `stream()`, `astream()`, `update_state()`, and `aupdate_state()` to merge configurable fields with previous checkpoint config.

### To Do
- [x] Double check that `update_state()` and `aupdate_state()` are supposed to be updated.
- [x] ~~Maybe refactor the code to deduplicate the repeated code.~~

### Notes
1. This PR blocks https://github.com/langchain-ai/opengpts/pull/341.

### DIFF
--- a/langgraph/pregel/__init__.py
+++ b/langgraph/pregel/__init__.py
@@ -546,8 +546,19 @@ class Pregel(
             # apply to checkpoint and save
             _apply_writes(checkpoint, channels, task.writes)
             step = saved.metadata.get("step", -2) + 1 if saved else -1
+
+            # merge configurable fields with previous checkpoint config
+            checkpoint_config = config
+            if saved:
+                checkpoint_config = {
+                    "configurable": {
+                        **config.get("configurable", {}),
+                        **saved.config["configurable"],
+                    }
+                }
+
             return self.checkpointer.put(
-                saved.config if saved else config,
+                checkpoint_config,
                 create_checkpoint(checkpoint, channels, step),
                 {
                     "source": "update",
@@ -621,8 +632,19 @@ class Pregel(
             # apply to checkpoint and save
             _apply_writes(checkpoint, channels, task.writes)
             step = saved.metadata.get("step", -2) + 1 if saved else -1
+
+            # merge configurable fields with previous checkpoint config
+            checkpoint_config = config
+            if saved:
+                checkpoint_config = {
+                    "configurable": {
+                        **config.get("configurable", {}),
+                        **saved.config["configurable"],
+                    }
+                }
+
             return await self.checkpointer.aput(
-                saved.config if saved else config,
+                checkpoint_config,
                 create_checkpoint(checkpoint, channels, step),
                 {
                     "source": "update",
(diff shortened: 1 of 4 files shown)


tests/test_pregel.py
```diff
@@ -5439,3 +5442,143 @@ def router(state: AgentState) -> str:
...
+                "test_config_3": "foo",
+                "test_config_4": "bar",
+            }
+        },
+    )
+
+    # assert that checkpoint metadata contains the run's configurable fields
+    chkpnt_metadata_3 = checkpointer_2.get_tuple(config).metadata
+    assert chkpnt_metadata_3["thread_id"] == "2"
+    assert chkpnt_metadata_3["test_config_3"] == "foo"
```
andrewnguonly, line 5583: This is the real test. This assertion does not succeed without the code changes.

langgraph/pregel/__init__.py
```diff
@@ -546,8 +546,19 @@ def update_state(
             # apply to checkpoint and save
             _apply_writes(checkpoint, channels, task.writes)
             step = saved.metadata.get("step", -2) + 1 if saved else -1
+
+            # merge configurable fields with previous checkpoint config
+            checkpoint_config = config
```
andrewnguonly, line 551: Double check: Are we supposed to update `update_state()` and `aupdate_state()`?

langgraph/pregel/__init__.py
```diff
@@ -621,8 +632,19 @@ async def aupdate_state(
...
+            if saved:
+                checkpoint_config = {
```
andrewnguonly, line 639: We're constructing a new "RunnableConfig" that doesn't have all of the keys defined in `RunnableConfig`. If a developer is implementing the `put()`/`aput()` API for a custom checkpointer, excluding the keys might be unexpected. Is this what we want?

langgraph/pregel/__init__.py
nfcampos: Yes always update both

tests/memory_assert.py
```diff
@@ -46,3 +49,63 @@ def put(
         )
         # call super to write checkpoint
         return super().put(config, checkpoint, metadata)
+
+
+class MemorySaverAssertCheckpointMetadata(MemorySaver):
```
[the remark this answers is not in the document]
nfcampos: We m don't want to mock checkpointers because we want to test them too

langgraph/pregel/__init__.py
andrewnguonly: Confirmed, we need this. When a tool query is manually updated, `update_state()` or `aupdate_state()` is called. The client needs to pass the same configurable fields for all thread update actions.

langgraph/pregel/__init__.py
nfcampos: All keys are optional

nfcampos (approved):

PR #520, opened 2024-05-22
ReWoo with persistence

lgesuellip:
### Summary

This PR implements the ReWoo example with persistence.
Based on the [LangChain ReWoo example](https://github.com/langchain-ai/langgraph/blob/main/examples/rewoo/rewoo.ipynb), I implemented a version that includes checkpoint saving. This implementation follows the approach used in the [Agents persistence example](https://github.com/langchain-ai/langgraph/blob/main/examples/persistence.ipynb).

### To Do
* [X]  Run the `rewoo.ipynb` notebook.


hwchase17:
this is great, but i dont think adds much. it is kind of a combination of two existing tutorials. if we do that for all tutorials, that won't scale. we'd be happy to highlight this separately!

PR #523, opened 2024-05-22
fix: getpass module call

deepankarm:
In a few notebooks, `getpass` call was incorrect resulting in a `TypeError: 'module' object is not callable` error. This PR fixes it.


hinthornw (approved):
Thanks for catching!

PR #524, opened 2024-05-22
Explicitly use config

hinthornw:


hinthornw:
I could add tests i guess

nfcampos (approved):

PR #528, opened 2024-05-23
Langgraph Code Generation Example

rlancemartin:


hinthornw:
FYI this won't show up in the docs unless we explicitly list it in the appropriate page (tutorials/index.md + mkdocs.yml)

I think the mistral specific one shouldn't be listed unless it's really demonstrating a new concept. The generic one probably worth keeping though

rlancemartin:
> FYI this won't show up in the docs unless we explicitly list it in the appropriate page (tutorials/index.md + mkdocs.yml)
> 
> I think the mistral specific one shouldn't be listed unless it's really demonstrating a new concept. The generic one probably worth keeping though

right. we don't want these in the docs. docs make sense only for canonical examples, like you said. this is reference for a video we're doing w/ mistral.

PR #534, opened 2024-05-26
docs: specify that configurables are added to LangSmith as metadata

Xtrah:
- Emphasize by explicitly stating that configurables are added to LangSmith as metadata. I spent hours trying different solutions here before I discovered that this was possible.
- Add link to conceptual guides in readme


hinthornw (approved):
Seems reasonable - at some point i may just make a separate tracing section, but won't have time to do that yet

PR #535, opened 2024-05-26
Mongo integratation to persist Data

Mahdi-Ba:
By following these steps, you can efficiently manage a stateful graph in LangGraph, leveraging its upsert capabilities to maintain a dynamic and consistent graph database. This approach is particularly useful for applications that require real-time updates and historical state tracking, such as social networks, recommendation systems,Human decision, and knowledge graphs.

### DIFF
--- a/examples/integrate_with_mongo/action.ipynb
+++ b/examples/integrate_with_mongo/action.ipynb
@@ -1,0 +1,247 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
+    "### Upsert in LangGraph\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


hinthornw:
Thanks for the PR, @Mahdi-Ba ! I think it'll be a couple more  weeks before we move the custom checkpointer workflows out of Beta, at which point something like this may be more viable. As it stands (and likely because of the lack of that documentation), this example doesn't actually use the LangGraph API as it is intended, so I wouldn't recommend people use this specific approach

hwchase17:
echoing @hinthornw , this is best done by creating a subclass of the langgraph checkpointer, something we will do once the interface is more stable

PR #537, opened 2024-05-27
Updated the JsonPlus class to handle generator objects within dictionaries, lists, and tuples.

shiwenbin1617:
- Updated the JsonPlus class to handle generator objects within dictionaries, lists, and tuples.
- This fixes the TypeError: Object of type generator is not JSON serializable issue.
image

### DIFF
--- a/langgraph/serde/jsonplus.py
+++ b/langgraph/serde/jsonplus.py
@@ -95,8 +95,20 @@ class JsonPlusSerializer(SerializerProtocol):
 
         return LC_REVIVER(value)
 
-    def dumps(self, obj: Any) -> bytes:
-        return json.dumps(obj, default=self._default, sort_keys=True).encode()
+    def dumps(self, obj):
+        return json.dumps(self._convert_generators(obj), default=self._default, sort_keys=True).encode()
+
+    def _convert_generators(self, obj):
+        import types
+        if isinstance(obj, types.GeneratorType):
+            return list(obj)
+        elif isinstance(obj, dict):
+            return {k: self._convert_generators(v) for k, v in obj.items()}
+        elif isinstance(obj, list):
+            return [self._convert_generators(elem) for elem in obj]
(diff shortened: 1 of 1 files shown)


hinthornw:
Thanks for the PR! will have to discuss this with @nfcampos . I think it would be counter-intuitive to consume the generator in the serializer, so at a glance it seems preferable to crash (aka not support generators in the state) and let the user decide transparently how to handle vs. trying to guess for them.

There's a few gotchas:

1. Generator could be infinite - we may never finish this / crash anyway
2. Generators may take long to consume - if the developer is returning a generator vs. returning a list, it's likely they're doing that for performance reasons
3. If we consume while serializing, need to ensure that the value in the state is still in the same location (would need to `tee` it at least I think)

Basically, though I may be overthinking this, right now I'm biased to prefer handling generators only explicitly via annotations or something else vs. automatically consuming them.


This PR does raise some interesting questions though. Are you using generators so you can do something more akin to LATS where you start processing in node T+1 while node T is still doing work?

shiwenbin1617:
Thank you for your reply. I thought about it 👌(｡•̀ᴗ-)✧ verification at the first time after receiving the reply. I seem to have found that this problem is due to the fact that I still stored the result of the streaming output in the state at the last node, which is a design problem of mine. Thank you for your reply and I will correct my problem. 
After that, I have some suggestions. 
1. Remind this problem as early as possible, because I have debugged for a long time because of this problem, sad. 
2. Can the official documentation add some examples for us to learn more advanced usages (because I have found many interesting implementations in continuous research). Finally, thank you again for your reply and look forward to langgraph getting better and better.

hinthornw:
@shiwenbin1617 
> Can the official documentation add some examples for us to learn more advanced usages 

What kind of advanced usages do you want more of? I have an incomplete list of things here: https://github.com/orgs/langchain-ai/projects/8 along with a longer personal list

PR #549, opened 2024-05-29
Add utilities to extract node and graph test cases from a thread

nfcampos:
- node test cases are input/output pairs for each node in the graph
- graph test cases are multi-turn test cases with sequences of inputs, outputs and trajectories

These can eg. be saved to a LangSmith dataset, or used for evals directly

### DIFF
--- a/langgraph/checkpoint/memory.py
+++ b/langgraph/checkpoint/memory.py
@@ -1,6 +1,7 @@
 import asyncio
 from collections import defaultdict
 from functools import partial
+from operator import itemgetter
 from typing import AsyncIterator, Iterator, Optional
 
 from langchain_core.runnables import RunnableConfig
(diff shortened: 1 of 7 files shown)


langgraph/checkpoint/memory.py
```diff
@@ -1,6 +1,7 @@
 import asyncio
```
nfcampos, line 1: MemorySaver didn't implement the entire checkpointer spec correctly, this diff fixes it

hinthornw: lol

langgraph/eval/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+    return dict(test_cases)
+
+
+def extract_node_test_cases_from_thread(
+    graph: Pregel, config: RunnableConfig
+) -> Dict[str, List[TestCase]]:
+    return _node_test_cases(graph.get_state_history(config))
+
+
+async def aextract_node_test_cases_from_thread(
```
hinthornw, line 60: Could we add docstrings and all that jazz for this? Kinda hard to know what to expect / how to use just from the signature

PR #553, opened 2024-05-30
Update introduction.ipynb

mltsy:
Replace old references to an "action" node with a "tools" node (since that is the name of the note in the first section of the tutorial)


hinthornw (approved):
Thank you!

PR #557, opened 2024-05-30
Implement map-reduce api for StateGraph/Pregel

nfcampos:
- current Pregel primitive is pull-based, ie. nodes write to channels, and it's up to other nodes to subscribe to those channels to "pull" updates
- this adds a "push" primitive where a node can directly schedule a node (more than once if desired) for execution in the next step, with additional kwargs to be passed in. Nodes scheduled in this way get called with both the current state and any kwargs passed to Packet


tests/test_pregel_async.py
```diff
@@ -2496,6 +2508,7 @@ def filter_by_source(config: RunnableConfig) -> Dict[str, Any]:
 
     class BaseState(TypedDict):
         messages: Annotated[list[AnyMessage], add_messages]
+        # tool_results: Annotated[list[str], operator.add]
```
hwchase17, line 2511: stray comment

nfcampos: oops

langgraph/pregel/__init__.py
```diff
@@ -1497,12 +1524,10 @@ def _apply_writes(
                 channels[chan].update(vals)
             except InvalidUpdateError as e:
                 raise InvalidUpdateError(
-                    f"Invalid update for channel {chan}: {e}"
+                    f"Invalid update for channel {chan} with values {vals}"
                 ) from e
             checkpoint["channel_versions"][chan] = max_version + 1
             updated_channels.add(chan)
-        else:
```
nfcampos, line 1504: this is done a little earlier now

langgraph/pregel/__init__.py
```diff
@@ -1550,6 +1575,64 @@ def _prepare_next_tasks(
...
+                    step,
+                    packet.node,
+                    processes[packet.node],
+                    managed,
+                    channels,
+                    catch=True,
+                )
+            )
+        except StopIteration:
+            logger.warn('No input for node "%s" in packet, skipping', packet.node)
```
nfcampos, line 1592: this can't happen in state graph, but could when using pregel directly

langgraph/pregel/__init__.py
```diff
@@ -1627,6 +1678,7 @@ def _prepare_next_tasks(
                                         "metadata": {
                                             "langgraph_step": step,
                                             "langgraph_node": name,
+                                            "langgraph_triggers": triggers,
```
nfcampos, line 1681: moar metadata

langgraph/pregel/__init__.py
```diff
@@ -1636,20 +1688,73 @@ def _prepare_next_tasks(
...
-                            sorted(triggers),
+                            triggers,
                         )
                     )
             else:
                 tasks.append(PregelTaskDescription(name, val))
     return checkpoint, tasks
 
 
+def _proc_input(
```
nfcampos, line 1707: pulling this logic out of prepare_next_tasks to share between pull and push code paths

langgraph/pregel/read.py
```diff
@@ -134,9 +140,12 @@ def get_node(self) -> Optional[Runnable[Any, Any]]:
...
             return RunnableSequence(*writers)
         elif writers:
-            return RunnableSequence(self.bound, *writers)
+            return RunnableSequence(
+                self.bound.bind(**kwargs) if kwargs is not None else self.bound,
+                *writers,
+            )
         else:
-            return self.bound
+            return self.bound.bind(**kwargs) if kwargs is not None else self.bound
```
nfcampos, line 148: need to do .bind() as RunnableSequence doesn't propagate kwargs to each step (nor should it I think)

langgraph/pregel/read.py
```diff
@@ -121,11 +121,17 @@ def get_writers(self) -> list[Runnable]:
             and isinstance(writers[-2], ChannelWrite)
         ):
             # we can combine writes if they are consecutive
-            writers[-2].writes += writers[-1].writes
+            # careful to not modify the original writers list or ChannelWrite
```
nfcampos, line 124: this was a bad bug, which was harmless because of a different bug fixed in write.py

langgraph/pregel/write.py
```diff
@@ -62,47 +76,69 @@ def config_specs(self) -> list[ConfigurableFieldSpec]:
         ]
 
     def _write(self, input: Any, config: RunnableConfig) -> None:
+        # split packets and entries
```
nfcampos, line 79: this code used to misbehave if the same node sent multiple updates for the same channel (not possible with state graph, but possible in pregel), it would keep only the last update for each channel (per node)

langgraph/utils.py
```diff
@@ -72,47 +72,47 @@ def __repr__(self) -> str:
         }
         return f"{self.get_name()}({', '.join(f'{k}={v!r}' for k, v in repr_args.items())})"
 
-    def invoke(self, input: Any, config: Optional[RunnableConfig] = None) -> Any:
+    def invoke(
+        self, input: Any, config: Optional[RunnableConfig] = None, **kwargs: Any
```
nfcampos, line 76: runnable callable didn't support runtime kwargs before

PR #579, opened 2024-06-03
Rm isinstance(..., NamedTuple)

hinthornw:
It's a function (not a type), so this line would raise an error.

If we want to custom serialize named tuples, would have to use orjson or separately recursively check for named tuples, since the json stdlib module doesn't permit overriding of stdlib types that are in the default conversion table: https://discuss.python.org/t/allowing-override-of-json-dumps-serialization-for-standard-types/16259/6


nfcampos:
thanks, superseded by #580

PR #608, opened 2024-06-06
fix: name must not container spaces

goern:
SSID


hinthornw (changes requested):
Will need to fix the other references too

PR #624, opened 2024-06-07
allow path_map keys to be non-strings in add_conditional_edges

### LINKED ISSUE #620: add_conditional_edges Type Hint
### Checked other resources

- [X] I searched the LangChain documentation with the integrated search.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```python
from typing import Annotated, TypedDict

from langchain_core.messages import AnyMessage


class State(TypedDict):


def branch(state: State) -> bool:
    return 1 + 1 == 3


graph_builder.add_node("foo", lambda state: {"messages": [("ai", "foo")]})
graph_builder.add_node("bar", lambda state: {"messages": [("ai", "bar")]})

graph_builder.add_conditional_edges(
    START,
    branch,   # type hint error
    path_map={True: "foo", False: "bar"},   # type hint error
    then=END,
)

app = graph_builder.compile()
print(app.invoke({"messages": []}))
```

It works fine, but the type hint is wrong.

### Description

When the return type of path is not str or the key of path_map is not str, the IDE will complain and mypy will complain.

```python
def add_conditional_edges(
        self,
        source: str,
        path: Union[
            Callable[..., Union[str, list[str]]],
            Callable[..., Awaitable[Union[str, list[str]]]],
            Runnable[Any, Union[str, list[str]]],
        ],
        path_map: Optional[Union[dict[str, str], list[str]]] = None,
        then: Optional[str] = None,
    ) -> None: ...
   ```
Is there a better Type Hint?
   
   

### System Info

langgraph==0.0.64
python_version==3.12.3

### PR DESCRIPTION
**Pull request by vbarda**
Fixes #620

### DIFF
--- a/langgraph/graph/graph.py
+++ b/langgraph/graph/graph.py
@@ -174,9 +174,9 @@ class Graph:
         self,
         source: str,
         path: Union[
-            Callable[..., Union[str, list[str]]],
-            Callable[..., Awaitable[Union[str, list[str]]]],
-            Runnable[Any, Union[str, list[str]]],
+            Callable[..., Union[Any, list[Any]]],
+            Callable[..., Awaitable[Union[Any, list[Any]]]],
+            Runnable[Any, Union[Any, list[Any]]],
         ],
         path_map: Optional[Union[dict[Any, str], list[str]]] = None,
         then: Optional[str] = None,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: langgraph/graph/graph.py [hunk 1]
```diff
@@ -235,9 +235,11 @@ def set_entry_point(self, key: str) -> None:
     def set_conditional_entry_point(
         self,
         path: Union[
-            Callable[..., str], Callable[..., Awaitable[str]], Runnable[Any, str]
+            Callable[..., Union[Hashable, list[Hashable]]],
```
hwchase17, line 238: i think this now changes functionality to allow entry point to multiple, which i think is fine, but probably deserves a test?

vbarda: added a unittest!

PR #625, opened 2024-06-08
Remove obsolete dependency groups from Dockerfile

sushengloong:


Dockerfile
```diff
@@ -7,6 +7,6 @@ WORKDIR /app
 COPY . .
 
 # Install any needed packages specified in requirements.txt
-RUN pip install poetry && poetry config virtualenvs.create false && poetry install --with test,lint,typing,dev
```
sushengloong, line 10: These 3 groups no longer exist in pyproject.toml

PR #643, opened 2024-06-11
first draft

isahers1:


hinthornw (commented):
Awesome start - thanks for leading on this! Reading over this as a user, here are my thoughts:

Structure:

Right now it feels like we could focus / frame the narrative more around user needs.

Maybe instead of just tool node -> execute manually -> non-explicit tools, we could frame it based on what the user wants to accomplish. These approaches aren’t mutually exclusive but illustrate ways we use tool calling to accomplish different tasks.

1. ToolNode: Directly execute tool functions.
    1. When it’s useful: if you can define utilities in specific 
    2.  Routing is between two nodes (llm and tool node). End condition is if no tool calls are present. Always execute values in the tool calls of the last ai message.
2. Customize the tool node: 
    1. Routing can still be between two nodes (llm and tool node). End condition is if no tool calls are present.
    2. Can combine the LLM’s provided values with other state or custom variables. Or could stream nested inputs within a node, etc.
3. Customize Routing: 
    1. “Tools” don’t have to be single functions, they are just ways for an LLM to generate structured output that your application can use. How it uses it is up to you. Some tools may not even correspond with a single scoped “tool” - they can be used for routing or for responding in a structured format. 
    2. We could actually maybe provide two examples here:
        1. If you want your app to “remember” the previous actions it has taken: maintain the `messages` list: The agent / llm node is the same as above and returns  messages
        2. If you want complete control and don’t want your LLM to keep track of all the previous action turns as messages: The llm uses `with_structured_output` or something similar and we store that schema in the state rather than as a message

And then remind the user again that these aren’t exhaustive or mutually exclusive.




Finally,  my take on general /style points for the final draft:

- Remove unnecessary adjectives (“simple”) and adverbs (“very”, “extremely” )  - either be specific about what the user gains or what something does, or remove entirely if it’s not meant to communicate something specific.
- For built-ins like “ToolNode”, link to the reference doc (https://langchain-ai.github.io/langgraph/reference/prebuilt/#toolnode)
- Try to avoid “lots of”  - this is intimidating to a reader. The reader needs it cleanly spelled out based on the distinct jobs they do. If there are two ways of doing something, just show one, or motivate the second example with a real need “I need to be able to combine the LLM output with other variables and so will choose to customize how I execute the tool calling a node”

PR #651, opened 2024-06-12
langgraph: add support for deleting messages

vbarda:
This change allows users or graph nodes to remove messages by `id` via `langchain_core.messages.RemoveMessage`

Examples:

* allow users to delete messages from state by calling

```python
graph.update_state(config, values=[RemoveMessage(id=state.values[-1].id)])
```

* allow nodes to delete messages

```python
graph.add_node("delete_messages", lambda state: [RemoveMessage(id=state[-1].id)])
```


langgraph/graph/message.py
```diff
@@ -1,18 +1,27 @@
...
     message_chunk_to_message,
 )
+from langchain_core.messages.base import BaseMessage
 
 from langgraph.graph.state import StateGraph
 
 Messages = Union[list[MessageLikeRepresentation], MessageLikeRepresentation]
 
 
+class DeleteMessage(BaseMessage):
```
hinthornw, line 17: ooc: Why not just make this a named tuple or typed dict and update the add_messages signature?

vbarda: yea, it's ephemeral. no specific reason other than for consistency, i don't think it gives us any benefit (and might actually add more headache). let me see if i can make it work

hinthornw: I had been doing this previously by having my reducer accept List[Union[AnyMessage, dict]] and then having the dict have an action -> IDK if we need to update our typing to let you differentiate between the channel type and the update type or anything though

jacoblee93: Seconding Will's suggestion if it's not massively more painful

vbarda: what are some other types of action just ooc? the obvious ones seem to be just `delete` / `update`, anything else?

langgraph/graph/message.py
```diff
@@ -10,6 +11,13 @@
 
 from langgraph.graph.state import StateGraph
 
+
+class MessageModifier(TypedDict):
```
vbarda, line 15: @hinthornw is this similar to what you had in mind? as an alternative to `DeleteMessage`

hinthornw: Ya!

nfcampos (approved):
Thanks!

PR #660, opened 2024-06-13
add example for persistence to DB

sameermahajan:
LangGraph has persistence. It persists state across all nodes and edges throughout the workflow. However, it has another subtle aspect that can persist it on disk / in DB so that when the application restarts (say after a crash or regular shutdown/maintenance) it can pick up where it left off. We will look at how it works in this example.

### DIFF
--- a/examples/langgraph_persist_to_db.ipynb
+++ b/examples/langgraph_persist_to_db.ipynb
@@ -1,0 +1,359 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "bbe13d39-aed8-45b0-b0d5-d64c750f6757",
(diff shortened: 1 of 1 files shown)


sameermahajan:
Did a YouTube video detailing this: https://youtu.be/GqKUmrd5FEk

hinthornw:
Hi @sameermahajan - thanks for making the video and opening the PR! Excited to hear you trying LangGraph out.

As far as I can tell, it seems like the provided notebook re-articulates ideas already well represented in our documentation, so I don't think we'll land this PR (since we would then become responsible for maintaining it) - you can feel free to put it in a separate repo, however!

PR #661, opened 2024-06-13
fix: update versions_seen when acting as a node

dqbd:
When updating the state via `update_state` / `aupdate_state`, the `versions_seen` is not updated, thus the next execution of the pregel will have include more tasks than expected


nfcampos (commented):
Could you add a test, ideally something that wouldn't pass before?

nfcampos:
No longer needed

PR #662, opened 2024-06-13
[docs]: add create_react_agent how-to notebook

vbarda:


docs/docs/how-tos/index.md
```diff
@@ -6,6 +6,7 @@ Welcome to the LangGraph how-to guides! These guides provide practical, step-by-
 
 The core guides show how to address common needs when building out AI workflows, with special focus placed on [ReAct](https://arxiv.org/abs/2210.03629)-style agents with [tool calling](https://python.langchain.com/docs/modules/model_io/chat/function_calling/).
 
+- [ReAct agent](create-react-agent.ipynb): How to create a simple ReAct agent
```
hinthornw, line 9: For the description, it may be more helpful to word it as something like? 

"How to create a simple tool-calling agent"
Less redundant . Maybe we even say what react means (How to create a tool-calling agent that **Re**asons and **Act**s to accomplish tasks)

examples/create-react-agent.ipynb
hinthornw, line 1: Do we need Tavily for these examples? Every dependency that requires the user to sign up for an API key makes the notebook harder to use

Also see a `HTTPError('502 Server Error: Bad Gateway for url: https://api.tavily.com/search')` in one of the searches anyway

hinthornw: Think `_set_env` should needs fixed (uncomment that line)

hinthornw: We have received feedback that people want fewer cells (to make it easier to copy code) - especially relevant for How-tos

hinthornw: Style nit/question: Is the colon in subtitles necessary? It adds no meaning and makes my brain pause every time I read it

vbarda: all great points

- re: tools, will just use some dummy
- re `set_env` good catch, thanks
- will merge more!
- totally not necessary, will remove

hinthornw (approved):
Nice - I'd probably split the first description so there's a little bit of welcome text at the top before running the installation, but am good with this

PR #688, opened 2024-06-18
libs: add separate langgraph cli, sdk-py, sdk-js libraries and move core langgraph

vbarda:
This change introduces several new langgraph libraries and a new repository structure:
- `langgraph` and other libraries will now live under `/libs` directory
- adds `langgraph_cli` for starting/managing LangGraph API server `/libs/cli`
- adds Python SDK for interacting with LangGraph API (`libs/sdk-py`)
- adds JS SDK for interacting with LangGraph API (`libs/sdk-js`)
- update CI that runs lint and test checks across all libraries
- add a new `pyproject.toml` for managing requirements for publishing docs


nfcampos:
@vbarda I made some updates to sdk today, lets make sure we dont lose those

vbarda:
> @vbarda I made some updates to sdk today, lets make sure we dont lose those

yup, i'll update it

libs/sdk-js/index.d.cts
```diff
@@ -0,0 +1 @@
+export * from './dist/index.cjs'
```
hinthornw, line 1: noob question - aren't these generated by `yarn build`? Or are we intending to check them in?

vbarda: actually great catch, i accidentally checked them in, let me remove

nfcampos (approved):
lgtm pending adding root licence

pyproject.toml
```diff
@@ -1,84 +1,26 @@
 [tool.poetry]
-name = "langgraph"
-version = "0.0.69"
-description = "langgraph"
+name = "langgraph-monorepo"
```
hinthornw, line 2:
```
Warning: The current project could not be installed: No file/folder found for package langgraph-monorepo
...

If you want to use Poetry only for dependency management but not for packaging, you can disable package mode by setting package-mode = false in your pyproject.toml file.
In a future version of Poetry this warning will become an error!
```

vbarda: nice, fixed!

vbarda: ugh, unfortunately it doesn't work w/ older versions of poetry, will make a TODO to bump poetry version in our CI and then fix up

PR #695, opened 2024-06-18
first draft

isahers1:


docs/.gitignore
```diff
@@ -1,2 +1 @@
-*.ipynb
```
andrewnguonly, line 1: @hinthornw is this a safe change to commit?

PR #703, opened 2024-06-19
Update pypi info for each package

nfcampos:


libs/sdk-py/pyproject.toml
```diff
@@ -1,9 +1,11 @@
 [tool.poetry]
 name = "langgraph-sdk"
-version = "0.1.21"
-description = ""
-authors = ["Nuno Campos <nuno@langchain.dev>"]
+version = "0.1.22"
+description = "CLI for interacting with LangGraph API"
```
vbarda, line 4: typo, should be SDK?

nfcampos: oops

vbarda (approved):

PR #705, opened 2024-06-19
add release workflows

vbarda:


.github/workflows/release.yml
```diff
@@ -191,19 +215,19 @@ jobs:
...
+      - name: Create Tag
         uses: ncipollo/release-action@v1
         with:
           artifacts: "dist/*"
           token: ${{ secrets.GITHUB_TOKEN }}
-          draft: false
-          generateReleaseNotes: true
-          tag: v${{ needs.build.outputs.version }}
-          commit: master
+          generateReleaseNotes: false
```
baskaryan, line 231: ooh why turn off release notes

baskaryan (commented):
looks reasonable to me!

.github/workflows/release.yml
vbarda: will do that in a separate change later

nfcampos (approved):

PR #723, opened 2024-06-20
cli: fix windows issues

vbarda:


libs/cli/langgraph_cli/exec.py
```diff
@@ -100,12 +107,13 @@ def signal_handler():
...
             if proc.returncode is None:
                 try:
-                    os.killpg(os.getpgid(proc.pid), signal.SIGINT)
+                    proc.terminate()
                 except (ProcessLookupError, KeyboardInterrupt):
                     pass
 
-            loop.remove_signal_handler(signal.SIGINT)
-            loop.remove_signal_handler(signal.SIGTERM)
+            if sys.platform != "win32":
```
nfcampos, line 114: we need to remove handler on windows too

vbarda: it's not added on windows?

PR #754, opened 2024-06-21
cli: add analytics

vbarda:


libs/cli/langgraph_cli/analytics.py
```diff
@@ -0,0 +1,72 @@
...
+
+    if kwargs.get("debugger_port"):
+        params["debugger_port"] = True
+
+    if kwargs.get("postgres_uri"):
+        params["postgres_uri"] = True
+
+    # pick up exact values for boolean flags
+    for boolean_param in ["recreate", "pull", "watch", "wait", "verbose"]:
+        if kwargs.get(boolean_param):
```
nfcampos, line 37: Should this have "is not None"

libs/cli/langgraph_cli/analytics.py
```diff
@@ -0,0 +1,72 @@
...
+            "os": platform.system(),
+            "os_version": platform.version(),
+            "python_version": platform.python_version(),
+            "cli_version": __version__,
+            "cli_command": func.__name__,
+            "params": get_anonymized_params(kwargs),
+        }
+
+        # TODO: update w/ LangChain supabase key
+        headers = {"apikey": os.environ["SUPABASE_PUBLIC_API_KEY"]}
```
nfcampos, line 59: I think this needs to be hardcoded in the code? The cli won't ship with an env file

libs/cli/langgraph_cli/analytics.py
```diff
@@ -0,0 +1,72 @@
...
+        supabase_url = os.environ["SUPABASE_URL"]
+        try:
+            httpx.post(
```
nfcampos, line 62: This should be done in background thread? Currently this would block until it returns

vbarda: good point, yes!

libs/cli/langgraph_cli/analytics.py
vbarda: depends on whether we actually want to log `False` for booleans -- I also considered in the beginning but decided against it. we can just always assume that if the flag is not in the `params` it's set to `False`

libs/cli/langgraph_cli/analytics.py
```diff
@@ -0,0 +1,89 @@
...
+    params: dict[str, Any]
+
+
+def get_anonymized_params(kwargs: dict[str, Any]) -> dict[str, bool]:
+    params = {}
+
+    # anonymize params with values
+    if config := kwargs.get("config"):
+        if config != pathlib.Path(DEFAULT_CONFIG).resolve():
+            params["config"] = config
```
nfcampos, line 34: Should this be True

libs/cli/langgraph_cli/analytics.py
```diff
@@ -0,0 +1,89 @@
...
+def log_data(data: LogData) -> None:
+    headers = {"apikey": SUPABASE_PUBLIC_API_KEY}
+    supabase_url = SUPABASE_URL
+    try:
+        httpx.post(
+            f"{supabase_url}/rest/v1/logs",
+            json=data,
+            headers=headers,
+        )
+    except httpx.HTTPStatusError:
```
nfcampos, line 66: This would actually never be raised, only if you were calling raise_for_status. Look at https docs to confirm but I think you want to catch like connection error or something

vbarda: nice, didn't realize this -- updated!

nfcampos (approved):

PR #772, opened 2024-06-23
bug fix: expected a list got a tuple

yassinech-99:
image
it should have returned a list not a tuple, our state has past_steps: Annotated[List[Tuple], operator.add], yet this node returns a tuple


vbarda:
Thanks for your contribution, this issue has already been fixed!

PR #781, opened 2024-06-24
Wfh/craggycrag

hinthornw:
@rlancemartin - i still don't love the react structure for this, but this seems like a semi-reasonable way of getting it to grade itself if we want to restrict it to a simple loop with static tools.

If i wanted a single step up, could maybe include a method to check "which react state it's in" and then restrict the agent to a single tool for the next turn:

1. Iniitally in "retrieve" stage: only show retrieve
2. then in "grade" stage: only show "Grade" as the tool
3. then in check and transition to whichever state based on the results there

etc.


rlancemartin (approved):
Ya. This seems clever. 

1/ We want the agent to explicitly call a `grade documents` tool. 
2/ But, we don't want the tool arguments to include the raw documents. This has high token cost.
3/ So, the tool call simply asks the LLM to show its work with `step_by_step_reasoning` and provide the grade, `score`.
4/ We then use this to route to the next tool.
so you simply have the LLM compress its grading into the tool call itself.

PR #782, opened 2024-06-24
docs: Create LangGraph Cloud API reference page

andrewnguonly:
### Summary
The LangGraph Cloud API reference page is built using `neoteroi-mkdocs`.


andrewnguonly:
Not sure if I want to merge this yet. Pending discussion about `neoteroi-mkdocs`.

nfcampos:
Why not just use the same we use in the server (scalar)?

andrewnguonly:
> Why not just use the same we use in the server (scalar)?

Just figured out how to do this. Looks way better! Closing this PR.

PR #785, opened 2024-06-24
[CI] Add formatting

hinthornw:


Makefile
```diff
@@ -16,5 +16,10 @@ clean-docs:
 	find ./docs/docs -name "*.ipynb" -type f -delete
 	rm -rf docs/site
 
+format format_diff:
+	poetry run ruff format libs/ examples/
```
vbarda, line 20: we have individual formatters in each library, so probably let's remove `libs` from here

hinthornw: Hm ok

PR #804, opened 2024-06-25
sdk-py: added Cron search functionality

isahers1:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -687,6 +687,17 @@ async def create(
...
+    async def list(
+        self, assistant_id: Optional[str], thread_id: Optional[str] = None
+    ) -> ThreadState:
+        """Get a list of all the cron jobs."""
+        if thread_id:
+            return await self.http.get(f"/threads/{thread_id}/runs/crons")
+        elif assistant_id:
+            return await self.http.get(f"/assistants/{assistant_id}/runs/crons")
+        else:
+            return await self.http.get(f"runs/crons")
```
akira, line 700: assistant_id should be required for cron also (similar to run)

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -687,6 +687,20 @@ async def create(
     async def delete(self, cron_id: str) -> None:
         """Delete a cron."""
         await self.http.delete(f"/runs/crons/{cron_id}")
+    
+    async def list(
```
akira, line 691: can you call this search

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -687,6 +687,20 @@ async def create(
...
+        assistant_id: Optional[str] = None,
+        thread_id: Optional[str] = None
+    ) -> ThreadState:
+        """Get a list of all the cron jobs."""
+        payload = {
+            "assistant_id" : assistant_id,
+            "thread_id": thread_id,
+        }
+        payload = {k: v for k, v in payload.items() if v is not None}
+        return await self.http.post(f"/runs/crons/search", json=payload)
```
akira, line 703: don't need f"" since there's no replacements

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -412,8 +412,7 @@ def stream(
         interrupt_after: Optional[list[str]] = None,
         feedback_keys: Optional[list[str]] = None,
         multitask_strategy: Optional[MultitaskStrategy] = None,
-    ) -> AsyncIterator[StreamPart]:
-        ...
+    ) -> AsyncIterator[StreamPart]: ...
```
andrewnguonly, line 415: This is causing the lint error. The ellipse (`...`) should be on a new line. Undo this change and all of the other instances and you should be good to go!
```python
) -> AsyncIterator[StreamPart]:
    ...
```

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -15,6 +15,7 @@
 from langgraph_sdk.schema import (
     Assistant,
     Config,
+    Cron,
```
vbarda, line 18: does this exist?

isahers1: totally forgot! it should now

vbarda (approved):

PR #809, opened 2024-06-25
Fix bug in add_conditional_edges when no path_map is provided

aaw:
When an instance of a callable class is passed as the path arg to add_conditional_edges but no path_map is provided, get_type_hints(path) is called, which raises a TypeError (since get_type_hints only accepts a module, class, method, or function).

This patch fixes the error by trying to get type hints from path.\_\_call\_\_ first, which should work for instances of callable classes.

Tested: Added a test that raises TypeError without the fix in this patch but passes with the fix.


nfcampos (approved):
Thanks!

PR #816, opened 2024-06-26
cli: update CAT -> RUN in the dockerfile

vbarda:


nfcampos (approved):

andrewnguonly (approved):
Verified that this works in `langchain-test` GCP project.

PR #826, opened 2024-06-26
Move studio how-to to top level outside of deployments

jacoblee93:
Testing/iteration is important and doesn't seem to belong nested under "deployment"

CC @hwchase17


hwchase17:
i think should be handled in isaacs refactor, will tal

PR #839, opened 2024-06-26
docs: add postgres checkpointer example

vbarda:


nfcampos:
can you share the sql strings between sync and async methods

vbarda:
> can you share the sql strings between sync and async methods

sure!

PR #863, opened 2024-06-27
sdk-js: added cron functionality to match langgraph-api

isahers1:


libs/sdk-js/src/client.mts
```diff
@@ -104,6 +106,96 @@ class BaseClient {
...
+      metadata: payload?.metadata,
+      assistant_id: assistantId,
+      interrupt_before: payload?.interruptBefore,
+      interrupt_after: payload?.interruptAfter,
+      webhook: payload?.webhook,
+    };
+    return this.fetch<Run>(`/runs/crons`, {
+      method: "POST",
+      json,
+      signal: payload?.signal,
```
vbarda, line 162: do we need signal here and above? cc @nfcampos i think we only need for streaming methods?

libs/sdk-js/src/schema.ts
```diff
@@ -68,6 +68,16 @@ export interface Thread {
   metadata: Metadata;
 }
 
+export interface Cron {
+  cron_id: string;
+  thread_id: Optional<string>;
+  end_time: Optional<Date>;
+  schedule: string;
+  created_at: Date;
```
vbarda, line 76: i think it will be just string unless you explicitly convert on return

vbarda (approved):

libs/sdk-js/src/client.mts
```diff
@@ -104,6 +106,94 @@ class BaseClient {
...
+      method: "DELETE",
+    });
+  }
+
+  /**
+   *
+   * @param query Query options.
+   * @returns List of crons.
+   */
+  async search(query?: {
```
nfcampos, line 179: these haven't been merged yet, so this pr needs to wait

PR #867, opened 2024-06-27
[Docs] How to pass other params to tools

hinthornw:
Be it in state or by configuration, it's common for tools to require information beyond what the LLM explicitly populates.

Here is a how-to on doing so.


hinthornw:
Gonna merge this with an existing guide.

PR #888, opened 2024-06-30
Add missing pip install libraries that are used in the example 

hayorov:
Add missing pip install libraries that are used in the example (Python imports).


vbarda:
they're actually installed below in the example, but we'll probably update / add comments for clarity! thank you!

PR #891, opened 2024-06-30
Interupt during tool call

Kirushikesh:
There is no break-statement which required to achieve the proposed technique, of asking for the user approval before the tool invocation. The current implementation doesn't stop when the LLM returns a tool_call.


nfcampos:
What determines interruption is the argument `interrupt_before=["action"]` to `.compile()`

PR #898, opened 2024-07-01
fix: typo

yoogle96:
fix typo `# How to stream from the final node node` -> `# How to stream from the final node`


vbarda:
thank you for fixing!

PR #900, opened 2024-07-01
feat(cli): add debugger host CLI flag to specify default base URL

dqbd:


libs/cli/langgraph_cli/cli.py
```diff
@@ -509,6 +521,7 @@ def prepare(
         port=port,
         watch=watch,
         debugger_port=debugger_port,
+        debugger_host=debugger_host or f"http://127.0.0.1:{DEFAULT_PORT}",
```
vbarda, line 524: hm, why is the default host not just `127.0.0.1`?

vbarda: (read more, i think it's a bit confusing that we have both host and port, but we're configuring whether the debugger is turned on or not based on port being provided. not sure if this is standard practice, but that would be a bit confusing to me as a user

libs/cli/langgraph_cli/docker.py
```diff
@@ -117,11 +105,36 @@ def check_capabilities(runner) -> DockerCapabilities:
     )
 
 
+def debugger_compose(*, port: Optional[int] = None, host: Optional[str] = None) -> str:
+    if port is None:
```
vbarda, line 109: IMO this is not very intuitive. could we do this check on host instead?

- cli option for host: no default in `click`
- cli option for port: default to `DEFAULT_PORT` in `click`

and below add environment only `if host != localhost`?

dqbd: This is not really for changing the host of the debugger image, rather to change the default URL used by the debugger to access the LangGraph API. 

Renamed the CLI flag to closely match the intent.

vbarda: ah, that makes sense, thanks for clarifying!

vbarda (approved):

PR #902, opened 2024-07-02
[Docs] Update notebooks to use START

hinthornw:
Migrated using gritql to use START instead of the set_entry_point

So we advertise one way of doing things


deesumblip:
Do we need to swap out `create_openai_functions_agent` with `create_openai_tools_agent` to keep this running in the next few months? Unfortunately a one-to-one swap is not working and the agent workflow seems to fall apart when using the latter.

RemanenetSpy:
Transitioning from naive checkpointing to true episodic memory requires building and tuning custom NLP extraction pipelines (like SpaCy or LLM extractors) to reliably parse raw text into Subject-Verb-Object (SVO) tuples before vectorizing them.

If your roadmap includes actively training extraction classifiers and maintaining a custom graph mapping layer, building this NLP pipeline internally is appropriate.

However, if your objective is to achieve SVO-based semantic search immediately without building the extraction layer, Smriti resolves this natively. It automatically extracts, anchors, and indexes SVO tuples upon text ingestion via its Managed API. This allows you to perform exact episodic semantic retrieval immediately, without building a custom NLP extraction pipeline.

PR #910, opened 2024-07-03
langgraph: add tool error handling

vbarda:


hinthornw (approved):
Ya this is what I had in mind. I'd also be a fan of letting the user pass a callable in:

```
handle_tool_errors = Optional[Union[
bool,
Callable[[BaseException, ToolCall, Optional[BaseTool]], str]
]] = None
```

But perhaps yagni

PR #939, opened 2024-07-08
feat: support using `add_node` as a decorator

Ahmad-Alsaleh:
This pull request adds support for using the `add_node` method as a decorator in the `StateGraph` class. Previously, the method could only be used to add nodes by passing the node name and action as arguments. With this change, developers can now use the `add_node` method as a decorator to define nodes directly within the class definition. This provides a more concise and intuitive way to add nodes to the state graph.

Examples:
```python
graph = Graph(...)

# All of the following are supported

@graph.add_node("search node")
def search():
    pass

# Uses search.__name__ as the name of the node

@graph.add_node()
def search():
    pass

@graph.add_node
def search():
    pass
```

The default behavior is still supported as well

```python
def search():
    pass

graph.add_node(search)

graph.add_node("search tool", search)
```

### DIFF
--- a/libs/langgraph/langgraph/graph/graph.py
+++ b/libs/langgraph/langgraph/graph/graph.py
@@ -1,5 +1,6 @@
 import logging
 from collections import defaultdict
+from functools import partial
 from typing import (
     Any,
     Awaitable,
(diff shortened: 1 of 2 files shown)


Ahmad-Alsaleh:
Note: in #209, @nfcampos supported a PR that implements similar functionality to the one in this PR.

vbarda:
Closing since a duplicate of #940

PR #940, opened 2024-07-08
feat: support using `add_node` as a decorator

Ahmad-Alsaleh:
This pull request adds support for using the `add_node` method as a decorator in the `StateGraph` class. Previously, the method could only be used to add nodes by passing the node name and action as arguments. With this change, developers can now use the `add_node` method as a decorator to define nodes directly within the class definition. This provides a more concise and intuitive way to add nodes to the state graph.

Examples:
```python
graph = Graph(...)

# All of the following are supported

@graph.add_node("search node")
def search():
    pass

# Uses search.__name__ as the name of the node

@graph.add_node()
def search():
    pass

@graph.add_node
def search():
    pass
```

The default behavior is still supported as well

```python
def search():
    pass

graph.add_node(search)

graph.add_node("search tool", search)
```


vbarda:
Thanks for your contribution! Going to close since this is already supported in the library https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/tests/test_pregel.py#L2189

Ahmad-Alsaleh:
Hello @vbarda 👋🏻

The current implementation is slightly different and less convenient than the implementation of the proposed change in this PR. For example, while developers can use `@graph.add_node`, they can't use `@graph.add_node()` (with parentheses) or `@graph.add_node("search node")` (with a custom name).

PR #958, opened 2024-07-09
docs: add memory how-tos + update add_messages

hwchase17:


libs/langgraph/langgraph/graph/message.py
```diff
@@ -78,10 +78,11 @@ def add_messages(left: Messages, right: Messages) -> Messages:
     # merge
     left_idx_by_id = {m.id: i for i, m in enumerate(left)}
     merged = left.copy()
+    ids_to_remove = set()
```
vbarda, line 81: nice!

vbarda (approved):
lgtm, thanks for adding these! (pending CI issues)

libs/langgraph/tests/test_messages_state.py
```diff
@@ -0,0 +1,126 @@
...
+
+def test_missing_ids():
+    left = [HumanMessage(content="Hello")]
+    right = [AIMessage(content="Hi there!")]
+    result = add_messages(left, right)
+    assert len(result) == 2
+    assert all(isinstance(m.id, str) and UUID(m.id, version=4) for m in result)
+
+
+def test_remove_message():
```
vbarda, line 56: should we also add a test where we're sending multiple `RemoveMessage` w/ the same ID (assuming the issue you ran into)

PR #974, opened 2024-07-09
Add as_node= arg to invoke/stream

nfcampos:
- This acts as a shortcut for the current flow of update_state(as_node=) then invoke(None)

### DIFF
--- a/libs/langgraph/langgraph/pregel/__init__.py
+++ b/libs/langgraph/langgraph/pregel/__init__.py
@@ -17,6 +17,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Tuple,
     Type,
     Union,
     cast,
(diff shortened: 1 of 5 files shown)


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -762,8 +764,8 @@ def stream(
                 updates: Emit only the updates to the state for each step.
                     Output is a dict with the node name as key and the updated values as value.
                 debug: Emit debug events for each step.
-            output_keys: The keys to stream, defaults to all non-context channels.
-            input_keys: The keys to use from the input, defaults to all input channels.
+            outputs: The keys to stream, defaults to all non-context channels.
+            as_node: The keys to use from the input, defaults to all input channels.
```
nfcampos, line 768: update

PR #976, opened 2024-07-10
lib: Checkpoint pending writes whenever a node finishes

nfcampos:
- Whenever a node finishes, checkpoint pending writes
- [x] Use pending_writes at beginning of stream (where we deal with input writes?)
- [x] Add test for two concurrent nodes, the longer one fails, the shorter one isn't re-run on resume
- [x] in sqlite/aiosqlite savers read pending writes in get_tuple

### DIFF
--- a/libs/langgraph/langgraph/checkpoint/base.py
+++ b/libs/langgraph/langgraph/checkpoint/base.py
@@ -118,7 +118,7 @@ class CheckpointTuple(NamedTuple):
     checkpoint: Checkpoint
     metadata: CheckpointMetadata
     parent_config: Optional[RunnableConfig] = None
-    pending_writes: Optional[List[Tuple[str, Any]]] = None
+    pending_writes: Optional[List[Tuple[str, str, Any]]] = None
 
 
 CheckpointThreadId = ConfigurableFieldSpec(
--- a/libs/langgraph/langgraph/checkpoint/base.py
+++ b/libs/langgraph/langgraph/checkpoint/base.py
@@ -185,7 +185,9 @@ class BaseCheckpointSaver(ABC):
         writes: List[Tuple[str, Any]],
         task_id: str,
     ) -> None:
-        raise NotImplementedError
+        raise NotImplementedError(
+            "This method was added in langgraph 0.1.7. Please update your checkpointer to implement it."
+        )
 
     async def aget(self, config: RunnableConfig) -> Optional[Checkpoint]:
         if value := await self.aget_tuple(config):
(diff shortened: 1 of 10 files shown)


libs/langgraph/langgraph/checkpoint/base.py
```diff
@@ -203,6 +213,14 @@ async def aput(
     ) -> RunnableConfig:
         raise NotImplementedError
 
+    async def aput_writes(
+        self,
+        config: RunnableConfig,
+        writes: List[Tuple[str, Any]],
+        task_id: str,
+    ) -> None:
+        raise NotImplementedError
```
hinthornw, line 222: Should we default this to running put_writes in executor or does that mess up the error?

nfcampos: No, let's make consistent with the other methods

libs/langgraph/tests/test_pregel.py
```diff
@@ -944,6 +956,96 @@ def raise_if_above_10(input: int) -> int:
...
+        one = AwhileMaker(0.2, {"value": 2})
+        two = AwhileMaker(0.6, ValueError("I'm not good"))
+        builder = StateGraph(State)
+        builder.add_node("one", one)
+        builder.add_node("two", two)
+        builder.add_edge(START, "one")
+        builder.add_edge(START, "two")
+        graph = builder.compile(checkpointer=checkpointer)
+
+        # test interrupting astream
```
hinthornw, line 1002: ubernit: dont see astream

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1825,6 +1893,14 @@ def _prepare_next_tasks(
...
         if for_execution:
             if node := processes[packet.node].get_node():
+                triggers = [TASKS]
+                metadata = {
+                    "langgraph_step": step,
+                    "langgraph_node": packet.node,
+                    "langgraph_triggers": triggers,
+                    "langgraph_task_idx": len(tasks),
+                }
+                task_id = str(uuid5(UUID(checkpoint["id"]), json.dumps(metadata)))
```
hinthornw, line 1903: Is it guaranteed that metadata is json serializable?  I think it is but wanted to check

hinthornw (approved):

libs/langgraph/langgraph/pregel/__init__.py
nfcampos: I created it just above, right? it's just strings and numbers

libs/langgraph/tests/test_pregel.py
nfcampos: oops copied comment from elsewhere

libs/langgraph/langgraph/pregel/__init__.py
hinthornw: ooops

HoangNguyen689:
@nfcampos 
Can I ask what the `writes` are used for?
I can't find any documents about it.

bieschke:
`put_writes` and `aput_writes` were added with a `NotImplementedError` referencing langgraph 0.1.7

This is confusing because it appears this change was created well after 0.1.7...

PR #977, opened 2024-07-10
langgraph[patch]: ToolNode support for tools outputting msg

baskaryan:
After https://github.com/langchain-ai/langchain/pull/24038, tools will be able to take in typed ToolCalls directly and will output a ToolMessage when invoked with a ToolCall

Requires `langchain-core==0.2.16` to be released


vbarda (approved):
nice, thank you!

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -112,16 +101,13 @@ async def _afunc(
...
+                return await self.tools_by_name[call["name"]].ainvoke(input, config)
             except Exception as e:
                 if not self.handle_tool_errors:
                     raise e
-                output = f"Error: {repr(e)}\n Please fix your mistakes."
-            return ToolMessage(
-                content=str_output(output), name=call["name"], tool_call_id=call["id"]
-            )
+                content = f"Error: {repr(e)}\n Please fix your mistakes."
+                return ToolMessage(content, name=call["name"], tool_call_id=call["id"])
```
hinthornw, line 110: why are we no lnoger passing content by keyword?

hinthornw:
So tools **always** return a tool message in this formulation? I must have missed where we guarantee this; it' sin the TOolCall itself?

libs/langgraph/langgraph/prebuilt/tool_node.py
baskaryan: cleaner this way (imo), messages accept content as pos arg

baskaryan:
> So tools **always** return a tool message in this formulation? I must have missed where we guarantee this; it' sin the TOolCall itself?

tools always return tool message if input is a ToolCall with "type" and "id" set

vbarda (approved):

PR #998, opened 2024-07-11
Remove broken edit button, view source button

jacoblee93:
CC @hinthornw


hinthornw (approved):
think it'll still be broken for the ntoebooks

jacoblee93:
> think it'll still be broken for the ntoebooks

Should we just remove altogether?

PR #999, opened 2024-07-11
Fix example

jacoblee93:
Getting `Type arguments for "Literal" must be None, a literal value (int, bool, str, or bytes), or an enum value`

CC @vbarda

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ tool_node = ToolNode(tools)
 model = ChatAnthropic(model="claude-3-5-sonnet-20240620", temperature=0).bind_tools(tools)
 
 # Define the function that determines whether to continue or not
-def should_continue(state: MessagesState):
+def should_continue(state: MessagesState) -> Union[Literal["tools"], type(END)]:
     messages = state['messages']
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -84,7 +82,7 @@ tool_node = ToolNode(tools)
 model = ChatAnthropic(model="claude-3-5-sonnet-20240620", temperature=0).bind_tools(tools)
 
 # Define the function that determines whether to continue or not
-def should_continue(state: MessagesState) -> Literal["tools", END]:
+def should_continue(state: MessagesState) -> Union[Literal["tools"], type(END)]:
```
vbarda, line 85: let's instead replace END w/ `__end__` (also might as well update in all docs with sed!)

README.md
```diff
@@ -57,8 +57,6 @@ export LANGSMITH_API_KEY=lsv2_sk_...
 ```
 
 ```python
-from typing import Annotated, Literal, TypedDict
```
vbarda, line 60: i think we still need typing import for Literal, right?

PR #1004, opened 2024-07-12
Support multiple interruptions after resuming execution

mme:
We noticed that it is currently not possible to interrupt a graph multiple times.

Once the graph resumes execution after an interruption, it just continues executing, ignoring `interrupt_before` and `interrupt_after`.

The reason is that after resuming this condition in `_should_interrupt` seems to always return false:
```
any(
    checkpoint["channel_versions"].get(chan, null_version)
    > seen.get(chan, null_version)
    for chan in snapshot_channels
)
```

In this PR I added a unit test in `test_interruption.py` to spec the desired behavior. I modified the code to pass the unit test, but since I do not understand what this code does it's probably not the right thing.

It would be great to get some guidance on how to fix this properly.

### DIFF
--- a/libs/langgraph/langgraph/pregel/__init__.py
+++ b/libs/langgraph/langgraph/pregel/__init__.py
@@ -1799,7 +1799,7 @@ def _should_interrupt(
     # defaultdicts are mutated on access :( so we need to copy
     seen = checkpoint["versions_seen"].copy()[INTERRUPT]
     return (
-        # interrupt if any of snapshopt_channels has been updated since last interrupt
+        # interrupt if any channel has been updated since last interrupt
         any(
             version > seen.get(chan, null_version)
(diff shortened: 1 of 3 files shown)


nfcampos (approved):
Thanks a lot for reporting this. I've updated the fix, but this was a great find.

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1801,9 +1801,8 @@ def _should_interrupt(
     return (
         # interrupt if any of snapshopt_channels has been updated since last interrupt
         any(
-            checkpoint["channel_versions"].get(chan, null_version)
-            > seen.get(chan, null_version)
-            for chan in snapshot_channels
+            version > seen.get(chan, null_version)
```
nfcampos, line 1804: the fix was to check all channels, not only the state key channels

PR #1012, opened 2024-07-12
Add guide on handling tool calling errors

jacoblee93:


vbarda:
thanks for adding! looks good overall, 2 comments:

* i would probably use the same example in both cases for consistency, otherwise feels a bit disconnected
* for the last example, can we use streaming instead, to show what was happening? i know we have a langsmith trace, but people might ignore that and it's not-obvious from the notebook outputs

examples/tool-calling-errors.ipynb
hinthornw, line 1: "Calling tools with an LLM is generally more reliable than pure prompting, but it isn't perfect" idk if this take is common, but I don't typically think of tools as a replacement for pure prompting. They're just the way the LLM interfaces with the rest of your program and with the outside world.

jacoblee93: What would you suggest?

jacoblee93:
> thanks for adding! looks good overall, 2 comments:
> 
> * i would probably use the same example in both cases for consistency, otherwise feels a bit disconnected
> * for the last example, can we use streaming instead, to show what was happening? i know we have a langsmith trace, but people might ignore that and it's not-obvious from the notebook outputs

1. They're two separate failure modes - I guess I could try to say "generate a haiku about a specific place" or three different cities but it'd be tough to prompt engineer.
2. Done

PR #1017, opened 2024-07-14
Update agent_supervisor.ipynb

RamziRebai:
If you choose to make custom agents and use the llm.bind_tools (), the previous code will generate an error! because we can either call a llm by bind_tools() or bind_functions(). We can not make a combination of them


vbarda:
Thanks for your contribution - this has already been resolved in the current documentation

PR #1023, opened 2024-07-15
langgraph: allow passing custom state & let message modifiers access state in create_react_agent

vbarda:
Addressing #991 

This change allows passing custom, user-defined `agent_state` schema to `StateGraph`, as well as a new, custom `state_modifier` parameter that takes in the whole graph state and prepares inputs to the LLM

```python
SYSTEM_INIT_PROMPT = """
You are a helpful assistant.

Today is {today}.
"""

prompt = ChatPromptTemplate.from_messages(
    [
        ("system", SYSTEM_INIT_PROMPT),
        ("placeholder", "{messages}"),
    ]
)

class AgentState(TypedDict):
    today: str
    messages: Annotated[list[BaseMessage], add_messages]
    is_last_step: str


new_react_agent = create_react_agent(model, [search], state_schema=AgentState, state_modifier=prompt)
new_react_agent.invoke({
    "messages": [("human", "what's the weather in sf? make sure to mention today's date")], 
    "today": "July 12, 2004"
})
```


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -164,10 +180,55 @@ async def acall_tool(state: AgentState, config: RunnableConfig):
...
+        model_runnable = state_modifier | model
+    else:
+        raise ValueError(
+            f"Got unexpected type for `state_modifier`: {type(state_modifier)}"
+        )
+
+    return model_runnable
+
+
+def convert_messages_modifier_to_state_modifier(
```
vbarda, line 205: @hwchase17 added this helper, so should be easy to handle the messages_modifier-style functions from langchain

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -164,10 +180,73 @@ async def acall_tool(state: AgentState, config: RunnableConfig):
...
+        return state_modifier
+    elif isinstance(messages_modifier, Runnable):
+        state_modifier = (lambda state: state["messages"]) | messages_modifier
+        return state_modifier
+    raise ValueError(
+        f"Got unexpected type for `messages_modifier`: {type(messages_modifier)}"
+    )
+
+
+def _get_model_preprocessing_runnable(
```
hwchase17, line 227: not really sure worth being its own function, could just combine with _convert...

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -421,28 +518,16 @@ def should_continue(state: AgentState):
...
-        model_runnable = (lambda messages: [_system_message] + messages) | model
-    elif isinstance(messages_modifier, SystemMessage):
-        model_runnable = (lambda messages: [messages_modifier] + messages) | model
-    elif isinstance(messages_modifier, (Callable, Runnable)):
-        model_runnable = messages_modifier | model
-    else:
-        raise ValueError(
-            f"Got unexpected type for `messages_modifier`: {type(messages_modifier)}"
-        )
+    model_runnable = (
```
hwchase17, line 521: nit: i think more readable just to do two lines
```
get_messages_runnable = ....
model_runnable = get_messages_runnable | model
```

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: disagree here actually, i think better to split for readability

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -164,10 +180,73 @@ async def acall_tool(state: AgentState, config: RunnableConfig):
     return workflow.compile()
 
 
+def _get_state_modifier_runnable(state_modifier: Optional[StateModifier]) -> Runnable:
+    if state_modifier is None:
+        state_modifier_runnable = RunnableLambda(lambda state: state["messages"])
```
hinthornw, line 185: Would be good to have a name for these so the trace looks nicer

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -164,10 +180,73 @@ async def acall_tool(state: AgentState, config: RunnableConfig):
...
+    elif isinstance(state_modifier, str):
+        _system_message: BaseMessage = SystemMessage(content=state_modifier)
+        state_modifier_runnable = RunnableLambda(
+            lambda state: [_system_message] + state["messages"]
+        )
+    elif isinstance(state_modifier, SystemMessage):
+        state_modifier_runnable = RunnableLambda(
+            lambda state: [state_modifier] + state["messages"]
+        )
+    elif isinstance(state_modifier, types.FunctionType):
```
hinthornw, line 195: This won't trigger for:

```python
class Foo:
    def __call__(self, state):
        ...
```

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -34,6 +35,21 @@ class AgentState(TypedDict):
...
+    SystemMessage,
+    str,
+    Callable[[Sequence[BaseMessage]], Sequence[BaseMessage]],
+    Runnable[Sequence[BaseMessage], Sequence[BaseMessage]],
+]
+
+StateModifier = Union[
+    SystemMessage,
+    str,
+    Callable[[AgentState], Sequence[BaseMessage]],
```
hinthornw, line 48: Should this use generics? wouldn't mypy complain if I use a custom state?

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -164,10 +180,73 @@ async def acall_tool(state: AgentState, config: RunnableConfig):
...
+
+    return _get_state_modifier_runnable(state_modifier)
+
+
+@deprecated_parameter("messages_modifier", "0.1.8", "state_modifier", removal="0.2.0")
 def create_react_agent(
     model: LanguageModelLike,
     tools: Union[ToolExecutor, Sequence[BaseTool]],
-    messages_modifier: Optional[Union[SystemMessage, str, Callable, Runnable]] = None,
+    agent_state: Optional[AgentState] = None,
```
hinthornw, line 247: Shouldn't this be `Type[AgentState]`? Or use a TypeVar like:

```
S = TypeVar("S", bound="AgentState")
state: Optional[Type[S]] = None
```
And then update the state modifier type to accept either agent state or S? (if that's possible)

hinthornw: Also personal nit: could this just be `state_schema` ? That's what it is and i'd rather it be the same as in STateGraph.

Right now, the term "state" is overloaded betwen the schema, the instance of that state, and then the general "location" the state machine is in

hinthornw: Also, adding these in the current order is technically a breaking change since someone may have done `create_react_agent(model, tools, lambda messages: ...)`. Would prefer if we use `*` and make these all keyword only

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: great point! will fix

hwchase17 (approved):

PR #1027, opened 2024-07-15
langgraph: add nested interrupts/checkpoints

nfcampos:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -878,6 +890,12 @@ def stream(
                     },
                 }
 
+            # if we're in a subgraph and have a saved checkpoint,
```
nfcampos, line 893: this doesn't seem right, means subgraphs never accept new input after their first run?

vbarda: great point, let me fix

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -984,12 +1019,7 @@ def put_checkpoint(metadata: CheckpointMetadata) -> Iterator[Any]:
                     # increment start to 0
                     start += 1
                 else:
-                    # no input is taken as signal to proceed past previous interrupt
-                    checkpoint = copy_checkpoint(checkpoint)
-                    for k in channels:
-                        if k in checkpoint["channel_versions"]:
-                            version = checkpoint["channel_versions"][k]
-                            checkpoint["versions_seen"][INTERRUPT][k] = version
+                    raise ValueError("Received no input")
```
vbarda, line 1022: @nfcampos technically we shouldn't even get here, not sure what's the cleanest thing to do here

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1176,6 +1216,9 @@ def put_checkpoint(metadata: CheckpointMetadata) -> Iterator[Any]:
 
                 # set final channel values as run output
                 run_manager.on_chain_end(read_channels(channels, output_keys))
+        except GraphInterrupt as e:
+            run_manager.on_chain_end(e)
```
vbarda, line 1220: @nfcampos i don't think this is actually working, i still see error getting logged in langsmith (even though we are catching the interrupt correctly here). am i doing anything wrong?

image

nfcampos: I think passing an error to on_chain_end is wrong?

vbarda: yea, fair enough -- what should we pass instead?
i'll update that, but the issue is unrelated

nfcampos: pass `read_channels(channels, output_keys)` as we do in the line above?

nfcampos (commented):
Can you update the checkpointer reference implementations in docs too?

libs/langgraph/langgraph/pregel/__init__.py
nfcampos: Let's create a new exception class for this and use run line 1061 too. eg. `EmptyInputError`

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1108,7 +1143,13 @@ def put_checkpoint(metadata: CheckpointMetadata) -> Iterator[Any]:
                             del fut, task
 
                     # panic on failure or timeout
-                    _panic_or_proceed(done, inflight, step)
+                    # NOTE: for subgraphs we'll raise GraphInterrupt exception on interrupt
+                    exceptions_to_handle = () if is_subgraph else (GraphInterrupt,)
+                    try:
```
nfcampos, line 1150: can we do an if statement instead? ie don't use a try statement without exceptions to catch

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -945,8 +964,27 @@ def put_checkpoint(metadata: CheckpointMetadata) -> Iterator[Any]:
                             ),
                         )
 
+                if is_subgraph and not checkpoint["channel_values"]:
```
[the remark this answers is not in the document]
vbarda: this is needed to handle the case of parent graph interrupt before subgraph node with interrupt: when we pass `input=None` to resume _parent_ graph, subgraph hasn't run anything yet at that point. so if we set `is_resuming` for subgraph to True at that point, we would try to resume with an empty subgraph checkpoint and break. so we do this check to actually pass inputs to the subgraph in that scenario

let me know if there's a better way to do it!

libs/langgraph/langgraph/pregel/__init__.py
vbarda: but we don't have access to `ChannelsManager` at that point, should we add it there?

libs/langgraph/langgraph/pregel/__init__.py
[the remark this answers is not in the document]
nfcampos: I see, okay, I'll reorganize the code a bit, but makes sense

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1616,12 +1615,14 @@ def put_checkpoint(metadata: CheckpointMetadata) -> Iterator[Any]:
...
-                    try:
-                        _panic_or_proceed(done, inflight, step, asyncio.TimeoutError)
-                    except exceptions_to_handle:
-                        break
+                    if is_subgraph:
+                        _panic_or_proceed(done, inflight, step)
+                    else:
+                        # NOTE: for subgraphs we'll raise GraphInterrupt exception on interrupt
+                        try:
+                            _panic_or_proceed(done, inflight, step)
```
nfcampos, line 1623: missing asyncio.TimeoutError

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1691,8 +1692,10 @@ def put_checkpoint(metadata: CheckpointMetadata) -> Iterator[Any]:
 
                 # set final channel values as run output
                 await run_manager.on_chain_end(read_channels(channels, output_keys))
-        except GraphInterrupt as e:
-            await asyncio.shield(run_manager.on_chain_end(e))
+        except GraphInterrupt:
+            await asyncio.shield(
```
nfcampos, line 1696: don't think we need shield here? only when handling cancelled error

libs/langgraph/tests/test_pregel.py
```diff
@@ -7514,7 +7514,18 @@ def side(state: State):
...
+    [
+        MemorySaverAssertImmutable(),
+        SqliteSaver.from_conn_string(":memory:"),
+    ],
+    ids=[
+        "memory",
+        "sqlite",
+    ],
+)
+def test_nested_graph_interrupts(checkpointer: BaseCheckpointSaver) -> None:
```
nfcampos, line 7528: line 7567 shadows this

libs/langgraph/tests/test_pregel.py
```diff
@@ -7640,7 +7651,18 @@ def outer_2(state: State):
...
+    [
+        MemorySaverAssertImmutable(),
+        SqliteSaver.from_conn_string(":memory:"),
+    ],
+    ids=[
+        "memory",
+        "sqlite",
+    ],
+)
+def test_nested_graph_interrupts_parallel(checkpointer: BaseCheckpointSaver) -> None:
```
nfcampos, line 7665: line 7705 shadows

libs/langgraph/langgraph/pregel/__init__.py
vbarda: oops, something with went wrong when i merged, thanks

nfcampos:
To fix issue when adding call to put_checkpoint: move all checkpoint updates from prepare_next_tasks to apply_writes

libs/langgraph/tests/test_pregel_async.py
```diff
@@ -6006,6 +6006,381 @@ async def side(state: State):
...
+    [
+        MemorySaverAssertImmutable(),
+        AsyncSqliteSaver.from_conn_string(":memory:"),
+    ],
+    ids=[
+        "memory",
+        "sqlite",
+    ],
+)
+async def test_nested_graph_interrupts(checkpointer: BaseCheckpointSaver) -> None:
```
nfcampos, line 6020: whenever using sqlite saver you need to close it at end of test, see other tests

nfcampos:
Superseded by #1075

PR #1056, opened 2024-07-18
docs: langgraph state arg

baskaryan:


examples/pass-run-time-values-to-tools.ipynb
```diff
@@ -5,18 +5,13 @@
    "id": "51466c8d-8ce4-4b3d-be4e-18fdbeda5f53",
    "metadata": {},
    "source": [
-    "# How to pass runtime values to tools\n",
+    "# How to pass graph state to tools\n",
```
vbarda, line 8: could you please the notebook name too? and put it into tool calling section https://langchain-ai.github.io/langgraph/how-tos/tool-calling/

examples/pass-run-time-values-to-tools.ipynb
```diff
@@ -67,52 +72,171 @@
...
+      ]
+     },
+     "execution_count": 65,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "get_context.tool_call_schema.schema()"
+   ]
```
vbarda, line 239: not from your change, but could you please update `StatefulGraph` in the cell below to `StateGraph`?

examples/pass-run-time-values-to-tools.ipynb
baskaryan: changing notebook name will break existing links

PR #1059, opened 2024-07-18
Make Pregel loop runnable step-by-step

nfcampos:
This lays the groundwork for lots of new ways to run graphs, eg
- run nodes in different machines
- place hard rate limits on particular nodes globally across all runs
- run loop as the target for a queue topic subscription (eg. event driven architecture)


libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -0,0 +1,315 @@
...
+            self.checkpoint, _ = prepare_next_tasks(
+                self.checkpoint,
+                self.graph.nodes,
+                self.channels,
+                self.managed,
+                self.config,
+                self.step,
+                for_execution=True,
+                get_next_version=self.get_next_version,
+                # TODO missing run_manager
```
nfcampos, line 239: add a test that fails without this

libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -0,0 +1,315 @@
...
+            for_execution=True,
+            get_next_version=self.get_next_version,
+        )
+
+        # if no more tasks, we're done
+        if not self.tasks:
+            self.status = "done"
+            return False
+
+        # TODO how to make this work for both
```
nfcampos, line 210: figure this out

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1288,228 +1061,70 @@ async def astream(
...
-                        interrupt_before,
-                        self.stream_channels_list,
-                        next_tasks,
-                    ):
-                        break
-                    else:
-                        checkpoint = next_checkpoint
-
+                # with channel updates applied only at the transition between steps
+                while loop.tick(
```
hinthornw, line 1084: And I guess loop.tick can be sync because we aren't currently doing io-bound work in it?

libs/langgraph/langgraph/checkpoint/base.py
```diff
@@ -25,6 +25,7 @@
 from langgraph.serde.jsonplus import JsonPlusSerializer
 
 V = TypeVar("V", int, float, str)
+PendingWrite = Tuple[str, str, Any]
```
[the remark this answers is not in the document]
nfcampos: yep

libs/langgraph/langgraph/checkpoint/base.py
```diff
@@ -53,6 +54,10 @@ class CheckpointMetadata(TypedDict, total=False):
     """
 
 
+class TaskInfo(TypedDict):
+    status: Literal["scheduled", "success", "error"]
```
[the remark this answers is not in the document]
nfcampos: it's more because scheduled the transition from scheduled to processing is handled by your queueing system, so we don't need to track here too. we're not using this yet (until we write the first queue adapter)

libs/langgraph/langgraph/pregel/__init__.py
nfcampos: yep, just python logic. all io things (like writing a checkpoint) get sent to background thread/task in both sync/async

metaimagine:
Cool work! Have a runable example? :)

PR #1062, opened 2024-07-18
Enable configuring retry policy for each node of StateGraph

nfcampos:


libs/langgraph/langgraph/pregel/types.py
```diff
@@ -46,14 +46,14 @@ class RetryPolicy(NamedTuple):
...
     max_interval: float = 128.0
     """Maximum amount of time that may elapse between retries. In seconds."""
-    max_attempts: int = 10
+    max_attempts: int = 3
     """Maximum number of attempts to make before giving up, including the first."""
     jitter: bool = True
     """Whether to add random jitter to the interval between retries."""
     retry_on: Union[
-        tuple[Exception, ...], Callable[[Exception], bool]
+        Type[Exception], tuple[Type[Exception], ...], Callable[[Exception], bool]
```
vbarda, line 54: should the last one be`Callable[[Type[Exception]], bool]` ?

nfcampos: No, the last one receives an exception instance

PR #1064, opened 2024-07-18
[docs]Fix runtime error README.md example

KYUUUW:
## version
```
langchain==0.2.5
langchain-anthropic==0.1.20
langgraph==0.1.9
```

## AS-IS (Error)

```python
# Define the tools for the agent to use
@tool
def search(query: str):
    """Call to surf the web."""
    # This is a placeholder, but don't tell the LLM that...
    if "sf" in query.lower() or "san francisco" in query.lower():
        
        return ["It's 60 degrees and foggy."]
    return ["It's 90 degrees and sunny."]
```

`BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.2.content.0.tool_result.content.0: Input should be a valid dictionary or object to extract fields from'}}`

## To-be

```python
# Define the tools for the agent to use
@tool
def search(query: str):
    """Call to surf the web."""
    # This is a placeholder, but don't tell the LLM that...
    if "sf" in query.lower() or "san francisco" in query.lower():
        
        return "It's 60 degrees and foggy."
    return "It's 90 degrees and sunny."
```
Result:
`"Based on the search results, I can tell you that the current weather in San Francisco is:\n\nTemperature: 60 degrees Fahrenheit\nConditions: Foggy\n\nSan Francisco is known for its microclimates and frequent fog, especially during the summer months. The cool temperature and foggy conditions are quite typical for the city. Remember that the weather can change throughout the day, so if you're planning any activities, it's a good idea to dress in layers and be prepared for potential changes in temperature and visibility.\n\nIs there anything else you'd like to know about the weather in San Francisco or any other information I can help you with?"
`

## Done😎


vbarda:
great catch, thanks! could you please also update in libs/langgraph/README?

KYUUUW:
Done :)

PR #1067, opened 2024-07-19
langgraph[patch]: InjectedState annotation 

baskaryan:
```python

    def tool3(some_val: int, foo: Annotated[str, InjectedState("foo")]) -> str:
        """Tool 1 docstring."""
        return foo

    tool_call = {
        "name": "tool3",
        "args": {"some_val": 1},
        "id": "some 0",
        "type": "tool_call",
    }
    msg = AIMessage("hi?", tool_calls=[tool_call])
    result = node.invoke({"messages": [msg], "foo": "bar"})
    tool_message: ToolMessage = result["messages"][-1]
    assert tool_message.content == "bar"

```


libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -127,6 +144,37 @@ def _validate_tool_call(self, call: ToolCall) -> Optional[ToolMessage]:
...
+            else:
+                err_msg = (
+                    f"Invalid input to ToolNode. Tool {tool_call['name']} requires "
+                    f"graph state dict as input."
+                )
+                if any(state_field for state_field in state_args.values()):
+                    required_fields_str = ", ".join(f for f in required_fields if f)
+                    err_msg += f" State should contain fields {required_fields_str}."
+                raise ValueError(err_msg)
+        tool_call_copy: ToolCall = copy(tool_call)
```
hinthornw, line 168: Instead of copy and mutate should we do:

```
tool_call_copy = {
  ***tool_call,
  "args": {
    ...
    }
}
```

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -347,3 +340,58 @@ def check_results(messages: list):
...
+        "args": {"some_val": 1},
+        "id": "some 0",
+        "type": "tool_call",
+    }
+    msg = AIMessage("hi?", tool_calls=[tool_call])
+    result = node.invoke({"messages": [msg]})
+    tool_message = result["messages"][-1]
+    assert tool_message.content == "hi?"
+
+    result = node.invoke([msg])
```
hinthornw, line 395: Could we check the other tools for `MessageGraph` as well?

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -183,3 +231,34 @@ def tools_condition(
     if hasattr(ai_message, "tool_calls") and len(ai_message.tool_calls) > 0:
         return "tools"
     return "__end__"
+
+
+class InjectedState(InjectedToolArg):
```
hinthornw, line 236: Could we add a docstring and include hte path in `prebuilt.md` so it has a ref doc rendered? Include examples

hinthornw (commented):
Excited for this!

libs/langgraph/langgraph/prebuilt/tool_node.py
baskaryan: whats the diff? or do you mean stylistically

libs/langgraph/tests/test_prebuilt.py
baskaryan: as in pass in just messages to the others?

libs/langgraph/langgraph/prebuilt/tool_node.py
hinthornw: nvm

hinthornw (approved):

PR #1073, opened 2024-07-19
Don't include lambdas in __eq__ for BinOp

nfcampos:
When using forward refs, inline lambdas in Annotated are re-evaluated for every subclass, thus making the comparison fail

### DIFF
--- a/libs/langgraph/langgraph/channels/binop.py
+++ b/libs/langgraph/langgraph/channels/binop.py
@@ -56,9 +56,11 @@ class BinaryOperatorAggregate(Generic[Value], BaseChannel[Value, Value, Value]):
             pass
 
     def __eq__(self, value: object) -> bool:
-        return (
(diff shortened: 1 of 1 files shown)


libs/langgraph/langgraph/channels/binop.py
```diff
@@ -56,9 +56,11 @@ def __init__(self, typ: Type[Value], operator: Callable[[Value, Value], Value]):
             pass
 
     def __eq__(self, value: object) -> bool:
-        return (
-            isinstance(value, BinaryOperatorAggregate)
-            and value.operator == self.operator
+        return isinstance(value, BinaryOperatorAggregate) and (
+            value.operator is self.operator
```
hinthornw, line 60: we're sure we don't want to keep the `==` condition here?

hinthornw (approved):

PR #1076, opened 2024-07-20
[Docs] Format notebooks

hinthornw:


OmarHory:
I believe there should be a put_writes and aput_writes functions to Redis. Because it is giving me an error whenever I am trying to invoke the graph using `langgraph==0.1.14`

PR #1108, opened 2024-07-24
langgraph: update get_state to handle nested subgraph state

vbarda:


libs/langgraph/tests/test_pregel.py
```diff
@@ -8673,6 +8674,360 @@ def parent_2(state: State):
...
+            metadata={
+                "source": "loop",
+                "writes": {"outer_1": {"my_key": "hi my value"}},
+                "step": 1,
+            },
+            created_at=AnyStr(),
+            parent_config={"configurable": {"thread_id": "1", "thread_ts": AnyStr()}},
+            subgraph_state_snapshots=None,
+        )
+        assert app.get_state(config, include_subgraph_state=True) == StateSnapshot(
```
vbarda, line 8753: @nfcampos @hwchase17 see an example of nested graph state here

libs/langgraph/tests/test_pregel.py
```diff
@@ -8673,6 +8674,360 @@ def parent_2(state: State):
...
+                "source": "loop",
+                "writes": {"outer_1": {"my_key": "hi my value"}},
+                "step": 1,
+            },
+            created_at=AnyStr(),
+            parent_config={"configurable": {"thread_id": "1", "thread_ts": AnyStr()}},
+            subgraph_state_snapshots={
+                "inner": StateSnapshot(
+                    values={"my_key": "hi my value here"},
+                    next=(),
```
vbarda, line 8767: this should probably be `next=("inner_2",)`

libs/langgraph/langgraph/checkpoint/base.py
```diff
@@ -231,6 +231,12 @@ def list(
         """
         raise NotImplementedError
 
+    def list_subgraph_checkpoints(
```
vbarda, line 234: alternatively this can just be in `def list` w/ an additional parameter

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -432,28 +515,44 @@ def get_state_history(
...
-                yield StateSnapshot(
-                    read_channels(channels, self.stream_channels_asis),
-                    tuple(name for name, _ in next_tasks),
-                    config,
-                    metadata,
-                    checkpoint["ts"],
-                    parent_config,
+            # is there a way to do this more efficiently?
+            if include_subgraph_state:
+                checkpoint_tuples = self.checkpointer.list_subgraph_checkpoints(config)
```
vbarda, line 520: don't love this as there is a separate list call here, but not sure what's a better way besides somehow combining w/ `.list`

vbarda: but if this looks fine, will implement the same for async and add async tests

libs/langgraph/langgraph/checkpoint/sqlite.py
```diff
@@ -548,6 +549,7 @@ def search_where(
     config: Optional[RunnableConfig],
     filter: Optional[Dict[str, Any]],
     before: Optional[RunnableConfig] = None,
+    as_prefix: bool = False,
```
vbarda, line 552: TODO: document kwarg

libs/langgraph/tests/test_pregel.py
```diff
@@ -9093,6 +9109,704 @@ def parent_2(state: State):
...
+                "configurable": {
+                    "thread_id": "1",
+                    "checkpoint_ns": "",
+                    "checkpoint_id": AnyStr(),
+                }
+            },
+            subgraph_state_snapshots={
+                "inner": StateSnapshot(
+                    values={"my_key": "hi my value here"},
+                    next=(),
```
vbarda, line 9278: my only concern is that `next` here is not populated and i think it should. @nfcampos

libs/langgraph/langgraph/constants.py
```diff
@@ -67,23 +70,26 @@ class Send:
...
             node (str): The name of the target node to send the message to.
             arg (Any): The state or message to send to the target node.
+            id (str): ID associated with the Send.
         """
         self.node = node
         self.arg = arg
+        self.id = id or str(uuid4())
 
     def __hash__(self) -> int:
         return hash((self.node, self.arg))
```
nfcampos, line 89: the id should be included?

libs/langgraph/langgraph/constants.py
```diff
@@ -22,6 +23,7 @@
 END = "__end__"
 
 CHECKPOINT_NAMESPACE_SEPARATOR = "|"
+SEND_CHECKPOINT_NAMESPACE_SEPARATOR = ":"
```
nfcampos, line 26: why is this a different separator?

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -352,86 +361,287 @@ def managed_values_dict(self) -> dict[str, ManagedValueSpec]:
             if is_managed_value(v)
         }
 
-    def get_state(self, config: RunnableConfig) -> StateSnapshot:
-        """Get the current state of the graph."""
-        if not self.checkpointer:
-            raise ValueError("No checkpointer set")
+    def _get_nodes_and_channels(
```
nfcampos, line 364: in langgraph we don't add private methods to the class, sorry. if you utility methods they should be functions outside of the class

libs/langgraph/langgraph/constants.py
vbarda: so that we can distinguish between node names and IDs, otherwise we don't know if the last item in the `checkpoint_ns` is a node name or an ID

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -789,19 +982,44 @@ def _defaults(
...
         if config and config.get("configurable", {}).get(CONFIG_KEY_READ) is not None:
             # if being called as a node in another graph, always use values mode
             stream_mode = ["values"]
+
+            if (
+                (interrupt_before or interrupt_after)
+                and parent_checkpointer is not None
+                and self.checkpointer is None
+            ):
+                raise ValueError(
```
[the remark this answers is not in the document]
vbarda: ah, so you're saying we should always inherit it automatically if we use interrupts and otherwise just add it via `INHERIT_CHECKPOINTER` as an opt-in?

libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -293,6 +293,21 @@ def prepare_next_tasks(
         if for_execution:
             proc = processes[packet.node]
             if node := proc.get_node():
+                triggers = [TASKS]
```
nfcampos, line 296: bad merge, this same code is immediately above

libs/langgraph/langgraph/pregel/types.py
```diff
@@ -92,6 +92,8 @@ class StateSnapshot(NamedTuple):
     """Config used to fetch the parent snapshot, if any"""
     tasks: tuple[PregelTask, ...]
     """Tasks to execute in this step. If already attempted, may contain an error."""
+    subgraph_state_snapshots: Optional[dict[str, "StateSnapshot"]] = None
+    """State snapshots of subgraphs represented as a mapping from thread ID suffix to snapshot."""
```
nfcampos, line 96: I think you mean checkpoint namespace not thread suffix?

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -962,8 +1113,7 @@ def stream(
                         )
                         if not done:
                             break  # timed out
-                        for fut in done:
-                            task = futures.pop(fut)
+                        for fut, task in zip(done, [futures.pop(fut) for fut in done]):
```
nfcampos, line 1116: undo, I fixed this in a different way

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1212,8 +1363,8 @@ async def astream(
                         )
                         if not done:
                             break  # timed out
-                        for fut in done:
-                            task = futures.pop(fut)
+
+                        for fut, task in zip(done, [futures.pop(fut) for fut in done]):
```
nfcampos, line 1367: undo

libs/langgraph/langgraph/pregel/types.py
vbarda: oh great catch that was a leftover from a few weeks back, pre namespacing

nfcampos (approved):

nfcampos (approved):

PR #1130, opened 2024-07-25
sdk-py: add docstrings

isahers1:


docs/docs/cloud/concepts/api.md
```diff
@@ -12,6 +12,10 @@ An assistant is a configured instance of a [`CompiledGraph`][compiledgraph]. It
 
 The LangGraph Cloud API provides several endpoints for creating and managing assistants. See the <a href="../reference/api/api_ref.html#tag/assistantscreate" target="_blank">API reference</a> for more details.
 
+#### Configuring Assistants
+
+One of the benefits of LangGraph API is that it lets you create agents with different configurations. This is useful when you want to allow users to use different models or prompts when using your graph. This also allows you the ability to quickly test out different configurations without having to rewrite your graph every time. See <a href="https://langchain-ai.github.io/langgraph/cloud/how-tos/cloud_examples/configuration_cloud/">this</a> how-to for information on how to configure a deployed graph.
```
hinthornw, line 17: style preference (that I often deviate from): be very direct.

Instaed of "One of the benefits of LangGraph API is that it lets you create agents with different configurations. This is useful when you want to allow users to use different models or prompts when using your graph."

"You can save custom *assistants* from the same graph to set different default prompts, models, and other *configurations* without changing a line of code in your graph."

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -239,15 +239,57 @@ def __init__(self, http: HttpClient) -> None:
...
 
     async def get_graph(self, assistant_id: str) -> dict[str, list[dict[str, Any]]]:
-        """Get the graph of an assistant by ID."""
+        """Get the graph of an assistant by ID.
+
+        Args:
+            assistant_id (str): The ID of the assistant to get the graph of.
+
+        Returns:
+            dict[str, list[dict[str, Any]]]: The graph information for the assistant.
```
hinthornw, line 266: Would be good to include what that information actually is. Would be nice to type it as a TypedDict or something too.

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -239,15 +239,57 @@ def __init__(self, http: HttpClient) -> None:
...
-        """Get an assistant by ID."""
+        """Get an assistant by ID.
+
+        Args:
+            assistant_id (str): The ID of the assistant to get.
+
+        Returns:
+            Assistant: The returned assistant.
+
+        Example Usage:
```
hinthornw, line 250: Could you check that this renders nicely? I think mkdocstrings checks for `Example:` and is very sensitive to spacing

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -259,7 +301,31 @@ async def create(
         assistant_id: Optional[str] = None,
         if_exists: Optional[OnConflictBehavior] = None,
     ) -> Assistant:
-        """Create a new assistant."""
+        """Create a new assistant.
+
+        Useful when graph is configurable and you want to create different assistants based on different configurations.
+
+        Args:
+            graph_id (str): The ID of the graph the assistant should use.
```
hinthornw, line 309: Could you clarify this is usually set in the `langgraph.json` file?

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -413,7 +640,28 @@ async def update_state(
         as_node: Optional[str] = None,
         checkpoint_id: Optional[str] = None,
     ) -> None:
-        """Update the state of a thread."""
+        """Update the state of a thread.
+
+        Args:
+            thread_id (str): The ID of the thread to get the state of.
```
hinthornw, line 646: nit: "to update" instaedof "to get the state of"

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -510,7 +794,50 @@ def stream(
...
+                If None a random UUID will be generated.
+            assistant_id (str): The assistant ID or graph name to stream from.
+                If using graph name, will default to first assistant created from that graph.
+            input (optional, dict): The input to the graph. Defaults to None.
+            stream_mode (optional, StreamMode or list[StreamMode]): The stream mode(s) to use.
+                Defaults to "values".
+            metadata (optional, dict): Metadata to assign to the run. Defaults to None.
+            config (optional, Config): The configuration for the assistant. Defaults to None.
+            checkpoint_id (optional, str): The checkpoint to start streaming from. Defaults to None.
+            interrupt_before (optional, list[str]): Nodes to interrupt streaming on immediately they get executed.
```
hinthornw, line 810: nit `immediately they get executed.`

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -320,6 +423,15 @@ async def search(
 
         Returns:
             list[Assistant]: A list of assistants.
+
```
hinthornw, line 426: COuld we clarify how the m etadata filter works? Assuming it's an AND of exact match filters for each KV pair??

hinthornw (commented):
Thanks for adding!

docs/docs/cloud/concepts/api.md
isahers1: Yup, this sounds much better - thanks for the feedback!

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -239,15 +239,168 @@ def __init__(self, http: HttpClient) -> None:
...
 
     async def get(self, assistant_id: str) -> Assistant:
-        """Get an assistant by ID."""
+        """Get an assistant by ID.
+
+        Args:
+            assistant_id (str): The ID of the assistant to get.
+
+        Returns:
+            Assistant: The returned assistant.
```
vbarda, line 248: nit: `The returned assistant` -> `Assistant object`

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -239,15 +239,168 @@ def __init__(self, http: HttpClient) -> None:
         self.http = http
 
     async def get(self, assistant_id: str) -> Assistant:
-        """Get an assistant by ID."""
+        """Get an assistant by ID.
+
+        Args:
+            assistant_id (str): The ID of the assistant to get.
```
vbarda, line 245: do we even need the types if the functions have type annotations?

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -259,7 +412,31 @@ async def create(
...
+        """Create a new assistant.
+
+        Useful when graph is configurable and you want to create different assistants based on different configurations.
+
+        Args:
+            graph_id (str): The ID of the graph the assistant should use. The graph ID is normally set in your langgraph.json configuration.
+            config (Config, optional): Configuration to use for the graph. Defaults to None.
+            metadata (dict, optional): Metadata to add to assistant. Defaults to None.
+            assistant_id (str, optional): Assistant ID to use, will default to a random UUID if not provided.
+            if_exists (OnConflictBehavior, optional): How to handle duplicate creation. Defaults to None.
```
vbarda, line 424: can we document this better:

(1) it actually will default to `raise` under the hood
(2) `raise` = raise an error if there is a duplicate
(3) `do_nothing` = return existing assistant

vbarda: same for threads

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -378,7 +672,29 @@ async def search(
...
+                Must be one of 'idle', 'busy', or 'interrupted'.
+            limit (int, optional): Limit on number of threads to return. Defaults to 10.
+            offset (int, optional): Offset in threads table to start search from. Defaults to 0.
+
+        Returns:
+            list[Thread]: List of the threads matching the search parameters.
+
+        Example Usage:
+
+            threads = await client.threads.searcb(
```
vbarda, line 690: typo: `searcb`

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -445,7 +905,27 @@ async def get_history(
         before: Optional[str] = None,
         metadata: Optional[dict] = None,
     ) -> list[ThreadState]:
-        """Get the history of a thread."""
+        """Get the history of a thread.
```
vbarda, line 908: `history of a thread` -> `state history of a thread`

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -579,7 +1107,91 @@ async def create(
...
+        Args:
+            thread_id (optional,str): the thread ID to assign to the thread. Defaults to None.
+                If None a random UUID will be generated.
+            assistant_id (str): The assistant ID or graph name to stream from.
+                If using graph name, will default to first assistant created from that graph.
+            input (optional, dict): The input to the graph. Defaults to None.
+            metadata (optional, dict): Metadata to assign to the run. Defaults to None.
+            config (optional, Config): The configuration for the assistant. Defaults to None.
+            checkpoint_id (optional, str): The checkpoint to start streaming from. Defaults to None.
+            interrupt_before (optional, list[str]): Nodes to interrupt streaming on immediately before they get executed.
```
vbarda, line 1121: `Nodes to interrupt streaming on immediately before they get executed.` can we change this to just `Nodes to interrupt immediately before they get executed.`

streaming is confusing in the context of a background run

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -672,28 +1348,113 @@ async def wait(
     async def list(
         self, thread_id: str, *, limit: int = 10, offset: int = 0
     ) -> List[Run]:
-        """List runs."""
+        """List runs.
+
+        Args:
+            thread_id (str): The thread ID to delete.
```
vbarda, line 1354: typo -- should be thread ID of the thread to list runs for or the like

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -672,28 +1348,113 @@ async def wait(
...
         return await self.http.get(
             f"/threads/{thread_id}/runs?limit={limit}&offset={offset}"
         )
 
     async def get(self, thread_id: str, run_id: str) -> Run:
-        """Get a run."""
+        """Get a run.
+
+        Args:
+            thread_id (str): The thread ID to delete.
```
vbarda, line 1378: typo - delete

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -672,28 +1348,113 @@ async def wait(
...
+            run_id (str): The run ID to delete.
+
+        Returns:
+            Run: The returned Run.
```
vbarda, line 1382: nit Run object instead of "returned Run"

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -672,28 +1348,113 @@ async def wait(
...
             f"/threads/{thread_id}/runs/{run_id}/cancel?wait={1 if wait else 0}",
             json=None,
         )
 
     async def join(self, thread_id: str, run_id: str) -> None:
-        """Block until a run is done."""
+        """Block until a run is done.
+
+        Args:
+            thread_id (str): The thread ID to delete.
```
vbarda, line 1424: this is a copy-pasted typo, need to fix everywhere

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -672,28 +1348,113 @@ async def wait(
...
+        """  # noqa: E501
+
         return await self.http.get(f"/threads/{thread_id}/runs/{run_id}")
 
     async def cancel(self, thread_id: str, run_id: str, *, wait: bool = False) -> None:
-        """Get a run."""
+        """Get a run.
+
+        Args:
+            thread_id (str): The thread ID to delete.
```
vbarda, line 1399: typo this and below

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -650,7 +1262,71 @@ async def wait(
         webhook: Optional[str] = None,
         multitask_strategy: Optional[MultitaskStrategy] = None,
     ) -> Union[list[dict], dict[str, Any]]:
-        """Create a run, wait for and return the final state."""
+        """Create a run, wait for and return the final state.
```
vbarda, line 1265: `"Create a run, wait until it finishes and return the final state."`

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -650,7 +1262,71 @@ async def wait(
...
+
+        Args:
+            thread_id (optional,str): the thread ID to create the run on. Defaults to None.
+                If None a random UUID will be generated.
```
vbarda, line 1269: this is actually not the point in this case -- this is for creating stateless runs. e.g. if you pass thread_id = None, we won't store the thread state but will still return the output

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -579,7 +1107,91 @@ async def create(
         webhook: Optional[str] = None,
         multitask_strategy: Optional[MultitaskStrategy] = None,
     ) -> Run:
-        """Create a background run."""
+        """Create a background run.
+
+        Args:
+            thread_id (optional,str): the thread ID to assign to the thread. Defaults to None.
+                If None a random UUID will be generated.
```
vbarda, line 1114: this is actually not the point in this case -- this is for creating stateless runs. e.g. if you pass thread_id = None, we won't store the thread state. the difference from `wait` is that it won't return any output, but can still be useful for side effects, i.e. agents modifying some external database

docs/docs/cloud/concepts/api.md
```diff
@@ -12,6 +12,10 @@ An assistant is a configured instance of a [`CompiledGraph`][compiledgraph]. It
 
 The LangGraph Cloud API provides several endpoints for creating and managing assistants. See the <a href="../reference/api/api_ref.html#tag/assistantscreate" target="_blank">API reference</a> for more details.
 
+#### Configuring Assistants
+
+You can save custom assistants from the same graph to set different default prompts, models, and other configurations without changing a line of code in your graph. This allows you the ability to quickly test out different configurations without having to rewrite your graph every time, and also give users the flexibility to select different configurations when sing your LangGraph application. See <a href="https://langchain-ai.github.io/langgraph/cloud/how-tos/cloud_examples/configuration_cloud/">this</a> how-to for information on how to configure a deployed graph.
```
vbarda, line 17: typo `sing` -> `using`

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -378,7 +672,29 @@ async def search(
         limit: int = 10,
         offset: int = 0,
     ) -> list[Thread]:
-        """Search for threads."""
+        """Search for threads.
+
```
vbarda, line 684: nit extra line

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -361,13 +626,42 @@ async def create(
         return await self.http.post("/threads", json=payload)
 
     async def update(self, thread_id: str, *, metadata: dict[str, Any]) -> Thread:
-        """Update a thread."""
+        """Update a thread.
+
+        Args:
+            thread_id: ID of thread to update.
+            metadata: Metadata to add to thread.
```
vbarda, line 641: nit: metadata to add/update

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -510,7 +990,54 @@ def stream(
         webhook: Optional[str] = None,
         multitask_strategy: Optional[MultitaskStrategy] = None,
     ) -> AsyncIterator[StreamPart]:
-        """Create a run and stream the results."""
+        """Create a run and stream the results.
+
+        Args:
+            thread_id: the thread ID to assign to the thread.
+                If None a random UUID will be generated.
```
vbarda, line 1007: also stateless here

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -744,7 +1540,41 @@ async def create(
...
+
+        Args:
+            assistant_id: The assistant ID or graph name to use for the cron job.
+                If using graph name, will default to first assistant created from that graph.
+            schedule: The cron schedule to execute this job on.
+            input: The input to the graph.
+            metadata: Metadata to assign to the cron job runs.
+            config: The configuration for the assistant.
+            interrupt_before: Nodes to interrupt immediately before they get executed.
+
```
vbarda, line 1569: extra lines -- let's remove

PR #1154, opened 2024-07-27
State Inject 'Any' types

hinthornw:


libs/langgraph/tests/test_prebuilt.py
```diff
@@ -498,7 +498,7 @@ def tool3(
         return foo
 
     def tool4(
-        some_val: int, msgs: Annotated[List[AnyMessage], InjectedState("messages")]
+        some_val: int = 5, *, msgs: Annotated[Any, InjectedState("messages")]
```
hinthornw, line 501: Pydantic infers this as `typing.Optional[Annotated[typing.Optional[Any], InjectedState("messages")]]` or somehting like that, which made our inference miss the annotation

PR #1159, opened 2024-07-29
fix: update the function for node colors

akshseh:
`NodeColors` does not exist anymore in[ Langchain_core ](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/runnables/graph.py) -> updated to `NodeStyles` and changed the param names. 

Comment:
- Additional point I noticed is that the graph outputs do not match the code. Let me know if it's intentional or you want me to fix it.

### DIFF
--- a/examples/visualization.ipynb
+++ b/examples/visualization.ipynb
@@ -268,7 +268,7 @@
    ],
    "source": [
     "from IPython.display import Image, display\n",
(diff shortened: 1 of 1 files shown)


nfcampos (approved):

nfcampos:
Thanks!

PR #1160, opened 2024-07-29
Fix issue when cond edge visited after multiple executions of Send

nfcampos:
- cond edge will run for each execution of Send, so target channels need to support multiple publishes

Fixes https://github.com/langchain-ai/langgraph/discussions/1157


nfcampos:
Fixes #1157

hwchase17 (approved):

PR #1185, opened 2024-08-01
checkpoint: switch thread_ts -> checkpoint_id, add checkpoint_ns, change serializer protocol

vbarda:


libs/checkpoint/README.md
```diff
@@ -10,16 +10,20 @@ Checkpoint is a snapshot of the graph state at a given point in time. Checkpoint
...
+Threads enable the checkpointing of multiple different runs, making them essential for multi-tenant chat applications and other scenarios where maintaining separate states is necessary. A thread is a unique ID assigned to a series of checkpoints saved by a checkpointer. When using a checkpointer, you must specify a `thread_id` and optionally `checkpoint_ns` / `checkpoint_id` when running the graph.
 
 - `thread_id` is simply the ID of a thread. This is always required
-- `thread_ts` can optionally be passed. This identifier refers to a specific checkpoint within a thread. This can be used to kick of a run of a graph from some point halfway through a thread.
+- `checkpoint_ns` can optionally be passed. This is a path to the subgraph the checkpoint originates from, separated by `|` character. Defaults to "" (root graph). Examples:
```
nfcampos, line 16: checkpoint_ns should never be passed manually

libs/checkpoint/README.md
```diff
@@ -10,16 +10,20 @@ Checkpoint is a snapshot of the graph state at a given point in time. Checkpoint
...
+  - for a parent graph node with a child subgraph node "child" that itself has a subgraph node "grandchild", `checkpoint_ns` for the grandchild subgraph will be `"child|grandchild"`
+- `checkpoint_id` can optionally be passed. This identifier refers to a specific checkpoint within a thread. This can be used to kick of a run of a graph from some point halfway through a thread.
 
 You must pass these when invoking the graph as part of the configurable part of the config, e.g.
 
 ```python
 {"configurable": {"thread_id": "1"}}  # valid config
-{"configurable": {"thread_id": "1", "thread_ts": "0c62ca34-ac19-445d-bbb0-5b4984975b2a"}}  # also valid config
+{"configurable": {"thread_id": "1", "checkpoint_id": "0c62ca34-ac19-445d-bbb0-5b4984975b2a"}}  # also valid config
+{"configurable": {"thread_id": "1", "checkpoint_ns": "child"}}  # also valid config
```
nfcampos, line 26: not valid, see above

libs/checkpoint/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -203,62 +203,76 @@ def get_tuple(self, config: RunnableConfig) -> Optional[CheckpointTuple]:
...
+                        checkpoint_ns,
+                        str(config["configurable"]["checkpoint_id"]),
                     ),
                 )
                 # deserialize the checkpoint and metadata
                 return CheckpointTuple(
                     config,
-                    self.serde.loads(value[3]),
-                    self.serde.loads(value[4]) if value[4] is not None else {},
+                    self.serde.loads(("json", checkpoint)),
```
nfcampos, line 259: what's the idea here? we should be storing the `type` string too

libs/checkpoint/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -354,23 +384,27 @@ def put(
...
+                "INSERT OR REPLACE INTO checkpoints (thread_id, checkpoint_ns, checkpoint_id, parent_checkpoint_id, checkpoint, metadata) VALUES (?, ?, ?, ?, ?, ?)",
                 (
                     str(config["configurable"]["thread_id"]),
+                    checkpoint_ns,
                     checkpoint["id"],
-                    config["configurable"].get("thread_ts"),
-                    self.serde.dumps(checkpoint),
-                    self.serde.dumps(metadata),
+                    config["configurable"].get("checkpoint_id"),
+                    self.serde.dumps(checkpoint)[1],
```
nfcampos, line 399: see comment above, the whole point is to store the type string

libs/checkpoint/langgraph/checkpoint/sqlite/aio.py
```diff
@@ -209,20 +207,22 @@ async def setup(self) -> None:
                 PRAGMA journal_mode=WAL;
                 CREATE TABLE IF NOT EXISTS checkpoints (
                     thread_id TEXT NOT NULL,
-                    thread_ts TEXT NOT NULL,
-                    parent_ts TEXT,
+                    checkpoint_ns TEXT NOT NULL,
```
nfcampos, line 210: for backwards compat let's make this default to empty string?

libs/checkpoint/langgraph/checkpoint/sqlite/utils.py
```diff
@@ -99,6 +68,9 @@ def search_where(
     if config is not None:
         wheres.append("thread_id = ?")
         param_values.append(config["configurable"]["thread_id"])
+        checkpoint_ns = config["configurable"].get("checkpoint_ns", "")
```
nfcampos, line 71: I think when you finish up nested checkpoints we'll want the ability to call `list` and get back checkpoints for all namespace ids

libs/checkpoint/langgraph/checkpoint/serde/base.py
```diff
@@ -10,8 +10,8 @@ class SerializerProtocol(Protocol):
     Valid implementations include the `pickle`, `json` and `orjson` modules.
     """
 
-    def dumps(self, obj: Any) -> bytes:
+    def dumps(self, obj: Any) -> tuple[str, bytes]:
```
nfcampos, line 13: if we want to minimize breaking changes we could do this change in a fully backwards compatible way I think, by making these new methods, and any old serializer with only old methods that is passed in we wrap it to implement the new methods delegating to old ones

libs/checkpoint/langgraph/checkpoint/memory.py
```diff
@@ -70,10 +71,11 @@ def get_tuple(self, config: RunnableConfig) -> Optional[CheckpointTuple]:
             Optional[CheckpointTuple]: The retrieved checkpoint tuple, or None if no matching checkpoint was found.
         """
         thread_id = config["configurable"]["thread_id"]
-        if ts := config["configurable"].get("thread_ts"):
-            if saved := self.storage[thread_id].get(ts):
-                checkpoint, metadata, parent_ts = saved
-                writes = self.writes[(thread_id, ts)]
+        checkpoint_ns = config["configurable"].get("checkpoint_ns", "")
+        if checkpoint_id := config["configurable"].get("checkpoint_id"):
```
nfcampos, line 75: we can make this change backwards compat by reading thread_ts if checkpoint_id not found, we probably should? Even though we are making breaking changes always good to minimize them as much as we can

libs/langgraph/langgraph/pregel/debug.py
```diff
@@ -97,7 +98,8 @@ def map_debug_task_results(
             continue
 
         metadata = config["metadata"].copy()
-        metadata.pop("thread_ts", None)
+        metadata.pop("checkpoint_id", None)
+        metadata.pop("checkpoint_ns", None)
```
nfcampos, line 102: this doesn't need to be removed, given it is a stable id

libs/checkpoint/langgraph/checkpoint/sqlite/__init__.py
vbarda: i was thinking of only using json serializer for sqlite, but it makes sense to make more generic!

libs/checkpoint/langgraph/checkpoint/sqlite/utils.py
vbarda: sounds good to me

libs/checkpoint/langgraph/checkpoint/memory.py
vbarda: sounds good to me

libs/checkpoint/langgraph/checkpoint/base/utils.py
```diff
@@ -0,0 +1,10 @@
+from typing import Optional
+
+from langchain_core.runnables import RunnableConfig
+
+
+def get_checkpoint_id(config: RunnableConfig) -> Optional[str]:
```
nfcampos, line 6: should we export this from the file where base class is defined, I think makes more sense for something a lot of subclasses will want to use?

vbarda: sure

libs/checkpoint/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -132,20 +133,22 @@ def setup(self) -> None:
             PRAGMA journal_mode=WAL;
             CREATE TABLE IF NOT EXISTS checkpoints (
                 thread_id TEXT NOT NULL,
-                checkpoint_ns TEXT NOT NULL,
+                checkpoint_ns TEXT NOT NULL DEFAULT '',
                 checkpoint_id TEXT NOT NULL,
                 parent_checkpoint_id TEXT,
+                type TEXT,
```
nfcampos, line 139: I think in this version without blobs table you'd need a checkpoint_type and metadata_type?

nfcampos: Or maybe better to serialize metadata always with `json.dumps`

libs/langgraph/langgraph/pregel/debug.py
```diff
@@ -72,7 +72,6 @@ def map_debug_tasks(
 
         metadata = config["metadata"].copy()
         metadata.pop("checkpoint_id", None)
-        metadata.pop("checkpoint_ns", None)
```
nfcampos: so you remove it in one function but not the other?

vbarda: discussed offline -- bug in core, will update when fixed!

libs/checkpoint/langgraph/checkpoint/sqlite/__init__.py
vbarda: updated w/ `dumps`!

PR #1187, opened 2024-08-01
langgraph[patch]: Use _schema instead of BaseModel.schema() for pydantic 2 migration

eyurtsev:
This PR introduces a module with some helper utilities for the pydantic 1 -> 2 migration.

They're meant to be used in the following way:

1) Use the utility code to get unit tests pass without requiring modification to the unit tests
2) (If desired) upgrade the unit tests to match pydantic 2 output
3) (If desired) stop using the utility code

Currently, this module contains a way to map `schema()` generated by pydantic 2 to (mostly) match the output from pydantic v1.

Mirrors this PR:  https://github.com/langchain-ai/langchain/pull/24930

### DIFF
--- a/libs/langgraph/tests/pydantic_utils.py
+++ b/libs/langgraph/tests/pydantic_utils.py
@@ -1,0 +1,94 @@
+"""Helper utilities for pydantic.
+
+This module includes helper utilities to ease the migration from pydantic v1 to v2.
+
+They're meant to be used in the following way:
+
+1) Use utility code to help (selected) unit tests pass without modifications
+2) Upgrade the unit tests to match pydantic 2
+3) Stop using the utility code
+"""
+
+from typing import Any
+
+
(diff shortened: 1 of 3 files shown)


libs/langgraph/tests/pydantic_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    if isinstance(schema, dict):
+        # If the schema has an allOf key with a single item that contains a $ref
+        if (
+            "allOf" in schema
+            and len(schema["allOf"]) == 1
+            and "$ref" in schema["allOf"][0]
+        ):
+            schema["$ref"] = schema["allOf"][0]["$ref"]
+            del schema["allOf"]
+            if "default" in schema and schema["default"] is None:
```
nfcampos, line 27: why delete default?

PR #1196, opened 2024-08-02
langgraph: remove deprecations and add new warnings

vbarda:


libs/langgraph/langgraph/_api/deprecation.py
```diff
@@ -1,38 +1,64 @@
...
 
 class LangGraphDeprecationWarning(DeprecationWarning):
     pass
 
 
 F = TypeVar("F", bound=Callable[..., Any])
+C = TypeVar("C", bound=Type[Any])
 
 
 def deprecated(
```
nfcampos, line 14: you've confirmed this decorator doesn't strip types?

vbarda: yup

PR #1204, opened 2024-08-02
Add interrupt info to graph repr

nfcampos:


libs/langgraph/langgraph/graph/graph.py
```diff
@@ -493,6 +493,11 @@ def add_edge(
 
         for key, n in self.builder.nodes.items():
             node = n.runnable
+            metadata = n.metadata or {}
+            if key in self.interrupt_before_nodes:
+                metadata["__interrupt"] = "before"
```
vbarda, line 498: what if it's set both before and after?

nfcampos: only show before, on purpose for now, as I don't think that's a very valid use case

PR #1228, opened 2024-08-05
checkpoint: stop using sqlite checkpointers as context managers, make memorysaver a context manager

vbarda:


libs/langgraph/tests/test_pregel.py
```diff
@@ -7516,21 +7506,20 @@ def side(state: State):
 
 @pytest.mark.repeat(10)
 @pytest.mark.parametrize(
-    "checkpointer_fct",
+    "checkpointer",
```
nfcampos, line 7509: this one was a factory so we could repeat the test 10 times

PR #1240, opened 2024-08-06
Add UntrackedValue to mark a state key as not checkpointable

nfcampos:


libs/langgraph/langgraph/channels/untracked_value.py
```diff
@@ -0,0 +1,62 @@
...
+                del empty.value
+            except AttributeError:
+                pass
+
+    def update(self, values: Sequence[Value]) -> bool:
+        if len(values) == 0:
+            return False
+        if len(values) != 1 and self.guard:
+            raise InvalidUpdateError(
+                "EphemeralValue can only receive one value per step."
```
hinthornw, line 52: Naming here

libs/langgraph/langgraph/channels/untracked_value.py
```diff
@@ -0,0 +1,62 @@
...
+from langgraph.channels.base import BaseChannel, Value
+from langgraph.errors import EmptyChannelError, InvalidUpdateError
+
+
+class UntrackedValue(Generic[Value], BaseChannel[Value, Value, Value]):
+    """Stores the last value received, never checkpointed."""
+
+    def __init__(self, typ: Type[Value], guard: bool = True) -> None:
+        self.typ = typ
+        self.guard = guard
```
hinthornw, line 16: Could we document what *guard* means in this context?

PR #1255, opened 2024-08-07
checkpoint-postgres: use correct metadata serialization

vbarda:


libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -202,6 +204,19 @@ def _dump_writes(
             for idx, (channel, value) in enumerate(writes)
         ]
 
+    def _load_metadata(self, metadata: dict[str, Any]) -> dict[str, Any]:
+        return self.jsonplus_serde.loads(self.jsonplus_serde.dumps(metadata))
```
nfcampos, line 208: what?

vbarda: discussed offline, needed for LC message deserialization

PR #1263, opened 2024-08-07
langgraph: more checkpointer tests

vbarda:


libs/langgraph/tests/test_prebuilt.py
```diff
@@ -76,30 +68,12 @@ def test_no_modifier(checkpointer: Optional[BaseCheckpointSaver]):
...
-                "agent": 3,
-            },
-            "versions_seen": {
-                "__input__": {},
-                "__start__": {"__start__": 1},
-                "agent": {"start:agent": 2},
-            },
-            "pending_sends": [],
-            "current_tasks": {},
+        assert saved.checkpoint["channel_values"] == {
```
nfcampos, line 71: why does this need to change?

vbarda: i removed because of the different version types/formats across different checkpointers, but i can also make something like `AnyVersion` that will just ignore those fields / customize memory saver for tests to return strings. wdyt?

nfcampos: yea I think that's better

nfcampos (approved):

nfcampos:
I think there's still a bunch of tests where we test a single checkpointer. The end goal here should be to remove the imports of MemorySaver and SqliteSaver from test_pregel and test_pregel_async

PR #1270, opened 2024-08-08
langgraph, checkpoint-postgres: propagate new versions in update_state

vbarda:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -629,6 +634,7 @@ async def aupdate_state(
         # get last checkpoint
         saved = await self.checkpointer.aget_tuple(config)
         checkpoint = copy_checkpoint(saved.checkpoint) if saved else empty_checkpoint()
+        checkpoint_previous_versions = checkpoint["channel_versions"].copy()
```
nfcampos, line 637: Why is this being copied?

vbarda: cause it's mutated below, but i updated to avoid copying altogether

nfcampos (approved):

PR #1281, opened 2024-08-08
Fix message duplication in GraphState messages

hamidriasat:
### Context:
In the `GraphState` class, the `messages` variable is of type list, annotated with `add_messages`. This ensures that new messages are appended to the list without needing to resend old messages, as they are already present.

### Problem:
In the `generate` and `code_check` methods, the existing list was being extracted and, after appending a new message, sent back to update the `messages` variable. This approach caused duplication since the old messages were already present in the list. The correct behavior should only send back the new message to be appended.

### Solution:
This pull request fixes the issue by ensuring that only the new message is sent from the `generate` and `code_check` methods, preventing duplication.

###  Additional Notes:
- I verified the issue through debugging the GraphState variables.
- The rest of the script remains unchanged.

### DIFF
--- a/examples/code_assistant/langgraph_code_assistant_mistral.ipynb
+++ b/examples/code_assistant/langgraph_code_assistant_mistral.ipynb
@@ -33,7 +33,9 @@
    "id": "e501686f-323f-4b87-8f9c-8ba89133078b",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


hamidriasat:
@vbarda Kindly review this pull request.

vbarda:
Thank you! This wouldn't actually be an issue since `add_messages` deduplicates messages based on the ID

PR #1282, opened 2024-08-08
add task_id to node logs

isahers1:


libs/langgraph/langgraph/pregel/debug.py
```diff
@@ -90,9 +90,10 @@ def map_debug_task_results(
     step: int,
     tasks: list[PregelExecutableTask],
     stream_channels_list: Sequence[str],
+    step_start_time: datetime,
 ) -> Iterator[DebugOutputTaskResult]:
-    ts = datetime.now(timezone.utc).isoformat()
-    for name, _, _, writes, config, _, _, _ in tasks:
+    ts = datetime.now(timezone.utc)
+    for name, _, _, writes, config, _, _, task_id in tasks:
```
vbarda, line 96: should we just use attributes directly since it's a namedtuple? cc @nfcampos

libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -241,7 +241,9 @@ def _output_writes(
                 self.stream.extend(
                     (self.config["configurable"].get("checkpoint_ns", ""), "debug", v)
                     for v in map_debug_task_results(
-                        self.step, [(task, writes)], self.stream_keys
+                        self.step,
+                        [(task, writes)],
+                        self.stream_keys,
```
nfcampos, line 246: undo this change

nfcampos (approved):

PR #1289, opened 2024-08-09
docs: correct mistakes in customer-support.ipynb

CJack812:
In examples/customer-support/customer-support.ipynb, I found two same mistakes in 'Part3' and 'Part4'.
In the call method of the Assistant class, the lines `messages = state["messages"] + [("user", "Respond with a real output.")] `and `state = {**state, "messages": messages}` are repeated twice. I believe this is a mistake and could cause confusion for beginners. I have modified this part of the code.

original code:

```python
class Assistant:
    def __init__(self, runnable: Runnable):
        self.runnable = runnable

    def __call__(self, state: State, config: RunnableConfig):
        while True:
            result = self.runnable.invoke(state)

            if not result.tool_calls and (
                not result.content
                or isinstance(result.content, list)
                and not result.content[0].get("text")
            ):
                messages = state["messages"] + [("user", "Respond with a real output.")]
                state = {**state, "messages": messages}
                messages = state["messages"] + [("user", "Respond with a real output.")]
                state = {**state, "messages": messages}
            else:
                break
        return {"messages": result}
```

modified code:

```python
class Assistant:
    def __init__(self, runnable: Runnable):
        self.runnable = runnable

    def __call__(self, state: State, config: RunnableConfig):
        while True:
            result = self.runnable.invoke(state)

            if not result.tool_calls and (
                not result.content
                or isinstance(result.content, list)
                and not result.content[0].get("text")
            ):
                messages = state["messages"] + [("user", "Respond with a real output.")]
                state = {**state, "messages": messages}
            else:
                break
        return {"messages": result}
```


vbarda:
Thank you! this is already fixed in the current version of documentation

Issue #1314, opened 2024-08-12
DOC: filter_message code is wrong


streamnsight:
### Issue with current documentation:

simple issue, but worth pointing out

the code here:
https://langchain-ai.github.io/langgraph/how-tos/memory/manage-conversation-history/#filtering-messages

```def filter_messages(messages: list):
    # This is very simple helper function which only ever uses the last two messages
    return messages[-1:]

```
actually returns the last message, not the last 2.
last 2 would be `messages[-2:]`

vbarda:
this actually needs to be `messages[-1:]`

streamnsight:
@vbarda 

messages[-1:] gives the last message
last 2 is messages[-2:]
```
>>> a = [1,2,3,4,5]
>>> a[-2:]
[4, 5]
>>> a[-1:]
[5]
```

so if the function is to return last 2 as described, then is is messages[-2:]

vbarda:
the description was wrong

vbarda:
see https://github.com/langchain-ai/langgraph/pull/1438

gbaian10:
I previously thought that this tutorial example was meant to include only the first two sentences of the historical dialogue record, not all the dialogue, until I saw that the final run included this:

```py
# This will now not remember the previous messages
# (because we set `messages[-1:]` in the filter messages argument)
input_message = HumanMessage(content="what's my name?")
for event in app.stream({"messages": [input_message]}, config, stream_mode="values"):
    event["messages"][-1].pretty_print()
```

In that case, indeed, just updating the comments within the filter is enough.

shaikhn1750:
how to use 

```
def filter_messages(messages: list):
    # This is very simple helper function which only ever uses the last message
    return messages[-1:]


# Define the function that calls the model
def call_model(state: MessagesState):
    messages = filter_messages(state["messages"])
    response = bound_model.invoke(messages)
    # We return a list, because this will get added to the existing list
    return {"messages": response}
```

in multi agent supervisor scenario?

PR #1315, opened 2024-08-12
docs: fix filter_messages error in manage-conversation-history.ipynb

gbaian10:
fix: #1314

### DIFF
--- a/examples/memory/manage-conversation-history.ipynb
+++ b/examples/memory/manage-conversation-history.ipynb
@@ -269,7 +269,7 @@
     "\n",
     "def filter_messages(messages: list):\n",
     "    # This is very simple helper function which only ever uses the last two messages\n",
-    "    return messages[-1:]\n",
+    "    return messages[-2:]\n",
     "\n",
     "\n",
     "# Define the function that calls the model\n",


vbarda:
by the way this was actually incorrect -- we actually needed to update only the comment :) https://github.com/langchain-ai/langgraph/pull/1438

gbaian10:
I forgot to verify the purpose of the subsequent runs and blindly trusted the comments.
It turned out that the program was correct, but the comments were wrong.

Subsequent runs:
```py
# This will now not remember the previous messages
# (because we set `messages[-1:]` in the filter messages argument)
input_message = HumanMessage(content="what's my name?")
for event in app.stream({"messages": [input_message]}, config, stream_mode="values"):
    event["messages"][-1].pretty_print()
```

Sorry for causing the mistake. 🥹

vbarda:
> I forgot to verify the purpose of the subsequent runs and blindly trusted the comments. It turned out that the program was correct, but the comments were wrong.
> 
> Subsequent runs:
> 
> ```python
> # This will now not remember the previous messages
> # (because we set `messages[-1:]` in the filter messages argument)
> input_message = HumanMessage(content="what's my name?")
> for event in app.stream({"messages": [input_message]}, config, stream_mode="values"):
>     event["messages"][-1].pretty_print()
> ```
> 
> Sorry for causing the mistake. 🥹

@gbaian10 no worries! i just wanted to flag more so you know which behavior should be expected :)

PR #1317, opened 2024-08-12
updated-codes

Swastik-Swarup-Dash:
DOC: filter_message code is wrong #1314 

    # This is very simple helper function which only ever uses the last two messages
    return messages[-1:]

       # after updating
    return messages[-2:]


vbarda:
Thank you for your contribution -- this was fixed in #1315 already!

Swastik-Swarup-Dash:
Thats cool loved to work on it

PR #1328, opened 2024-08-13
docs: fix persistence_postgres.ipynb

gbaian10:
1. According to the [make start-postgres](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/Makefile#L8) command in the context, execute docker compose, and the configured port is [5441](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/tests/compose-postgres.yml#L5).

1. remove unuse import


vbarda:
thank you!

Issue #1331, opened 2024-08-13
The graph incorrectly obtained the input schema


gbaian10:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
from langchain_core.runnables import RunnableConfig
from langgraph.graph import MessagesState, StateGraph


def foo(state: MessagesState) -> MessagesState:
    print(f"foo: {state}")
    return {"messages": "foo"}


def error1(state, config: RunnableConfig) -> MessagesState:
    # It read the first type_hint,
    # thus incorrectly obtaining the type_hint of the `config` variable.
    print(f"error1: {state}")
    return {"messages": "this is config"}


def error2(state, config) -> MessagesState:
    # It actually obtained the type_hint of the `return`,
    # which just happened to be the same as the `state`.
    print(f"error2: {state}")
    return {"messages": "this is return"}


graph = StateGraph(MessagesState)
graph.add_node(foo)
graph.add_node(error1)
graph.add_node(error2)

graph.set_entry_point("foo")
graph.add_edge("foo", "error1")
graph.add_edge("error1", "error2")
graph.set_finish_point("error2")

graph = graph.compile()
graph.invoke({"messages": "Hello"})
```

### Error Message and Stack Trace (if applicable)

```shell
foo: {'messages': [HumanMessage(content='Hello', id='d00c79c6-3c77-4a1b-8e68-806334f1dd36')]}

error1: {'tags': None, 'metadata': None, 'callbacks': None, 'run_name': None, 'max_concurrency': None, 'recursion_limit': None, 'configurable': None, 'run_id': None}

error2: {'messages': [HumanMessage(content='Hello', id='d00c79c6-3c77-4a1b-8e68-806334f1dd36'), HumanMessage(content='foo', id='6ac49f5c-e8be-4d93-a5e0-bcce19cf9ca3'), HumanMessage(content='this is config', id='91bc0d28-3095-4d27-a667-0bb0aab09123')]}
```

### Description

[...]

PR #1332, opened 2024-08-13
langgraph: fix add_node input schema error

gbaian10:
fix: #1331

I used `inspect.signature` to read the name of the first parameters and determine whether this variable has a type_hint.

And used `StopIteration` to prevent situations where there are no parameters.
Although in this situation, executing invoke will still result in an error.

If want to catch errors before running the graph, maybe you can check that the number of parameters is not zero?

### DIFF
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -1,3 +1,4 @@
+import inspect
 import logging
 import typing
 import warnings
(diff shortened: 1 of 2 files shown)


gbaian10:
Regarding first_parameter_name = next(iter(inspect.signature(action).parameters.keys()))

Also can use list() and then check the length before accessing [0].

I'm not sure which one is better.

nfcampos (approved):
Thanks, do you want to add a test case?

gbaian10:
> Thanks, do you want to add a test case?

I plan to add tests using this logic. Do you have any better suggestions?

```py
from typing import TypedDict

# from libs.langgraph.langgraph.graph.state import StateGraph


class InputState(TypedDict):
    question: str


class OutputState(TypedDict):
    input_state: dict


def foo(state: InputState) -> OutputState:
    return {"input_state": state}


def bar(state, config: RunnableConfig) -> OutputState:
    # The expected `state` result is `InputState`, but received `RunnableConfig`.
    return {"input_state": state}


def baz(state, config) -> OutputState:
    # The expected `state` result is `InputState`, but received `OutputState`.
    return {"input_state": state}


graph = StateGraph(input=InputState, output=OutputState)
graph.add_node(foo)
graph.add_node(bar)
graph.add_node(baz)

graph.set_entry_point(foo.__name__)
graph.add_edge(foo.__name__, bar.__name__)
graph.add_edge(bar.__name__, baz.__name__)
graph.set_finish_point(baz.__name__)

graph = graph.compile()

for v in graph.stream({"question": "foo"}, stream_mode="values"):
    print(v)
    if "input_state" in v:
        assert v["input_state"] == {"question": "foo"}
```

vbarda:
Thank you @gbaian10  -- going to merge!

PR #1334, opened 2024-08-13
[docs]: cloud hil review tool calls

isahers1:


docs/docs/cloud/how-tos/human_in_the_loop_review_tool_calls.md
```diff
@@ -0,0 +1,575 @@
...
+=== "Python"
+
+    ```python
+    input = {"messages": [{"role": "user", "content": "what's the weather in sf?"}]}
+
+    async for chunk in client.runs.stream(
+        thread["thread_id"],
+        "agent",
+        input=input,
+        stream_mode="values",
```
vbarda, line 233: i think we need to pass  an interrupt here

vbarda: ah it's set on the graph. i think it's worth mentionig up top

isahers1: graph was defined as such: `builder.compile(checkpointer=memory, interrupt_before=["human_review_node"])` so I think we are OK? I ran the code and it worked, but idk

vbarda: yup!, i think it's just worth mentioning that interrupt is defined in the example that we link to, so people understand where it's coming from

vbarda (approved):

docs/docs/cloud/how-tos/human_in_the_loop_review_tool_calls.md
isahers1: ah yeah sorry missed your second comment - will add a mention rn

PR #1347, opened 2024-08-14
langgraph: allow updating subgraph state on interrupts

vbarda:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -676,30 +677,60 @@ def update_state(
...
+        if not checkpointer:
             raise ValueError("No checkpointer set")
 
+        has_nested_interrupts = _has_nested_interrupts(self)
+        if (
+            has_nested_interrupts
+            and as_node is not None
+            and not isinstance(as_node, list)
+        ):
+            raise ValueError(
```
nfcampos, line 698: what is this for? we should accept string as having the same meaning as a list of length 1

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -709,11 +740,23 @@ def update_state(
                     **saved.config["configurable"],
                 }
             }
+
+        # if we're writing a subgraph update for a key that's not in the parent graph channels,
```
nfcampos, line 744: why?

vbarda:
Closed in favor of https://github.com/langchain-ai/langgraph/pull/1539

PR #1350, opened 2024-08-14
[docs]: copying threads

isahers1:


docs/docs/cloud/how-tos/copy_threads.md
```diff
@@ -0,0 +1,165 @@
...
+    ```bash
+    curl --request POST \
+      --url <DEPLOYMENT_URL>/threads \
+      --header 'Content-Type: application/json' \
+      --data '{
+        "metadata": {}
+      }'
+    ```
+
+### Thread creation
```
vbarda, line 43: Don't think we need this section at all, we can just add a note to the section below saying "This assumes that a thread you'd like to copy already exists"

docs/docs/cloud/how-tos/copy_threads.md
```diff
@@ -0,0 +1,165 @@
...
+    ```python
+    def remove_thread_id(d):
+      if 'metadata' in d and 'thread_id' in d['metadata']:
+          del d['metadata']['thread_id']
+      return d
+
+    original_thread_history = list(map(remove_thread_id,await client.threads.get_history(<THREAD_ID>)))
+    copied_thread_history = list(map(remove_thread_id,await client.threads.get_history(copied_thread['thread_id'])))
+
+    if original_thread_history == copied_thread_history:
```
vbarda, line 116: (nit) maybe just do assert original == copied? hopefully we will never see "histories are different" :)

docs/docs/cloud/how-tos/copy_threads.md
```diff
@@ -0,0 +1,165 @@
...
+        return d;
+    }
+
+    // Assuming `client.threads.getHistory(threadId)` is an async function that returns a list of dicts
+    async function compareThreadHistories(threadId, copiedThreadId) {
+        const originalThreadHistory = (await client.threads.getHistory(threadId)).map(removeThreadId);
+        const copiedThreadHistory = (await client.threads.getHistory(copiedThreadId)).map(removeThreadId);
+
+        // Compare the two histories
+        if (JSON.stringify(originalThreadHistory) === JSON.stringify(copiedThreadHistory)) {
```
vbarda, line 138: same -- maybe just console.assert?

docs/docs/cloud/how-tos/copy_threads.md
```diff
@@ -0,0 +1,165 @@
+# Copying Threads
+
+You may wish to copy an existing thread in order to keep the existing history and experiment with different runs in the future.
```
vbarda, line 3: Maybe something like "You may wish to copy (i.e. "fork") an existing thread in order to keep the existing thread's history and create independent runs that do not affect the original thread. This guide shows how you can do that."

docs/docs/cloud/how-tos/copy_threads.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Setup
+
+To copy a thread, we must first set up our client and also create a thread we wish to copy.
```
vbarda, line 7: don't think we need to actually create one, we can just say "we assume you already have a thread to copy. if not, see this guide on how to make one"

docs/docs/cloud/how-tos/copy_threads.md
isahers1: yeah was going to do, but I did this to ensure exact compatibility with CURL command. If you think important to use assert I can do that and just leave the CURL command a little different.

vbarda (approved):

PR #1354, opened 2024-08-15
Enable raising Interrupt from inside a node, add list of current interrupts to get_state

nfcampos:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -550,7 +553,7 @@ def update_state(
         saved = self.checkpointer.get_tuple(config)
         checkpoint = copy_checkpoint(saved.checkpoint) if saved else empty_checkpoint()
         checkpoint_previous_versions = (
-            saved.checkpoint["channel_versions"] if saved else {}
+            saved.checkpoint["channel_versions"].copy() if saved else {}
```
nfcampos, line 556: important bug fix

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -576,7 +579,7 @@ def update_state(
                 create_checkpoint(checkpoint, None, step),
                 {
                     "source": "update",
-                    "step": step,
+                    "step": step + 1,
```
nfcampos, line 582: also bug fix, less important

PR #1401, opened 2024-08-20
sdk-py: add sync client

vbarda:


libs/sdk-py/langgraph_sdk/client/__init__.py
```diff
@@ -0,0 +1,3 @@
+from langgraph_sdk.client.async_client import get_client
+
+__all__ = ["get_client"]
```
nfcampos, line 3: breaking change I think, some people would have been importing AsyncClient from here

PR #1421, opened 2024-08-22
checkpoint-*: add pending writes to .list()

vbarda:


libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -318,7 +318,9 @@ def list(
         ORDER BY checkpoint_id DESC"""
         if limit:
             query += f" LIMIT {limit}"
-        with self.cursor(transaction=False) as cur:
+        with self.cursor(transaction=False) as cur, self.cursor(
```
nfcampos, line 321: two cursors?

libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/aio.py
```diff
@@ -346,6 +346,14 @@ async def alist(
                 checkpoint,
                 metadata,
             ) in cursor:
+                writes_cur = await self.conn.execute(
```
nfcampos, line 349: this cursor is never closed it seems

vbarda: let me adjust!

libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
vbarda: not sure what's a better way to do it other than just put everything into a single query and join aggregated writes (not sure if doable in sqlite)

nfcampos:
Superseded by #1436

PR #1434, opened 2024-08-22
Update langgraph

Swastik-Swarup-Dash:
Updated the Langraph definition for easier undersatnding

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


vbarda:
@Swastik-Swarup-Dash thank you for your contribution! we will keep the paragraph as is at this time, but we welcome other improvement suggestions / docs PRs!

PR #1436, opened 2024-08-22
Fix semantics of put_writes/list

nfcampos:
- put_writes(error) should not prevent saving future successful if task is retried successfully
- put_writes(writes) should be a no-op if non-error writes already exist for that task (this prevents tasks executed more than once from modifying writes previously saved / acted on)
- checkpoints should not include channel default values (ie those without a version)
- list() should fetch and return writes for each checkpoint

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -156,8 +156,6 @@ class BasePostgresSaver(BaseCheckpointSaver):
         if not versions:
             return []
 
-        print("_dump_blobs", versions, values)
-
         return [
             (
                 thread_id,
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -6,6 +6,7 @@ from langchain_core.runnables import RunnableConfig
 from psycopg.types.json import Jsonb
 
 from langgraph.checkpoint.base import (
+    WRITES_IDX_MAP,
     BaseCheckpointSaver,
     Checkpoint,
     EmptyChannelError,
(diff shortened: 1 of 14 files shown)


libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -165,6 +156,8 @@ def _dump_blobs(
         if not versions:
             return []
 
+        print("_dump_blobs", versions, values)
```
vbarda, line 159: print statement

libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/aio.py
```diff
@@ -329,14 +330,14 @@ async def alist(
...
-        where, param_values = search_where(config, filter, before)
+        where, params = search_where(config, filter, before)
         query = f"""SELECT thread_id, checkpoint_ns, checkpoint_id, parent_checkpoint_id, type, checkpoint, metadata
         FROM checkpoints
         {where}
         ORDER BY checkpoint_id DESC"""
         if limit:
             query += f" LIMIT {limit}"
-        async with self.conn.execute(query, param_values) as cursor:
+        async with self.conn.execute(query, params) as cur, self.conn.cursor() as wcur:
```
vbarda, line 340: i think this needs to be `aclosing(self.conn.cursor()) as wcur`

libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -25,6 +25,7 @@
     ChannelProtocol,
     SendProtocol,
 )
+from langgraph.constants import ERROR
```
vbarda, line 28: shouldn't import from langgraph here

libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/aio.py
nfcampos: No, aiosqliue cursor actually is a context manager

vbarda: ah cool

PR #1451, opened 2024-08-23
Allow passing `ToolNode` as `tools` in `create_react_agent`

waleko:
In the current version of `create_react_agent`, execution cannot be halted when a tool exception occurs. Instead, exceptions are intercepted and passed to the agent as text in a `ToolMessage`. This differs from the `AgentExecutor`, which does not handle exceptions and leaves tool error handling to the tools themselves.

To address this limitation, `create_react_agent` signature is adjusted to take `ToolNode` in the `tools` argument, which allows modifying how tool errors are handled


hinthornw:
@waleko I'd prefer if, instead of adding more arguments to the function, we permit passing in a `ToolNode` directly via the `tools` argument. Then you can configure that however you'd like

vbarda:
Thank you for your contribution @waleko !

PR #1452, opened 2024-08-23
checkpoint-postgres: allow passing pool

vbarda:


libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
```diff
@@ -66,21 +72,24 @@ def setup(self) -> None:
...
-                    version = -1
-                for v, migration in zip(
-                    range(version + 1, len(self.MIGRATIONS)),
-                    self.MIGRATIONS[version + 1 :],
-                ):
-                    cur.execute(migration)
-                    cur.execute(f"INSERT INTO checkpoint_migrations (v) VALUES ({v})")
-            if self.pipe:
-                self.pipe.sync()
+            with self._connection() as conn:
```
nfcampos, line 75: let's use the same with statement

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -64,24 +70,25 @@ async def setup(self) -> None:
...
-                    range(version + 1, len(self.MIGRATIONS)),
-                    self.MIGRATIONS[version + 1 :],
-                ):
-                    await cur.execute(migration)
-                    await cur.execute(
-                        f"INSERT INTO checkpoint_migrations (v) VALUES ({v})"
-                    )
-            if self.pipe:
-                await self.pipe.sync()
+            async with self._connection() as conn:
```
nfcampos, line 73: use same with statement

libs/langgraph/tests/test_pregel.py
```diff
@@ -590,7 +590,7 @@ def test_invoke_two_processes_in_out(mocker: MockerFixture) -> None:
 
 @pytest.mark.parametrize(
```
nfcampos, line 591: should we move this to conftest ?

nfcampos: ALL_CHECKPOINTERS_SYNC
ALL_CHECKPOINTERS_ASYNC

vbarda: yah sure, good idea

libs/checkpoint-postgres/pyproject.toml
```diff
@@ -13,6 +13,7 @@ python = "^3.9.0,<4.0"
 langgraph-checkpoint = "^1.0.1"
 orjson = ">=3.10.1"
 psycopg = {extras = ["binary"], version = ">=3.1.19"}
+psycopg-pool = "^3.2.2"
```
nfcampos, line 16: let's relax the version constraint here

libs/checkpoint-postgres/pyproject.toml
```diff
@@ -13,6 +13,7 @@ python = "^3.9.0,<4.0"
 langgraph-checkpoint = "^1.0.1"
 orjson = ">=3.10.1"
 psycopg = {extras = ["binary"], version = ">=3.1.19"}
```
nfcampos, line 15: I don't think this should be installing the binary extra out of the box, can conflict with user's dependency on one of the other psycopg variants

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -288,27 +291,38 @@ async def aput_writes(
                 ),
             )
 
+    @asynccontextmanager
+    async def _connection(self) -> AsyncIterator[AsyncConnection]:
```
nfcampos, line 295: Why do we want one more private method that's used only in one place

vbarda: moved outside the class but lmk if that's not what you wanted!

nfcampos (approved):

PR #1455, opened 2024-08-23
lib: Context values never stored in checkpoints

nfcampos:
- Convert Context to a ManagedValue
- Add shim for old Context constructor
- Add `runtime` flag for managed values, which, prior to serialization, replaces the value with a placeholder, and replaces it back with the actual value on resuming from checkpoint


libs/langgraph/tests/test_pregel.py
```diff
@@ -4117,11 +4118,19 @@ def should_continue(data: AgentState) -> str:
...
         else:
             return END
 
-    def tools_node(tool_call: ToolCall, config: RunnableConfig) -> AgentState:
+    class ToolInput(TypedDict):
+        call: ToolCall
+        my_session: httpx.Client
+
+    def tools_node(input: ToolInput, config: RunnableConfig) -> AgentState:
+        tool_call = input["call"]
```
hinthornw, line 4133: Also assert the `my_session` looks correct?

libs/langgraph/langgraph/managed/base.py
```diff
@@ -103,3 +106,34 @@ def is_writable_managed_value(value: Any) -> TypeGuard[Type[WritableManagedValue
 
 ChannelKeyPlaceholder = object()
 ChannelTypePlaceholder = object()
+
+
+class ManagedValueMapping(dict[str, ManagedValue]):
+    def replace_runtime_values(self, step: int, values: Union[dict[str, Any], Any]):
+        print("replace_runtime_values", values)
```
hinthornw, line 113: nit (and below)

libs/langgraph/langgraph/managed/base.py
```diff
@@ -16,11 +16,16 @@
...
 
+from langgraph.constants import RUNTIME_PLACEHOLDER
+
 V = TypeVar("V")
 U = TypeVar("U")
 
 
 class ManagedValue(ABC, Generic[V]):
+    runtime: bool = False
+    """Whether the managed value is always created at runtime, ie. never stored."""
```
hinthornw, line 27:
```suggestion
    """Whether the managed value is always created at runtime, i.e. never stored."""
```

libs/langgraph/langgraph/managed/context.py
```diff
@@ -0,0 +1,85 @@
...
+        if ctx is None and actx is None:
+            raise ValueError("Must provide either sync or async context manager.")
+        return ConfiguredManagedValue(Context, {"ctx": ctx, "actx": actx})
+
+    @classmethod
+    @contextmanager
+    def enter(cls, config: RunnableConfig, **kwargs: Any) -> Iterator[Self]:
+        with super().enter(config, **kwargs) as self:
+            if self.ctx is None:
+                raise ValueError("Cannot enter sync context manager.")
```
hinthornw, line 39: Out of context, this could be confusing. Perhaps: 

```suggestion
                raise ValueError("Synchronous context manager not found. Please initialize Context value with a sync context manager, or invoke your graph asynchronously.")
```

or something less verbose but more contextual

libs/langgraph/langgraph/managed/context.py
```diff
@@ -0,0 +1,85 @@
...
+                    self.value = v
+                    yield self
+            else:
+                with ctx as v:
+                    self.value = v
+                    yield self
+
+    def __init__(
+        self,
+        config: RunnableConfig,
```
hinthornw, line 76: oh interesting so we ignore the config? I guess it's not really that useful

nfcampos: No, not for this one, it's part of the base class signature

libs/langgraph/tests/test_pregel.py
nfcampos: Yep

Issue #1469, opened 2024-08-26
DOC: Missing import in `langgraph/examples/docs/quickstart.ipynb`


danwild:
### Issue with current documentation:

I'm new here, so I started with: https://github.com/danwild/langgraph/blob/main/examples/docs/quickstart.ipynb

The 3rd python cell fails to import `START` from `langgraph.graph` before it's used, which breaks notebook execution.

```
NameError                                 Traceback (most recent call last)
Cell In[6], [line 14](vscode-notebook-cell:?execution_count=6&line=14)
     [10](vscode-notebook-cell:?execution_count=6&line=10) graph.add_node("oracle", model)
     [11](vscode-notebook-cell:?execution_count=6&line=11) graph.add_edge("oracle", END)
---> [14](vscode-notebook-cell:?execution_count=6&line=14) graph.add_edge(START, "oracle")
     [16](vscode-notebook-cell:?execution_count=6&line=16) runnable = graph.compile()

NameError: name 'START' is not defined
```

<img width="668" alt="Screenshot 2024-08-26 at 12 30 49 pm" src="https://github.com/user-attachments/assets/9a696144-f29c-40d0-ade3-bd48dfc62dce">

PR #1470, opened 2024-08-26
1469: add missing langgraph.graph import

danwild:
Hi,

New here 👋 

Addresses [1469](https://github.com/langchain-ai/langgraph/issues/1469)

Hopefully this one's trivial ;)

Before:

```python
from langgraph.graph import END, MessageGraph
```

After:

```python
from langgraph.graph import START, END, MessageGraph
```


vbarda:
Thank you for reporting and for the contribution -- it's already fixed in https://github.com/langchain-ai/langgraph/pull/1472 :)

PR #1479, opened 2024-08-26
Override with_config to store config in Pregel instance

nfcampos:
- This enables eg customizing callbacks/metadata in langgraph cloud deployments

### DIFF
--- a/libs/langgraph/langgraph/pregel/__init__.py
+++ b/libs/langgraph/langgraph/pregel/__init__.py
@@ -38,6 +38,7 @@ from langchain_core.runnables.config import (
     ensure_config,
     get_async_callback_manager_for_config,
     get_callback_manager_for_config,
+    merge_configs,
     patch_config,
 )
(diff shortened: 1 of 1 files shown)


hinthornw (approved):

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -219,11 +220,20 @@ class Pregel(
...
+    config: Optional[RunnableConfig] = None
+
     name: str = "LangGraph"
 
     class Config:
         arbitrary_types_allowed = True
 
+    def with_config(
+        self, config: RunnableConfig | None = None, **kwargs: Any
+    ) -> Runnable[Dict[All, Any] | Any, Dict[All, Any] | Any]:
```
hinthornw, line 232: Would it make sense to type as CompiledGraph type?

naaive:
create_react_agent() , them call with_config, it not works

PR #1485, opened 2024-08-26
[docs]: remove agent executor 

isahers1:
also removed a bunch of empty python cells to make docs look slightly nicer


nfcampos (commented):
Do we need to remove import of `create_agent`?

isahers1:
> Do we need to remove import of `create_agent`?

Not 100% sure what this means, I don't think there are any imports of `create_agent` but I may be missing something

PR #1501, opened 2024-08-27
[docs]: recursion limit info

isahers1:


docs/docs/how-tos/index.md
```diff
@@ -15,7 +15,7 @@ These how-to guides show how to achieve that controllability.
 - [How to create subgraphs](subgraph.ipynb)
 - [How to create branches for parallel execution](branching.ipynb)
 - [How to create map-reduce branches for parallel execution](map-reduce.ipynb)
-
+- [How to control graph recurison limit](recursion-limit.ipynb)
```
vbarda, line 18: typo recurison

examples/recursion-limit.ipynb
```diff
@@ -0,0 +1,172 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# How to control graph recursion limit\n",
+    "\n",
+    "You can set the graph recurison limit when invoking or streaming the graph. Let's see an example of this in a siple graph with parallel branches to better understand exactly how the recursion limit works.\n",
```
vbarda, line 9: i think it's worth explaining briefly what it does here and linking to the concepts docs
also, typo "siple"

docs/docs/concepts/low_level.md
```diff
@@ -401,7 +401,17 @@ def node_a(state, config):
     ...
 ```
 
-See [this guide](../how-tos/configuration.ipynb) for a full breakdown on configuration
+See [this guide](../how-tos/configuration.ipynb) for a full breakdown on configuration.
+
+### Recursion Limit Configuration
+
+The recursion limit sets the amount of super-steps (read more about what a super-step is [here](#graphs)) your graph can execute before failing. By default this value is set to 25. The recursion limit can be set on any graph, and is passed in the outer `config` dictionary, not inside of the `configurable` key like user defined configurations as in the example below:
```
vbarda, line 408: (nit) let's either just link on `[super-steps](...url)` or shorten to something like (read more `[here[(...)`)

consider something like this


The recursion limit sets the maximum number of [super-steps](...url) the graph can execute during a single execution. Once the limit is reached, LangGraph will raise `GraphRecursionError`. By default this value is set to 25 steps. The recursion limit can be set on any graph at runtime, and is passed to `.invoke` / `.stream` via the `config` dictionary. Importantly, `recursion_limit` is a standalone config key and should **not** be passed inside the `configurable` key as all other user-defined configuration. See the example below:

examples/recursion-limit.ipynb
```diff
@@ -0,0 +1,172 @@
...
+    "from IPython.display import Image, display\n",
+    "\n",
+    "display(Image(graph.get_graph().draw_mermaid_png()))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As we can see, our graph will execute nodes `b` and `c` in parallel (i.e. in a single super-step), which means that if we run this graph it should take exactly 3 steps. We can set the recursion limit to 3 first to check that it fails (the recursion limit is inclusive, so if the limit is 3 the graph will fail when it reaches step 3) as expected: "
```
vbarda, line 82: let's say "raise error" instead of "fail"

examples/recursion-limit.ipynb
```diff
@@ -0,0 +1,172 @@
...
+    "    print(\"Recursion Error\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Perfect, just as we expected the graph runs succesfully in this case. \n",
+    "\n",
+    "Setting the correct recursion limit can be important for minimizing unneccesary API costs, and also for avoiding undesired node loops. "
```
vbarda, line 147: how about "Setting recursion limit is important for avoiding graph runs with potentially infinite loops and thus minimizing unnecessary costs."

examples/recursion-limit.ipynb
```diff
@@ -0,0 +1,172 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# How to control graph recursion limit\n",
+    "\n",
+    "You can set the graph recursion limit when invoking or streaming the graph. The recursion limit sets the amount of supersteps that the graph is allowed to execute before it fails. Read more about the concept of recursion limits [here](https://langchain-ai.github.io/langgraph/concepts/low_level/#recursion-limit). Let's see an example of this in a simple graph with parallel branches to better understand exactly how the recursion limit works.\n",
```
vbarda, line 9: (nit) amount -> number, fails -> raises an error

PR #1503, opened 2024-08-28
feat(cli): add building support for JS

dqbd:


andrewnguonly (approved):

libs/cli/langgraph_cli/config.py
```diff
@@ -17,27 +18,46 @@ class Config(TypedDict):
...
+            "pip_config_file": config.get("pip_config_file"),
+            "dockerfile_lines": config.get("dockerfile_lines", []),
+            "dependencies": config.get("dependencies", []),
+            "graphs": config.get("graphs", {}),
+            "env": config.get("env", {}),
+        }
+    )
+
+    if config.get("node_version"):
+        if config["node_version"] not in ("20"):
```
vbarda, line 40: `("20")` needs to be `("20",)` -- i am surprised this is not raising an error right now

Issue #1504, opened 2024-08-28
ToolNode output is broken for non-english


sangmandu:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
langgraph.prebuilt.tool_node.py

def str_output(output: Any) -> str:
    if isinstance(output, str):
        return output
    else:
        try:
            return json.dumps(output)
        except Exception:
            return str(output)
```

### Description

I found the tool node output is encoded.
And when I check _run_one and _arun_one function, they use str_output function.
And I checked the arg, the input is alreadt encoded before return str or json.dumps.
I don't know about that well...

this is not completely suitable code but I can solve my problem temporarily

```python
def str_output(output: Any) -> str:
    return json.dumps(json.loads(output), ensure_ascii=False)
```

And I think the origin problem presents here: langchain_core.tools.base

```python
def _stringify(content: Any) -> str:
    try:
        return json.dumps(content)
    except Exception:
        return str(content)
```

they just got a dumps without ensure_ascii=False.

Thank you for reading and I am forwarding to fixing this quickly

### System Info

Just simple error

gbaian10:
Test Case

```py
from enum import Enum

from flask.cli import load_dotenv
from langchain.pydantic_v1 import BaseModel
from langchain_core.tools import tool
from langchain_openai import ChatOpenAI
from langgraph.graph import MessagesState, StateGraph
from langgraph.prebuilt import ToolNode

load_dotenv()


class Day(str, Enum):
    MONDAY = "星期一"
    TUESDAY = "星期二"
    WEDNESDAY = "星期三"
    THURSDAY = "星期四"
    FRIDAY = "星期五"
    SATURDAY = "星期六"
    SUNDAY = "星期日"


class Week(BaseModel):
    """choose days"""

    days: list[Day]


@tool(args_schema=Week)
def get_day_list(days: list[Day]) -> list[Day]:
    return days


tools = [get_day_list]
[...]

PR #1505, opened 2024-08-28
fix: ToolNode output isn't a str and contains non-ASCII characters

gbaian10:
fix: #1504
Solve the issue where non-ASCII characters can't be restored when the tool's output is not a string.

- #1504

### DIFF
--- a/libs/langgraph/langgraph/prebuilt/tool_node.py
+++ b/libs/langgraph/langgraph/prebuilt/tool_node.py
@@ -34,7 +34,7 @@ def str_output(output: Any) -> str:
         return output
     else:
         try:
-            return json.dumps(output)
+            return json.dumps(output, ensure_ascii=False)
         except Exception:
             return str(output)
 


vbarda:
@gbaian10 thank you -- could you please adda a unittest to `tests/test_prebuilt.py`?

sangmandu:
I think this code does not have effect bcuz tool node receive already encoded string.
You should look langchain_core.tools.base

```python
def _stringify(content: Any) -> str:
    try:
        return json.dumps(content)
    except Exception:
        return str(content)
```

gbaian10:
@sangmandu

The `content` of `ToolMessage` itself accepts `str` or `list[str]`, so if your tool outputs a `list[str]`, it won't execute `_stringify`.

image
image

At this point, the content of `tool_message` remains in UTF-8. Then,
`ToolNode` forcibly converts all results to `str`, ultimately causing the output to become ASCII.

image

---

I also noticed the TODO marker and agree that the `_stringify` function in `langchain_core/tool/base.py` should consider incorporating the same changes.

image

However, this change to the core won't affect the issue of `list[str]` becoming ASCII here.
As mentioned earlier, it bypassed execution of `_stringify` due to detection.

gbaian10:
> @gbaian10 thank you -- could you please adda a unittest to `tests/test_prebuilt.py`?

@vbarda I added a test.

vbarda:
Thank you!

PR #1549, opened 2024-08-30
Support optional types from typeddict

hinthornw:
In the input schema


libs/langgraph/langgraph/graph/state.py
```diff
@@ -33,14 +31,7 @@
...
 from langgraph.errors import InvalidUpdateError
-from langgraph.graph.graph import (
-    END,
-    START,
-    Branch,
-    CompiledGraph,
-    Graph,
-    Send,
-)
+from langgraph.graph.graph import END, START, Branch, CompiledGraph, Graph, Send
```
hinthornw, line 34: not sure why my ruff formatter did this. lmk if you want me to change back

libs/langgraph/langgraph/utils.py
```diff
@@ -183,3 +189,20 @@ def coerce_to_runnable(thing: RunnableLike, *, name: str, trace: bool) -> Runnab
...
+        origin = get_origin(type_)
+        if origin is Optional:
+            return True
+        if origin is Union:
+            return any(
+                arg is type(None) or is_optional_type(arg) for arg in type_.__args__
+            )
+        return origin is None
+    if hasattr(type_, "__bound__") and type_.__bound__ is not None:
+        return is_optional_type(type_.__bound__)
```
hinthornw, line 207: This currently does *not* type Any as optional, which maybe feels contradictory given that Optional[Any] equals Union[Any, None] and Any covers the None case?

libs/langgraph/langgraph/utils.py
```diff
@@ -183,3 +189,20 @@ def coerce_to_runnable(thing: RunnableLike, *, name: str, trace: bool) -> Runnab
...
+def is_optional_type(type_: Any) -> bool:
+    """Check if a type is Optional."""
+
+    if hasattr(type_, "__origin__") and hasattr(type_, "__args__"):
+        origin = get_origin(type_)
+        if origin is Optional:
+            return True
+        if origin is Union:
+            return any(
+                arg is type(None) or is_optional_type(arg) for arg in type_.__args__
```
nfcampos, line 203: why `arg is type(None)` instead of `arg is None`

hinthornw: ya i think i typoed that when i was inpsecting and seeing "NoneType"

hinthornw: [edit] It actually *would* have to be nonetype rather than None check directly

nfcampos (approved):

PR #1556, opened 2024-08-30
docs: update `Postgres checkpointer for persistence`

gbaian10:
1. Directly send the `AsyncConnectionPool` to the `AsyncPostgresSaver`
2. Rename the variable used with `ConnectionPool`
3. Remove unuse import


gbaian10:
Is it necessary for `ConnectionPool` to use `connection()`?

Since I noticed that `AsyncConnectionPool` doesn't use a `connection()` method, should we also add the use of `connection()` to `AsyncConnectionPool`?

vbarda:
thanks for updating and good question. we don't need to use `pool.connection()` -- we can just pass the `pool` directly for either sync or async checkpointer. do you mind updating that as well?

PR #1557, opened 2024-08-30
Optional types for dataclass state defs

hinthornw:
Mark those with default or default_factory as being optional


libs/langgraph/langgraph/utils.py
```diff
@@ -273,6 +274,16 @@ def get_field_default(name: str, type_: Any, schema: Type[Any]) -> Any:
             return ...
         # Handle NotRequired[<type>] for earlier versions of python
         return None
+    if hasattr(schema, "__dataclass_fields__"):
```
nfcampos, line 277: I think we should be using `fields` function from dataclasses module no?

libs/langgraph/tests/test_state.py
```diff
@@ -120,11 +122,59 @@ class State(InputState):  # this would be ignored
 
         expected_optional = {"val2"}
 
-    # The others should always have precedence based on the required annotation
+    # The others shoud always have precedence based on the required annotation
```
nfcampos, line 125: Typo

PR #1560, opened 2024-08-31
patch(docs): Update docs with JS cloud instructions

jacoblee93:
@isahers1

### DIFF
--- a/docs/docs/cloud/deployment/setup_javascript.md
+++ b/docs/docs/cloud/deployment/setup_javascript.md
@@ -1,0 +1,200 @@
+# How to Set Up a LangGraph Application for Deployment
+
+A LangGraph.js application must be configured with a [LangGraph API configuration file](../reference/cli.md#configuration-file) in order to be deployed to LangGraph Cloud (or to be self-hosted). This how-to guide discusses the basic steps to setup a LangGraph.js application for deployment using `package.json` to specify project dependencies.
+
+This walkthrough is based on [this repository](https://github.com/langchain-ai/langgraphjs-studio-starter), which you can play around with to learn more about how to setup your LangGraph application for deployment.
+
+The final repo structure will look something like this:
+
+```bash
+my-app/
+├── src # all project code lies within here
+│   ├── utils # optional utilities for your graph
+│   │   ├── tools.ts # tools for your graph
+│   │   ├── nodes.ts # node functions for you graph
+│   │   └── state.ts # state definition of your graph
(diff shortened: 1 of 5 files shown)


docs/docs/cloud/quick_start.md
```diff
@@ -34,14 +46,51 @@ This tutorial will use:
...
+    const tools = [
+      new TavilySearchResults({ maxResults: 3, }),
+    ];
+
+    export const graph = createReactAgent({ llm: model, tools });
+    ```
+
+3.  The `requirements.txt`/`package.json` file should contain any dependencies for your graph(s). In this case we only require four packages for our graph to run:
+
+=== "Python"
```
isahers1, line 69: Think you need tripe back slash for the python imports below

isahers1, line 69: think you want triple backslashes for the python imports below

docs/docs/cloud/quick_start.md
```diff
@@ -34,14 +46,51 @@ This tutorial will use:
...
+          "name": "my-app",
+          "packageManager": "yarn@1.22.22",
+          "dependencies": {
+            "@langchain/community": "^0.2.31",
+            "@langchain/core": "^0.2.31",
+            "@langchain/langgraph": "0.2.0",
+            "@langchain/openai": "^0.2.8"
+          }
+        }
+        ```
```
isahers1: This tab renders weirdly, you can see the backslashes and the json

docs/docs/cloud/deployment/setup_javascript.md
```diff
@@ -0,0 +1,200 @@
...
+    "agent": "./src/agent.ts:graph"
+  },
+  "env": ".env"
+}
+```
+
+Note that the variable name of the `CompiledGraph` appears at the end of the value of each subkey in the top-level `graphs` key (i.e. `:<variable_name>`).
+
+!!! warning "Configuration Location"
+The LangGraph API configuration file must be placed in a directory that is at the same level or higher than the TypeScript files that contain compiled graphs and associated dependencies.
```
isahers1, line 196: Doesn't render properly, think you need to indent the text you want in the box

isahers1 (commented):
I think we are missing tabs in the "Use with sdk" section of the quick start ([here](https://langchain-ai.github.io/langgraph/cloud/quick_start/#use-with-the-sdk)) but other than that lgtm

isahers1 (approved):

PR #1573, opened 2024-09-02
docs: fix list appending and grammar

Doge-is-Dope:
* Fixed grammatical errors and refined the document.
* Added list brackets, as tuples cannot be concatenated with lists.


vbarda:
Thank you! The issue has already been fixed in the current version of the documentation, so will close

PR #1582, opened 2024-09-03
Remove pydantic base class from Pregel/PregelNode

nfcampos:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -232,36 +225,50 @@ class Pregel(
...
+        channels: Mapping[str, Union[BaseChannel, ManagedValueSpec]] = None,
+        auto_validate: bool = True,
+        stream_mode: StreamMode = "values",
+        output_channels: Union[str, Sequence[str]],
+        stream_channels: Optional[Union[str, Sequence[str]]] = None,
+        interrupt_after_nodes: Union[All, Sequence[str]] = (),
+        interrupt_before_nodes: Union[All, Sequence[str]] = (),
+        input_channels: Union[str, Sequence[str]],
+        step_timeout: Optional[float] = None,
+        debug: bool = False,
```
vbarda, line 241: (nit) should `debug` be None here to match the previous behavior of using `get_debug` by default?

nfcampos: Good point

PR #1583, opened 2024-09-03
Use a lock for all operations on sqlite checkpointer

nfcampos:
- Otherwise when used in multiple subgraphs in parallel separate queries can interfere w each other

Closes #1586

### DIFF
--- a/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
+++ b/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
@@ -163,14 +163,15 @@ class SqliteSaver(BaseCheckpointSaver):
         Yields:
             sqlite3.Cursor: A cursor for the SQLite database.
         """
-        self.setup()
-        cur = self.conn.cursor()
-        try:
-            yield cur
-        finally:
-            if transaction:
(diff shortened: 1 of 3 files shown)


libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -163,14 +163,15 @@ def cursor(self, transaction: bool = True) -> Iterator[sqlite3.Cursor]:
...
         """
-        self.setup()
-        cur = self.conn.cursor()
-        try:
-            yield cur
-        finally:
-            if transaction:
-                self.conn.commit()
-            cur.close()
+        with self.lock:
```
vbarda, line 166: should we just move the lock into `self.cursor()` and use it everywhere?

nfcampos: that's what I did no?

vbarda: yup, my bad on the first part of the question, but saw a few places where we still explicitly do `async with self.lock, self.conn.execute(` -- think we can just use cursor there

PR #1594, opened 2024-09-03
langgraph: bump to langchain-core==0.3 / pydantic v2

vbarda:


libs/langgraph/tests/test_pregel_async.py
```diff
@@ -6214,194 +6214,6 @@ async def qa(data: State) -> State:
         ]
 
 
-@pytest.mark.parametrize("checkpointer_name", ALL_CHECKPOINTERS_ASYNC)
-async def test_in_one_fan_out_state_graph_waiting_edge_custom_state_class(
```
vbarda, line 6218: TODO: bring this back and test it without snapshot assertions

libs/langgraph/langgraph/prebuilt/tool_validator.py
```diff
@@ -175,9 +174,7 @@ def __init__(
                         f"Tool {schema.name} does not have an args_schema defined."
                     )
                 self.schemas_by_name[schema.name] = schema.args_schema
-            elif isinstance(schema, type) and issubclass(
```
nfcampos, line 178: we don't want to remove support for pydantic model v1

libs/langgraph/tests/__snapshots__/test_pregel_async.ambr
```diff
@@ -1,226 +1,4 @@
 # serializer version: 1
-# name: test_in_one_fan_out_state_graph_waiting_edge_custom_state_class
```
nfcampos, line 2: why are we removing tests?

vbarda: this was before i realized that we wanted to support both, i have a todo above to bring this back

libs/langgraph/langgraph/prebuilt/tool_validator.py
vbarda: for tool validation? i thought we wanted v1 and v2 support only for state schema?

libs/langgraph/langgraph/utils.py
```diff
@@ -56,19 +56,21 @@ def __init__(
         recurse: bool = True,
         **kwargs: Any,
     ) -> None:
+        runnable_name = None
```
nfcampos, line 59: why is this changing?

libs/langgraph/tests/__snapshots__/test_pregel_async.ambr
```diff
@@ -1204,15 +1204,15 @@
   	weather_graph_model_node(model_node)
   	weather_graph_weather_node(weather_node<hr/><small><em>__interrupt = before</em></small>)
   	__end__([__end__]):::last
-  	subgraph weather_graph
```
nfcampos, line 1207: why is this changing?

libs/langgraph/langgraph/utils.py
vbarda: otherwise there is an error about `self.name` not being defined, cause it's defined only in some branches without a default value

nfcampos: then just add an `else`?

libs/langgraph/tests/__snapshots__/test_pregel_async.ambr
nfcampos: okay cool

libs/langgraph/langgraph/graph/state.py
```diff
@@ -473,11 +473,7 @@ class CompiledStateGraph(CompiledGraph):
     def get_input_schema(
         self, config: Optional[RunnableConfig] = None
     ) -> type[BaseModel]:
-        from pydantic import BaseModel as BaseModelP
-
-        if isclass(self.builder.input) and issubclass(
-            self.builder.input, (BaseModel, BaseModelP)
-        ):
+        if isclass(self.builder.input) and issubclass(self.builder.input, BaseModel):
```
nfcampos, line 476: why is this changing? removing compat with pydantic v1 models

libs/langgraph/langgraph/graph/state.py
```diff
@@ -508,11 +504,7 @@ def get_input_schema(
     def get_output_schema(
         self, config: Optional[RunnableConfig] = None
     ) -> type[BaseModel]:
-        from pydantic import BaseModel as BaseModelP
```
nfcampos, line 511: same as above

.github/workflows/_test.yml
```diff
@@ -21,7 +21,11 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
-    name: "test #${{ matrix.python-version }}"
+        core-version:
+          - ">=0.2.0,<0.3.0"
+          - "latest"
```
nfcampos, line 26: this won't test with core 0.3 until it's out of pre release I think

libs/langgraph/langgraph/prebuilt/tool_validator.py
```diff
@@ -218,7 +220,7 @@ def run_one(call: ToolCall):
                     name=call["name"],
```
nfcampos, line 220: In line 217 you should use non deprecated methods for v2 models

libs/checkpoint/pyproject.toml
```diff
@@ -10,7 +10,7 @@ packages = [{ include = "langgraph" }]
 
 [tool.poetry.dependencies]
 python = "^3.9.0,<4.0"
-langchain-core = ">=0.2.22,<0.3"
+langchain-core = ">=0.2.38,<0.4"
```
nfcampos, line 13: can we remove this dependency?

libs/langgraph/tests/test_pregel.py
```diff
@@ -7632,10 +7663,6 @@ def decider(data: State) -> str:
 
     app = workflow.compile()
 
-    assert app.get_graph().draw_mermaid(with_styles=False) == snapshot
```
nfcampos, line 7635: why is this being removed?

.github/workflows/_test.yml
vbarda: yea, it needs to be flipped, will update

libs/checkpoint/pyproject.toml
vbarda: don't think so, we're using it in jsonplus serializer

```python
from langchain_core.load.load import Reviver
from langchain_core.load.serializable import Serializable
```

libs/langgraph/tests/test_pregel.py
vbarda: can add it back, originally removed because it was a pydantic v1 test, but we can keep it

libs/langgraph/langgraph/graph/state.py
```diff
@@ -21,6 +21,7 @@
 from langchain_core.runnables.base import RunnableLike
 from langchain_core.runnables.utils import create_model
 from pydantic import BaseModel
+from pydantic.v1 import BaseModel as BaseModelV1
```
nfcampos, line 24: does this crash if the user has pydantic v1 installed instead?

vbarda: no, tested on 1.10.18

alin-golumbeanu:
Hello,

Since this changes I get the following error:

image

PR #1607, opened 2024-09-04
docs: Correct the reference repo in the `Cloud - Setup App (pyproject.toml)` documents

gbaian10:
1. fix: Point the reference repo to `langgraph-example-pyproject`
1. fmt


vbarda:
nice, thank you

PR #1630, opened 2024-09-06
Implement LangGraph Scheduler for Kafka

nfcampos:
- Orchestrator and Executor classes to run LangGraph in a distributed fashion using Kafka as a message bus for communication
- Orchestrator and Executor run on-demand when a new message is published to the topic they listen to
- Orchestrator is responsible for running the Pregel algorithm (deciding next tasks to run) and sending messages to the executor topic
- Executor is responsible for executing each task (node), and sending messages to the orchestrator topic when done


libs/scheduler-kafka/langgraph/scheduler/kafka/orchestrator.py
```diff
@@ -0,0 +1,94 @@
...
+                                task=ExecutorTask(
+                                    id=task.id,
+                                    path=task.path,
+                                    step=loop.step,
+                                    resuming=loop.input is INPUT_RESUMING,
+                                ),
+                            ),
+                        )
+                    # flush producer
+                    await self.producer.flush()
```
nfcampos, line 89: @akira is this needed?

libs/scheduler-kafka/testss/test_scheduler.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+async def test_fanout_graph(topics: Topics, checkpointer: BaseCheckpointSaver) -> None:
+    graph = mk_fanout_graph(checkpointer)
+    n_orch_msgs = 0
+    n_exec_msgs = 0
+
+    async def orchestrator() -> None:
+        nonlocal n_orch_msgs
+        async with KafkaOrchestrator(graph, topics) as orch:
```
nfcampos, line 90: @hinthornw is this a good api? the iterator just serves to inspect the messages, it does all the actual work internally

libs/scheduler-kafka/langgraph/scheduler/kafka/orchestrator.py
akira: This would wait until the messages are all delivered rather than do async and be handled by periodic flush - so depends on what the semantics that you need?  Would more performance to use async but there's discussion here:
https://github.com/confluentinc/confluent-kafka-python/issues/137

nfcampos: I was using flush to ensure that I don't mark as "scheduled" messages that actually weren't

Issue #1251, opened 2024-08-07
ToolNode is not seting  the `status` field of the ToolMessage


thiagotps:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
import requests
from langgraph.prebuilt import ToolNode
from langchain_core.messages import AIMessage

def binance(symbol: str):
    """This tool get the symbol information from binance"""
    raise Exception("Network crashed !!!")


ToolNode([binance]).invoke({"messages": [
AIMessage(content=[{'type': 'tool_use', 'name': 'binance', 'input': {'symbol': 'BTCBRL'}, 'id': 'tooluse_82v7QpgYQv6uL2stXaFaQw'}], response_metadata={'ResponseMetadata': {'RequestId': '2bad89c2-9451-41ce-a360-221509ee434f', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Wed, 07 Aug 2024 13:03:07 GMT', 'content-type': 'application/json', 'content-length': '275', 'connection': 'keep-alive', 'x-amzn-requestid': '2bad89c2-9451-41ce-a360-221509ee434f'}, 'RetryAttempts': 0}, 'stopReason': 'tool_use', 'metrics': {'latencyMs': 748}}, id='run-808b9eb2-887d-43ab-addd-b77d37ecb026-0', tool_calls=[{'name': 'binance', 'args': {'symbol': 'BTCBRL'}, 'id': 'tooluse_82v7QpgYQv6uL2stXaFaQw', 'type': 'tool_call'}], usage_metadata={'input_tokens': 337, 'output_tokens': 56, 'total_tokens': 393})
    ]})
```

### Error Message and Stack Trace (if applicable)

```shell
{'messages': [ToolMessage(content="Error: Exception('Network crashed !!!')\n Please fix your mistakes.", name='binance', tool_call_id='tooluse_82v7QpgYQv6uL2stXaFaQw')]}
```

### Description

[...]

PR #1636, opened 2024-09-06
Set tool message status

hinthornw:
Closes https://github.com/langchain-ai/langgraph/issues/1251


hinthornw:
I hear we're supposed to let tools do this now. cool.

PR #1647, opened 2024-09-06
langgraph: check if model passed as runnable binding with tools in create_react_agent

vbarda:
This should allow users to customize models they pass to `create_react_agent`, including enabling/disabling parallel tool calling on the model.

If a model is passed as `RunnableBinding` with tools, we'll validate that the tools match the list of tools in `create_react_agent` and use the bound tools instead of binding inside `create_react_agent`


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -426,7 +465,9 @@ class Agent,Tools otherClass
     else:
         tool_classes = tools
         tool_node = ToolNode(tool_classes)
```
hinthornw, line 467: Unrelated to this but thoughts on getting the tool_classes from the tool node here too so we can let you pass in raw functions ?
```
        tool_classes = tools.tools_by_name.values()
```

vbarda: i think in `tools_by_name` it will also be a tool as we're calling `create_tool` under the hood https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/prebuilt/tool_node.py#L95

nfcampos (approved):

PR #1648, opened 2024-09-07
Adding postgres cache support for node

zhouyuf6741:
Summary:
We currently have writes result stored to the mapping data storage. however, during each different execution, we still rerun the execution to get the writes result.
To solve this issue, we provided a CachePolicy which allows people pass in a cache_key which specify how they want to cache this node. It can based up the step name or user id etc.

This version currently only support postgres sync request, but in the future we would like to expand this to support all the data storages and async.

Test:
Added test_cache file.


libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -334,6 +334,17 @@ def tick(
             self.status = "done"
             return False
 
+        if self.checkpointer:
+            for task in self.tasks.values():
+                # if there are cached writes, apply them
+                cached_writes = self.checkpointer.get_writes_by_cache_key(task.id)
+                if cached_writes and not task.writes:
```
zhouyuf6741, line 341: only do this when task.cache_policy is existing

zhouyuf6741:
@nfcampos looks like `test_pregel.py::test_message_graph` is failing for langchain-core: 0.3.0.dev1 even on main branch. Let me know if you want me to update the snapshot.

vbarda:
> @nfcampos looks like `test_pregel.py::test_message_graph` is failing for langchain-core: 0.3.0.dev1 even on the main branch. Let me know if you want me to update the snapshot.

this should be fixed now -- merged the PR fixing the snapshots

PR #1665, opened 2024-09-10
sdk-py: add versioning for assistants

isahers1:
- Lists all the versions for a given assistant
- Allows you to change assistant to point to specific version

### DIFF
--- a/libs/sdk-js/src/client.ts
+++ b/libs/sdk-js/src/client.ts
@@ -245,7 +245,7 @@ export class AssistantsClient extends BaseClient {
     metadata?: Metadata;
     assistantId?: string;
     ifExists?: OnConflictBehavior;
-    assistantName?: string;
+    name?: string;
   }): Promise<Assistant> {
     return this.fetch<Assistant>("/assistants", {
       method: "POST",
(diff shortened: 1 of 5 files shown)


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -563,6 +568,50 @@ async def search(
...
+
+            assistant_versions = await client.assistants.get_versions(
+                assistant_id="my_assistant_id"
+            )
+
+        """  # noqa: E501
+
+        return await self.http.get(f"/assistants/{assistant_id}/versions")
+
+    async def change_version(self, assistant_id: str, version: int) -> Assistant:
```
nfcampos, line 590: maybe should be called set_latest or something?

isahers1: added to my todo, will make sure to change before merging

vbarda: might make sense to keep these consistent in the API as well. maybe `update_version` / `set_version`?

dqbd (approved):

libs/sdk-js/src/client.ts
```diff
@@ -313,6 +318,47 @@ export class AssistantsClient extends BaseClient {
...
+  }
+
+  /**
+   * Change the version of an assistant.
+   *
+   * @param assistantId ID of the assistant.
+   * @param version The version to change to.
+   * @returns The updated assistant.
+   */
+  async setVersion(assistantId: string, version: number): Promise<Assistant> {
```
dqbd, line 356: Should be `setLatest` to mirror Python

PR #1667, opened 2024-09-10
langgraph: expand handle_tool_errors in ToolNode

isahers1:
This change expands error-handling functionality of the `ToolNode` by introducing more options for `handle_tool_errors`. Default behavior of the `ToolNode` is unchanged -- all errors are handled and wrapped in a `ToolMessage` to be sent back to LLM.

With this change, users have flexibility to only handle the exceptions that they need to pass back to the LLM:

* they can specify exceptions to handle by passing a tuple of exceptions in `handle_tool_errors`
* specify `handle_tool_errors=True/str/callable`
  * when `handle_tool_errors` is a callable, the signature will be inspected and exceptions from the signature will be handled


libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +71,83 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
     a list of ToolMessages, one for each tool call.
 
+    Args:
+        tools: A sequence of tools that can be invoked by the ToolNode.
+        name: The name of the ToolNode in the graph. Defaults to "tools".
+        tags: Optional tags to associate with the node. Defaults to None.
+        handle_tool_errors: How to handle tool errors raised by tools inside the node. Defaults to "all".
+            Must be one of the following:
+
+            - "all": all non-validation errors will be caught and
```
baskaryan, line 130: why have "all" and True?

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +71,109 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
 
     It can be used either in StateGraph with a "messages" key (or a custom key passed via ToolNode's 'messages_key').
     If multiple tool calls are requested, they will be run in parallel. The output will be
     a list of ToolMessages, one for each tool call.
 
+    Args:
+        tools: A sequence of tools that can be invoked by the ToolNode.
+        name: The name of the ToolNode in the graph. Defaults to "tools".
+        tags: Optional tags to associate with the node. Defaults to None.
+        handle_tool_errors: How to handle tool errors raised by tools inside the node. Defaults to "all".
```
eyurtsev, line 153: defaults to "all" but all is not documented?

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +71,109 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
+                a ToolMessage with a default error message (TOOL_CALL_ERROR_TEMPLATE) will be returned.
+            - str: all non-validation errors will be caught and
+                a ToolMessage with the string value of 'handle_tool_errors' will be returned.
+            - tuple[type[Exception], ...]: exceptions in the tuple will be caught and
+                a ToolMessage with a default error message (TOOL_CALL_ERROR_TEMPLATE) will be returned.
+            - Callable[..., str]: exceptions from the signature of the callable will be caught and
+                a ToolMessage with the string value of the result of the 'handle_tool_errors' callable will be returned.
+            - False: none of the errors raised by the tools will be caught
+
+        handle_validation_errors: How to handle validation errors raised by tools inside the node. Defaults to True.
```
eyurtsev, line 166: This logic assumes that a validation error b/c the tool call arguments do not match the function signature of the tool, but a validation error can also be raised from inside the tool (e.g., a tool that uses pydantic models inside).

Is there a way to consolidate into a parameter maybe that's more generic?

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: typo, thanks for catching (leftover from previous version)

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: yea, let's consolidate into `handle_tool_errors`

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -368,6 +374,39 @@ def tool2(some_val: int) -> str:
...
+    expected = (ValueError, ToolException)
+    actual = _infer_handled_types(handle3)
+    assert expected == actual
+
+    expected = (ValueError,)
+    actual = _infer_handled_types(handle4)
+    assert expected == actual
+
+
 async def test_tool_node():
```
eyurtsev, line 410: This is a unit test with 300 lines of stuff in it? Any way to break apart? It's hard to follow right now to understand if the API works as expected

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -394,6 +433,20 @@ async def tool4(some_val: int, some_other_val: str) -> str:
             {"type": "image_url", "image_url": {"url": "abdc"}},
         ]
 
+    @dec_tool
```
eyurtsev, line 436: Does the tool decorator not support `handle_tool_error` as a parameter ?

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -368,6 +374,39 @@ def tool2(some_val: int) -> str:
...
 
 
+def test__infer_handled_types() -> None:
+    def handle(e):  # type: ignore
+        return ""
+
+    def handle2(e: Exception) -> str:
+        return ""
+
+    def handle3(e: Union[ValueError, ToolException]) -> str:
```
eyurtsev, line 384: What happens if the function is misstyped?

```Union[Union[ValueError, TypeError], ToolException]]```

or
```Tuple[ValueError, TypeError]```

**Review by eyurtsev — APPROVED** [review]

**Review by eyurtsev — CHANGES_REQUESTED** [review]
one sec

File: libs/langgraph/langgraph/prebuilt/tool_node.py [hunk 7]
```diff
@@ -67,13 +69,75 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
 
     It can be used either in StateGraph with a "messages" key (or a custom key passed via ToolNode's 'messages_key').
     If multiple tool calls are requested, they will be run in parallel. The output will be
     a list of ToolMessages, one for each tool call.
 
+    Args:
+        tools: A sequence of tools that can be invoked by the ToolNode.
+        name: The name of the ToolNode in the graph. Defaults to "tools".
+        tags: Optional tags to associate with the node. Defaults to None.
+        handle_tool_errors: How to handle tool errors raised by tools inside the node. Defaults to True.
```
eyurtsev, line 128: Is `True` the safe default? I know it's convenient, but what's the contents of `error` in the error message?

Is it the full stack trace? Just the error message? Could something sensitive be leaked by mistake as part of the error message? (e.g., API Key)

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +69,75 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
+            Must be one of the following:
+
+            - True: all errors will be caught and
+                a ToolMessage with a default error message (TOOL_CALL_ERROR_TEMPLATE) will be returned.
+            - str: all errors will be caught and
```
eyurtsev, line 133: Does string support a template variable? {error} ?

vbarda: Nope, i think at that point users should just use callable. do you think we should remove?

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: it's the current default, so i think it would be problematic to change it as it would be breaking. the content is just the error message, no stack trace

hm, that's an interesting point about a sensitive info in error message -- possible? but IMO this is the same class of issue as "what if someone invokes a tool/graph with their API key as input"

libs/langgraph/tests/test_prebuilt.py
vbarda: will break -- good catch

vbarda: added a better error message for those situations

libs/langgraph/langgraph/prebuilt/tool_node.py
eyurtsev: if it's default can leave as is

libs/langgraph/langgraph/prebuilt/tool_node.py
eyurtsev: I think it's fine either way

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +69,81 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
+        if params[0].name in ["self", "cls"] and len(params) == 2:
+            first_param = params[1]
+        else:
+            first_param = params[0]
+
+        type_hints = get_type_hints(handler)
+        if first_param.name in type_hints:
+            origin = get_origin(first_param.annotation)
+            if origin is Union:
+                return tuple(get_args(first_param.annotation))
```
eyurtsev, line 110: What if it's 

`Union[Tuple[str, str]]` ?

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +69,81 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
+            elif origin is not None:
+                raise ValueError(
+                    "Generic types are not supported in the error handler signature. "
```
eyurtsev, line 113: Generic is a thing already

```suggestion
                    "Arbitrary types are not supported in the error handler signature."
```

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +69,81 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
+                    "Please use only Exception types or a union of Exception types."
```
eyurtsev, line 114:
```suggestion
                    "Please annotate the error with either a specific Exception type or a union of Exception types. "
```

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -67,13 +69,81 @@ def msg_content_output(output: Any) -> str | List[dict]:
...
+                )
```
eyurtsev, line 115: If you like could do something like this to be even more helpful:

```suggestion
   "For example, `def custom_handler(e: ValueError)` or `def custom-handler(e: Union[ValueError, TypeError])`
                )
```

and could also try to output the parameter signature

eyurtsev (approved):

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: added more restrictive checks & better error messages

PR #1683, opened 2024-09-11
Performance improvements in core library

nfcampos:
- Avoid creating new callback manager when received one as arg
- Avoid looking for config when already received one as arg
- Avoid copies of values in ensure_config/merge_configs
- Implement version of ensure_config that accepts multiple configs (avoids calling merge_configs first)
- Avoid calling merge_configs when we only need to attach extra tags/metadata

### DIFF
--- a/libs/langgraph/langgraph/pregel/algo.py
+++ b/libs/langgraph/langgraph/pregel/algo.py
@@ -46,6 +46,8 @@ from langgraph.pregel.read import PregelNode
 from langgraph.pregel.types import All, PregelExecutableTask, PregelTask
 from langgraph.utils.config import merge_configs, patch_config
 
+EMPTY_SEQ = tuple()
+
 
 class WritesProtocol(Protocol):
     name: str
@@ -78,7 +80,10 @@ def should_interrupt(
             task
             for task in tasks
             if (
-                (not task.config or TAG_HIDDEN not in task.config.get("tags"))
+                (
+                    not task.config
+                    or TAG_HIDDEN not in task.config.get("tags", EMPTY_SEQ)
+                )
                 if interrupt_nodes == "*"
                 else task.name in interrupt_nodes
             )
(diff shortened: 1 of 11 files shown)


libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -46,6 +46,8 @@
 from langgraph.pregel.types import All, PregelExecutableTask, PregelTask
 from langgraph.utils.config import merge_configs, patch_config
 
+EMPTY_SEQ = tuple()
```
hinthornw, line 49: Do you have an idea of how helpful this was? I guess this is sort of an inner loop

libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -182,11 +187,10 @@ def apply_writes(
         for chan in task.triggers
         if chan not in RESERVED and chan in channels
     }:
-        if channels[chan].consume():
-            if get_next_version is not None:
-                checkpoint["channel_versions"][chan] = get_next_version(
-                    max_version, channels[chan]
-                )
+        if channels[chan].consume() and get_next_version is not None:
```
hinthornw, line 190: is this just stylistic or does it actually improve perf?

libs/langgraph/langgraph/pregel/io.py
```diff
@@ -97,6 +97,9 @@ def __radd__(self, other: dict[str, Any]) -> "AddableUpdatesDict":
         raise TypeError("AddableUpdatesDict does not support right-side addition")
 
 
+EMPTY_SEQ = tuple()
```
hinthornw, line 100: Should we just make a single empty_seq in an internal _constants file or something

libs/langgraph/langgraph/pregel/algo.py
nfcampos: No this was just style

libs/langgraph/langgraph/pregel/io.py
nfcampos: I guess so?

hinthornw (approved):

libs/langgraph/langgraph/pregel/algo.py
nfcampos: This was more because the previous code was actually incorrect, as it could end up trying to do `<str> in None`

PR #1685, opened 2024-09-11
More performance improvements in checkpointing and channels

nfcampos:


libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -245,7 +245,7 @@ def get_next_version(self, current: Optional[str], channel: ChannelProtocol) ->
             current_v = int(current.split(".")[0])
         next_v = current_v + 1
         try:
-            next_h = md5(self.serde.dumps_typed(channel.checkpoint())[1]).hexdigest()
+            next_h = sha1(self.serde.dumps_typed(channel.checkpoint())[1]).hexdigest()
```
hinthornw, line 248: I'm naively assuming since there's more entropy in sha1 it's at least as unlikely to have a collision as md5 ?

nfcampos: yea less likely, but this isn't that important here tbh, even sha1 is way more entropy than we'd ever need here

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -58,14 +58,14 @@ def _default(self, obj):
...
             )
+        elif hasattr(obj, "_asdict") and callable(obj._asdict):
+            return self._encode_constructor_args(obj.__class__, kwargs=obj._asdict())
         elif isinstance(obj, pathlib.Path):
             return self._encode_constructor_args(pathlib.Path, args=obj.parts)
         elif isinstance(obj, re.Pattern):
             return self._encode_constructor_args(
                 re.compile, args=[obj.pattern, obj.flags]
             )
-        elif isinstance(obj, UUID):
```
hinthornw, line 67: Why not just `str()`?

hinthornw (approved):

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
nfcampos: Because then when we get it back it's not a UUID anymore, but will actually change, as that's the only option with orjson

PR #1699, opened 2024-09-12
checkpoint postgres: allow passing custom serde

vermapratyush:
Enhance AsyncPostgresSaver.from_conn_string constructor method to accept custom jsonserializer.


vbarda:
@vermapratyush thank you for your contribution -- it's okay to pass custom serde, but the metadata serializer should remain jsonplus. could you please make the changes accordingly?

libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -224,7 +224,7 @@ def _dump_writes(
         ]
 
     def _load_metadata(self, metadata: dict[str, Any]) -> dict[str, Any]:
-        return self.jsonplus_serde.loads(self.jsonplus_serde.dumps(metadata))
```
nfcampos, line 227: I think this change also needs to be undone?

vbarda: yup, needs to be undone

nfcampos (approved):

PR #1708, opened 2024-09-13
langgraph: use create_model_v2 and fallback on create_model

vbarda:


libs/langgraph/langgraph/utils/pydantic.py
```diff
@@ -0,0 +1,40 @@
...
+        return create_model_v2(
+            model_name,
+            field_definitions=field_definitions,
+            root=root,
+        )
+    except ImportError:
+        # for langchain-core < 0.3.0
+        from langchain_core.runnables.utils import create_model
+
+        v1_kwargs = {**field_definitions} if field_definitions else {}
```
nfcampos, line 33: Let's skip this copy here

nfcampos (approved):

PR #1716, opened 2024-09-13
Implement serialization with msgpack library

nfcampos:
- encode custom python objects with a msgpack extension type, with constructor path string, and args encoded as nested msgpack doc


zzaym:
Hey @nfcampos, it seems this change breaks redis checkpointer https://github.com/langchain-ai/langgraph/discussions/1123#discussioncomment-10681449.

PR #1750, opened 2024-09-17
langgraph[patch]: rename state_modifier -> preprocess_model_inputs in create_react_agent

baskaryan:


vbarda:
@baskaryan i think we should just make it work with `state_modifier` instead of adding another kwarg

baskaryan:
> @baskaryan i think we should just make it work with `state_modifier` instead of adding another kwarg

sticking to state_modifier makes our code cleaner but think it makes the user experience more confusing. think if prompt was available ppl wouldn't use state_modifier. and makes the transition from langchain agent executor much simpler

vbarda:
> > @baskaryan i think we should just make it work with `state_modifier` instead of adding another kwarg
> 
> sticking to state_modifier makes our code cleaner but think it makes the user experience more confusing. think if prompt was available ppl wouldn't use state_modifier. and makes the transition from langchain agent executor much simpler

updated this to use a new name - `preprocess_model_inputs`

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -49,89 +49,110 @@ class AgentState(TypedDict):
...
     raise ValueError(
         f"Got unexpected type for `messages_modifier`: {type(messages_modifier)}"
     )
 
 
-def _get_model_preprocessing_runnable(
-    state_modifier: Optional[StateModifier],
+def _get_model_input_preprocessor(
+    preprocess_model_inputs: Optional[ModelInputPreprocessor],
+    state_modifier: Optional[ModelInputPreprocessor],
```
baskaryan, line 115: why do we need both (for a private method)

vbarda: we don't necessarily, but putting the logic for picking the preprocessor into a single private function keeps the actual `create_react_agent` a bit cleaner imo

PR #1753, opened 2024-09-18
[Bug Fix] retry_on handling and add support for lists in RetryPolicy

Etelis:
This PR addresses two issues in the run_with_retry function:

* Exception Classes Misidentified as Callables:
    Issue: Exception classes are callable, leading callable(retry_policy.retry_on) to mistakenly treat them as callables for custom retry logic.

* No Support for Lists in retry_on:
    Issue: retry_on accepted only exception classes, tuples, or callables, not lists of exceptions (as defined in the documentation)

Summary of Changes:
* Modified retry_on type annotation to include List[Type[Exception]].
* Updated run_with_retry to handle lists and tuples when checking exceptions.
* Ensured exception classes are not incorrectly treated as callables.
* Added a TypeError if retry_on is of an unsupported type.

### DIFF
--- a/libs/langgraph/langgraph/pregel/retry.py
+++ b/libs/langgraph/langgraph/pregel/retry.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 import random
 import time
-from typing import Optional
+from typing import Optional, Sequence
 
 from langgraph.constants import CONFIG_KEY_RESUMING
 from langgraph.errors import GraphInterrupt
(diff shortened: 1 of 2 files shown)


nfcampos:
Thanks for this! Made a few updates, will merge it once CI finishes

nfcampos (approved):

Etelis:
Thank you.
Also, what are your thoughts about introducing type checking for the `types.py `file?
@nfcampos

nfcampos:
@Etelis in the past we faced some issues with getting mypy to behave with our codebase. I can try to see if mypy has improved since, but would probably be a fairly involved task

PR #1759, opened 2024-09-18
docs: replaying from subgraph node

isahers1:


docs/docs/how-tos/subgraphs-manage-state.ipynb
```diff
@@ -382,22 +382,120 @@
...
+    "parent_graph_state_before_subgraph = next(h for h in graph.get_state_history(config) if h.next == ('weather_graph',))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 63,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "subgraph_state_before_model_node = next(h for h in graph.get_state_history(parent_graph_state_before_subgraph.tasks[0].state) if h.next == ('model_node',))\n",
```
apoorvkk, line 432: what if we are in a loop in the graph? Would it make sense to get the last state that has its next action as `model_node` instead? I get that this particular example it's not necessary but I am just thinking of scenarios where there could be a circular loop and we might want to retrieve the **latest state** that has the targeted next action.

apoorvkk: Unless state history yields the latest states at the start of the list rather than at the end. If that is the case, my comment is redundant :)

isahers1: Yes, the most recent state is returned first :)

PR #1775, opened 2024-09-19
docs: add docs on disabling streaming

isahers1:


nfcampos (commented):
`## With disabling` sounds pretty strange

isahers1:
> `## With disabling` sounds pretty strange

renamed to "Disabling streaming" - lmk if that sounds better

VKudlay:
Hey @isahers1. Just stumbled upon this and tried it out (didn't quite work ootb for my case, but no matter). Are there any more weedy details about how stream gets enforced in the nodes, and maybe a more general (maybe scope-based?) disabling technique? Asking since also this is just relying on a very specific client supporting a very specific argument which might take a while to bleed into other clients.

PR #1776, opened 2024-09-19
Add stream_mode=custom

nfcampos:
- adds the ability for nodes (including in subgraphs) to emit chunks directly to the output stream, emitted chunks can have any type
- when stream_mode=custom isnt requested by the caller emitted chunks are ignored

### DIFF
--- a/libs/langgraph/langgraph/constants.py
+++ b/libs/langgraph/langgraph/constants.py
@@ -1,11 +1,13 @@
 from dataclasses import dataclass
-from typing import Any, Literal
+from types import MappingProxyType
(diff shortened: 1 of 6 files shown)


hinthornw (approved):
What does this look like in the API? Or if you do stream_events?  
When should I use this instead of `adispatch_custom_event `?

nfcampos:
@hinthornw IMO astream_events is an escape hatch no one should have to use under normal circumstances. I think in a world with multimodal / continuous IO having custom streaming output from your graph should be a first class citizen. Thoughts?

efriis (approved):

PR #1820, opened 2024-09-24
docs: fix quick start example code

Yuxuan1998:
Encountered following error when running the example:

TypeError: can only concatenate list (not "tuple") to list


vbarda:
@Yuxuan1998 thanks for the PR -- the original code works fine for me, did you modify it in any way?

PR #1835, opened 2024-09-25
Support stringized annotations

hinthornw:


hinthornw:
Hm seems eval is too broad

libs/langgraph/langgraph/utils/runnable.py
```diff
@@ -65,14 +65,14 @@ class StrEnum(str, enum.Enum):
 ] = (
     (
         sys.intern("writer"),
-        (StreamWriter, inspect.Parameter.empty),
+        (StreamWriter, "StreamWriter", inspect.Parameter.empty),
```
hinthornw, line 68: Open to doing something more specific to eval annotations only if it matches this condition or something, but I feel like this should be precise enough...?

nfcampos (approved):

PR #1836, opened 2024-09-25
Add store to pregel config

hinthornw:
Apologies for resorting the test imports

### DIFF
--- a/libs/langgraph/langgraph/pregel/__init__.py
+++ b/libs/langgraph/langgraph/pregel/__init__.py
@@ -61,6 +61,7 @@ from langgraph.constants import (
     CONFIG_KEY_READ,
     CONFIG_KEY_RESUMING,
     CONFIG_KEY_SEND,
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1237,6 +1240,8 @@ def output() -> Iterator:
                 config[CONF][CONFIG_KEY_STREAM_WRITER] = lambda c: stream.put(
                     ((), "custom", c)
                 )
+            if self.store:
```
hinthornw, line 1243: this is the acutal change.

Very open to other suggestions.

Issue was that the store wasn't being set anymore now that we aren't necessarily deifning managed channels

nfcampos: Ah

hinthornw:
Closing since Nuno made a better fix that works in the Kafka scheduler

PR #1843, opened 2024-09-25
docs: add guide for entity memory

ccurme:
This guide is intended to help users transition from a class of legacy memory abstractions. See guides in v0.1 LangChain docs here:
- https://python.langchain.com/v0.1/docs/modules/memory/types/entity_summary_memory/
- https://python.langchain.com/v0.1/docs/modules/memory/types/kg/

Note: the existing [shared state](https://langchain-ai.github.io/langgraph/how-tos/memory/shared-state/) guide already demonstrates this functionality. Here we add a guide focusing on the "conversation memory" problem that:

1. Covers the same solution;
2. Adds a section showing how to use an external knowledge base, instead of the LangGraph persistence layer, if desired.

If we'd rather not maintain such similar content in two places, we can also direct users to the existing shared state guide.


ccurme:
Converting to draft pending discussion with @vbarda about merging into a single tutorial with vectorstore memory.

ccurme:
Closing in favor of https://github.com/langchain-ai/langgraph/pull/1861

PR #1849, opened 2024-09-25
ci: fix notebook runner & VCR cassette recorder

isahers1:


docs/docs/how-tos/disable-streaming.ipynb
```diff
@@ -113,9 +113,12 @@
...
    "source": [
     "input = {\"messages\": {\"role\":\"user\", \"content\":\"how many r's are in strawberry?\"}}\n",
-    "async for event in graph.astream_events(input, version=\"v2\"):\n",
-    "    if event[\"event\"] == \"on_chat_model_end\":\n",
-    "        print(event[\"data\"][\"output\"].content, end=\"\", flush=True)"
+    "try:\n",
+    "    async for event in graph.astream_events(input, version=\"v2\"):\n",
+    "        if event[\"event\"] == \"on_chat_model_end\":\n",
+    "            print(event[\"data\"][\"output\"].content, end=\"\", flush=True)\n",
+    "except:\n",
```
vbarda, line 120: why is this needed?

docs/docs/how-tos/subgraphs-manage-state.ipynb
```diff
@@ -636,7 +636,6 @@
     }
    ],
    "source": [
-    "graph = \n",
```
vbarda, line 639: also need to fix comment for this notebook

docs/_scripts/prepare_notebooks_for_ci.py
```diff
@@ -107,8 +112,11 @@ def add_vcr_to_notebook(
         "",
         "def filter_cassette_data(cassette_dict):",
         "    for interaction in cassette_dict['interactions']:",
-        "        if len(interaction['response']['body']['string']) > 1000:",
-        "            interaction['response']['body']['string'] = interaction['response']['body']['string'][:1000] + '... (truncated)'",
+        "        body = interaction['response']['body']['string']",
```
vbarda, line 115: are you sure this code is not messing up anything?

docs/docs/tutorials/extraction/retries.ipynb
```diff
@@ -755,18 +734,21 @@
...
-    "            (\n",
-    "                \"user\",\n",
-    "                f\"Extract the summary from the following conversation:\\n\\n<convo>\\n{formatted}\\n</convo>\"\n",
-    "                \"\\n\\nRemember to respond using the TranscriptSummary function.\",\n",
-    "            )\n",
-    "        ]\n",
-    "    },\n",
-    ")\n",
-    "results.pretty_print()"
+    "try:\n",
```
vbarda, line 737: why is this needed?

docs/docs/tutorials/multi_agent/hierarchical_agent_teams.ipynb
```diff
@@ -291,7 +291,16 @@
     "from langchain_openai import ChatOpenAI\n",
     "\n",
     "from langgraph.graph import END, StateGraph, START\n",
-    "from langchain_core.messages import HumanMessage\n",
+    "from langchain_core.messages import HumanMessage, trim_messages\n",
+    "\n",
+    "llm = ChatOpenAI(model=\"gpt-4o-mini\")\n",
+    "\n",
+    "trimmer = trim_messages(\n",
+    "    max_tokens=100000,\n",
```
vbarda, line 299: ooc -- does it not work without the trimmer?

isahers1: context window gets exceeded

docs/docs/how-tos/disable-streaming.ipynb
isahers1: because it throws an error - on purpose, but the CI doesn't know that. All errors that we throw on purpose to show something failing (usually we show the way to fix it after) need to be wrapped in try/excepts for the CI

vbarda: cool, makes sense

ccurme:
Not sure if you are running into this, but an issue I had was that running locally, tiktoken embeddings are cached, but in CI it might not be cached leading to extra HTTP calls that are inconsistent with cassettes. I ended up adding a [small script](https://github.com/langchain-ai/langchain/blob/cc/run_notebooks/docs/scripts/download_tiktoken.py) to pre-download the embeddings and running it as a step in the workflow.

vbarda:
> Not sure if you are running into this, but an issue I had was that running locally, tiktoken embeddings are cached, but in CI it might not be cached leading to extra HTTP calls that are inconsistent with cassettes. I ended up adding a [small script](https://github.com/langchain-ai/langchain/blob/cc/run_notebooks/docs/scripts/download_tiktoken.py) to pre-download the embeddings and running it as a step in the workflow.

haven't run into this yet but this is great -- we might need to do that in the future, thanks!

docs/_scripts/execute_notebooks.sh
```diff
@@ -23,10 +23,10 @@ execute_notebook() {
 export -f execute_notebook
 
 # Find all notebooks and filter out those in the skip list
-notebooks=$(find docs/docs/tutorials docs/docs/how-tos -name "*.ipynb" | grep -v ".ipynb_checkpoints" | grep -vFf <(echo "$SKIP_NOTEBOOKS"))
+# notebooks=$(find docs/docs/tutorials docs/docs/how-tos -name "*.ipynb" | grep -v ".ipynb_checkpoints" | grep -vFf <(echo "$SKIP_NOTEBOOKS"))
+notebooks=$(find docs/docs/tutorials -name "*.ipynb" | grep -v ".ipynb_checkpoints" | grep -vFf <(echo "$SKIP_NOTEBOOKS"))
 
-# Run notebooks in parallel
-if ! parallel execute_notebook ::: $notebooks; then
```
ccurme, line 29: Out of curiosity why remove `parallel`?

vbarda: mainly because it has unexpected side effects when running with VCR locally

vbarda: (and it actually has no effect on the runner since it doesn't have multiple core, although maybe it's possible to enable that)

PR #1850, opened 2024-09-25
docs: add long-term memory tutorial

vbarda:


docs/docs/tutorials/memory/long_term_memory_agent.ipynb
hinthornw, line 1: We should replace chroma with InMemoryVectorStore https://python.langchain.com/docs/integrations/vectorstores/memorydb/#create-memorydb-vector-store

hinthornw: I think Isaac is going through and converting the "__end__" things to instead not have a response type and do `["tools", END]` as the path map

hinthornw: We can probably load the tiktoken model outside the function (i believe it's cached but even so - better to do on app startup i think)

hinthornw (approved):
Nice! Made a couple comments - I think we'll eventually update this to use the message store logic as well but this is good

docs/docs/tutorials/memory/long_term_memory_agent.ipynb
hinthornw: I think rather than "long-term-memory" we should emphasize it as more like multi-conversation memory or "shared memory" or something

Regular checkpointing is equally persistent, it just is thread-scoped rather than user-scoped

eyurtsev:
Looks great! I only have very minor nit:

1. Is it possible to define what a "memory" is at the very top? Is it a key value pair? a sentence?
2.  When we're printing recall memories it formats the list in a way that make it look like a sentence, would be nicer to show the list maybe:
{'recall_memories': "John loves pizza.\n\nJohn's favorite pizza topping is pepperoni.\n\nJohn recently moved to New York."}

PR #1851, opened 2024-09-25
docs: remove shared state how-to

vbarda:


Revigatech:
can I ask why you deleted this section as it's really cool feature to share state between threads ? 
thanks in adavance

PR #1864, opened 2024-09-26
Add tests for MemoryStore in graph

hinthornw:


libs/checkpoint/langgraph/store/base.py
```diff
@@ -0,0 +1,191 @@
...
+from typing import Any, Iterable, NamedTuple, Optional, Sequence, Union
+
+
+@dataclass
+class Item:
+    """Represents a stored item with metadata."""
+
+    value: dict[str, Any]
+    """The stored data."""
+    scores: dict[str, float]
```
hinthornw, line 19: Could remove until we add the searching functionality

libs/checkpoint/langgraph/store/base.py
```diff
@@ -0,0 +1,191 @@
...
+    """Abstract base class for key-value stores."""
+
+    __slots__ = ("__weakref__",)
+
+    @abstractmethod
+    def batch(self, ops: Iterable[Op]) -> list[Result]:
+        """Execute a batch of operations synchronously."""
+
+    @abstractmethod
+    async def abatch(self, ops: Iterable[Op]) -> list[Result]:
```
hinthornw, line 80: Iterable input, list otuput type.

Did we want to do `Sequence` input instead? I don't see a need to support generators here

PR #1871, opened 2024-09-26
docs: update typing -> typing_extensions for TypedDict

vbarda:


hinthornw (approved):
Nice - can we add a lint rule?

vbarda:
> Nice - can we add a lint rule?

we don't currently lint notebooks, but i think worth adding a linter! will add next week

PR #1880, opened 2024-09-27
feat(sdk): expose xray kwarg to get_graph/getGraph

dqbd:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -304,11 +304,14 @@ async def get(self, assistant_id: str) -> Assistant:
         """  # noqa: E501
         return await self.http.get(f"/assistants/{assistant_id}")
 
-    async def get_graph(self, assistant_id: str) -> dict[str, list[dict[str, Any]]]:
+    async def get_graph(
```
vbarda, line 307: @dqbd could you update sync method as well?

nfcampos (approved):

PR #1884, opened 2024-09-27
docs: reorganize API refs

vbarda:
This change reorganizes the API ref docs in the following ways:

* Removed any additional text, all of the information is pulled in from the docstrings
* Expanded coverage for all files
* New pages:
  * `Types`
  * `Channels`
  *  `Constants` (only public ones)
* Style changes:
  * Show object types
  * Show type annotations

Example screenshots:

* Graphs:
image

* Errors
image


docs/docs/reference/errors.md
```diff
@@ -1,6 +1,3 @@
 # Errors
 
-While you may not want to see them, informative errors help you design better workflows.
```
hinthornw, line 3: Worth adding a module-level docstring here?

docs/docs/reference/channels.md
```diff
@@ -0,0 +1,15 @@
+# Channels
+
+::: langgraph.channels.base
```
hinthornw, line 3: Maybe worth a Module-level docstring saying this section is really primarily for internal or advanced usage?

cc @nfcampos e.g., do we treat changes to these APIs as semver breaking changes?

docs/docs/reference/errors.md
vbarda: maybe, what do you think it should say? to me it feels a bit self-explanatory

eyurtsev:
Is this going to break any existing links?

eyurtsev:
This is the code in langchain right now that links to langgraph api reference: https://github.com/langchain-ai/langchain/pull/26877/files

vbarda:
> This is the code in langchain right now that links to langgraph api reference: https://github.com/langchain-ai/langchain/pull/26877/files

@eyurtsev I don't think so, but worth double-checking once we merge

PR #1892, opened 2024-09-27
Add SDK

hinthornw:


libs/checkpoint/langgraph/store/base.py
```diff
@@ -53,6 +61,15 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self) -> int:
         return hash((self.namespace, self.key))
 
+    def dict(self) -> dict:
```
hinthornw, line 64: This is convenient in the server - am fine moving it out

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -1874,6 +1879,195 @@ async def search(
...
+            None
+
+        Example Usage:
+
+            await client.store.delete_item(
+                ["documents", "user123"],
+                key="item456",
+            )
+        """
+        await self.http.post(
```
nfcampos, line 1963: Shouldn't this use DELETE method?

libs/sdk-py/langgraph_sdk/schema.py
```diff
@@ -197,6 +197,27 @@ class RunCreate(TypedDict):
     multitask_strategy: Optional[MultitaskStrategy]
 
 
+class Item(TypedDict):
+    namespace: list[str]
+    """The namespace of the item."""
+    id: str
+    """The unique identifier of the item within its namespace."""
+    value: Any
```
nfcampos, line 205: Don't we enforce this to be dict?

nfcampos (approved):

libs/sdk-py/langgraph_sdk/schema.py
hinthornw: Ah ya good good point

libs/sdk-py/langgraph_sdk/client.py
hinthornw: Updated this and GET and PUT

PR #1906, opened 2024-09-30
Add Postgres Store Implementation

hinthornw:
Both sync and async savers.

Will improve the docstrings with examples in a followup PR


libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
```diff
@@ -17,9 +17,9 @@
     CheckpointTuple,
     get_checkpoint_id,
 )
-from langgraph.checkpoint.postgres.base import (
-    BasePostgresSaver,
-)
+from langgraph.checkpoint.postgres.aio_store import AsyncPostgresStore
+from langgraph.checkpoint.postgres.base import BasePostgresSaver
+from langgraph.checkpoint.postgres.base_store import PostgresStore
```
vbarda, line 22: should this be `langgraph.checkpoint.postgres.store.base` and the same for others?

hinthornw: ya

libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
```diff
@@ -384,3 +386,6 @@ def _cursor(self, *, pipeline: bool = False) -> Iterator[Cursor[DictRow]]:
             else:
                 with self.lock, conn.cursor(binary=True, row_factory=dict_row) as cur:
                     yield cur
+
+
+__all__ = ["PostgresSaver", "Conn", "PostgresStore", "AsyncPostgresStore"]
```
vbarda, line 391: we need to remove/edit

libs/checkpoint-postgres/langgraph/store/postgres/base_store.py
```diff
@@ -0,0 +1,393 @@
+import asyncio
```
vbarda, line 1: (nit) can be renamed as base / aio

libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
hinthornw: Aya thanks! I'm a tad surprised this isn't caught by linting

vbarda (approved):
Might be good to add a test that uses this in langgraph tests

PR #1910, opened 2024-09-30
Add JS SDK Store Client

hinthornw:


libs/sdk-js/src/client.ts
```diff
@@ -983,6 +986,137 @@ export class RunsClient extends BaseClient {
...
+  }
+
+  /**
+   * Retrieve a single item.
+   *
+   * @param namespace A list of strings representing the namespace path.
+   * @param key The unique identifier for the item.
+   * @returns Promise<Item>
+   */
+  async getItem(namespace: string[], key: string): Promise<Item> {
```
bracesproul, line 1030: In the `BaseStore` api, get returns `Item` or `null`, is that not the case for this api?

libs/sdk-js/src/client.ts
```diff
@@ -983,6 +986,137 @@ export class RunsClient extends BaseClient {
...
+   *
+   * @param namespace A list of strings representing the namespace path.
+   * @param key The unique identifier for the item.
+   * @returns Promise<void>
+   */
+  async deleteItem(namespace: string[], key: string): Promise<void> {
+    return this.fetch<void>("/store/items", {
+      method: "DELETE",
+      json: { namespace, key },
+    });
```
bracesproul, line 1055: No period check?

libs/sdk-js/src/schema.ts
```diff
@@ -219,3 +219,19 @@ export interface Checkpoint {
...
+export interface SearchItemsResponse {
+  items: Item[];
+}
+
+export interface Item {
+  namespace: string[];
+  key: string;
+  value: Record<string, any>;
+  created_at: string;
+  updated_at: string;
```
bracesproul, line 236: In JS I added these as `createdAt` and `updatedAt`...

To keep parity we should probably re-map these values after fetching so that they aren't returned in the camel case format

libs/sdk-js/src/client.ts
```diff
@@ -983,6 +986,137 @@ export class RunsClient extends BaseClient {
...
+      if (label.includes(".")) {
+        throw new Error(
+          `Invalid namespace label '${label}'. Namespace labels cannot contain periods ('.')`,
+        );
+      }
+    });
+
+    return this.fetch<Item>("/store/items", {
+      params: { namespace: namespace.join("."), key },
+    });
```
bracesproul, line 1041: Here:
```suggestion
    const getRes = await this.fetch<Item>("/store/items", {
      params: { namespace: namespace.join("."), key },
    });
    
    if (!getRes.ok) {
      throw new Error("...")
    }
    return {
      ...getRes,
      createdAt: getRes.created_at,
      updatedAt: getRes.updated_at,
    }
```

Not exactly sure if this typing is all correct, but roughly this idea

bracesproul (commented):
On last thing, I'm not entirely sure the result of a fetch request will be the data object, I would double check it's not nested inside some other field in the response

libs/sdk-js/src/schema.ts
hinthornw: cool can do

libs/sdk-js/src/client.ts
hinthornw: ack - thanks!

libs/sdk-js/src/client.ts
hinthornw: good catch

PR #1920, opened 2024-09-30
docs: add shared memory how to

vbarda:


docs/docs/how-tos/memory/shared-state.ipynb
hinthornw, line 1: Let's note the minimum required version?

hinthornw (approved):

docs/docs/how-tos/memory/shared-state.ipynb
```diff
@@ -0,0 +1,439 @@
...
+   "id": "a7f303d6-612e-4e34-bf36-29d4ed25d802",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from typing import Literal, TypedDict, Annotated\n",
+    "import uuid\n",
+    "from langchain_openai import ChatOpenAI\n",
+    "\n",
+    "from langchain_core.runnables import RunnableConfig\n",
+    "from langgraph.graph.message import MessagesState\n",
```
hwchase17, line 111: you can just import this and the below line from `langgraph.graph`

docs/docs/how-tos/memory/shared-state.ipynb
```diff
@@ -0,0 +1,439 @@
...
+    "# This is the in memory store needed to save the memories (i.e. user preferences)\n",
+    "in_memory_store = InMemoryStore()\n",
+    "\n",
+    "# Construct this relatively simple graph\n",
+    "graph = StateGraph(MessagesState)\n",
+    "graph.add_node(call_model)\n",
+    "graph.add_node(update_memory)\n",
+    "graph.add_edge(\"update_memory\", END)\n",
+    "graph.add_edge(START, \"call_model\")\n",
+    "graph.add_conditional_edges(\"call_model\", route, [END, \"update_memory\"])\n",
```
hwchase17, line 214: i dont think you need the third argument (can just type the conditional edge better)

if you do want to keep here, at least pass it as a kwarg (since it is)

vbarda: we're actually moving away from types in conditional edges to use this format instead in the guides, since you can't use `END` in `Literal` for type annotations, i.e. have to use `"__end__"`. so this is just part of the bigger change to standardize everything and avoid users having to type `__end__` everywhere

but lmk if you disagree with this!

PR #1924, opened 2024-09-30
tutorial part6 introduction.ipynb fix imports

jrwren:


vbarda:
Thank you -- this import is actually not necessary, will update separately and close this PR

PR #1931, opened 2024-10-01
Add rollback to postgres saver setup

nickderobertis:
When I try to use the `PostgresSaver` I'm getting the following error:

```
  File "/<redacted>/.venv/lib/python3.12/site-packages/langgraph/checkpoint/postgres/__init__.py", line 106, in setup
    cur.execute(migration)
  File "/<redacted>/.venv/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block
```

Adding the rollbacks in the `except UndefinedTable` clause seems to fix it and the tables get created successfully.

I consistently hit this error and the change consistently fixes it. I'm not sure how others have avoided this issue. But it doesn't seem like it should cause an issue for anyone as doing an extra rollback shouldn't hurt when the only thing before it is a select query.

Let me know if you'd like me to contribute a unit test or file a separate issue, it just seemed like a small change.

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
@@ -94,6 +94,7 @@ class PostgresSaver(BasePostgresSaver):
                     version = row["v"]
             except UndefinedTable:
                 version = -1
+                cur.connection.rollback()
             for v, migration in zip(
                 range(version + 1, len(self.MIGRATIONS)),
                 self.MIGRATIONS[version + 1 :],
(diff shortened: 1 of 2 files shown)


vbarda:
@nickderobertis thanks for your contribution! the reason you're running into the issue is because you're likely using a connection that is not in [autocommit mode](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py#L71). i am going to close this PR in the meantime, but please do let me know if that resolves your issue!

nickderobertis:
Hey @vbarda thank you for your quick response and clarification about this. Indeed if I add the `autocommit` kwarg to my connection pool it starts working.

I ran into this after upgrading from langgraph 0.1 where the Postgres saver was just example code in the docs that I copy-pasted into my project rather than something directly supported in the code base. For that example code, the connection pool was working without autocommit. 

Perhaps I could submit a PR to update the docs [here](https://langchain-ai.github.io/langgraph/how-tos/persistence_postgres/) to be a little more clear that autocommit is needed for the connection? It is shown in connection kwargs in the example but nothing explicitly says it is needed from what I can see.

PR #1941, opened 2024-10-01
ci: run notebook checks for changed notebooks

vbarda:


.github/workflows/ci.yml
```diff
@@ -78,6 +90,13 @@ jobs:
       - name: Run check_sdk_methods script
         run: python .github/scripts/check_sdk_methods.py
 
+  check-notebooks:
```
eyurtsev, line 93: Before adding this for every CI run -- should probably figure out how to fix VCR set up? since if it breaks on individual import additions it'll mostly get in the way?

eyurtsev: Also `deploy_docs` currently runs on every PR and does some checks (i.e., attempts to build the documentation) -- so adding stuff to CI will now mean that there are two different workflows that check documentation

vbarda: yea could potentially be moved there

PR #1942, opened 2024-10-01
langgraph, checkpoint: update versions

vbarda:
Postgres will follow, need to publish `checkpoint` first


hinthornw (approved):

libs/langgraph/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "langgraph"
-version = "0.2.31"
+version = "0.2.29"
```
hinthornw, line 3: This needs changed right?

vbarda: yea if we wanna publish right away

PR #1948, opened 2024-10-01
Update PG Implementation

hinthornw:
LTree supports too small a subset of labels (especially pre-postgres 16) and provides very little benefits over text columns (only improves less common namespace listing operations)

Rewrite to use a regular text column with a btree here.

### DIFF
--- a/libs/checkpoint-postgres/Makefile
+++ b/libs/checkpoint-postgres/Makefile
@@ -16,10 +16,10 @@ test:
 	EXIT_CODE=$$?; \
 	make stop-postgres; \
 	exit $$EXIT_CODE
-
+TEST ?= .
 test_watch:
 	make start-postgres; \
-	poetry run ptw .; \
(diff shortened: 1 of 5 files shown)


libs/checkpoint-postgres/Makefile
```diff
@@ -16,10 +16,10 @@ test:
 	EXIT_CODE=$$?; \
 	make stop-postgres; \
 	exit $$EXIT_CODE
-
+TEST ?= .
 test_watch:
 	make start-postgres; \
```
vbarda, line 21: would it make sense to test on both versions of postgres (15 and 16)? or is it an overkill?

hinthornw: hm - maybe a good idea

hinthornw: I think better for sep PR though

PR #1955, opened 2024-10-01
docs: update tutorial for long-term memory

vbarda:


hinthornw:
Brilliant. Will review now

docs/docs/tutorials/memory/long_term_memory_agent.ipynb
hinthornw, line 1: 1. For core_memories, we should fetch that from the store in the node (rather than saving to state) to emphasize that it's NOT thread-scoped
2. Hmmmmmmmmmmmmmmmmm - I think we should add support for injecting the store in the tool node for this since in the cloud deployment, you wouldn't be creating the memory store outside the graph and calling - you'd be accepting it within a node. 

Gonna mark that as a TODO. 

I don't think (2) needs to block this initial PR - will add to release checklist

vbarda: yup (1) sounds good -- will address!

vbarda: (and will add support for (2)!)

hinthornw (approved):

PR #1958, opened 2024-10-02
Shared state conceptual docs

rlancemartin:


docs/docs/concepts/persistence.md
```diff
@@ -216,6 +216,106 @@ The final thing you can optionally specify when calling `update_state` is `as_no
 
 ![Update](img/persistence/checkpoints_full_story.jpg)
 
+## Memory Store
```
hwchase17, line 219: i think this is missing some description of what is returned from `store.search`. eg we access a `value` attribute... what??? what is that?

docs/docs/concepts/memory.md
```diff
@@ -160,7 +160,9 @@ config = {"configurable": {"thread_id": "1"}}
...
+Persistence is critical sustaining a long-running chat sessions. For example, a chat between a user and an AI assistant may have interruptions. Persistence ensures that a user can continue that particular chat session at any later point in time. However, what happens if a user initiates a new chat session with an assistant? This spawns a new thread, and the information from the previous session (thread) is not retained. This motivates the need for a memory that can maintain data across chat sessions (threads). 
+
+For this, we can use LangGraph's `Store` interface to save and retrieve information across threads. Shared information can be namespaced by, for example, `user_id` to retain user-specific information across threads. See more detail in the [persistence conceptual guide](https://langchain-ai.github.io/langgraph/concepts/persistence/#persistence) and this [how-to guide on shared state](../how-tos/memory/shared-state.ipynb).
```
hwchase17, line 165: if you have a code snippet for the checkpointer, i would do the same for here

also, now that we have this, i would add more concrete examples for the two below

also id maybe rename "metaprompting" to "update own instructions" or something?

docs/docs/concepts/memory.md
```diff
@@ -160,9 +160,77 @@ config = {"configurable": {"thread_id": "1"}}
...
+Persistence is critical sustaining a long-running chat sessions. For example, a chat between a user and an AI assistant may have interruptions. Persistence ensures that a user can continue that particular chat session at any later point in time. However, what happens if a user initiates a new chat session with an assistant? This spawns a new thread, and the information from the previous session (thread) is not retained. This motivates the need for a memory that can maintain data across chat sessions (threads). 
 
-## Meta-prompting
+For this, we can use LangGraph's `Store` interface to save and retrieve information across threads. Shared information can be namespaced by, for example, `user_id` to retain user-specific information across threads. Let's show how to use the `Store` interface to save and retrieve information.
```
eyurtsev, line 165: Do we have an API reference that we can link `Store` interface to?

docs/docs/concepts/memory.md
```diff
@@ -160,9 +160,77 @@ config = {"configurable": {"thread_id": "1"}}
...
+```python
+from langgraph.store.memory import InMemoryStore
+in_memory_store = InMemoryStore()
+
+# Namespace for memories
+user_id = "1"
+namespace_for_memory = (user_id, "memories")
+
+# Save memories 
+memory_id = str(uuid.uuid4())
```
eyurtsev, line 176: If there's a single sentence answer  to the question why is the key not a tuple? Why has the key been broken into a namespace (tuple) and a string.

docs/docs/concepts/persistence.md
```diff
@@ -216,6 +216,139 @@ The final thing you can optionally specify when calling `update_state` is `as_no
...
+
+With checkpointers alone, we cannot share information across threads. This motivates the need for the `Store` interface. As an illustration, we can define an `InMemoryStore` to store information about a user across threads. We simply compile our graph with a checkpointer, as before, and will our new in_memory_store.
+First, let's showcase this in isolation without using LangGraph.
+
+```python
+from langgraph.store.memory import InMemoryStore
+in_memory_store = InMemoryStore()
+```
+
+Memories are namespaced by a `tuple`, which in our case will be `(<user_id>, "memories")`. We can think about this namespace as a directory, where each `user_id` can have various sub-directories of things that we want to store (e.g., `memories`, `preferences`, etc.).
```
eyurtsev, line 235: Is the first key in the tuple required to be `user_id`? Can the tuple be of arbitrary length?

docs/docs/concepts/persistence.md
```diff
@@ -216,6 +216,139 @@ The final thing you can optionally specify when calling `update_state` is `as_no
...
+```
+
+Memories are namespaced by a `tuple`, which in our case will be `(<user_id>, "memories")`. We can think about this namespace as a directory, where each `user_id` can have various sub-directories of things that we want to store (e.g., `memories`, `preferences`, etc.). 
+
+```python 
+user_id = "1"
+namespace_for_memory = (user_id, "memories")
+```
+
+We use the `store.put` to save memories to our namespace in the store. When we do this, we specify the namespace, as defined above, and a key-value pair for the memory: the key is simply a unique identifier for the memory (`memory_id`) and the value (a dictionary) is the memory itself.
```
eyurtsev, line 242: The full key words out to be:

```python
(user_id, "memories", "some_random_uuid", "food_preference")
```

Does structuring memories as:

```python
{
   'type': "food_preference",
  "value": "I like pizza"
}
```

work with respect to the memory store API? (e.g., does search work as expected etc.)


cc @hinthornw

docs/docs/concepts/persistence.md
```diff
@@ -216,6 +216,139 @@ The final thing you can optionally specify when calling `update_state` is `as_no
...
+
+```python
+memory_id = str(uuid.uuid4())
+memory = {"food_preference" : "I like pizza"}
+in_memory_store.put(namespace_for_memory, memory_id, memory)
+```
+
+We can read out memories in our namespace using `store.search`, which will return all memories for a given user as a list. The most recent memory is the last in the list.
```
eyurtsev, line 250: * This looks like a `list` API. Does `search` support queries?
* Is it all memories or is there a page size?

eyurtsev:
@rlancemartin I think this is looking great! Just a few nits and tagged @hinthornw in a few places for clarification on API

docs/docs/concepts/persistence.md
```diff
@@ -216,6 +216,139 @@ The final thing you can optionally specify when calling `update_state` is `as_no
...
+
+If we create a new thread, we can still access the same memories so long as the `user_id` is the same. 
+
+```python
+# Invoke the graph
+config = {"configurable": {"thread_id": "2", "user_id": "1"}}
+
+# Let's say hi again
+for update in graph.stream(
+    {"messages": [{"role": "user", "content": "hi, tell me about my memories"}]}, config, stream_mode="updates"
```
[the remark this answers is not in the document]
hinthornw: I just ran into mypy anger from this type of thing where BaseMessgae isn't included either.

I think we can use `from langgraph.graph.mesages import Message`

docs/docs/concepts/persistence.md
hinthornw: not yet

docs/docs/concepts/persistence.md
hinthornw: You can either put food preference in the namespace (so you can search in that directory) or as a key and do filter of {"type", "food_preference"}

More flexible if it's in the body

docs/docs/concepts/persistence.md
hinthornw: I think the tuple can go up to like 1000 labels? Depends on the DB restrictions.

No the keys can be anything. Main consideration is how you want to be querying, since it's easier to do filters by namespace than content right now.

docs/docs/concepts/memory.md
```diff
@@ -160,9 +160,77 @@ config = {"configurable": {"thread_id": "1"}}
...
+memory = {"food_preference" : "I like pizza"}
+in_memory_store.put(namespace_for_memory, memory_id, memory)
+
+# Retrieve memories
+memories = in_memory_store.search(namespace_for_memory)
```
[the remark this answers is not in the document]
hinthornw: it always returns a list of items. Default is listing

docs/docs/concepts/memory.md
```diff
@@ -132,44 +155,124 @@ def summarize_conversation(state: State):
...
+**Do you want to update memory "in the hot path" or "as a separate process"**
 
-See this how-to [video](https://www.youtube.com/watch?v=37VaU7e7t5o) for example usage of dynamic few-shot example selection in LangSmith. Also, see this [blog post](https://blog.langchain.dev/few-shot-prompting-to-improve-tool-calling-performance/) showcasing few-shot prompting to improve tool calling performance and this [blog post](https://blog.langchain.dev/aligning-llm-as-a-judge-with-human-preferences/) using few-shot example to align an LLMs to human preferences.
+Memory can be updated either as part of the application logic (e.g. "in the hot path" of the application) or as a separate process.
```
eyurtsev, line 171: Why do we call it "separate process" vs. "in parallel" (vs. in the background)

Is the fact that it's a process vs. a thread important or just an implementation detail?

docs/docs/concepts/memory.md
```diff
@@ -132,44 +155,124 @@ def summarize_conversation(state: State):
...
 
 Meta-prompting can use past information to update the prompt. As an example, this [Tweet generator](https://www.youtube.com/watch?v=Vn8A3BxfplE) uses meta-prompting to iteratively improve the summarization prompt used to generate high quality paper summaries for Twitter. In this case, we used a LangSmith dataset to house several papers that we wanted to summarize, generated summaries using a naive summarization prompt, manually reviewed the summaries, captured feedback from human review using the LangSmith Annotation Queue, and passed this feedback to a chat model to re-generate the summarization prompt. The process was repeated in a loop until the summaries met our criteria in human review.
 
-## Retrieving relevant information from long-term storage
+This will utilize the memory store concept above to store the updated instructions in a shared namespace. This namespace will have only a single item (unless you want to update instructions specific for each user, but that's a separate issue). This will look something like:
+
+```python
+# Node that *uses* the instructions
+def call_model(state: State, store: BaseStore):
+    instructions = store.search(("instructions",))[0]
```
eyurtsev, line 233: nit: `("instructions",)` -- may trip users new to python. Should the API support taking either a list or else handling lifting a string to a tuple of strings?

docs/docs/concepts/memory.md
```diff
@@ -132,44 +155,124 @@ def summarize_conversation(state: State):
...
+    # Memory logic
+    prompt = prompt_template.format(instructions=instructions.value["instructions"], conversation=state["messages"])
+    output = llm.invoke(prompt)
+    new_instructions = output['new_instructions']
+    store.put(("instructions",), current_instructions.key, {"instructions": new_instructions})
+    ...
+```
+
+![](img/memory/update-instructions.png)
+### Remember a profile
```
eyurtsev, line 251: nit: Remember vs. learn? (for foreign speakers remember and recall can be synonymous)

docs/docs/concepts/memory.md
```diff
@@ -132,44 +155,124 @@ def summarize_conversation(state: State):
...
+
+The profile is generally just a JSON blob with various key-value pairs. When remembering a profile, you will want to make sure that you are **updating** the profile each time. As a result, you will want to pass in the previous profile and ask the LLM to generate a new profile (or some JSON patch to apply to the old profile).
+
+If the profile is large, this can get tricky. You may need to split the profile into subsections and update each one individually. You also may need to fix errors if the LLM generates incorrect JSON.
+
+![](img/memory/update-profile.png)
+### Remember a list
```
eyurtsev, line 258: nit: list of information => multiple facts? Is there any other way of phrasing this?

It feels like we're naming the concept after a data structure rather than the concept that we use it to represent

docs/docs/concepts/memory.md
```diff
@@ -132,44 +155,124 @@ def summarize_conversation(state: State):
...
+
+This has several benefits. First of all, it happens realtime, so if the user starts a new thread right away that memory will be present. It also makes it possible to show the user that memory has been updated, and so can be a bit more transparent that way.
 
-Persistence is critical sustaining a long-running chat sessions. For example, a chat between a user and an AI assistant may have interruptions. Persistence ensures that a user can continue that particular chat session at any later point in time. However, what happens if a user initiates a new chat session with an assistant? This spawns a new thread, and the information from the previous session (thread) is not retained. This motivates the need for a memory service that can maintain data across chat sessions (threads). 
+This also has several downsides. It may slow down the final response since it needs to decide what to remember. It also means your application not only needs to think about the application logic, but also what to remember (which could result in more complicated instructions to the LLM).
```
eyurtsev, line 211: `but also what to remember (which could result in more complicated instructions to the LLM).`

Is this issue not at least partially shared by the application logic? In this case it's a bit less about what to remember, and more about what to recall or how to incorporate memories into the conversation. But both the learn and the recall path have to agree on the schema of memories

eyurtsev (approved):

docs/docs/concepts/memory.md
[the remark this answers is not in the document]
eyurtsev: 1) What is the order by if there's pagination without query?
2) Is it possible for client code to see the same record twice during pagination? (e.g., probably yes if memory is being updated from a separate conversation -- is it possible for this to occur if there's only one conversation going on?)

docs/docs/concepts/memory.md
```diff
@@ -2,23 +2,44 @@
...
+
+Short-term memory primarily refers to the ability of an application to accurately remember interactions that previously occurred in the same conversation with the user.
+Most conversations are represented as a list of messages and so the best first thing to do is to store a list of those messages and pass those in to future invocations of the LLM.
+LangGraph's [persistence layer](persistence.md#persistence) has checkpointers that enable [thread](persistence.md#threads)-level memory.
+
+As the conversation grows in length, you will need to think about how to manage that list of messages.
+The reason you will need to think about this is that too many messages will either (a) not fit inside an LLMs context window and will throw an error, or (b) will "overwhelm" the LLM and cause it to perform poorly.
+Therefore, it becomes crucial to think about what parts (or representations) of the conversation to pass to the LLM.
+
+**Note:** we say "messages" and "conversation" above, but short-term memory is not limited to conversations.
```
vbarda, line 26: i would make this more prominent with an admonition. that was one of my questions when i was reading the above

docs/docs/concepts/memory.md
hinthornw: @eyurtsev what about `namespace = ("instructions",)` ?
It's a directory/namespace/file path basically

docs/docs/concepts/memory.md
hinthornw: it's actually a sep thread not process; mainly conceptually

vbarda: no, the difference is that it's either part of the app or a call to update from outside of the app logic, worth clarifying this language

docs/docs/concepts/memory.md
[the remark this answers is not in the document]
hinthornw: It should be by "updated_at" by default right now, though we haven't made an acceptance test suite for community stores so in theory they could make a non-compliant one

docs/docs/concepts/memory.md
hinthornw: Kinda like files and folders, but there's not a super strong distinction in some cases

PR #1959, opened 2024-10-02
docs: clean up and standardize more how-tos

vbarda:


eyurtsev:
fwiw i've been placing the concept link first thing immediately after the title

PR #1968, opened 2024-10-02
langgraph: add support for store in ToolNode

vbarda:


libs/langgraph/langgraph/utils/runnable.py
```diff
@@ -71,7 +71,7 @@ class StrEnum(str, enum.Enum):
         sys.intern("store"),
         (BaseStore, "BaseStore", inspect.Parameter.empty),
         CONFIG_KEY_STORE,
-        inspect.Parameter.empty,
+        None,
```
vbarda, line 74: @nfcampos i am a bit concerned about this one, but without this change you would get
```python
ValueError: Missing required config key '__pregel_store' for 'tools'.
```
when running `tool_node.invoke()` without passing store kwarg

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -391,23 +425,78 @@ def __init__(self, field: Optional[str] = None) -> None:
...
+        from langchain_core.tools import tool
+
+        from langgraph.store.memory import InMemoryStore
+        from langgraph.prebuilt import InjectedStore, ToolNode
+
+        store = InMemoryStore()
+        store.put(("values",), "foo", {"bar": 2})
+
+        @tool
+        def store_tool(x: int, my_store: Annotated[Any, InjectedStore()]) -> str:
```
hinthornw, line 453: Does it have to be `Any`?
Thoughts on it being

```
Annotated[BaseStore,  InjectedToolArg]
```

(so we don't require duplicate annotation type things?)

libs/langgraph/langgraph/utils/runnable.py
hinthornw: Right now I think raising an error is good actually?  

Would be better to have a specific user-facing one though. Maybe even handle better in the tool node? (like saying "please invoke within a graph where you've provided a store" or something of that nature

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -223,8 +232,6 @@ def _inject_state(
             BaseModel,
         ],
     ) -> ToolCall:
-        if tool_call["name"] not in self.tools_by_name:
-            return tool_call
         state_args = _get_state_args(self.tools_by_name[tool_call["name"]])
```
hinthornw, line 235: We should be doing this once in the constructor and caching so we don't have to do all this meta-programming every time the node is called

vbarda: nice, good idea

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: let me improve

vbarda: there is an issue in `core`, will update once fixed up

hinthornw: 🉑

hinthornw (approved):

PR #1974, opened 2024-10-02
[Docs] Add SDK ref docs for the Store client

hinthornw:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -151,11 +183,20 @@ def __init__(self, client: httpx.AsyncClient) -> None:
...
+    Attributes:
+        client (httpx.AsyncClient): Underlying HTTPX async client.
+    """
+
     def __init__(self, client: httpx.AsyncClient) -> None:
         self.client = client
 
     async def get(self, path: str, *, params: Optional[QueryParamTypes] = None) -> Any:
-        """Make a GET request."""
+        """Send a GET request."""
```
vbarda, line 199: (nit) probably need to update this for sync as well

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -1,3 +1,11 @@
+"""The LangGraph client implementations connect to the LangGraph API.
```
vbarda, line 1: let's add a sentence that there is both sync and async in this file?

docs/docs/cloud/reference/sdk/python_sdk_ref.md
```diff
@@ -1,79 +1,4 @@
...
-```python
-from langgraph_sdk import get_client
-
-client = get_client(url="http://localhost:8123")
-await client.crons.<method_name>()
-```
-
-::: langgraph_sdk.client.CronClient
-    handler: python
+::: langgraph_sdk.client
```
vbarda, line 3: nice!

vbarda (approved):

PR #1984, opened 2024-10-02
docs: update how-to for passing runtime values

vbarda:


docs/docs/how-tos/pass-run-time-values-to-tools.ipynb
hinthornw, line 1: > Sometimes we need to pass values to our tools at runtime. For example, you might want to give tools access to:

I think language like this may contribute to people thinking you "have to" use langchain, etc. 

I don't have a perfect ording but all this is more like:

"Tools" are functions you've exposed to the model. Sometimes, you want to let the LLM populate only a subset of the functions' arguments; your program can provide the rest. If you're using LangChain-style tools, an easy way to handle this is by annotating function parameters with `InjectedArg`. This excludes that parameter from being shown to the LLM.

If you're using LangGraph's prebuilt ToolNode, we have a couple additional special `InejectedArg` types you can use to tell the graph to populate the tool with values from the state, store, or configuration.

hinthornw:
image

PR #2000, opened 2024-10-03
docs: clean up memory docs

vbarda:


docs/docs/how-tos/memory/shared-state.ipynb
hinthornw, line 1: Skimming this, I actually now think it's not great that we pre-populate the store. I don't think this is how it'll be used in practice, and in the studio/cloud setup people won't be sure what to do. 
What if we do something silly like "if 'remember' in state['messages'][-1].content: put'

- I think worth naming a "namespace" variable; then I read the code and think "namespace", and I can search "namespace" in the docs and get related things:
```
user_id = config["configurable"].get("user_id")
namespace = ("memories", user_id)
memories = store.search(namespace)
```

- I know we say this already, but for "how-tos", I (and others) often don't read, so I think it is good to say it's unnecessary in studio/Cloud inline the first code block:
```python
# NOTE: we're passing the store object here when compiling the graph
graph = builder.compile(checkpointer=MemorySaver(), store=in_memory_store)
# In LangGraph Studio & LangGraph Cloud, the store is provided automatically;
# no need to manually define above
```

- I think we shouldn't import message types just do `{"role": "system", "content": ...}`

docs/docs/how-tos/persistence.ipynb
hinthornw, line 1: - When creating any LangGraph graph, you can set it up to persist its state by adding a checkpointer when compiling the graph. 
I think in the code snippet 
- I don't like importing message classes but if we're doing it everyehwere then I suppose we should remain consistent

docs/docs/how-tos/index.md
```diff
@@ -18,17 +18,22 @@ These how-to guides show how to achieve that controllability.
...
-LangGraph makes it easy to persist state across graph runs. The guide below shows how to add persistence to your graph.
+LangGraph makes it easy to persist state across graph runs (thread-level persistence) and across threads (cross-thread persistence). These how-to guides show how to add persistence to your graph.
 
-- [How to add persistence ("memory") to your graph](persistence.ipynb)
-- [How to manage conversation history](memory/manage-conversation-history.ipynb)
-- [How to delete messages](memory/delete-messages.ipynb)
-- [How to add summary conversation memory](memory/add-summary-conversation-history.ipynb)
-- [How to share state between threads](memory/shared-state.ipynb)
+- [How to add thread-level persistence to your graph](persistence.ipynb)
+- [How to add cross-thread persistence to your graph](memory/shared-state.ipynb)
```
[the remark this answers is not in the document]
vbarda: do you think it should live under memory?

docs/docs/how-tos/memory/shared-state.ipynb
vbarda: sounds good -- will update

docs/docs/how-tos/index.md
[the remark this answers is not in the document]
hinthornw: I think we could put under /persistence

docs/docs/how-tos/memory/shared-state.ipynb
hinthornw, line 1: Oh I was even going to not use structured output and just put the raw content in if they say "remember"

Issue #1950, opened 2024-10-01
Passing private state with Class Node


b-yukky:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
from typing import Annotated
from langgraph.graph import StateGraph, START, END
from typing_extensions import TypedDict


# The overall state of the graph
class OverallState(TypedDict):
    question: str
    answer: str


# This is what the node that generates the query will return
class QueryOutputState(TypedDict):
    query: str


# This is what the node that retrieves the documents will return
class DocumentOutputState(TypedDict):
    docs: list[str]


# This is what the node that generates the final answer will take in
class GenerateInputState(OverallState, DocumentOutputState):
    pass


# Node to generate query
def generate_query(state: OverallState) -> QueryOutputState:
    # Replace this with real logic
    return {"query": state["question"][:2]}


class RetrieveDocumentsNode:
    def __init__(self):
        self.docs = []

    def __call__(self, state: QueryOutputState) -> DocumentOutputState:
        self.docs = [state["query"]] * 2
        return {"docs": self.docs}


# Node to generate answer
def generate(state: GenerateInputState) -> OverallState:
    return {"answer": "\n\n".join(state["docs"] + [state["question"]])}


graph = StateGraph(OverallState)
graph.add_node(generate_query)
graph.add_node("retrieve_documents", RetrieveDocumentsNode())
graph.add_node(generate)
graph.add_edge(START, "generate_query")
graph.add_edge("generate_query", "retrieve_documents")
graph.add_edge("retrieve_documents", "generate")
graph.add_edge("generate", END)
graph = graph.compile()

graph.invoke({"question": "foo"})
```

### Error Message and Stack Trace (if applicable)

```shell
{
	"name": "InvalidUpdateError",
	"message": "Must write to at least one of ['question', 'answer', 'docs']",
	"stack": "---------------------------------------------------------------------------
InvalidUpdateError                        Traceback (most recent call last)
Cell In[14], line 57
[...]

PR #2014, opened 2024-10-05
Support read type hints from the method in `add_node`

gbaian10:
- #1950

Add function to read type hints from the `__call__` method to resolve issue #1950.


libs/langgraph/langgraph/graph/state.py
```diff
@@ -830,3 +819,23 @@ def _get_schema(
...
                 },
             )
+
+
+def _get_input_schema_from_type_hint(
+    action: Optional[RunnableLike],
+) -> Optional[Type[Any]]:
+    try:
+        if not hasattr(action, "__call__"):
+            raise AttributeError
```
vbarda, line 829: could you populate these with descriptive error messages?

vbarda:
Looks good, thanks for adding the support!

libs/langgraph/langgraph/graph/state.py
gbaian10: Hi @vbarda, You're right.
I adjusted the code into the following two versions.
Is it better than the original? Which style do you prefer?
Both have the same logic, with only slight differences.

```py
def _get_input_schema_from_type_hint(
    action: Optional[RunnableLike],
) -> Optional[Type[Any]]:
    if not isfunction(action) and not ismethod(getattr(action, "__call__", None)):
        return None
    action = cast(Callable, action)

    hints = get_type_hints(getattr(action, "__call__")) or get_type_hints(action)
    try:
        first_parameter_name = next(iter(inspect.signature(action).parameters.keys()))
    except StopIteration:
        return None
    else:
        input_hint = hints.get(first_parameter_name)
        if isinstance(input_hint, type) and get_type_hints(input_hint):
            return input_hint
    return None
```

---

```py
def _get_input_schema_from_type_hint(
    action: Optional[RunnableLike],
) -> Optional[Type[Any]]:
    if not isfunction(action) and not ismethod(getattr(action, "__call__", None)):
        return None
    action = cast(Callable, action)

    hints = get_type_hints(getattr(action, "__call__")) or get_type_hints(action)
    try:
        first_parameter_name = next(iter(inspect.signature(action).parameters.keys()))
        input_hint = hints.get(first_parameter_name)
        if isinstance(input_hint, type) and get_type_hints(input_hint):
            return input_hint
    except StopIteration:
        pass        
    return None
```

vbarda: thank you -- personal preference for the second one :)

gbaian10: Updated!

PR #2027, opened 2024-10-07
Fix: is last step logic to run per execution

abprime:
Use both step and stop value to decide the last step (per execution)

Current logic for last step checks the step number against recursion_limit in the config.
The step number keeps incrementing every execution, so it won't set the is_last_step correctly every execution.

Discussion on the same:
https://github.com/langchain-ai/langgraph/discussions/2025


abprime:
Hi @nfcampos , 
Could you please review this and let me know your feedback. 
I was able to find this solution for the issue.

abprime:
Hi @hinthornw ,

could you please take a look at this PR and give your feedback?

antonioalegria:
May be worthwhile to open an issue from that discussion.

abprime:
https://github.com/langchain-ai/langgraph/issues/2081

nfcampos:
Thanks, we can't modify the checkpoint format for this purpose, fixed in a different way in #2105

PR #2034, opened 2024-10-07
Implement `RemotePregel` class and `PregelProtocol`

andrewnguonly:
### Summary
This PR is a follow-up/reimplementation of [this PR](https://github.com/langchain-ai/langgraph/pull/1541).

### To Do
- [x] Add tests
- [x] Fix `mypy` errors


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -0,0 +1,360 @@
+from typing import (
```
vbarda, line 1: should this live in the SDK?

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -0,0 +1,360 @@
...
+        self.graph_id = graph_id
+
+    def copy(self, update: dict[str, Any]) -> Self:
+        attrs = {**self.__dict__, **update}
+        return self.__class__(**attrs)
+
+    def with_config(
+        self, config: Optional[RunnableConfig] = None, **kwargs: Any
+    ) -> Self:
+        return self.copy({})
```
vbarda, line 45: this seems a bit strange

libs/langgraph/langgraph/pregel/remote.py
andrewnguonly: I didn't consider it, but that's a fair thought.

1. If I think of `RemotePregel` as a client of the SDK, then it makes sense to live in the SDK.
2. If I think of `RemotePregel` as a primitive of `langgraph` or as a "prebuilt" (kind of like a `ToolNode`), then it feels more appropriate to live in `langgraph` and using the SDK is just an implementation detail. For example, we could just use vanilla `requests` to make the remote API calls.

I don't have a strong opinion. I could go either way.

libs/langgraph/langgraph/pregel/remote.py
andrewnguonly: I wasn't sure what to do here because there's no concept of "cloning" a remote Pregel instance (i.e. creating a new Assistant which is a clone of the current one).

To "clone" or "copy" an Assistant, we have to...
1. Call `GET /assistants/<graph_id>`.
1. Update the `configs` (via merge).
1. Call `POST /assistants` to a create a new assistant, the clone/copy (this endpoint doesn't exist right now).

libs/langgraph/langgraph/pregel/remote.py
nfcampos: I think it needs to be in langgraph

libs/langgraph/langgraph/pregel/remote.py
nfcampos: I don't think you need to call any endpoints, just save the config in the instance and then use it on every other method?

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -0,0 +1,360 @@
...
+        for task in state["tasks"]:
+            interrupts = []
+            for interrupt in task["interrupts"]:
+                interrupts.append(Interrupt(**interrupt))
+
+            tasks.append(
+                PregelTask(
+                    id=task["id"],
+                    name=task["name"],
+                    path=tuple(),
```
nfcampos, line 136: why empty tuple?

andrewnguonly: I was mapping the fields from `ThreadTask` to `PregelTask`, but there's no `path` field in `ThreadTask`.
```python
class ThreadTask(TypedDict):
    """Represents a task within a thread."""

    id: str
    name: str
    error: Optional[str]
    interrupts: list[dict]
    checkpoint: Optional[Checkpoint]
    state: Optional["ThreadState"]


class PregelTask(NamedTuple):
    id: str
    name: str
    path: tuple[Union[str, int], ...]
    error: Optional[Exception] = None
    interrupts: tuple[Interrupt, ...] = ()
    state: Union[None, RunnableConfig, "StateSnapshot"] = None
```

What's supposed to be in `path` or how do we get the `path`?

libs/langgraph/langgraph/pregel/remote.py
andrewnguonly: > I don't think you need to call any endpoints, just save the config in the instance and then use it on every other method?

Using `aupdate_state()` as an example... What do we do with `config` passed in the method and `self.config` on the instance. Merge them?

```python
 async def aupdate_state(
    self,
    config: RunnableConfig,
    values: Optional[Union[dict[str, Any], Any]],
    as_node: Optional[str] = None,
) -> RunnableConfig:

    # what do we do with self.config and config?
    self.config

    response: dict = await self.client.threads.update_state(  # type: ignore
        thread_id=config["configurable"]["thread_id"],
        values=values,  # type: ignore
        as_node=as_node,
        checkpoint=self._get_checkpoint(config),
    )
    return self._get_config(response["checkpoint"])
 ```

Comment (nfcampos) [reply libs/langgraph/langgraph/pregel/remote.py:45]: yes exactly merge them. exactly what the real pregel class does

File: libs/langgraph/langgraph/pregel/remote.py (hunk 3 above)
Comment (nfcampos) [reply libs/langgraph/langgraph/pregel/remote.py:136]: ah, because its an internal thing I removed from the api, I guess now that comes back to bite me

File: libs/langgraph/langgraph/pregel/remote.py [hunk 4]
```diff
@@ -0,0 +1,458 @@
...
+        self.config = config
+
+    def copy(self, update: dict[str, Any]) -> Self:
+        attrs = {**self.__dict__, **update}
+        return self.__class__(**attrs)
+
+    def with_config(
+        self, config: Optional[RunnableConfig] = None, **kwargs: Any
+    ) -> Self:
+        self.config = config
```
nfcampos, line 53: can we return a copy instead?

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -0,0 +1,458 @@
...
+            xray=xray,
+        )
+
+        # construct DrawableNodes
+        nodes = {}
+        for node in graph["nodes"]:
+            node_id = node["id"]
+            nodes[node_id] = DrawableNode(
+                id=node_id,
+                name=node.get("name", node_id),
```
nfcampos, line 100: why do we pass this node_id default here?

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -0,0 +1,458 @@
...
+
+class RemotePregel(PregelProtocol, Runnable):
+    def __init__(
+        self,
+        client: LangGraphClient,
+        sync_client: SyncLangGraphClient,
+        graph_id: str,
+        config: Optional[RunnableConfig] = None,
+    ):
+        self.client = client
```
nfcampos, line 41: is there some way we can accept a single client and build the other one from it?, or make only one of them required at least

nfcampos: Most people will be using only sync or only async

libs/langgraph/langgraph/pregel/remote.py
andrewnguonly: I just needed to pass some default value. I can set the default to an empty string. `name` in `DrawableNode` is not `Optional`.

PR #2044, opened 2024-10-08
Update name to InMemorySaver

hinthornw:
(Keep MemorySaver around for backwards compatibility)

"MemorySaver" is ambiguous: is it saving memories? Where is it saving memories to?

InMemorySaver aligns naming InMemoryStore as well as similar LangChain objects (InMemoryVectorStore, etc.)


eyurtsev:
InMemorySaver is much better

PR #2046, opened 2024-10-08
Handle pandas DataFrame conversion to JSON-friendly format

zoltankoos:
This pull request enhances the `JsonPlusSerializer` class to handle the serialization and deserialization of `pandas.DataFrame` objects. Previously, the class could not serialize `pandas.DataFrame` objects. 

This update improves support for data science workflows where `pandas.DataFrame` objects need to be serialized or passed between different components.

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -3,6 +3,7 @@ import decimal
 import importlib
 import json
 import pathlib
+import pandas as pd
 import re
 from collections import deque
 from datetime import date, datetime, time, timedelta, timezone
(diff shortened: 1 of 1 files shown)


vbarda:
@zoltankoos thanks for your contribution! we're planning to add support for numpy and pandas serialization but the solution will be a bit different from the one you offered with `.to_dict()` -- we will be serializing to a binary format and we would like to avoid an external dependency on pandas. going to close this PR for now and will update here once we release the version with the support!

zoltankoos:
Thank you for your thoughtful response and for considering my contribution! I completely understand the approach you're planning to take with binary serialization and the goal to minimize dependencies. I look forward to seeing the future release with the support, and appreciate the update! 😊

PR #2048, opened 2024-10-08
feat(debug): send tasks info

dqbd:


libs/langgraph/langgraph/pregel/utils.py
```diff
@@ -17,3 +23,29 @@ def get_new_channel_versions(
         new_versions = current_versions
 
     return new_versions
+
+
+def find_subgraph_pregel(candidate: Runnable) -> Optional[Runnable]:
+    candidates: list[Runnable] = [candidate]
+
+    for c in candidates:
+        # Cannot do isinstance(c, Pregel) due to circular imports
```
nfcampos, line 32: you can import Pregel inside the function?

dqbd: ah nice

PR #2054, opened 2024-10-08
sdk-py: Add `Sequence[dict]` type to `values` param type for `update_state()`

andrewnguonly:
### Summary
The LangGraph API supports a list of `dict` for the `values` field for the `POST /threads/<thread_id>/state` endpoint.

Reference: https://github.com/langchain-ai/langgraph-api/blob/main/api/openapi.json#L3005-L3021

### DIFF
--- a/libs/sdk-py/langgraph_sdk/client.py
+++ b/libs/sdk-py/langgraph_sdk/client.py
@@ -1056,7 +1056,7 @@ class ThreadsClient:
     async def update_state(
         self,
         thread_id: str,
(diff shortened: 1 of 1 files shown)


andrewnguonly:
I don't think this will resolve the `mypy` errors with the `RemotePregel` interface, but at least it's more precise with respect to the API implementation.

The `RemotePregel` interface accepts `Any` for `values`.

vbarda (approved):

PR #2071, opened 2024-10-10
fix(docs): Small nits & typo fixes

bracesproul:


docs/docs/concepts/human_in_the_loop.md
```diff
@@ -120,7 +120,7 @@ See [our guide](../how-tos/human_in_the_loop/breakpoints.ipynb) for a detailed h
 
 Sometimes we want to review and edit the agent's state. 
  
-As with approval, we can interrupt our agent at a [breakpoint](./low_level.md#breakpoints) prior the the step we want to check. 
+As with approval, we can interrupt our agent at a [breakpoint](./low_level.md#breakpoints) prior the step we want to check.
```
vbarda, line 123:
```suggestion
As with approval, we can interrupt our agent at a [breakpoint](./low_level.md#breakpoints) prior to the step we want to check. 
```

docs/docs/concepts/persistence.md
```diff
@@ -259,16 +259,16 @@ memories[-1].dict()
  'updated_at': '2024-10-02T17:22:31.590605+00:00'}
 ```
 
-Each memory type is a Python class with certain attributes. We can access it as a dictionary by converting via `.dict` as above.
+Each memory type is a Python class ([`Item`](https://langchain-ai.github.io/langgraph/cloud/reference/sdk/python_sdk_ref/?h=item#langgraph_sdk.schema.Item)) with certain attributes. We can access it as a dictionary by converting via `.dict` as above.
```
vbarda, line 262: this should link to the regular `Item` here https://langchain-ai.github.io/langgraph/reference/store/#langgraph.store.base.Item, not the SDK Item

nfcampos (approved):

PR #2079, opened 2024-10-11
docs: update subgraph how-to

isahers1:


eyurtsev:
0. Could you add the answer to the very top of the how-to guide before creating an example? Can be in the form of a sentence or a short code snippet if it's short.

My understanding is that to add persistence you use the `checkpointer` as you always did, but you need to only add it to the root level graph.

The snippet below is the most important information for the entire how-to guide, and should be mentioned at the very top of the guide as the answer

```python
checkpointer = MemorySaver()
# You must only pass checkpointer when compiling the parent graph.
# LangGraph will automatically propagate the checkpointer to the child subgraphs.
!!! note You shouldn't provide a checkpointer when compiling a subgraph. Instead, you must define a single checkpointer that you pass to parent_graph.compile(), and LangGraph will automatically propagate the checkpointer to the child subgraphs.
```

1. The guide should help me to (a) add persistence to the subgraphs, (b) confirm that we have persistence. 

To confirm that we have persistence, is "view subgraph state" the best way to do this? I think we need 1-2 sentences the explain what we're doing (i.e., why are we viewing subgraph state) and why it confirms that we have persistence.

-----

nits:

5. Could you add a link here to the API reference for stream and explain what "subgraphs" does?
```python
for chunk in graph.stream({"foo": "foo"}, config, subgraphs=True):
```


6. In the first cell, consider moving all the sub-graph logic to the top, and the main graph logic to the bottom (e.g., including the `State` schema). So we don't see anything associated w/ the main graph until after the sub-graph has been fully defined. And maybe including a comment `######### Using as a sub-graph inside another graph #### ` (or something else, but just an inline comment that servers as a horizontal separator). An alternative is splitting into 2 code blocks.

7. Is this the best way to illustrate flow? Maybe a sentence explanation about what we're seeing in the output or what we should be seeing?

image

8. For prerequisis I think we might be able to use !!! tip syntax right now with markdown inside of it so we don't need to use HTML everywhere (@vbarda is that correct?)

9. Could we use a for loop instead? it's a bit more code, but easier to understand especially for folks not familiar with comprehensions or `next` + unfortunately we have `.next` as an attribute that shadows `next` and that makes things a bit visually harder to follow

```python
state_with_subgraph = next(
    s for s in graph.get_state_history(config) if s.next == ("node_2",)
)
```

eyurtsev:
At least the count is monotonic...

eyurtsev:
# subgraphs notebook

0. "When parent and subgraph share schema keys" vs. "When parent and subgraph have different schemas"  -- There needs to be more precise terminology for this or way to think about what's going on.

Would something like add a subgraph as a node to the graph vs. invoke a subgraph from a node directly work in terms of terminology?

2. If possible could you include the answer as part of the overview into the how-to guide?

A how-to guide should reduce the amount of time it takes a user to get the answer to the question they're looking for. The detail / in depth explanation can follow

3. In the 2nd pattern, how does persistence work? (subgraph.invoke({"bar"..."}, config=??) Is it consistent with our recommendation in the subgraph persistence graph about not providing a check-pointer? Is the thread id automatically propagated?

```
[Subgraphs](https://langchain-ai.github.io/langgraph/concepts/low_level/#subgraphs) allow you to build complex systems with multiple components that are themselves graphs. A common use case for using subgraphs is building [multi-agent systems](https://langchain-ai.github.io/langgraph/concepts/multi_agent).

[ 1-2 sentence answer here -- could be a snippet or a description of what to do ]
```



--------------------------
nits:

* The comment which explains what the example does is not visible without scrolling

image


2. There should be a heading prior -- it shouldn't be after the Set Up heading

> The main question when adding subgraphs is how the parent graph and subgraph communicate, i.e. how they pass the [state](https://langchain-ai.github.io/langgraph/concepts/low_level/#state) between each other during the graph execution. There are two scenarios:

4. Re-order the content in the code cells so that the the subgraph is created prior to the main graph (e.g., `State` variable should shift down if possible)

nit: maybe `State` should be relabeled as `SharedState`? Just to make it easier to understand what graph it belongs to

5. consider adding a single sentence before to explain what we're doing (why are we streaming? what are we checking?) can probably move the content after the cell to be before the cell.


image

6. perhaps use caution level?
> !!! note You cannot invoke more than one subgraph inside the same node.

eyurtsev:
# How to manage state in subgraphs

From the description of this notebook in the overview, I don't know what to expect. What does it mean to manage state?

Is this meant for mostly human-in-the-loop? or is this showing different techniques associated with statement management -- if so maybe a bullet point list explaining what will be shown?

The challenge for me as a reader is that I'd like to know whether the how-to guide is relevant for the objective I'm trying to achieve without having to read the entire how-to guide. (i.e., I don't want to be forced to read a long code example, unless I know that it's useful/relevant)

PR #2092, opened 2024-10-12
Update `stream()` and `astream()` methods in `RemoteGraph` to process `updates` event types

andrewnguonly:
### Summary
This PR depends on/is blocked by [this PR](https://github.com/langchain-ai/langgraph/pull/2088).

In order to process interrupts from a remote graph, the `RemoteGraph` needs to read the `__interrupt__` key from the `updates` event type. This requires manually inserting the `updates` stream mode if it's not provided and excluding the `updates` events if it's not provided. This approach allows the client to _always_ receive `updates` events, but abstract the implementation for processing interrupts.

The downside of this approach is that it requires other `PregelProtocol` implementations to also implement this abstracted logic.


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -357,17 +389,24 @@ def stream(
...
             interrupt_after=interrupt_after,  # type: ignore
             stream_subgraphs=subgraphs,
         ):
+            if chunk.event == "updates":
+                if INTERRUPT in chunk.data:
+                    raise GraphInterrupt()
+                if not include_updates:
+                    continue
+
             yield chunk
```
nfcampos, line 410: I think this needs to unnest the tuple w stream_mode, payload if the caller had asked for a single stream mode

andrewnguonly: Clarifying my understanding:
1. `RunsClient.stream()` returns `Iterator[StreamPart]`.
1. `StreamPart` is a `NamedTuple` containing `event` and `data`.
1. If a client calls `RunsClient.stream()` with only 1 stream mode (or no stream mode, which means the default is `values`), then only `data` is returned. Is this correct?

If this is correct, should we make the change in the SDK or the API instead?

PR #2105, opened 2024-10-14
Fix IsLastStep counter for runs with checkpointers

nfcampos:
- Share step/stop logic with PregelLoop
- Add RemainingSteps value which contains the number of remaining steps
- Switch create_react_agent to use RemainingSteps, so that it behave correctly for return_direct tools

### DIFF
--- a/libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
+++ b/libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
@@ -14,7 +14,7 @@ from langgraph._api.deprecation import deprecated_parameter
 from langgraph.graph import StateGraph
 from langgraph.graph.graph import CompiledGraph
 from langgraph.graph.message import add_messages
-from langgraph.managed import IsLastStep
+from langgraph.managed import IsLastStep, RemainingSteps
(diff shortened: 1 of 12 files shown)


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -547,10 +567,14 @@ def call_model(state: AgentState, config: RunnableConfig) -> AgentState:
...
-            state["is_last_step"]
-            and isinstance(response, AIMessage)
-            and response.tool_calls
+        has_tool_calls = isinstance(response, AIMessage) and response.tool_calls
+        all_tools_return_direct = (
+            all(call["name"] in should_return_direct for call in response.tool_calls)
+            if isinstance(response, AIMessage)
+            else False
+        )
+        if (state["remaining_steps"] < 1 and all_tools_return_direct) or (
```
vbarda, line 576: shouldn't these be the same between the call_model and acall_model??

nfcampos: oops

PR #2136, opened 2024-10-18
feat: Add LangGraph error pages

jacoblee93:
Also needs redirect from LangChain docs


libs/langgraph/langgraph/errors.py
```diff
@@ -22,13 +22,22 @@ class GraphRecursionError(RecursionError):
         )
     """
 
-    pass
+    def __init__(
+        self,
+        message: str = "Recursion limit reached.\n\nTroubleshooting URL: https://python.langchain.com/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT/",
```
efriis, line 27: I don't think this will actually get used as-written. I think we need to either append to the message passed in, or log a message

docs/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT.ipynb
efriis, line 1: langgraph currently has docs for errors in reference. Maybe better to include these in those docstrings.

efriis (commented):
@vbarda thoughts

docs/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT.ipynb
vbarda: yup, i agree, i think this would be better

libs/langgraph/langgraph/errors.py
vbarda: why? it works for me

libs/langgraph/langgraph/errors.py
```diff
@@ -22,13 +22,22 @@ class GraphRecursionError(RecursionError):
...
+        self.message = message
+        super().__init__(self.message)
 
 
 class InvalidUpdateError(Exception):
     """Raised when attempting to update a channel with an invalid set of updates."""
 
-    pass
+    def __init__(self, message: str, code: Optional[str] = None):
+        if code is not None:
```
vbarda, line 37: this is problematic because we don't actually pass the code currently and we do use `message`, so we would lose specific messages we currently have for different invalid updates. for example, here https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/pregel/__init__.py#L848

libs/langgraph/langgraph/errors.py
```diff
@@ -22,13 +22,22 @@ class GraphRecursionError(RecursionError):
...
+            message = f"{message}\n\nTroubleshooting URL: https://python.langchain.com/docs/troubleshooting/errors/{code}/"
```
vbarda, line 38: also we need to fix the URLs here to point to langgraph docs

vbarda:
> @vbarda thoughts

I agree with you that we should put these into the docstrings and just surface in the API ref page here https://langchain-ai.github.io/langgraph/reference/errors/

docs/docs/troubleshooting/errors/INVALID_GRAPH_NODE_RETURN_VALUE.ipynb
```diff
@@ -0,0 +1,56 @@
+{
```
vbarda, line 1: i think this needs to be updated to the following example:

```python
class State(TypedDict):
    some_key: str

def node(state: State):
    return {"some_key": "whoops"}

def other_node(state: State):
    return {"some_key": "whoops"}


builder = StateGraph(State)
builder.add_node(node)
builder.add_node(other_node)
builder.add_edge(START, "node")
builder.add_edge(START, "other_node")
```

libs/langgraph/langgraph/errors.py
efriis: I think all instances of `raise GraphRecursionError` have a message set, so this wouldn't populate right?

https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/pregel/__init__.py#L1310

https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/pregel/__init__.py#L1516

libs/langgraph/langgraph/errors.py
jacoblee93: No we're adding a redirect so we don't have github.io URLs in code

libs/langgraph/langgraph/errors.py
jacoblee93: Yeah looks like a mistake

docs/docs/troubleshooting/errors/INVALID_GRAPH_NODE_RETURN_VALUE.ipynb
vbarda: (wrong notebook, needs to be for the above)

PR #2149, opened 2024-10-21
docs: remove Literal as it is not being used

sarthakgupta072:
This PR removes the use of `from typing import Literal` since it is not being used in the code implementation


vbarda (approved):

vbarda:
Thank you!

PR #2154, opened 2024-10-21
checkpoint: add DuckDB store

vbarda:


libs/checkpoint-duckdb/langgraph/store/duckdb/aio.py
```diff
@@ -0,0 +1,202 @@
...
+        tasks = []
+
+        if GetOp in grouped_ops:
+            tasks.append(
+                self._batch_get_ops(
+                    cast(Sequence[tuple[int, GetOp]], grouped_ops[GetOp]), results
+                )
+            )
+
+        if PutOp in grouped_ops:
```
eyurtsev, line 48: Are we worried at about race conditions? 

We have write operations mixed with read operations?

vbarda: mm, that's a good question. don't think we're currently calling them together anywhere, but it's probably good to avoid that -- i can add a warning?

eyurtsev: Is this exposed to end users?

vbarda: yes but we're not showing that anywhere in the docs. i.e. the APIs that are actually meant to be used are `.put()` / `.search()` / `.get`, which just call `.batch()` under the hood

libs/checkpoint-duckdb/langgraph/store/duckdb/base.py
```diff
@@ -0,0 +1,395 @@
...
+            results.append((query, params, namespace, items))
+        return results
+
+    def _get_batch_PUT_queries(
+        self,
+        put_ops: Sequence[tuple[int, PutOp]],
+    ) -> list[tuple[str, Sequence]]:
+        inserts: list[PutOp] = []
+        deletes: list[PutOp] = []
+        for _, op in put_ops:
```
andrewnguonly, line 81: I see this was copied over from the Postgres implementation.

I'm wondering if there's a scenario where breaking the ordering of `DELETE` and `INSERT` in the original list of `put_ops` causes an issue.

This method accepts an arbitrary list of `put_ops` and returns a reordered list of queries.

libs/checkpoint-duckdb/langgraph/store/duckdb/base.py
```diff
@@ -0,0 +1,395 @@
...
+                SELECT prefix, key, value, created_at, updated_at
+                FROM store
+                WHERE prefix LIKE ?
+            """
+            params: list = [f"{_namespace_to_text(op.namespace_prefix)}%"]
+
+            if op.filter:
+                filter_conditions = []
+                for key, value in op.filter.items():
+                    if isinstance(value, list):
```
andrewnguonly, line 139: Double checking: The code in both if-branches have the same code. Expected or should we update?
```python
for key, value in op.filter.items():
    filter_conditions.append(f"json_extract(value, '$.{key}') = ?")
    params.append(json.dumps(value))
```

andrewnguonly (approved):

libs/checkpoint-duckdb/langgraph/store/duckdb/base.py
vbarda: will update - thanks for catching!

libs/checkpoint-duckdb/langgraph/store/duckdb/base.py
vbarda: i think this is a valid concern, kind of related to Eugene's questions above. i think we'll need to rethink this a bit for both postgres and duckdb, but for now all of the publicly documented methods that we show in the docs are single ops https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint/langgraph/store/base/__init__.py#L260

PR #2163, opened 2024-10-23
Bugfix

gipde:
if multiple Configs contains a List of Callbacks, the last one overrides the previous ones (compare ensure_config:281


vbarda:
@gipde thanks for your contribution -- current behavior is actually as intended. let me know if this is causing any specific issues for you!

PR #2164, opened 2024-10-23
langgraph: add prebuilt pipeline graph

vbarda:


libs/langgraph/langgraph/prebuilt/__init__.py
```diff
@@ -1,6 +1,7 @@
 """langgraph.prebuilt exposes a higher-level API for creating and executing agents and tools."""
 
 from langgraph.prebuilt.chat_agent_executor import create_react_agent
+from langgraph.prebuilt.sequential_executor import create_sequential_executor
```
hwchase17, line 4: name is weird and not super intuitive

`pipeline`? `chain`?

libs/langgraph/langgraph/prebuilt/sequential_executor.py
```diff
@@ -0,0 +1,77 @@
...
+        store: An optional store object. This is used for persisting data
+            across multiple threads (e.g., multiple conversations / users).
+        interrupt_before: An optional list of step names to interrupt before execution.
+        interrupt_after: An optional list of step names to interrupt after execution.
+        debug: A flag to enable debug mode.
+
+    Returns:
+        A CompiledStateGraph object.
+    """
+    if len(steps) < 2:
```
hwchase17, line 51: why? id do `<1`

libs/langgraph/langgraph/prebuilt/sequential_executor.py
```diff
@@ -0,0 +1,77 @@
...
+        return step.name
+    elif callable(step):
+        return getattr(step, "__name__", step.__class__.__name__)
+    else:
+        raise TypeError(f"Unsupported step type: {step}")
+
+
+def create_sequential_executor(
+    *steps: Union[RunnableLike, tuple[str, RunnableLike]],
+    state_schema: Type[Any],
```
hwchase17, line 26: add `*,` after this to force passing as kwargs?

libs/langgraph/langgraph/prebuilt/sequential_executor.py
```diff
@@ -0,0 +1,77 @@
...
+    checkpointer: Optional[BaseCheckpointSaver] = None,
+    store: Optional[BaseStore] = None,
+    interrupt_before: Optional[list[str]] = None,
```
hwchase17, line 29: worth checking that these values are in the node names? or will the graph.compile flag that?

libs/langgraph/langgraph/prebuilt/sequential_executor.py
vbarda: yea wasn't sure which is a more sensible one here, will make it 1

libs/langgraph/langgraph/prebuilt/sequential_executor.py
vbarda: it's already forced to be passed as kwargs actually, learned this today

vbarda: (cause of `*steps` in the beginning)

libs/langgraph/langgraph/prebuilt/__init__.py
vbarda: yea chain might be best

libs/langgraph/langgraph/prebuilt/sequential_executor.py
vbarda: will be flagged by `graph.compile`

libs/langgraph/langgraph/prebuilt/sequential_executor.py
```diff
@@ -0,0 +1,77 @@
...
+    interrupt_after: Optional[list[str]] = None,
+    debug: bool = False,
+) -> CompiledStateGraph:
+    """Creates a sequential executor graph that runs a series of provided steps in order.
+
+    Args:
+        *steps: A sequence of RunnableLike objects or (name, RunnableLike) tuples.
```
eyurtsev, line 36: * what happens if the same name is re-used by mistake?
* What happens if the same name is re-used on purpose? (e.g., we want to make a sequence that re-uses the same node?)

libs/langgraph/langgraph/prebuilt/sequential_executor.py
```diff
@@ -0,0 +1,77 @@
...
+        builder.add_node(name, step)
+        if previous_name is None:
+            builder.add_edge(START, name)
+        else:
+            builder.add_edge(previous_name, name)
+
+        previous_name = name
+
+    builder.add_edge(cast(str, previous_name), END)
+    return builder.compile(
```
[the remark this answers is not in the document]
vbarda: > Should it be compiled or uncompiled?

Also not sure, had the same question. I think might be nice to have an option to return either (even though it's a bit icky to have multiple return types like this). But i think this needs to be consistent with `create_react_agent`

> Can it be added a sub-graph after being compiled?

it can definitely be added as a subgraph after being compiled :) you actually can't add it uncompiled (can either add a compiled subgraph as a node or invoke the compiled subgraph inside a node func)

> Will users be confused about how to interpret where to provide a checkpointer?

why do you think so? think should just document it well. it would probably be more confusing if we have options to return either compiled or uncompiled

libs/langgraph/langgraph/prebuilt/sequential_executor.py
vbarda: > what happens if the same name is re-used by mistake?

langgraph already handles this at graph.compile()

> What happens if the same name is re-used on purpose? (e.g., we want to make a sequence that re-uses the same node?)

it would have to have a different name, we don't allow multiple nodes with the same name in langgraph

libs/langgraph/langgraph/prebuilt/sequential_executor.py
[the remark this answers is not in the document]
hinthornw: I still feel strongly that it should be uncompiled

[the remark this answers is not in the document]
vbarda: added an option to return either, might need to add the same to `create_react_agent`

libs/langgraph/langgraph/prebuilt/sequential_executor.py
vbarda: added a check actually!

libs/langgraph/langgraph/prebuilt/sequential_executor.py
vbarda: added a check for the duplicate names actually!

libs/langgraph/langgraph/prebuilt/chain_executor.py
```diff
@@ -0,0 +1,89 @@
...
+    builder = StateGraph(state_schema)
+    previous_name: Optional[str] = None
+    for step in steps:
+        if isinstance(step, tuple) and len(step) == 2:
+            name, step = step
+        else:
+            name = _get_name(step)
+
+        if name in node_names:
+            raise ValueError(f"Node name {name} already exists.")
```
eyurtsev, line 68: Could we include more information in the error message that includes an action point for users on how to resolve the problem? It's nice for pre-built components do as much hand holding as possible.

---

Ideally the error message would explain that the function can be invoked with a 2-tuple to specify a name for each node.

---

Edge cases: 

1) functions with an identical name defined in different modules will end up mapping to the same name.
2) different lambdas get mapped to the same node name

libs/langgraph/langgraph/prebuilt/chain_executor.py
```diff
@@ -0,0 +1,89 @@
...
+        builder.add_node(name, step)
+        if previous_name is None:
+            builder.add_edge(START, name)
+        else:
+            builder.add_edge(previous_name, name)
+
+        previous_name = name
+
+    builder.add_edge(cast(str, previous_name), END)
+    if not return_compiled:
```
eyurtsev, line 80: if return_compiled=False, should we raise an error or log a warning on other arguments being provided?

libs/langgraph/langgraph/prebuilt/__init__.py
```diff
@@ -12,6 +13,7 @@
 
 __all__ = [
     "create_react_agent",
+    "create_chain_executor",
```
eyurtsev, line 16: `create_chain_executor` or `create_chain` to mirror `create_react_agent` ?

eyurtsev: I don't have a strong opinion here, but less typing is usually better if there's no ambiguity

libs/langgraph/langgraph/prebuilt/chain_executor.py
vbarda: probably, but removed the flag for now in favor of uncompiled

libs/langgraph/langgraph/prebuilt/sequential_executor.py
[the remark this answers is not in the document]
hinthornw: I personally think it should always be uncompiled, especially for this "lower-level" type method
and that we made the wrong decision for create_react_agent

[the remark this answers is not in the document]
vbarda: sounds good! i kept is uncompiled in the current version of the PR

eyurtsev (approved):

PR #2175, opened 2024-10-24
langgraph: always raise NodeInterrupt in ToolNode if raised from a tool

vbarda:


libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -267,6 +268,10 @@ def _run_one(self, call: ToolCall, config: RunnableConfig) -> ToolMessage:
                 Union[str, list], msg_content_output(tool_message.content)
             )
             return tool_message
+        # NodeInterrupt is a special exception that will always be raised
+        # if a tool is raising it
+        except NodeInterrupt as e:
```
nfcampos, line 273: switch to GraphInterrupt

PR #2181, opened 2024-10-24
Update memory concept doc

rlancemartin:
Add content on [memory types](https://blog.langchain.dev/memory-for-agents/).


docs/docs/concepts/memory.md
```diff
@@ -2,7 +2,9 @@
 
 ## What is Memory?
 
-Memory in AI applications refers to the ability to process, store, and effectively recall information from past interactions. With memory, your agents can learn from feedback and adapt to users' preferences. This guide is divided into two sections based on the scope of memory recall: short-term memory and long-term memory.
+Memory is the ability to encode, store, retain, and subsequently recall information and experiences. Consider the frustration of working with a colleague who forgets everything you tell them, requiring constant repetition! As AI agents undertake more complex tasks involving numerous user interactions, equipping them with memory becomes equally crucial for efficiency and user satisfaction. 
+
+Memory in AI agents refers to the ability to process, store, and effectively recall information from past interactions. With memory, your agents can learn from feedback and adapt to users' preferences. This guide is divided into two sections based on the scope of memory recall: short-term memory and long-term memory.
```
eyurtsev, line 7: nit: language of first sentence is very similar to first sentence in previous paragraph -- here it feels more like duplication than serving as an insight to drive an analogy
(not important to address)

docs/docs/concepts/memory.md
```diff
@@ -190,94 +194,73 @@ item = store.get(namespace, "a-memory")
...
 
-Managing long-term memory as a collection of documents, on the other hand, lets you store an unbounded amount of information. This technique is useful when you want to repeatedly extract & remember items over a long time horizon but can be more complicated to query and manage over time.
-Similar to the "profile" memory, you still define schema(s) for each memory. Rather than overwriting a single document, you instead will insert new ones (and potentially update or re-contextualize existing ones in the process). We explain this approach in more detail in ["managing a collection of memories"](#manage-a-collection-of-memories).
+**When do you want to update memories?**
 
-**Do you want to present memories to your agent as updated instructions or as few-shot examples?**
+Memory can be updated as part of an agent's application logic (e.g. "on the hot path"). In this case, the agent typically decides to remember facts before responding to a user. Alternatively, memory can be updated as a background task (as a separate function that generates memories based on the primary application's state). We explain the tradeoffs between these approaches in the [section below](#writing-memories).
```
eyurtsev, line 209: nit: "a separate function" a bit strange in terms of terminoloy? what doe it mean that it's a "function"? does it have to be a function or can i provide it as a method or a class? what does it mean that it's "separate"?

Maybe could paraphrase a bit to be more general "some logic that runs in the background / asynchronously and generates memories" ?

docs/docs/concepts/memory.md
```diff
@@ -190,94 +194,73 @@ item = store.get(namespace, "a-memory")
...
 
-Memories are typically provided to the LLM as a part of the system prompt. Some common ways to "frame" memories for the LLM include providing raw information as "memories from previous interactions with user A", as system instructions or rules, or as few-shot examples.
+## Memory types
 
-Framing memories as "learning rules or instructions" typically means dedicating a portion of the system prompt to instructions the LLM can manage itself. After each conversation, you can prompt the LLM to evaluate its performance and update the instructions to better handle this type of task in the future. We explain this approach in more detail in [this section](#update-own-instructions).
+Different applications require various types of memory. Although the analogy isn't perfect, examining [human memory types](https://www.psychologytoday.com/us/basics/memory/types-of-memory?ref=blog.langchain.dev) can be insightful. Some [research papers](https://arxiv.org/pdf/2309.02427) have even mapped these human memory types to those used in AI agents.
```
eyurtsev, line 213: "Some" research papers  only links to one (maybe include "e.g.,")

docs/docs/concepts/memory.md
```diff
@@ -190,94 +194,73 @@ item = store.get(namespace, "a-memory")
...
 
-A profile is generally just a JSON document with various key-value pairs you've selected to represent your domain. When remembering a profile, you will want to make sure that you are **updating** the profile each time. As a result, you will want to pass in the previous profile and ask the LLM to generate a new profile (or some JSON patch to apply to the old profile).
+Document collections also shift complexity to memory **search** over the list. The `Store` currently supports [filtering by metadata](https://langchain-ai.github.io/langgraph/reference/store/#storage) and will soon add [semantic search shortly](https://python.langchain.com/docs/concepts/vectorstores/), but selecting the most relevant documents can be tricky as the list grows.
```
eyurtsev, line 235: nit: Working with document collections? Representation with document collections?

docs/docs/concepts/memory.md
```diff
@@ -303,8 +285,24 @@ def update_instructions(state: State, store: BaseStore):
...
+
+While [humans often form long-term memories during sleep](https://medicine.yale.edu/news-article/sleeps-crucial-role-in-preserving-memory/), AI agents need a different approach. When and how should agents create new memories? There are at least two primary methods for agents to write memories: "on the hot path" and "in the background".
+
+![](img/memory/hot_path_vs_background.png)
+
+### Writing memories in the hot path
+
+Creating memories during runtime offers both advantages and challenges. On the positive side, this approach allows for real-time updates, making new memories immediately available for use in subsequent interactions. It also enables transparency, as users can be notified when memories are created and stored.
+
+However, this method also presents challenges. It increases complexity, as the agent must decide what to commit to memory alongside its primary task. This can impact performance, potentially slowing down response times due to additional decision-making and degrading tool-calling performance and task completion rates. Additionally, the multitasking involved may result in fewer memories being saved, affecting recall in future conversations.
```
eyurtsev, line 298: "It increases complexity, as the agent must decide what to commit to memory alongside its primary task."

I don't buy this. The agent already needs to know how to work with memory on the read path AND regardless the logic has to exist somewhere, so it just doesn't sound like a significant increase in complexity that's worth mentioning?

eyurtsev, line 298: 1. "performance" is generic (sounds like it might be quality) do we mean latency here?
2. Why is there a degradation of tool calling performance or task completion rate? (unclear to me)

docs/docs/concepts/memory.md
rlancemartin: Agree, let me smooth it over for better style.

rlancemartin: And I need to add a proper [citation](https://pmc.ncbi.nlm.nih.gov/articles/PMC10410470/) to the first sentence, anyway :P

docs/docs/concepts/memory.md
rlancemartin: > Maybe could paraphrase a bit to be more general "some logic that runs in the background / asynchronously and generates memories" ?

yes, function is too specific. will update.

docs/docs/concepts/memory.md
rlancemartin: This should be explained better: it means that an agent would need a tool (e.g., to write and / or format) memories, which is strictly an increase in overhead / complexity.

rlancemartin: > The agent already needs to know how to work with memory on the read

Arguably read is just a hard-coded call for the store rather than a tool that the agent needs to manage. But, I get the point that the argument needs to be a bit more crisp.

rlancemartin: This is specifically assuming that memory-creation is a tool. I will add that for clarity.

PR #2182, opened 2024-10-24
langgraph: add message list validation to create_react_agent + a troubleshooting guide

vbarda:


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -530,6 +561,7 @@ def should_continue(state: AgentState) -> Literal["tools", "__end__"]:
 
     # Define the function that calls the model
     def call_model(state: AgentState, config: RunnableConfig) -> AgentState:
+        _validate_messages(state["messages"])
```
eyurtsev, line 564: Is it possible for the messages to ever end up being dict representation of open ai format requests? (i.e., not base messages)

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -161,6 +161,37 @@ def _should_bind_tools(model: LanguageModelLike, tools: Sequence[BaseTool]) -> b
     return False
 
 
+def _validate_messages(
+    messages: Sequence[BaseMessage],
```
eyurtsev, line 165: is sequence base message guaranteed?

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -161,6 +161,37 @@ def _should_bind_tools(model: LanguageModelLike, tools: Sequence[BaseTool]) -> b
     return False
 
 
+def _validate_messages(
```
eyurtsev, line 164: Maybe more generic name? we're not validating the messages individually, but the sequence of messages / history

```suggestion
def _validate_chat_history(
```

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -161,6 +161,37 @@ def _should_bind_tools(model: LanguageModelLike, tools: Sequence[BaseTool]) -> b
...
+    messages: Sequence[BaseMessage],
+) -> None:
+    """Validate that all tool calls in AIMessages have a corresponding ToolMessage."""
+    all_tool_calls = [
+        tool_call
+        for message in messages
+        if isinstance(message, AIMessage)
+        for tool_call in message.tool_calls
+    ]
+    answered_tool_call_ids = {
```
eyurtsev, line 174: (nit) it's internal so doesn't super matter, but `answered` and `unanswered` are slightly confusing terms. 

Some better names would probably be: Unacknowledged/acknowledged

There's probably something even better

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -161,6 +161,37 @@ def _should_bind_tools(model: LanguageModelLike, tools: Sequence[BaseTool]) -> b
...
+        for tool_call in all_tool_calls
+        if tool_call["id"] not in answered_tool_call_ids
+    ]
+    if not unanswered_tool_calls:
+        return
+
+    error_message = (
+        "Found AIMessages with tool_calls that do not have a corresponding ToolMessage. "
+        f"Here are the first few of those tool calls: {unanswered_tool_calls[:3]}.\n\n"
+        "ALL tool calls in message history MUST have a corresponding ToolMessage - this is required by most LLM providers.\n\n"
```
eyurtsev, line 188: The **most** part is a problem in this error message. What if a user is working with an LLM that doesn't require it?

```suggestion
        "Every tool call (request to call a tool) in the message history MUST have a corresponding ToolMessage (result of tool invocation to be surfaced to the LLM) - this is required by most LLM providers.\n\n"
```

cc @baskaryan  -- should we enforce regardless?

eyurtsev (approved):

docs/docs/troubleshooting/errors/INVALID_CHAT_HISTORY.md
```diff
@@ -0,0 +1,30 @@
+# INVALID_CHAT_HISTORY
+
+This error is raised in the prebuilt [create_react_agent][langgraph.prebuilt.chat_agent_executor.create_react_agent] when the `call_model` graph node receives a malformed list of messages. Specifically, it is malformed when there are `AIMessages` with `tool_calls` that do not have a corresponding `ToolMessage`.
```
eyurtsev, line 3: Could be good to clarify:

`tool_calls` (representing requests to invoke tools) do not have a corresponding `ToolMessage` (representing results of tool call invocation)

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: no, the messages go through `add_messages` reducer

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: yea i don't love the imprecise language. i'd be in favor of enforcing regardless, but we could also do this validation only for specific LLMs too

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: yes

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: nice, will rename

docs/docs/troubleshooting/errors/INVALID_CHAT_HISTORY.md
vbarda: `ToolMessage` contains results of a single tool call (and has a corresponding tool call ID)

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: good point -- will double check this use case

vbarda: (although i suspect that might break react agent loop anyway)

PR #2186, opened 2024-10-24
cli: Passthrough args to docker build

nfcampos:
- langgraph build now passes through all other args/flags to docker build
- remove unused langgraph test command


hinthornw:
Copied over to other PR that was landed

PR #2194, opened 2024-10-25
Fix link that is causing docs build to fail

rlancemartin:


.github/workflows/deploy_docs.yml
```diff
@@ -82,7 +82,8 @@ jobs:
               --check-links-ignore "https://github\.com/.*" \
               --check-links-ignore "/.*\.(ipynb|html)$" \
               --check-links-ignore "https://python\.langchain\.com/.*" \
-              --check-links $(find docs/site -name "index.html" | grep -v 'storm/index.html')
+              --check-links $(find docs/site -name "index.html" | grep -v 'storm/index.html') \
+              --check-links-ignore "https://openai.com/index/memory-and-new-controls-for-chatgpt/"
```
vbarda, line 86: not sure if the order of flags is important - you might need to put it above the `—check-links`

vbarda (approved):

PR #2204, opened 2024-10-28
sdk: By default raise errors in /wait endpoint

nfcampos:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -159,7 +159,7 @@ def get_client(
     client = httpx.AsyncClient(
         base_url=url,
         transport=transport,
-        timeout=httpx.Timeout(connect=5, read=60, write=60, pool=5),
+        timeout=httpx.Timeout(connect=5, read=300, write=300, pool=5),
```
vbarda, line 162: confirming that this is intentional. if so, could we duplicate in the sync client too?

PR #2206, opened 2024-10-28
Add support for Docker arg passthrough in langgraph CLI

hinthornw:
Added a js-example to show it builds
Adapted integration tests after removing the test CLI command


libs/cli/langgraph_cli/cli.py
```diff
@@ -425,37 +328,33 @@ def _build(
...
+    docker_build_args: Sequence[str],
     base_image: Optional[str],
     pull: bool,
     tag: str,
 ):
     with Runner() as runner, Progress(message="Pulling...") as set:
-        # check docker available
-        langgraph_cli.docker.check_capabilities(runner)
-        # open config
+        if shutil.which("docker") is None:
```
vbarda, line 351: ooc, why not use `check_capabilities`?

nfcampos: Because that checks for docker compose etc, which more people don't have, and isn't needed to build images

PR #2211, opened 2024-10-29
 Create Local LLM Multi-Agent Using Langgraph

Barry1915:
I noticed that many people have the need to build agents using local LLMs, and I share that need as well. However, I found that currently, Langgraph only has tutorials based on closed-source LLMs. So, I created an example of an agent using a local LLM. I tried using Vllm, FastChat, and Ollama. The first two didn't work, but only the locally deployed model from Ollama was successful. I hope this example can help many people.

### DIFF
--- a/docs/docs/how-tos/create-local-llm-agent.ipynb
+++ b/docs/docs/how-tos/create-local-llm-agent.ipynb
@@ -29,7 +29,7 @@
     "    openai_api_base=inference_server_url,\n",
     "    max_tokens=500,\n",
     "    temperature=1,\n",
-    ")"
+    ")\n"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


vbarda:
@Barry1915 thank you for your contribution! You can start a discussion https://github.com/langchain-ai/langgraph/discussions and share the notebook there. The how-to guides are meant to be more general, without a focus on a specific model/provider, so will close this PR in the meantime.

Barry1915:
Thanks for reply，I want to express that this ipynb file is a way to connect a local LLM to Langgraph, and it is not specific to any particular LLM. I've noticed that many people in the issues are hoping to connect a local LLM to Langgraph.

PR #2217, opened 2024-10-29
Add docstring to `RemoteGraph`

andrewnguonly:


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -46,6 +46,14 @@ class RemoteException(Exception):
 
 
 class RemoteGraph(PregelProtocol):
+    """The RemoteGraph class is a client implementation for calling remote
+    APIs that implement the Open Agent API Specification.
+
+    For example, the RemoteGraph class can be used to call LangGraph Cloud
+    APIs.
+
+    RemoteGraph can be used directly as a node in Graph.
```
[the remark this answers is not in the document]
vbarda: also, i think we need to say that it behaves the same was as a regular `graph` _and_ can be used directly as a node in another graph

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -64,6 +74,15 @@ def __init__(
 
         If `client` or `sync_client` are provided, they will be used instead of the default clients.
         See `LangGraphClient` and `SyncLangGraphClient` for details on the default clients.
+
+        Args:
+            name: The name of the graph.
+            url: The URL of the remote API.
+            api_key: The API key to use for authentication.
```
vbarda, line 81: i would add that if not specified, we'll look for `LANGSMITH_API_KEY` etc, similar to the langgraph sdk

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -64,6 +74,15 @@ def __init__(
 
         If `client` or `sync_client` are provided, they will be used instead of the default clients.
```
vbarda, line 75: i would also add that if at least one of the 3 (url/client/sync_client) needs to be specified. something similar to how i describe in this how-to guide maybe https://github.com/langchain-ai/langgraph/pull/2218/files

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -268,6 +309,18 @@ def _sanitize_obj(obj: Any) -> Any:
     def get_state(
         self, config: RunnableConfig, *, subgraphs: bool = False
     ) -> StateSnapshot:
+        """Method to call POST /threads/{thread_id}/state/checkpoint or
```
vbarda, line 312: i would add a better description here and then mention which endpoint it calls below. i.e. as a user my intent is not to call the API endpoint but to get the state for a config

vbarda (approved):

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -46,6 +46,16 @@ class RemoteException(Exception):
 
 
 class RemoteGraph(PregelProtocol):
+    """The RemoteGraph class is a client implementation for calling remote
+    APIs that implement the Open Agent API Specification.
```
eyurtsev, line 50: "Open Agent API Specification"

Issue #2227, opened 2024-10-30
crfeate_react_agent does not support already bound tools with gemini models


vidit-bhatia:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
from langchain_google_genai import ChatGoogleGenerativeAI
from configs.configs import llm_api_key
from langchain.tools import tool, BaseTool, StructuredTool
from typing import Annotated, List, Dict  
from string import Template 
import logging
from langchain_core.runnables.config import RunnableConfig
from json import loads, dumps
from typing import Optional, Type

llm = ChatGoogleGenerativeAI(model="gemini-1.5-flash-8b",google_api_key=llm_api_key.gemini_api_key,temperature=1, tool_call  = True)

@tool
def authoring_tool(user_information: RunnableConfig,
           template: Annotated[str,"The template created by LLM to be used for personalization"]) -> str:
    """The tool gets a python string template as an input and the user information . Returns the result of personalizing the template with the user information"""
    # t = Template(template) 
    user = user_information.get("configurable", {}).get("user")
    # final_to_print = t.substitute(loads(user)) 
    final_to_print = template.format(user)
    logging.info(final_to_print)
    return final_to_print


authoring_tool_llm = llm.bind_tools(tools=[authoring_tool],tool_choice="any")
author_agent = create_react_agent(authoring_tool_llm, tools=[authoring_tool],state_modifier=author_agent_prompt)
```

### Error Message and Stack Trace (if applicable)

```shell
Traceback (most recent call last):
  File "/Users/vibhatia/anaconda3/lib/python3.11/runpy.py", line 198, in _run_module_as_main
    return _run_code(code, main_globals, None,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vibhatia/anaconda3/lib/python3.11/runpy.py", line 88, in _run_code
    exec(code, run_globals)
[...]

PR #2228, opened 2024-10-30
issue #2227 fixed with this change

vidit-bhatia:
issue #2227


vbarda:
@vidit-bhatia thanks for your contribution -- i responded in the original issue, the problem was actually in `langchain-google-genai`, and should be fixed by updating to the latest version

PR #2236, opened 2024-10-30
langgraph[patch]: fix annotations in inherited tool schemas

### LINKED ISSUE #2220: ToolNode doesn't detect `InjectedState` annotations for tools with Pydantic `args_schema`
### Checked other resources

- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
.
```

```shell
.
```

### Description

.

### System Info

.

baskaryan:
fix #2220


libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -34,6 +34,7 @@
 from langchain_core.runnables.utils import Input
 from langchain_core.tools import BaseTool, InjectedToolArg
 from langchain_core.tools import tool as create_tool
+from langchain_core.tools.base import _get_all_basemodel_annotations
```
vbarda, line 37: would be nice to make this a public API

vbarda (approved):

libs/langgraph/langgraph/prebuilt/tool_node.py
baskaryan, line 37: ah need to check when this was added to core

baskaryan: might need to just copy it over if we don't want to bump core dep

vbarda: i think it's fine to bump core dep

libs/langgraph/pyproject.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://www.github.com/langchain-ai/langgraph"
 
 [tool.poetry.dependencies]
 python = ">=3.9.0,<4.0"
-langchain-core = ">=0.2.39,<0.4"
+langchain-core = "^0.2.42 || ^0.3.14"
```
vbarda, line 12: TIL

vbarda (approved):

libs/langgraph/pyproject.toml
baskaryan: me too

PR #2265, opened 2024-10-31
disable more link checking

eyurtsev:


.github/workflows/deploy_docs.yml
```diff
@@ -82,7 +82,7 @@ jobs:
               --check-links-ignore "https://github\.com/.*" \
               --check-links-ignore "/.*\.(ipynb|html)$" \
               --check-links-ignore "https://python\.langchain\.com/.*" \
-              --check-links-ignore "https://www.\.langchain\.com/langgraph-platform-pricing.*" \
+              --check-links-ignore "https://www.langchain\.com/.*" \
```
eyurtsev, line 85: we'll need to revert this tomorrow after new version of main site is published

PR #2289, opened 2024-11-01
docs: update usage of claude-3-5-sonnet-2024062 to claude-3-5-sonnet-20241022

SokiKawashima:
Updates the default model used in documentation to claude-3-5-sonnet-20241022

https://docs.anthropic.com/en/docs/about-claude/models

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ tools = [search]
 
 tool_node = ToolNode(tools)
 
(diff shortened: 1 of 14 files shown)


vbarda:
Thank you! I think we should actually use `claude-3-5-sonnet-latest`. Since there are multiple merge conflicts, will close this one and we will update the guides to use the latest model. Feel free to open another pull request that does that

PR #2297, opened 2024-11-01
fix(docs): function call in partial

allenc97:
Docs use prompts with partial + function call, which would result in a static datetime as opposed to a dynamic datetime being rendered on use.


vbarda:
@allenc97 that wouldn't work, as `datetime.now` is a function, so it would need to be actually invoked

allenc97:
Partial takes in a string or a callable - this is the latter. The original behavior would result in an evaluation during the .partial function call and not on substitution.

vbarda:
you're right, my apologies - will reopen and merge!

vbarda (approved):

vbarda:
Thanks for adding!

PR #2301, opened 2024-11-01
docs: add more redirects & remove unused cloud pages

vbarda:


docs/mkdocs.yml
```diff
@@ -56,7 +56,11 @@ plugins:
   - autorefs
   - redirects:
       redirect_maps:
+        'cloud/index.md': 'concepts/index.md#langgraph-platform'
+        'cloud/how-tos/index.md': 'how-tos/index.md#langgraph-platform'
         'cloud/concepts/api.md': 'concepts/index.md#langgraph-platform'
+        'cloud/concepts/cloud.md': 'concepts/langgraph_cloud.md'
```
eyurtsev, line 62: does this not use the architecture picture?

docs/mkdocs.yml
```diff
@@ -56,7 +56,11 @@ plugins:
   - autorefs
   - redirects:
       redirect_maps:
+        'cloud/index.md': 'concepts/index.md#langgraph-platform'
+        'cloud/how-tos/index.md': 'how-tos/index.md#langgraph-platform'
         'cloud/concepts/api.md': 'concepts/index.md#langgraph-platform'
```
eyurtsev, line 61: api should probably map to langgraph server instead? and not to the concepts page?

docs/mkdocs.yml
vbarda: it does, good catch

docs/mkdocs.yml
vbarda: yep

docs/mkdocs.yml
vbarda: actually nevermind, i was correct, there was a separate image for this, this one was unused

PR #2304, opened 2024-11-01
RemoteGraph: If node name is not present, fallback to node id as the node name

andrewnguonly:


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -135,7 +135,7 @@ def _get_drawable_nodes(
             node_id = str(node["id"])
             nodes[node_id] = DrawableNode(
                 id=node_id,
-                name=node.get("name", ""),
+                name=node.get("name", node_id),
```
vbarda, line 138: oh i think this should also try to pull this in from the node data first

vbarda (approved):

PR #2323, opened 2024-11-04
langgraph: add config metadata to pregel loop

vbarda:


libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -494,6 +495,7 @@ def _put_checkpoint(self, metadata: CheckpointMetadata) -> None:
         # assign step and parents
         metadata["step"] = self.step
         metadata["parents"] = self.config[CONF].get(CONFIG_KEY_CHECKPOINT_MAP, {})
+        metadata.update(self.config[CONFIG_KEY_METADATA])
```
nfcampos, line 498: This would let users overwrite our default fields, which I don't want. This should come before

nfcampos (approved):

nfcampos (approved):

PR #2327, opened 2024-11-04
lib: Add two more tests for Send

nfcampos:


libs/langgraph/tests/test_pregel.py
```diff
@@ -1743,6 +1743,103 @@ def route_to_three(state) -> Literal["3"]:
     assert graph.invoke(["0"]) == ["0", "1", "2", "2", "3"]
 
 
+def test_concurrent_emit_sends() -> None:
```
vbarda, line 1746: should we add async tests as well?

nfcampos: Yea, added

PR #2339, opened 2024-11-05
[rfc] langgraph: allow tools / ToolNode to modify state

vbarda:


libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -221,8 +227,33 @@ def _func(
...
+        outputs, state_updates = zip(*outputs)
+        combined_state_updates = {}
+        for state_update in state_updates:
+            for k, v in state_update.items():
+                if k == self.messages_key:
+                    raise ValueError(
+                        "Cannot return state updates for the messages key."
+                    )
+
+                if k in combined_state_updates:
```
eyurtsev, line 239: Arguably should actually be allowed and follow the same semantics as the reducer. This would be solved by moving into a design where there's a node per tool. (In the context of the Tool Node, this might be the best that can be done.)

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -50,6 +51,11 @@
 TOOL_CALL_ERROR_TEMPLATE = "Error: {error}\n Please fix your mistakes."
 
 
+class StateUpdateArtifact(NamedTuple):
```
eyurtsev, line 54: * Does this serialize out of the box?
* Why limit to state_update rather vs. the full GraphCommand?

vbarda: this was before our most recent discussions -- would have to full `GraphCommand`. but yea, NamedTuple should serialize out of box

PR #2346, opened 2024-11-05
lib: Execute Sends in the superstep that originated them (feature-flagged)

nfcampos:
- Currently this behavior is opt-in, behind a feature flag
- Keep old code path for compatibility with existing checkpoints
- Keep a similar order of application of updates, in some cases there will be no visible change
- Update task path for Sends to contain the path of all the parent tasks (multiple parents when a Send task creates another Send)
- That lineage path is used to ensure order of application of updates respects their logical lineage (ie updates from parents always applied before their child tasks)
- Move Interrupt writes to use negative indexes, which allow replacing/shadowing (when task is re-run it may interrupt again, or succeed)
- Runner will now attempt to schedule new Send tasks as soon as the write is received (ie while the originating node is still running)
- Update kafka scheduler to support new Send behavior


libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -383,31 +452,75 @@ def prepare_single_task(
...
+            task_path_t = cast(tuple[str, tuple, int, str], task_path)
+            writes_for_path = [w for w in pending_writes if w[0] == task_path_t[3]]
+            if task_path_t[2] >= len(writes_for_path):
+                logger.warning(
+                    f"Ignoring invalid write index {task_path[2]} in pending writes"
+                )
+                return
+            packet = writes_for_path[task_path_t[2]][2]
+            if not isinstance(packet, Send):
+                print("packet", task_path_t, writes_for_path)
```
nfcampos, line 497: removed in https://github.com/langchain-ai/langgraph/pull/2393 which will be merged at same time

PR #2352, opened 2024-11-06
langgraph: add add_sequence to StateGraph

vbarda:


libs/langgraph/langgraph/graph/state.py
```diff
@@ -415,6 +424,57 @@ def add_edge(self, start_key: Union[str, list[str]], end_key: str) -> Self:
         self.waiting_edges.add((tuple(start_key), end_key))
         return self
 
+    def add_sequence(
+        self,
+        nodes: Sequence[Union[RunnableLike, tuple[str, RunnableLike]]],
```
eyurtsev, line 429:
```suggestion
        nodes: Sequence[Union[RunnableLike, tuple[str, RunnableLike]]],
        *,
```

libs/langgraph/langgraph/graph/state.py
```diff
@@ -415,6 +424,57 @@ def add_edge(self, start_key: Union[str, list[str]], end_key: str) -> Self:
...
+        from_nodes: Optional[list[str]] = None,
```
eyurtsev, line 430: Last questions about API:

1. Do we want to start off with from_nodes being required rather than having a default? The concern is that using this consecutively sounds like it should append, but instead it'll be creating branches.
2. If we are supporting a default, any reason not to do union[str, list[str]] == START so it's easy to see what the default is? Otherwise we're forcing folks to read the documentation to look up the default.

I have a minor concern about the API looking consistent. Do we think it makes sense to add `from_nodes` to `add_node`  as well? (not as part of this PR). Would it be helpful to just type out the entire API that we want as an end state?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -415,6 +424,57 @@ def add_edge(self, start_key: Union[str, list[str]], end_key: str) -> Self:
...
+    ) -> Self:
+        """Add a sequence of steps that will be executed in the provided order.
```
eyurtsev, line 432: Maybe one more sentence about from_nodes?

```suggestion
        """Add a sequence of nodes that will be executed in the provided order.
```

libs/langgraph/langgraph/graph/state.py
eyurtsev: my personal preference for APIs -- is to always use a `*` before a named argument with a default. 

It forces devs to type more code when using the API but generally makes the resulting code easier and with less variations/

libs/langgraph/langgraph/graph/state.py
vbarda: i think it's fine to make it required for now!

vbarda: we can discuss `add_node` + `from_nodes` separately -- i think we should be careful with having too many ways of doing the same thing

vbarda: (actually will remove `from_nodes` altogether based on the offline discussion)

PR #2366, opened 2024-11-07
Print statement is needed to print the output

saurabhvartak1982:
Print statement is needed to print the output of the below statement: final_state["messages"][-1].content


eyurtsev:
hi @saurabhvartak1982 thanks for the contribution. I'll close since the print isn't adding much here -- it's self-evident and lots of folks run these in interactive environments anyway

PR #2378, opened 2024-11-08
Imperative API

nfcampos:
- Whereas `Send` is for fire-and-forget type of calls, new `call` and `acall` functions are for flows where you want to wait for the node to finish before doing something else
- Because we return regular python future objects (concurrent.futures.Future or asyncio.Future) all the python primitives for working with futures work, eg. wait, gather, etc

```py
    @task()
    def mapper(input: str) -> str:
        print(f"mapper {input}")
        return input * 2

    @entrypoint(checkpointer=checkpointer)
    def graph(input: list[str]) -> list[str]:
        futures = [mapper(i) for i in input]
        mapped = [f.result() for f in futures]
        return mapped

    thread1 = {"configurable": {"thread_id": "1"}}
    assert graph.invoke(["0", "1"], thread1) == ["00", "11"]
```


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -540,6 +540,11 @@ def _prepare_state_snapshot(
                     next_tasks[tid].writes.append((k, v))
                 if tasks := [t for t in next_tasks.values() if t.writes]:
                     apply_writes(saved.checkpoint, channels, tasks, None)
+            print(
```
nfcampos, line 543: remove

libs/langgraph/langgraph/pregel/runner.py
```diff
@@ -152,6 +211,7 @@ def writer(
                 del fut, task
             # maybe stop other tasks
             if _should_stop_others(done):
+                print("stopping others")
```
nfcampos, line 214: remove

PR #2379, opened 2024-11-08
cli: add ability to output docker compose file

eyurtsev:
- qxqx
- qxqx


libs/cli/langgraph_cli/cli.py
```diff
@@ -360,14 +359,86 @@ def build(
...
+        "Add additional files for running the LangGraph API server with "
+        "docker-compose. These files include a docker-compose.yml, .env file, "
+        "and a .dockerignore file."
+    ),
+    is_flag=True,
+)
 @log_command
-def dockerfile(save_path: pathlib.Path, config: pathlib.Path):
-    with open(config) as f:
+def dockerfile(save_path: str, config: pathlib.Path, add_docker_compose: bool) -> None:
```
vbarda, line 433: is the type change from `pathlib.Path` to `str` correct? think it'll still be path

vbarda (approved):

libs/cli/langgraph_cli/cli.py
eyurtsev: yeah -- the types are str rather than Paths. click's path is not pathlib's Path

PR #2388, opened 2024-11-11
lib: update_state(values=None) should clear all tasks

nfcampos:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -478,6 +478,11 @@ def _prepare_state_snapshot(
                 checkpointer=self.checkpointer or None,
                 manager=None,
             )
+            print(
```
vbarda, line 481: we should remove print statements

libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -197,6 +197,7 @@ def apply_writes(
     # sort tasks on path
     tasks = sorted(tasks, key=lambda t: t.path)
 
+    print("versions_seen", checkpoint["versions_seen"], [task.name for task in tasks])
```
vbarda, line 200: another print statement

Issue #2407, opened 2024-11-13
AsyncPostgresSaver uses hardcoded pipeline=True even if no pipeline is available


epistoteles:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
any usage of the AsyncPostgresSaver where pipelines are not supported
```

### Error Message and Stack Trace (if applicable)

```shell
2024-11-13T13:41:45.856830355Z     await self.asgi_app(scope, receive, send)
2024-11-13T13:41:45.856836855Z   File "/tmp/8dd03de0847602f/antenv/lib/python3.11/site-packages/opentelemetry/instrumentation/asgi/__init__.py", line 735, in __call__
2024-11-13T13:41:45.856842156Z     await self.app(scope, otel_receive, otel_send)
2024-11-13T13:41:45.856847056Z   File "/tmp/8dd03de0847602f/antenv/lib/python3.11/site-packages/quart/app.py", line 1693, in asgi_app
2024-11-13T13:41:45.856851957Z     await asgi_handler(receive, send)
2024-11-13T13:41:45.856865358Z   File "/tmp/8dd03de0847602f/antenv/lib/python3.11/site-packages/quart/asgi.py", line 52, in __call__
2024-11-13T13:41:45.856870358Z     raise_task_exceptions(done)
2024-11-13T13:41:45.856875059Z   File "/tmp/8dd03de0847602f/antenv/lib/python3.11/site-packages/quart/utils.py", line 180, in raise_task_exceptions
2024-11-13T13:41:45.856880059Z     raise task.exception()
2024-11-13T13:41:45.856884659Z   File "/tmp/8dd03de0847602f/antenv/lib/python3.11/site-packages/quart/asgi.py", line 107, in handle_request
2024-11-13T13:41:45.856889560Z     await asyncio.wait_for(self._send_response(send, response), timeout=timeout)
[...]

PR #2413, opened 2024-11-14
checkpoint-postgres: handle cases when conn.pipeline is not supported

vbarda:
Fixes https://github.com/langchain-ai/langgraph/issues/2407


libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
```diff
@@ -379,10 +386,17 @@ def _cursor(self, *, pipeline: bool = False) -> Iterator[Cursor[DictRow]]:
             elif pipeline:
                 # a connection not in pipeline mode can only be used by one
                 # thread/coroutine at a time, so we acquire a lock
-                with self.lock, conn.pipeline(), conn.cursor(
-                    binary=True, row_factory=dict_row
-                ) as cur:
-                    yield cur
+                try:
```
nfcampos, line 389: We can't do this without trying every time?

vbarda: good point, will update

nfcampos (approved):

PR #2419, opened 2024-11-14
testing out routing

eyurtsev:
- experimental: add handoff tool support
- example
- update
- qxqx
- x
- qxqx
- x
- x


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -684,11 +696,119 @@ def route_tool_responses(state: AgentState) -> Literal["agent", "__end__"]:
...
+    # The GraphCommand cannot be used here.
+    # This is just a temporary hack to show ROUTING via visualization.
+    # Any chance we could provide soem API to show this kind of routing?!
+    def agent_node(
+        # Unclear on what to do here with the state.
+        state: AgentState,
+        config: RunnableConfig,
+    ):
+        """Agent node that calls the language model and tools."""
+        # Here it's pretty confusing since we need to route only the "new" messages.
```
vbarda, line 779: not necessarily, right? can just add input/output processor

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -684,11 +696,119 @@ def route_tool_responses(state: AgentState) -> Literal["agent", "__end__"]:
...
+
+# Not obvious we should be creating a graph rather than adding a node directly
+# to a graph, so that "name" is specified once
+
+
+def make_agent_v2(
+    model: LanguageModelLike,
+    tools: Union[ToolExecutor, Sequence[BaseTool], ToolNode],
+    # What do we do with things like is_last_step and remaining_steps?
+    # We definitely do not want them to be part of the shared state.
```
vbarda, line 760: why not? why is that an issue?

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -684,11 +696,119 @@ def route_tool_responses(state: AgentState) -> Literal["agent", "__end__"]:
...
+    route_names = [name for name, _ in routes]
+
+    if default_route not in route_names:
+        raise ValueError(
+            f"Default route '{default_route}' not found in routes {route_names}"
+        )
+
+    def router(state: graph.schema) -> Literal[*route_names, "__end__"]:
+        """Router node that determines where to go next."""
+        return Send(
```
vbarda, line 733: i think this should just be `return state.get("route_id", default_route)`

libs/langgraph/langgraph/prebuilt/handoff.py
```diff
@@ -1,37 +1,55 @@
...
+# This handoff is too limited. We want to make sure that we can update
+# graph states as part of the handoff or add a human-in-the-loop step w/ an interrupt.
+# THe primitive needs to look more like a regular tool.
+# We either push information into type annotation or we use a separate class.
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class Handoff:
+    node: str
+    parent_update: Optional[Any] = None
+    internal_update: Optional[Any] = None
```
vbarda, line 19: i don't actually understand what this one means -- the state update from the handoff will always be on the same level as the node we're handing off to

vbarda: nevermind, makes more sense now (updating the state on the level of a single react agent vs updating it on the level of the parent graph / multi-agent system)

PR #2475, opened 2024-11-19
Adding connection pool support to the PostgresStore

tyler-ball:
🚧 🚧 🚧 
I would like some guidance on how to update the tests here - the PostgresSaver does not use mocks and has a check for the type of connection. When I copied this code for the PostgresStore its causing test failures. Adding the mocks as accepted classes does not seem correct.

-----

This PR is in reference to https://github.com/langchain-ai/langgraph/discussions/2457

I went ahead and coded an implementation for accepting ConnectionPool when creating a PostgresStore. I copied the PostgresSaver implementation for guideance.

One question I have - for the batch code it was trying to execute all cursors and then collect all the results after all cursors have been executed. In order to use the context management block I had to switch the implementation to execute each cursors sequentially. Is this an unacceptable implementation?


hinthornw:
Oops i may have done duplicate work in https://github.com/langchain-ai/langgraph/pull/2494

hinthornw (approved):

dav3yblaz3:
great job team we test if fix

lukasugar:
Hey, thanks for adding this! Is there an example of usage of the component, with connection pooling? Thanks

dav3yblaz3 (commented):
ok to try for test thanks team and yes seems acceptable for now use

dav3yblaz3 (commented):
let us test

PR #2489, opened 2024-11-20
[CLI] Validate node version

hinthornw:


libs/cli/langgraph_cli/config.py
```diff
@@ -37,10 +39,13 @@ def validate_config(config: Config) -> Config:
     )
 
     if config.get("node_version"):
-        if config["node_version"] not in ("20",):
+        if config["node_version"] not in SUPPORTED_NODE_VERSIONS:
+            supported = ", ".join(
+                f'`node_version: "{v}"`' for v in SUPPORTED_NODE_VERSIONS
```
jacoblee93, line 44: If it were me I'd keep it a little bit more general and just disallow minor version pins rather than looking for specific values

Not sure how often this gets updated or if it'd be out of sync with the main LangGraph instances

nfcampos: just to confirm this is reading version from package.json or from langgraph.json?

hinthornw: package.json, but since we already had a check for `node_version == "20"` i made a constant for supported versions and used in both places.

hinthornw: Going to just assert:
- is only a major version (no minor or patch) if provided
- Is >= 20

jacoblee93 (approved):

hinthornw:
Discussed offline: in a future PR we'd like to rm the node/python_version field in the langgraph.json file

PR #2502, opened 2024-11-21
langgraph: fix issue w/ type annotations in tools_condition

vbarda:


hinthornw (approved):

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -35,15 +32,13 @@
 from langchain_core.tools import BaseTool, InjectedToolArg
 from langchain_core.tools import tool as create_tool
 from langchain_core.tools.base import get_all_basemodel_annotations
+from pydantic import BaseModel
```
eyurtsev, line 35: do we expect from pydantic.v1 import BaseModel to work?

PR #2516, opened 2024-11-22
langgraph: use input schema from conditional edge

vbarda:
Currently we ignore the input schema in the branch and instead use the input schema from the previous node (or overall graph schema)

This change makes the input schema to branches respected. This means that if you try to pass extra keys and they're NOT in the input schema, you will receive an error. If you don't provide an annotation in the router, it will fall back to the previous node's input schema / full graph state schema

Alternative solution is to just ignore the input schema in the router altogether (including ignoring the schema from previous node / full graph), but personally I find it more confusing.


libs/langgraph/tests/test_state.py
```diff
@@ -318,3 +319,38 @@ def class_method(self, state):
 
     # class method
     assert _get_node_name(MyClass().class_method) == "class_method"
+
+
+def test_input_schema_conditional_edge():
```
vbarda, line 324: this tests a related issue, will add another explicit test for different schema in conditional edge

libs/langgraph/langgraph/graph/state.py
```diff
@@ -760,12 +760,37 @@ def branch_writer(
                     config, cast(Sequence[Union[Send, ChannelWriteEntry]], writes)
                 )
 
+        # detect branch input schema
+        input = None
+
+        # detect input schema annotation in the branch callable
+        try:
+            if (
+                isinstance(branch.path, RunnableCallable)
```
hinthornw, line 769: What if it's a RUnnableLambda?

vbarda: yea might need to check for that too just in case

PR #2518, opened 2024-11-22
fix: Replace hardcoded PostgresSaver with cls in factory method

phoenixAja:
Replace hardcoded `PostgresSaver` with `cls` in the `from_conn_string` factory method to improve subclassing support.

## Changes
- Modified `from_conn_string` to use `cls` instead of `PostgresSaver` when instantiating new instances
- This change allows subclasses to properly inherit and override the factory method

## Why
This refactor makes the `PostgresSaver` class more extensible by following Python's convention of using `cls` in class methods. This enables proper inheritance patterns where subclasses can reuse the factory method without modification.

## Testing
The change is backward compatible and doesn't alter existing functionality. All existing tests should continue to pass.


vbarda:
Thanks -- could you please also do this for async postgres saver too?

phoenixAja:
> Thanks -- could you please also do this for async postgres saver too?

for sure!

PR #2524, opened 2024-11-25
Correct annotation for file_name in read_document function

geeekfa:
The current annotation states "File path to save the document." which implies that the function saves data to the file. However, read_document is only used to read the contents of the file, not save to it. This PR changes the annotation to "File path to read the document." to more accurately describe the parameter’s purpose.


vbarda:
Thank you! This should already be fixed in the docs

PR #2525, opened 2024-11-25
Fix Return Type for create_team_supervisor Function to Reflect Correc…

geeekfa:
…t Output Structure

I noticed that the return type of the create_team_supervisor function was incorrectly defined as -> str, but the function does not simply return a string. Instead, it returns a complex object involving a series of chained operations, including function bindings and JSON parsing. To reflect this behavior accurately, I updated the return type to -> Any, which is more appropriate for the dynamic nature of the return value.


vbarda:
Thank you for your contribution -- this code is actually outdated and is no longer used. See the latest version here https://langchain-ai.github.io/langgraph/tutorials/multi_agent/hierarchical_agent_teams/#create-tools

PR #2526, opened 2024-11-25
feat: Vector Search

hinthornw:
Right now you enable it by:
- Initializing the store with an 'embedding config' -> this contains the 'dims' (used to create the table) and the encoder object (rn langchain embeddings object, though that is ......) 
- Call setup() -> creates the vector table.

Each document has 1 or more vectors associated with it for each json path in the embedding config.

Would welcome critique and requests! 

Especially around handling migrations, whether we want to support multimedia (multimodal embeddings is old school already so kinda bad langchain embeddings classes don't support well), querying (do we need mmr or other strategies...?) and anything else that feels off.

Also happy to shelve the use of json path for saving embeddings if we think that's too complicated.

Also the default `__root__` embedding of the whole JSON object retains the keys. I'm guessing there's a bit of an impact here but probably better to have it remain self descriptive than to drop the keys before embedding the object.

### DIFF
--- a/libs/checkpoint-postgres/langgraph/store/postgres/aio.py
+++ b/libs/checkpoint-postgres/langgraph/store/postgres/aio.py
@@ -76,12 +76,6 @@ class AsyncPostgresStore(AsyncBatchedBaseStore, BasePostgresStore[_ainternal.Con
         conn: AsyncConnection[DictRow],
     ) -> None:
         async with self._cursor(conn, pipeline=True) as cur:
-            if PutOp in grouped_ops:
-                await self._batch_put_ops(
-                    cast(Sequence[tuple[int, PutOp]], grouped_ops[PutOp]),
-                    cur,
-                )
-
             if GetOp in grouped_ops:
                 await self._batch_get_ops(
                     cast(Sequence[tuple[int, GetOp]], grouped_ops[GetOp]),
@@ -103,6 +97,12 @@ class AsyncPostgresStore(AsyncBatchedBaseStore, BasePostgresStore[_ainternal.Con
                         grouped_ops[ListNamespacesOp],
                     ),
                     results,
+                    cur,
+                )
+
+            if PutOp in grouped_ops:
+                await self._batch_put_ops(
+                    cast(Sequence[tuple[int, PutOp]], grouped_ops[PutOp]),
                     cur,
                 )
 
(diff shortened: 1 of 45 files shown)


libs/checkpoint-postgres/langgraph/store/postgres/aio.py
```diff
@@ -260,33 +314,59 @@ async def setup(self) -> None:
...
+                        v INTEGER PRIMARY KEY
                     )
-                for v, migration in enumerate(
-                    self.MIGRATIONS[version + 1 :], start=version + 1
-                ):
-                    await cur.execute(migration)
+                """
+                )
+
+            for v, migration in enumerate(
```
hinthornw, line 337: Would love critique/suggestions on handling migrations.
Basically, the table needs the emb dimensionality & needs the extension to be installed. It breaks compatibility if we add pgvector as a hard requirement here since you have to install it on the target system.

On the bright side, it's a separate table, so not the worst thing to drop and then re-create if you're early on

hinthornw (commented):
several todos

libs/checkpoint-postgres/langgraph/store/postgres/aio.py
```diff
@@ -167,40 +217,42 @@ async def _batch_list_namespaces_ops(
 
     @asynccontextmanager
     async def _cursor(
-        self, conn: AsyncConnection[DictRow], *, pipeline: bool = False
-    ) -> AsyncIterator[AsyncCursor[Any]]:
+        self, *, pipeline: bool = False
```
hinthornw, line 220: TODO: This change should have been in parent branch

libs/checkpoint/langgraph/store/base/__init__.py
```diff
@@ -73,6 +73,52 @@ def dict(self) -> dict:
...
+    ) -> None:
+        """Initialize a result item.
+
+        Args:
+            namespace: Hierarchical path to the item.
+            key: Unique identifier within the namespace.
+            value: The stored value.
+            created_at: When the item was first created.
+            updated_at: When the item was last updated.
+            response_metadata: Optional metadata about the response/result.
```
hinthornw, line 105: Rn contains the similarity score.

hinthornw:
Transfered to #2526

PR #2535, opened 2024-11-26
feat: Add vector search

hinthornw:
- Initializing the store with an 'embedding config' -> this contains the 'dims' (used to create the table) and the encoder object (rn langchain embeddings object, though that is ......) 
- Call setup() -> creates the vector table.

Each document has 1 or more vectors associated with it for each json path in the embedding config.

Would welcome critique and requests! 

Leaving the params as the defaults for pgvector but open to feedback if you think it's important to be able to more transparently configure that in setup()

```python
from typing import TypedDict, List, Dict, Any, Optional

from langchain_openai import OpenAIEmbeddings
from langgraph.store.postgres import PostgresStore

emb_config = {
    "dims": 1536,  # OpenAI embedding dimensions
    "embed": OpenAIEmbeddings(model="text-embedding-3-small"),
    "distance_type": "cosine",
}
with PostgresStore.from_conn_string(
    "postgres://postgres:postgres@localhost:5441",
    embedding=emb_config,
) as store:
    store.setup()


# Define the state type for our graph
class State(TypedDict):
    query: str
    results: Optional[List[Dict[str, Any]]]


def put_stuff(state: State) -> State:
    docs = [
        ("doc1", {"text": "red apple in kitchen"}),
        ("doc2", {"text": "blue car in garage"}),
        ("doc3", {"text": "green apple on table"}),
    ]
    for key, value in docs:
        store.put(("docs",), key, value)


def search_stuff(state: State) -> State:
    """Search for documents using vector similarity."""
    results = store.search(("docs",), query=state["query"])

    return {"results": results}


builder = StateGraph(State)
builder.add_node(put_stuff)
builder.add_node(search_stuff)
builder.add_edge("__start__", "put_stuff")
builder.add_edge("put_stuff", "search_stuff")
# Compile
with PostgresStore.from_conn_string(
    "postgres://postgres:postgres@localhost:5441",
    embedding=emb_config,
) as store:
    chain = builder.compile(store=store)

    result = chain.invoke({"query": "sour apple"})

# Print results
for doc in result["results"]:
    print(doc.key)
    print(doc.value)
    print(doc.response_metadata)

```


libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -145,60 +325,162 @@ def _get_batch_PUT_queries(
...
+                    and self.embedding_config.get("distance_type") == "hamming"
+                ):
+                    score_expr = score_expr % ("%s", self.embedding_config["dims"])
+                else:
+                    score_expr = score_expr % ("%s", vector_type)
+
+                vectors_per_doc_estimate = self.embedding_config[
+                    "__estimated_num_vectors"
+                ]
+                expanded_limit = (op.limit * vectors_per_doc_estimate * 2) + 1
```
hinthornw, line 431: flagging this query as being approximate.

Ideally we:
1. Get all values matching filters (in namespace prefix, etc.)
2. Get all the store_vectors for those docs
3. Sort by score
4. Select the top K items based on max(score) over the vectors

If we materialize a large view  in (1), that makes the search index less useful (i think) and in general could use a ton of memory.

As a compromise right now I'm running this in. CTE and returning up to K * N where N is twice the estimated number of vectors per doc.

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -59,8 +75,56 @@
 -- For faster lookups by prefix
 CREATE INDEX IF NOT EXISTS store_prefix_idx ON store USING btree (prefix text_pattern_ops);
 """,
+    Migration(
```
hinthornw, line 78: Separate list & table

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -89,10 +153,76 @@ class PoolConfig(TypedDict, total=False):
     """
 
 
+class DBIndexConfig(TypedDict, total=False):
+    """Configuration for vector index in PostgreSQL store."""
+
+    kind: Literal["hnsw", "ivfflat"]
```
hinthornw, line 159: Check perf impact of initializing empty

hinthornw: See rebuild index for ivfflat

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -89,10 +153,76 @@ class PoolConfig(TypedDict, total=False):
...
+    """
+
+
+class PostgresIndexConfig(IndexConfig, total=False):
+    """Configuration for vector embeddings in PostgreSQL store with pgvector-specific options.
+
+    Extends EmbeddingConfig with additional configuration for pgvector index and vector types.
+    """
+
+    db_index_config: Union[HNSWConfig, IVFFlatConfig]
```
hinthornw, line 204: ann_config

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -144,60 +277,164 @@ def _get_batch_PUT_queries(
...
-                SET value = EXCLUDED.value, updated_at = CURRENT_TIMESTAMP
+                SET value = EXCLUDED.value,
+                    updated_at = CURRENT_TIMESTAMP
             """
             queries.append((query, insertion_params))
 
-        return queries
+            if vector_values:
+                values_str = ",".join(vector_values)
+                query = f"""
```
hinthornw, line 335: We aren't thresholding on the limit or anything here / not handling limit so could go over

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -144,60 +277,164 @@ def _get_batch_PUT_queries(
...
+                expanded_limit = (op.limit * vectors_per_doc_estimate * 2) + 1
+
+                # Direct query with DISTINCT ON to get best score per document
+                base_query = f"""
+                    with scored as (
+                        SELECT DISTINCT ON (s.prefix, s.key)
+                            s.prefix, s.key, s.value, s.created_at, s.updated_at,
+                            {score_expr} as score
+                        FROM store s
+                        JOIN store_vectors sv ON s.prefix = sv.prefix AND s.key = sv.key
```
hinthornw, line 394: should be inner join

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -144,60 +277,164 @@ def _get_batch_PUT_queries(
...
+                    score_expr = score_expr % ("%s", self.index_config["dims"])
+                else:
+                    score_expr = score_expr % ("%s", vector_type)
+
+                vectors_per_doc_estimate = self.index_config["__estimated_num_vectors"]
+                expanded_limit = (op.limit * vectors_per_doc_estimate * 2) + 1
+
+                # Direct query with DISTINCT ON to get best score per document
+                base_query = f"""
+                    with scored as (
```
hinthornw, line 389: cte unnecessary. do one statement

libs/checkpoint-postgres/langgraph/store/postgres/base.py
```diff
@@ -144,60 +277,164 @@ def _get_batch_PUT_queries(
...
+                        SELECT DISTINCT ON (s.prefix, s.key)
```
hinthornw, line 390: Does this execute before or after the order by

libs/checkpoint-postgres/langgraph/store/postgres/base.py
hinthornw: Going to defer ANN implementation

hinthornw:
Combining ANN indexing with b-tree indexing can be complicated so Ii'll add ANN support (HNSW/IVFFLAT) in a followup

PR #2548, opened 2024-11-27
[CLI] Inject langgraph.json's store config in docker

hinthornw:


libs/cli/langgraph_cli/config.py
```diff
@@ -399,6 +447,7 @@ def test_file(file_name):
 
 RUN cd {faux_path} && {install_cmd}
 
+ENV LANGGRAPH_STORE='{json.dumps(config["store"])}'
```
hinthornw, line 450: This is a lot of code changes to add a config thing. Thoughts on just doing `LANGGRAPH_CONFIG` or something and putting "store" in there along with potentially other things in the future to avoid all of this madness?

PR #2590, opened 2024-12-02
docs: Use edit mode as default to install template

Mingqi2:
Small change to install the dependencies with `edit` mode so that users or freshman can see the effect immediately when they change the template code. As below,
`pip install -e .`

It's very good to evaluate how agent works and easy to test & re-develop!

### DIFF
--- a/docs/docs/tutorials/langgraph-platform/local-server.md
+++ b/docs/docs/tutorials/langgraph-platform/local-server.md
@@ -35,10 +35,10 @@ Create a new app from the `react-agent` template. This template is a simple agen
 
 ## Install Dependencies
 
-In the root of your new LangGraph app, install the dependencies:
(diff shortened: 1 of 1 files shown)


Mingqi2:
Hi @eyurtsev , could you please help to review this PR? I think this is a more friendly change for template usage.

docs/docs/tutorials/langgraph-platform/local-server.md
```diff
@@ -35,10 +35,10 @@ Create a new app from the `react-agent` template. This template is a simple agen
 
 ## Install Dependencies
 
-In the root of your new LangGraph app, install the dependencies:
+In the root of your new LangGraph app, install the dependencies with `edit` mode (You can see the effect immediately when you change the template code) :
```
hinthornw, line 38:
```suggestion
In the root of your new LangGraph app, install the dependencies in `edit` mode so your local changes are used by the server:
```

PR #2635, opened 2024-12-04
docs: add Command docs

vbarda:


libs/langgraph/langgraph/graph/state.py
```diff
@@ -86,7 +86,18 @@ def _get_node_name(node: RunnableLike) -> str:
 
 @dataclasses.dataclass(**_DC_KWARGS)
 class GraphCommand(Generic[N], Command[N]):
-    """One or more commands to update a StateGraph's state and go to, or send messages to nodes."""
+    """One or more commands to update a StateGraph's state and go to, or send messages to nodes.
+
+    Args:
+        goto: name of the node to navigate to next.
```
eyurtsev, line 92: `goto`: can this go into an arbitrary node or not? What are the restrictions?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -86,7 +86,18 @@ def _get_node_name(node: RunnableLike) -> str:
...
+            If not specified, the graph will halt after executing the current superstep.
+        graph: graph to send the command to. Supported values are:
```
eyurtsev, line 94: `graph` feels like this should be the first parameter so it's clear that the rest of the things are sent to this graph

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,52 @@ def continue_to_jokes(state: OverallState):
 graph.add_conditional_edges("node_a", continue_to_jokes)
 ```
 
+## `GraphCommand`
+
+Typically, LangGraph separates control flow (edges) from state updates (nodes). However, it is often beneficial to combine the two. For example, you might want to BOTH perform state updates AND decide which node to go next in the SAME node. LangGraph provides a way to combine control flow and node state updates using [`GraphCommand`][langgraph.graph.state.GraphCommand]. To do so, you can return a `GraphCommand` object from a node instead of a state update or `Send` objects.
```
eyurtsev, line 327: grammar "decide which node to go next in the SAME node."

eyurtsev, line 327: nit: would see if possible to reduce text to 1-2 sentences

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,52 @@ def continue_to_jokes(state: OverallState):
...
+
+`GraphCommand` has the following properties:
+
+  - `goto`: optional, name of the node to navigate to next.
```
eyurtsev, line 331: Tabular format may be easier on the eye

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,52 @@ def continue_to_jokes(state: OverallState):
...
+        If not specified, the graph will halt after executing the current superstep.
+  - `graph`: optional, graph to send the command to. Supported values are:
```
eyurtsev, line 333: graph parameter should be first I think

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,52 @@ def continue_to_jokes(state: OverallState):
...
+    - `None`: the current graph (default)
+    - `GraphCommand.PARENT`: parent graph.
+  - `update`: optional, state update to apply to the graph's state at the current superstep.
+  - `send`: optional, list of [`Send`](#send) objects to send to other nodes.
+  - `resume`: optional, value to resume execution with. Will be used when `interrupt()` is called.
+
+```python
+from langgraph.graph import GraphCommand, StateGraph, START
+from typing_extensions import TypedDict, Literal
```
eyurtsev, line 342: sort order

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,52 @@ def continue_to_jokes(state: OverallState):
...
+
+With `GraphCommand` you can also achieve dynamic control flow behavior (identical to [conditional edges](#conditional-edges)):
+
+```python
+def my_node(state: State) -> GraphCommand[Literal["my_other_node", "__end__"]]:
+    if state["foo"] == "bar":
+        return GraphCommand(update={"foo": "baz"}, goto="my_other_node")
+    else:
+        return GraphCommand(goto="__end__")
+```
```
eyurtsev, line 369: link to how-to guide here for the example?

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,52 @@ def continue_to_jokes(state: OverallState):
 graph.add_conditional_edges("node_a", continue_to_jokes)
 ```
 
+## `GraphCommand`
```
eyurtsev, line 325: Could we somehow float this section to the top

```python
def my_node(state: State) -> GraphCommand[Literal["my_other_node"]]:
    return GraphCommand(update={"foo": "bar"}, goto="my_other_node")
```

These two lines is essentially everything a developer needs to know! Everything else is boiler plate

eyurtsev:
In the notebook:

1. Could you float this above the fold? 

```python
def my_node(state: State) -> GraphCommand[Literal["my_other_node"]]:
    return GraphCommand(update={"foo": "bar"}, goto="my_other_node")
```

These two lines is essentially everything a developer needs to know! Everything else is boiler plate


2. Remove the `Using edges` section. The how-to guide is supposed to show how to achieve something in the quickest possible way, not explain the alternatives. You could show other ways of doing it if you'd like but it would need to be below the solution not before it.

docs/docs/concepts/low_level.md
vbarda: yup!

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,63 @@ def continue_to_jokes(state: OverallState):
...
+        # state update
+        update={"foo": "bar"},
+        # control flow
+        goto="my_other_node"
+    )
+```
+
+`GraphCommand` has the following properties:
+
+| Property | Description |
```
eyurtsev, line 341: could you format?

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,63 @@ def continue_to_jokes(state: OverallState):
 graph.add_conditional_edges("node_a", continue_to_jokes)
 ```
 
+## `GraphCommand`
+
+It can be useful to combine control flow (edges) and state updates (nodes). For example, you might want to BOTH perform state updates AND  decide which node to go to next in the SAME node. LangGraph provides a way to do so by returning a [`GraphCommand`][langgraph.graph.state.GraphCommand] object from node functions:
```
nfcampos, line 327:
```suggestion
It can be useful to combine control flow (edges) and state updates (nodes). For example, you might want to BOTH perform state updates AND decide which node to go to next in the SAME node. LangGraph provides a way to do so by returning a [`GraphCommand`][langgraph.graph.state.GraphCommand] object from node functions:
```

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +322,63 @@ def continue_to_jokes(state: OverallState):
...
+        goto="my_other_node"
+    )
+```
+
+`GraphCommand` has the following properties:
+
+| Property | Description |
+| --- | --- |
+| `graph` | Graph to send the command to. Supported values:<br>- `None`: the current graph (default)<br>- `GraphCommand.PARENT`: parent graph |
+| `goto` | Name of the node to navigate to next. Can be any node that belongs to the specified `graph` (current or parent). If `goto` not specified, the graph will halt after executing the current superstep. |
```
nfcampos, line 344: `If `goto` not specified, the graph will halt after executing the current superstep.` this is only true if there are no other edges out of this node, and no other nodes running at same time navigated elsewhere

vbarda: yup good point will clarify!

docs/docs/concepts/low_level.md
vbarda: what do you mean? like bold?

docs/docs/concepts/low_level.md
```diff
@@ -283,6 +283,9 @@ You can optionally provide a dictionary that maps the `routing_function`'s outpu
 graph.add_conditional_edges("node_a", routing_function, {True: "node_b", False: "node_c"})
 ```
 
+!!! tip
+    Use [`Command`](#command) instead of conditional edges if you need to combine state updates and routing.
```
nfcampos, line 287:
```suggestion
    Use [`Command`](#command) instead of conditional edges if you want to combine state updates and routing in a single function.
```

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +325,63 @@ def continue_to_jokes(state: OverallState):
...
+```
+
+`Command` has the following properties:
+
+| Property | Description |
+| --- | --- |
+| `graph` | Graph to send the command to. Supported values:<br>- `None`: the current graph (default)<br>- `Command.PARENT`: closest parent graph |
+| `update` | State update to apply to the graph's state at the current superstep |
+| `resume` | Value to resume execution with. Will be used when `interrupt()` is called |
+| `goto` | Can be one of the following:<br>- name of the node to navigate to next (any node that belongs to the specified `graph`)<br>- list of node names to navigate to next<br>- `Send` object<br>- sequence of `Send` objects<br>If `goto` is not specified and there are no other tasks left in the graph, the graph will halt after executing the current superstep. |
```
nfcampos, line 349: list v sequence, best be consistent

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +325,63 @@ def continue_to_jokes(state: OverallState):
...
+```
+
+With `Command` you can also achieve dynamic control flow behavior (identical to [conditional edges](#conditional-edges)):
+
+```python
+def my_node(state: State) -> Command[Literal["my_other_node", "__end__"]]:
+    if state["foo"] == "bar":
+        return Command(update={"foo": "baz"}, goto="my_other_node")
+    else:
+        return Command(goto="__end__")
```
nfcampos, line 380: this doesn't do anything

docs/docs/concepts/low_level.md
```diff
@@ -322,6 +325,63 @@ def continue_to_jokes(state: OverallState):
 graph.add_conditional_edges("node_a", continue_to_jokes)
 ```
 
+## `Command`
+
```
nfcampos, line 329: I think you forgot to explain why it helps to decorate functions with Command type hint

libs/langgraph/langgraph/types.py
```diff
@@ -245,7 +245,20 @@ def __eq__(self, value: object) -> bool:
...
 class Command(Generic[N]):
-    """One or more commands to update the graph's state and send messages to nodes."""
+    """One or more commands to update the graph's state and send messages to nodes.
+
+    Args:
+        graph: graph to send the command to. Supported values are:
+            - None: the current graph (default)
+            - GraphCommand.PARENT: closest parent graph
+        update: state update to apply to the graph's state at the current superstep.
+        resume: value to resume execution with. Will be used when `interrupt()` is called.
```
nfcampos, line 255:
```suggestion
        resume: value to resume execution with. To be used together with `interrupt()`.
```
Can we link to interrupt() here?

libs/langgraph/langgraph/types.py
```diff
@@ -245,7 +245,20 @@ def __eq__(self, value: object) -> bool:
...
 @dataclasses.dataclass(**_DC_KWARGS)
 class Command(Generic[N]):
-    """One or more commands to update the graph's state and send messages to nodes."""
+    """One or more commands to update the graph's state and send messages to nodes.
+
+    Args:
+        graph: graph to send the command to. Supported values are:
+            - None: the current graph (default)
+            - GraphCommand.PARENT: closest parent graph
+        update: state update to apply to the graph's state at the current superstep.
```
nfcampos, line 254:
```suggestion
        update: update to apply to the graph's state.
```

libs/langgraph/langgraph/types.py
```diff
@@ -245,7 +245,20 @@ def __eq__(self, value: object) -> bool:
...
+        resume: value to resume execution with. Will be used when `interrupt()` is called.
+        goto: can be one of the following:
+            - name of the node to navigate to next (any node that belongs to the specified `graph`)
+            - list of node names to navigate to next
+            - `Send` object
```
nfcampos, line 259:
```suggestion
            - `Send` object (to execute a node with the input provided)
```

docs/docs/concepts/low_level.md
vbarda: yea, what's a better way to do this? just `return None`?

docs/docs/concepts/low_level.md
vbarda: ya did that in the how to -- will copy over

docs/docs/concepts/low_level.md
vbarda: one upside of this is that it allows for nicer rendering if you do want to explicitly indicate that something goes to the `END`

nfcampos: just don't return anything

nfcampos (approved):

PR #2636, opened 2024-12-04
lib: Add support for multiple interrupts per node

nfcampos:
- Includes support for interrupt loops

### DIFF
--- a/libs/langgraph/langgraph/pregel/algo.py
+++ b/libs/langgraph/langgraph/pregel/algo.py
@@ -37,13 +37,13 @@ from langgraph.constants import (
     CONFIG_KEY_CHECKPOINT_NS,
     CONFIG_KEY_CHECKPOINTER,
     CONFIG_KEY_READ,
-    CONFIG_KEY_RESUME_VALUE,
+    CONFIG_KEY_SCRATCHPAD,
     CONFIG_KEY_SEND,
     CONFIG_KEY_STORE,
     CONFIG_KEY_TASK_ID,
+    CONFIG_KEY_WRITES,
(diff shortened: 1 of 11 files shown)


libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -37,13 +37,13 @@
     CONFIG_KEY_CHECKPOINT_NS,
     CONFIG_KEY_CHECKPOINTER,
     CONFIG_KEY_READ,
-    CONFIG_KEY_RESUME_VALUE,
```
hinthornw, line 40: Is CONFIG_KEY_RESUME_VALUE used anymore?

libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -263,8 +264,28 @@ def put_writes(self, task_id: str, writes: Sequence[tuple[str, Any]]) -> None:
         """Put writes for a task, to be read by the next tick."""
         if not writes:
             return
+        # deduplicate writes to special channels, last write wins
```
hinthornw, line 267: ooc: why do we know we don't have to deduplicate if one of the writes is to a regular channel?

libs/langgraph/langgraph/pregel/algo.py
nfcampos: No, removed now

libs/langgraph/langgraph/pregel/loop.py
nfcampos: Regular channels can't be written to more than once (writes after the first are ignored)

PR #2638, opened 2024-12-04
lib: Merge GraphCommand and Command

nfcampos:
- Now we have only Command
- Command(goto=) combines the previous functionality of Command(send=) and GraphCommand(goto=)


hinthornw (commented):
nice - ya would need the string support to finalize the merge right?

nfcampos:
No I can merge before that, and fix in a later pr

PR #2643, opened 2024-12-05
docs: update multi-agent tutorials to use Command

vbarda:


mathislindner:
Not sure if this is all intended, but I would like to add a comment to make sure this was not a mistake:
There are only conditional edges now (the old mermaid graph used to have uninterrupted edges)

## should be:
This is the descriptive representation from the tutorial (where the agent **ALWAYS**  reports back):

image

## mermaid graph has changed to:
and here is the new mermaid graph that contradicts the intended behavior: it used to have uninterrupted arrows, but now it only has dotted arrows ( **but the agent should always report back to the supervisor** , no!?):

image

PR #2646, opened 2024-12-05
docs: document interrupt & HIL

eyurtsev:
- Document interrupt reference
- Update conceptual guides for HIL
- Split time-travel conceptual guide
- Split breakpoints into separate conceptual guide
- Update relevant how-tos
- Update how-to index page for HIL with more information and recommendations
- New how-to for multi turn conversation


docs/docs/concepts/breakpoints.md
```diff
@@ -0,0 +1,281 @@
...
+1. [**Specify a checkpointer**](persistence.md#checkpoints) to save the graph state after each step.
+2. [**Set breakpoints**](#setting-breakpoints) to specify where execution should pause.
+3. Run the graph with a [**thread ID**](./persistence.md#threads) to pause execution at the breakpoint.
+4. [**Resume execution**](#resuming) from the paused state.
+
+## Setting breakpoints
+
+There are two places where you can set breakpoints:
+
+1. **Inside** a node using the [`interrupt` function](#the-interrupt-function) (or the older [`NodeInterrupt` exception](#nodeinterrupt-exception)).
```
vbarda, line 18: should we still mention the `NodeInterrupt`?

docs/docs/concepts/breakpoints.md
```diff
@@ -0,0 +1,281 @@
...
+Breakpoints pause graph execution at specific points, enabling [**human-in-the-loop**](./human_in_the_loop.md) workflows and debugging. Breakpoints are powered by  LangGraph's [**persistence layer**](./persistence.md), which saves the state after each graph step. 
+
+## Requirements
+
+To use breakpoints, you will need to:
+
+1. [**Specify a checkpointer**](persistence.md#checkpoints) to save the graph state after each step.
+2. [**Set breakpoints**](#setting-breakpoints) to specify where execution should pause.
+3. Run the graph with a [**thread ID**](./persistence.md#threads) to pause execution at the breakpoint.
+4. [**Resume execution**](#resuming) from the paused state.
```
vbarda, line 12: this link is broken

docs/docs/concepts/breakpoints.md
eyurtsev: yes,i think it's used likely by some users?

docs/docs/concepts/breakpoints.md
```diff
@@ -0,0 +1,289 @@
...
+The `Command` primitive provides several options to control and modify the graph's state during resumption:
+
+1. **Pass a value to the `interrupt`**: Provide data, such as a user's response, to the graph using `Command(resume=value)`. Execution resumes from the beginning of the node where the `interrupt` was used, however, this time the `interrupt(...)` call will return the value passed in the `Command(resume=value)` instead of pausing the graph. The `resume` value is only used when using `interrupt` as a breakpoint.
+ 
+       ```python
+       # Resume graph execution with the user's input.
+       graph.invoke(Command(resume={"age": "25"}), thread_config)
+       ```
+
+2. **Update the graph state**: Modify the graph state using `Command(update=update)`. Note that resumption starts from the beginning of the node where the `interrupt` was used. Execution resumes from the beginning of the node where the `interrupt` was used, but with the updated state.
```
vbarda, line 201:
```suggestion
2. **Update the graph state**: Modify the graph state using `Command(update=update)`. Execution resumes from the beginning of the node where the `interrupt` was used, but with the updated state.
```

docs/docs/concepts/breakpoints.md
```diff
@@ -0,0 +1,289 @@
...
+        error=None, 
+        interrupts=(Interrupt(value='value_in_interrupt', resumable=True, ns=['node_foo:5d8ffc92-8011-0c9b-8b59-9d3545b7e553'], when='during'),), state=None, 
+        result=None
+    ),
+) # Pending tasks. interrupts 
+```
+
+## How does resuming from a breakpoint work?
+
+> Resuming from a breakpoint is **different** from traditional breakpoints or Python's `input()` function, where execution resumes from the exact point where the breakpoint was triggered or where the `input()` function was called.
```
vbarda, line 246: i think this needs to be more prominent, like a `note` or a `warning`, this shows up like this right now and is easy to skip over

image

docs/docs/concepts/breakpoints.md
```diff
@@ -0,0 +1,289 @@
...
+Upon **resuming** the graph, the counter will be incremented a second time, resulting in the following output:
+
+```pycon
+> Entered the node: 2 # of times
+The value of counter is: 2
+```
+
+Keep the following considerations in mind when using the `interrupt` function:
+
+1. **Side effects**: Place side-effecting code, such as API calls, **after** the `interrupt` to avoid duplication, as these are re-triggered every time the node resumes.
```
vbarda, line 275: nit: side-effecting -> code with side effects?

docs/docs/how-tos/index.md
```diff
@@ -48,12 +48,24 @@ LangGraph makes it easy to manage conversation [memory](../concepts/memory.md) i
...
-- [How to add dynamic breakpoints](human_in_the_loop/dynamic_breakpoints.ipynb)
-- [How to edit graph state](human_in_the_loop/edit-graph-state.ipynb)
-- [How to wait for user input](human_in_the_loop/wait-user-input.ipynb)
+
+Key workflows:
+
+- [How to wait for user input](human_in_the_loop/wait-user-input.ipynb): A basic example that shows how to implement a human-in-the-loop workflow in your graph using the `interrupt` function.
+- [How to review tool calls](human_in_the_loop/review-tool-calls.ipynb): Incorporate human-in-the-loop for reviewing/editing/accepting tool call requests before they executed using the `interrupt` function.
+ 
+
```
vbarda, line 57: should we add `interrupt()` howto?

docs/docs/concepts/breakpoints.md
```diff
@@ -0,0 +1,289 @@
...
+# Compile the graph with a checkpointer
+graph = graph_builder.compile(checkpointer=checkpointer)
+
+# Run the graph until the breakpoint
+thread_config = {"configurable": {"thread_id": "some_id"}}
+for event in graph.stream(inputs, thread_config, stream_mode="values"):
+    print(event)
+```
+
+```pycon
```
vbarda, line 60: this should probably be `python`

docs/docs/concepts/breakpoints.md
eyurtsev: I played with both variants, this was actually nicer looking for me, but can swap to warning

docs/docs/how-tos/index.md
eyurtsev: No, I don't think so I removed it instead. I think we should avoid documenting how to use a particular part of the API in a how-to. How-tos are not API references, but a way to achieve specific objectives. The API reference for `interrupt()` has a basic usage example.

The easiest way to see that we have a problem is by looking at looking at the nonsensical content that we have for the how-to for HIL (e.g., how does updating state have anything to do with HIL? -- it's super unclear from looking at the index)

docs/docs/concepts/breakpoints.md
eyurtsev: Using `pycon` for output

docs/docs/concepts/v0-human-in-the-loop.md
```diff
@@ -0,0 +1,322 @@
+# Human-in-the-loop
```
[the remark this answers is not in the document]
eyurtsev: Not sure yet -- let me merge and we can collect opinion from folks after we publish.

----

The main thing is that we need to retain some documentation that's consistent with **current** user code. It's possible that this doc ins't good enough due to how-to guides having changed as well

PR #2656, opened 2024-12-05
langgraph: allow tools to return Command in tool node

vbarda:


libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -214,6 +215,26 @@ def _func(
...
+
+        commands: list[Command] = [
+            output for output in outputs if isinstance(output, Command)
+        ]
+        # This can be relaxed by moving to a design where there is a single node per tool
+        # In that case multiple Commands can be handled natively by LangGraph
+        # (including concurrent state updates via reducers, multiple goto destinations, etc.)
+        if len(commands) > 1:
+            raise ValueError(
+                "Currently only one Command update per ToolNode is supported, got multiple Commands."
```
eyurtsev, line 227: Add better debug information? Which tools are the offending tools? What should a user do?

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -214,6 +215,20 @@ def _func(
         config_list = get_config_list(config, len(tool_calls))
         with get_executor_for_config(config) as executor:
             outputs = [*executor.map(self._run_one, tool_calls, config_list)]
+
+        commands: list[Command] = [
+            output for output in outputs if isinstance(output, Command)
+        ]
+        if len(commands) > 0:
```
nfcampos, line 222: why is this? you can just wrap non-command outputs with command

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -214,6 +215,20 @@ def _func(
...
+            if len(commands) != len(outputs):
+                raise ValueError(
+                    f"Cannot mix Command and non-command (message) tool outputs, got the following outputs: {outputs}"
+                )
+
+            # Users that want to include ToolMessages in the state update
```
nfcampos, line 228: this comment doesn't make sense imo. users that don't do this can never use this thread again as then all future llm calls will fail

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -254,10 +282,33 @@ def _run_one(self, call: ToolCall, config: RunnableConfig) -> ToolMessage:
...
 
         try:
-            input = {**call, **{"type": "tool_call"}}
-            tool_message: ToolMessage = self.tools_by_name[call["name"]].invoke(
-                input, config
-            )
+            tool = self.tools_by_name[call["name"]]
+            # check if the Tool.func / Tool._run method returns a Command in the type annotation
+            tool_func = getattr(tool, "func", tool._run)
+            if (return_type := tool_func.__annotations__.get("return")) and (
```
nfcampos, line 288: why is this code block needed? if we need to update something in langchain core lets do so. we should not be relying on type hints for actual runtime behavior

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -295,10 +346,38 @@ async def _arun_one(self, call: ToolCall, config: RunnableConfig) -> ToolMessage
             return invalid_tool_message
 
         try:
-            input = {**call, **{"type": "tool_call"}}
-            tool_message: ToolMessage = await self.tools_by_name[call["name"]].ainvoke(
-                input, config
+            tool = self.tools_by_name[call["name"]]
```
nfcampos, line 349: same as above, this needs to change

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: was also contemplating wrapping, will just do that, cause technically nothing wrong with it. i just can't imagine a use case where mixing them is valuable but i am sure someone would find it

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: that's true for using within the same graph, but you might also want to jump to a different node inside the parent graph without any state updates (i.e. navigate to a different agent) without adding tool message. so not sure if we should automatically add tool messages / enforce this somehow

do you have any ideas?

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: good point -- we're checking the typehint to decide whether we need to get raw tool output (e.g. Command) or a tool message. let me see if it's possible to simplify this

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -244,24 +264,92 @@ async def _afunc(
...
+
+                    channels, messages_updates = zip(*updated_command.update)
+                    if len(channels) != 1 or channels[0] != "__root__":
+                        raise ValueError(
+                            f"When using list of messages as ToolNode input, Command.update can only contain a single update in the following format: `[('__root__', update)]`, got: {updated_command.update} for tool '{call['name']}'"
+                        )
+
+                    messages_update = messages_updates[0]
+
+                if len(messages_update) != 1 or not isinstance(
```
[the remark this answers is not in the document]
vbarda: `The very first message in the update must be a ToolMessage` i go even further and just enforce a single message right now. but maybe that's too strict?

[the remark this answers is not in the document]
vbarda: I could potentially see people wanting to return other things like `RemoveMessage` (or update existing messages)

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -244,24 +264,92 @@ async def _afunc(
...
+                    raise ValueError(
+                        f"Expected exactly one ToolMessage in Command.update for tool '{call['name']}', got: {messages_update}"
+                    )
+
+                tool_message = messages_update[0]
+                tool_message.name = call["name"]
+                tool_message.tool_call_id = cast(str, call["id"])
+                return updated_command
+            else:
+                return ToolMessage(
```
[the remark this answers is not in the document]
vbarda: yea that's possible - do you think we should raise?

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -244,24 +264,92 @@ async def _afunc(
...
+            else:
+                combined_outputs.append(
+                    [output] if output_type == "list" else {self.messages_key: [output]}
+                )
+        return combined_outputs
+
+    def _run_one(
+        self,
+        call: ToolCall,
+        output_type: Literal["list", "dict"],
```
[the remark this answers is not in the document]
vbarda: agreed, will add comments

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -244,24 +264,92 @@ async def _afunc(
...
+                        raise ValueError(
+                            f"When using dict with '{self.messages_key}' key as ToolNode input, tools must provide a dict in Command.update, got: {response.update} for tool '{call['name']}'"
+                        )
+
+                    state_update = updated_command.update or {}
+                    messages_update = state_update.get(self.messages_key, [])
+                else:
+                    if output_type != "list":
+                        raise ValueError(
+                            f"When using list of messages as ToolNode input, tools must provide `[('__root__', update)]` in Command.update, got: {response.update} for tool '{call['name']}'"
```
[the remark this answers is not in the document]
vbarda: i updated a bit to clarify (need to actually double check that it's working as intended). this is also a niche use case cause i think most people use dicts as state schema anyway and i imagine that it'd be similar for ToolNode + Command

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: resolved this offline -- we'll raise an error if a tool message is not provided in `Command.update`

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -453,6 +531,54 @@ def _inject_tool_args(
...
+
+            messages_update = messages_updates[0]
+        else:
+            return command
+
+        # convert to message objects if updates are in a dict format
+        messages_update = convert_to_messages(messages_update)
+        if len(messages_update) != 1 or not isinstance(messages_update[0], ToolMessage):
+            raise ValueError(
+                f"Expected exactly one message (ToolMessage) in Command.update for tool '{call['name']}', got: {messages_update}. "
```
vbarda, line 570: i might relax this to allow multiple messages but only one tool message

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -876,7 +876,8 @@ def test_tool_node_individual_tool_error_handling():
 
     tool_message: ToolMessage = result_individual_tool_error_handler["messages"][-1]
     assert tool_message.type == "tool"
-    assert tool_message.status == "error"
+    # TODO: figure out how to propagate this properly
+    # assert tool_message.status == "error"
```
vbarda, line 880: will be fixed after some core changes

libs/langgraph/langgraph/prebuilt/tool_node.py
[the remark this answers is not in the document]
vbarda: removed this completely actually

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -244,24 +264,92 @@ async def _afunc(
...
-            tool_message: ToolMessage = self.tools_by_name[call["name"]].invoke(
-                input, config
-            )
-            tool_message.content = cast(
-                Union[str, list], msg_content_output(tool_message.content)
-            )
-            return tool_message
+            tool = self.tools_by_name[call["name"]]
+            if tool.response_format != "content":
+                # handle "content_and_artifact"
```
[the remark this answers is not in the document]
vbarda: this will actually not be needed once some changes in core are merged, will update

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -43,11 +45,15 @@
 TOOL_CALL_ERROR_TEMPLATE = "Error: {error}\n Please fix your mistakes."
 
 
+class InvalidToolCommandError(Exception):
```
eyurtsev, line 48: Are there other exceptions in pre-built?

Should it really be sub-classing from Exception?

What does it mean for a Command returned by a tool to be invalid?

libs/langgraph/langgraph/types.py
```diff
@@ -244,7 +252,7 @@ def __eq__(self, value: object) -> bool:
 
 
 @dataclasses.dataclass(**_DC_KWARGS)
-class Command(Generic[N]):
+class Command(Generic[N], ToolOutputMixin):
```
eyurtsev, line 255: @nfcampos could you confirm you're OK with this change?

libs/langgraph/langgraph/prebuilt/tool_node.py
vbarda: figured out how to avoid this altogether!

PR #2658, opened 2024-12-05
lib: Add support for returning multiple commands from a node

nfcampos:


libs/langgraph/tests/test_pregel.py
```diff
@@ -220,18 +220,6 @@ def logic(inp: str) -> str:
     class State(TypedDict):
         hello: str
 
-    def node_a(state: State) -> State:
```
vbarda, line 223: ooc, why remove this?

nfcampos: this isn't intended to raise exception, not sure why this was still here

PR #2670, opened 2024-12-06
docs: add a how-to on updating state from tools

vbarda:


eyurtsev:
for how to guide:


MAJOR:

1. The how-to guide makes it seem like it works for any tool, but it only works for tools that are executed using the create react agent. 

MINOR:

2. the snippet at the top of the how-to guide is doing the reverse of what the code does

- snippet shows something based on state and looks up customer id
- actual code takes id from config and looks up user info


NIT:

3. Perhaps separate the user info information into a separate cell so it doesn't appear above the imports (a bit jarring to see that)

vbarda:
> 1. The how-to guide makes it seem like it works for any tool, but it only works for tools that are executed using the create react agent.

Thanks for the feedback! addressed (2) and (3) and added a callout for (1)

ShetRo:
I tested the above code,  by commenting out the state modifier as below, and still get the same result.. also when i try to print out the state status, i only see Message , i dont see other parameter such as last_name and user_info. 

agent = create_react_agent(
    model,
    # pass the tool that can update state
    [lookup_user_info],
    state_schema=State,
    # pass dynamic prompt function
    #state_modifier=state_modifier,
)

vbarda:
> I tested the above code, by commenting out the state modifier as below, and still get the same result.. also when i try to print out the state status, i only see Message , i dont see other parameter such as last_name and user_info.
> 
> agent = create_react_agent( model, # pass the tool that can update state [lookup_user_info], state_schema=State, # pass dynamic prompt function #state_modifier=state_modifier, )

@ShetRo note that this notebook is on top of another PR that hasn't been merged and released yet. you'd need to install langgraph from source if you want to test this

ShetRo:
thanks my bad

PR #2681, opened 2024-12-09
Add rollback for aio setup

jcampbell:
ROLLBACK AND CHAIN if the checkpoint migrations table is missing.

I observed the current rollback attempt not to work on an uninitialized data source, because the transaction is left in an error state by the aio carry forward of the standard saver.


hinthornw:
Do you have a code snippet of how you're using the setup that makes this necessary?

jcampbell:
Thanks for the follow up. In looking at the docs more closely to confirm that I hadn't missed something, I realized of course that I had: the checkpointer assumes psycopg is set to autocommit. With autocommit, this is not needed; I'll close the PR.

PR #2682, opened 2024-12-09
lib: Treat Command as "resuming" signal

nfcampos:
- so it works w interrupt_before/after


eyurtsev:
Pushed an empty commit benchmarking code was in the 20+ mins?

PR #2683, opened 2024-12-09
lib: Add support for invoke(Command(goto=<str>))

nfcampos:


libs/langgraph/langgraph/graph/state.py
```diff
@@ -559,6 +559,7 @@ def compile(
         for key, node in self.nodes.items():
             compiled.attach_node(key, node)
 
+        compiled.attach_branch(START, SELF, CONTROL_BRANCH, with_reader=False)
```
jacoblee93, line 562: How would this trigger?

PR #2784, opened 2024-12-16
docs: reorg multi-agent howtos

vbarda:


eyurtsev:
MAJOR:

1) Consider improving the header titles. For a how-to guide, we're not trying to find a specific example, we're trying to find a specific implementation (e.g., I want to use tools with react).  `Travel Recommendations Example (ReAct)` -> Most users will not know what `react` is. They're looking for something that shows how tools can be used / added to agents.

We can create `examples` section at the very bottom if we want to add various examples.

2) What should users with custom agents do? How do custom agent implementations typically look?

3) Anyway to cut on complexity using 2 agents and still keep it meaningful?


MINOR:


3) Should we include a TOC at the top of the how-to guide? (maybe not necessary since it'll be included automatically)
4) Is there a way to have the agents use tools without the handoff having to be a tool? (I think not)
5) Some details seem like they belong in the conceptual guide? (Ok here fro now, but this is good material for the conceptual guide)

image

6)  Could you document the function with google style doc-strings:

This is what people will use to understand what the function does.

```python
def make_handoff_tool(*, agent_name: str):
    tool_name = f"transfer_to_{agent_name}"
```

7) missing transition between the two code cells:

image

eyurtsev:
Some paraphrasing if you like?

1. Implementing a handoff tool -> implementing a handoff using a tool

2. The output from ```for chunk in graph.stream(
    {
        "messages": [
            (
                "user",
                "what's (3 + 5) * 12"
            )
        ]
    },
    subgraphs=True
):
    print(chunk)
    print("\n")```
    
    is not readable I think


3. parallel_tool_calls=False -- does this work with all model providers?  Can we add this to create react agent?
4.  Using with a prebuilt ReAct agent. Do we need to note somewhere that the scratch pad is shared? -- that's an important thing about this implementation

PR #2826, opened 2024-12-19
checkpoint postgres: add a shallow checkpointer

vbarda:
This PR adds a "shallow" version of `PostgresSaver` checkpointer that ONLY stores the most recent checkpoint and does NOT retain any history. It is meant to be a light-weight drop-in replacement for the PostgresSaver that supports most of the LangGraph persistence functionality with the exception of time travel.


libs/checkpoint-postgres/langgraph/checkpoint/postgres/shallow.py
```diff
@@ -0,0 +1,391 @@
...
+    channel TEXT NOT NULL,
+    version TEXT NOT NULL,
+    type TEXT NOT NULL,
+    blob BYTEA,
+    PRIMARY KEY (thread_id, checkpoint_ns, channel, version)
+);""",
+    """CREATE TABLE IF NOT EXISTS checkpoint_writes (
+    thread_id TEXT NOT NULL,
+    checkpoint_ns TEXT NOT NULL DEFAULT '',
+    checkpoint_id TEXT NOT NULL,
```
nfcampos, line 55: this shouldn't be here right?

libs/checkpoint-postgres/langgraph/checkpoint/postgres/shallow.py
```diff
@@ -0,0 +1,391 @@
...
+        select array_agg(array[cw.type::bytea, cw.blob] order by cw.task_id, cw.idx)
+        from checkpoint_writes cw
+        where cw.thread_id = checkpoints.thread_id
+            and cw.checkpoint_ns = checkpoints.checkpoint_ns
+            and cw.channel = '{TASKS}'
+    ) as pending_sends
+from checkpoints """
+
+UPSERT_CHECKPOINT_BLOBS_SQL = """
+    INSERT INTO checkpoint_blobs (thread_id, checkpoint_ns, channel, version, type, blob)
```
nfcampos, line 107: this should be changed to store a single version per (thread_id, checkpoint_ns, channel), ie. remove version column, do on conflict do update

libs/checkpoint-postgres/langgraph/checkpoint/postgres/shallow.py
vbarda: not sure how we can implement without it

nfcampos (approved):

PR #2876, opened 2024-12-25
Add Node Caching Support

achrafmam2:
Users of the `StateGraph` API can optionally cache the outputs of their nodes by providing a `cache_policy` to `StateGraph::add_node`.

```
builder = StateGraph(State)
builder.add_node(
    node_fn, 
    cache_policy=CachePolicy(
        key=operator.itemgetter("x"), # Mandatory
        ttl=NODE_CACHE_TTL,           # Optional
    ),
)
```

Where `key` is a one argument function that takes the input state of the graph and returns a `cache_key` as a _string_.

Important: 

- Node caching requires a `checkpoint`.
- Caches are `thread` scoped

🪚 Sharp Edges 🪚:

- If two graphs use the same runnable name, the same thread id, and the same cache key, there will be cache contamination 😷. 
- In the same multi-step process, Send calls with identical input will be deduplicated. However, in the V2 API, this behavior is more complex—the node might still execute because it is scheduled in the same iteration it was created, rather than the next one.


libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -460,9 +501,27 @@ def put_writes(
...
+
+T = TypeVar("T")
+
+
+def _run_sync(coro: Awaitable[T], loop: asyncio.AbstractEventLoop) -> T:
+    try:
+        if asyncio.get_running_loop() is loop:
+            # Using `run_coroutine_threadsafe` will result in a deadlock.
+            # We first make the loop reentrant and then run the coroutine.
+            nest_asyncio.apply(loop)
```
nfcampos, line 518: we cannot use nest_asyncio in a library, that is the sort of thing which can only be used in an application

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -404,21 +441,7 @@ def get_tuple(self, config: RunnableConfig) -> Optional[CheckpointTuple]:
         Returns:
             Optional[CheckpointTuple]: The retrieved checkpoint tuple, or None if no matching checkpoint was found.
         """
-        try:
-            # check if we are in the main thread, only bg threads can block
-            # we don't check in other methods to avoid the overhead
-            if asyncio.get_running_loop() is self.loop:
-                raise asyncio.InvalidStateError(
```
nfcampos, line 411: we shouldn't be removing unrelated code in a pr adding a new feature

Issue #2909, opened 2024-12-31
PydanticUserError: Please use `typing_extensions.TypedDict` instead of `typing.TypedDict` on Python < 3.12.


mariomeissner:
### Checked other resources

- [X] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [X] I added a clear and detailed title that summarizes the issue.
- [X] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [X] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from langgraph.graph import MessagesState
from prisma.models import Document

class CustomState(MessagesState):
    document_store: dict[str, str]
```

### Error Message and Stack Trace (if applicable)

```shell
copilot-api:dev: │ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
copilot-api:dev: │ │                       annotation = <class 'langgraph.prebuilt.tool_node.InjectedState'>      │ │
copilot-api:dev: │ │                 get_inner_schema = <pydantic._internal._schema_generation_shared.CallbackGe… │ │
copilot-api:dev: │ │                                    object at 0x1048746d0>                                    │ │
copilot-api:dev: │ │              metadata_get_schema = <function                                                 │ │
copilot-api:dev: │ │                                    GenerateSchema._get_wrapped_inner_schema.<locals>.<lambd… │ │
copilot-api:dev: │ │                                    at 0x1235401f0>                                           │ │
copilot-api:dev: │ │ pydantic_js_annotation_functions = []                                                        │ │
copilot-api:dev: │ │                             self = <pydantic._internal._generate_schema.GenerateSchema       │ │
copilot-api:dev: │ │                                    object at 0x1056acf40>                                    │ │
copilot-api:dev: │ │                           source = <class 'copilot_api.state.CustomState'>                   │ │
copilot-api:dev: │ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
copilot-api:dev: │                                                                                                  │
copilot-api:dev: │ /Users/mario/Library/Caches/pypoetry/virtualenvs/copilot-api-7C_KWYzt-py3.10/lib/python3.10/site │
copilot-api:dev: │ -packages/pydantic/_internal/_generate_schema.py:1968 in <lambda>                                │
copilot-api:dev: │                                                                                                  │
[...]

PR #2910, opened 2024-12-31
Import `TypedDict` from `typing_extensions`, add ruff rule

mariomeissner:
Within `libs/langgraph`, change all `TypedDict` imports to come from `typing_extensions` rather than `typing`, as `pydantic` doesn't like the latter.

Additionally, add a ruff rule to ban these imports too (so this doesn't regress).

Solves  #2909.

### DIFF
--- a/libs/langgraph/README.md
+++ b/libs/langgraph/README.md
@@ -67,7 +67,8 @@ export LANGSMITH_API_KEY=lsv2_sk_...
 ```
 
 ```python
-from typing import Annotated, Literal, TypedDict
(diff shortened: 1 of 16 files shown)


mariomeissner:
Relevant ruff documentation:
https://docs.astral.sh/ruff/settings/#lint_flake8-tidy-imports_ban-relative-imports
https://docs.astral.sh/ruff/rules/banned-api/

nfcampos (approved):
Thanks for this

PR #2936, opened 2025-01-06
patch for interrupt in subgrahph

vigneshmj1997:
### Issue Description  
While working with interrupts in subgraphs, I noticed that the interrupt mechanism fails to function correctly when invoked for the second time. Specifically, the interrupt does not halt execution as expected during subsequent invocations, leading to unintended continuation of node execution.

---

### Solution  
To address this issue, I modified the logic to ensure that the `writes` variable is only updated if `__interrupt__` is explicitly present in `conf[__pregel_writes]`. If `__interrupt__` is not found, `writes` remains an empty list. This change ensures that the graph halts execution as expected when an interrupt is encountered and prevents the `writes` variable from being populated with invalid or unintended values.

---

### Changes Made  
1. Updated the logic in `writes` (see [code link](https://github.com/vigneshmj1997/langgraph/blob/interrupt_patch_for_subgraph/libs/langgraph/langgraph/types.py#L468)) to check for the presence of `__interrupt__` in `__pregel_writes`.  
2. If `__interrupt__` is found, `writes` is updated with the corresponding value.  
3. If `__interrupt__` is not found, `writes` remains an empty list.  

---

### Impact  
This change ensures that:  
- The graph halts execution as expected when an interrupt is encountered, improving the correctness of the subgraph behavior.

### DIFF
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -465,7 +465,7 @@ def interrupt(value: Any) -> Any:
     idx = scratchpad["interrupt_counter"]
     # find previous resume values
     task_id = conf[CONFIG_KEY_TASK_ID]
-    writes: list[PendingWrite] = conf[CONFIG_KEY_WRITES]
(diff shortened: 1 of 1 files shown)


nfcampos:
Thanks, this is fixed in #3054

PR #2941, opened 2025-01-06
tests: add test for multiple interrupts and tasks

eyurtsev:


libs/langgraph/tests/test_pregel.py
```diff
@@ -5300,3 +5300,61 @@ def node_b(state):
...
     ]
+
+
+def test_falsy_return_from_task() -> None:
+    """Test with a falsy return from a task."""
+    checkpointer = MemorySaver()
+
+    @task
+    def falsy_task() -> bool:
+        return False
```
eyurtsev, line 5311: triggers exception on any falsy value when used in conjunction with HIL

libs/langgraph/tests/test_pregel.py
```diff
@@ -5300,3 +5300,61 @@ def node_b(state):
...
+
+    configurable = {"configurable": {"thread_id": uuid.uuid4()}}
+    graph.invoke({}, configurable)
+    # Currently fails when double accepts a variable
+    graph.invoke(Command(resume="a"), configurable)
+    graph.invoke(Command(resume="b"), configurable)
+    result = graph.invoke(Command(resume="c"), configurable)
+    # `double` value should be cached appropriately when used w/ `interrupt`
+    assert result == {
+        "values": [2, "a", 4, "b", 6, "c"],
```
eyurtsev, line 5358: fails currently

PR #2947, opened 2025-01-07
feat(langgraph): Add interrupt schema to library

bracesproul:


libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,74 @@
...
+    action_request: ActionRequest
+    config: HumanInterruptConfig
+    description: Optional[str]
+
+
+class HumanResponse(TypedDict):
+    """The response provided by a human to an interrupt, which is returned when graph execution resumes.
+
+    Attributes:
+        type (Literal['accept', 'ignore', 'response', 'edit']): The type of response:
```
eyurtsev, line 62: nit: Could we stop documenting the type in the doc-string it's already documented as part of the type hints.

The convention to document the type predates type hints and isn't necessarily anymore these days -- LLMs still do it by default since most doucmented code predates type annotations

libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Contains the action type and any associated arguments needed for the action.
+
+    Attributes:
+        action (str): The type or name of action being requested (e.g., "Approve XYZ action")
+        args (dict): Key-value pairs of arguments needed for the action
+    """
+
+    action: str
+    args: dict
```
eyurtsev, line 39: what's the schema of args? is it a flat dict or can it be nested?

libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class HumanInterrupt(TypedDict):
```
eyurtsev, line 42: Consider adding an inline example of a human interrupt and then a corresponding example for human response

libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,74 @@
...
+    Attributes:
+        allow_ignore (bool): Whether the human can choose to ignore/skip the current step
+        allow_respond (bool): Whether the human can provide a text response/feedback
+        allow_edit (bool): Whether the human can edit the provided content/state
+        allow_accept (bool): Whether the human can accept/approve the current state
+    """
+
+    allow_ignore: bool
+    allow_respond: bool
+    allow_edit: bool
```
eyurtsev, line 24: What is the current scope of edit: 

1. does it allow editing the action?
2. does it allow editing any of the key values in the args?
3. can one delete a key from the args dict?

libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,74 @@
+from typing import (
+    Literal,
+    Optional,
+    Union,
```
eyurtsev, line 4: nit: for schema names could be nice to match `Request` with `Response` if there are good names just to make it easier to identify  the request / response pairs quickly  (right now it's HumanInterrupt vs. HumanResponse which isn't obvious from a first glance)

Do you have an example of usage w/ `interrupt()` ?

libs/langgraph/langgraph/prebuilt/interrupt.py
bracesproul: can be either-or!

libs/langgraph/langgraph/prebuilt/interrupt.py
bracesproul: it allows editing the action args (this is how the agent inbox manages editing, however your app could allow for other actions to be taken, up to you)

libs/langgraph/langgraph/prebuilt/interrupt.py
bracesproul: ya will add an example

eyurtsev:
@bracesproul would you mind linking in an example of the app that's consuming this right now?

i.e., a bit odd just reviewing schema or committing schema without it being used by any other code in the library?

bracesproul:
> @bracesproul would you mind linking in an example of the app that's consuming this right now?
> 
> i.e., a bit odd just reviewing schema or committing schema without it being used by any other code in the library?

[Email assistant](https://github.com/langchain-ai/executive-ai-assistant) (python code)
[Social media agent](https://github.com/langchain-ai/social-media-agent) (js code)

libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,94 @@
...
+    config: HumanInterruptConfig
+    description: Optional[str]
+
+
+class HumanResponse(TypedDict):
+    """The response provided by a human to an interrupt, which is returned when graph execution resumes.
+
+    Attributes:
+        type: The type of response:
+            - "accept": Approves the current state without changes
```
hinthornw, line 83: Why do accept and response have to be separate values?

libs/langgraph/langgraph/prebuilt/interrupt.py
```diff
@@ -0,0 +1,94 @@
...
+            - "ignore": Skips/ignores the current step
```
hinthornw, line 84: What is the difference between ignoring and rejecting?

bracesproul: rejecting isn't a valid response for this schema, what are you referencing here?

libs/langgraph/langgraph/prebuilt/interrupt.py
bracesproul: makes the implementation of the inbox a little simpler (separating out accept/response), and same with parsing the response in your graph. better dx if you can just check `interrupt_response.type` instead of checking the value of the response

PR #2948, opened 2025-01-07
langgraph[patch]: fix create_react_agent inspectability

baskaryan:


vbarda (approved):

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -2040,3 +2041,9 @@ def foo(a: str, b: int) -> float:
         return 0.0
 
     assert _get_state_args(foo) == {"a": None, "b": "bar"}
+
+
+def test_inspect_react() -> None:
```
baskaryan, line 2046: @vbarda @dqbd this test passes on main rn so not testing the right thing, do you know how to test the issue?

vbarda: updated!

PR #2972, opened 2025-01-09
layout

eyurtsev:
Some docs layout improvements to help guide user journey.

Currently we have `Home | Tutorials | How-tos | Concepts | Reference` in top-level horizontal navigation bar.

Here we make these updates:
- Top-level horizontal navigation bar is just `Home | API Reference`
- Add vertical sidebar to `Home` with sections:
  - Introduction
  - Get started
  - Guides
  - Resources

`Get Started` contains quickstarts for LG and LG Platform / deployment. These are tutorials in Diataxis terms.

`Guides` contains index pages for how-tos, concepts, tutorials.

Advantage of this organization is that users are directed naturally down the sidebar from Intro -> Get started -> How-tos, which is roughly how we expect them to proceed.

This also makes deployment info more accessible as it is highlighted in the "Getting started" section.

Screenshot 2025-01-14 at 1 37 01 PM


eyurtsev (commented):
gogogo

ccurme (approved):

PR #2993, opened 2025-01-10
Add support for custom fetch implementation

hinthornw:


libs/sdk-js/package.json
```diff
@@ -9,7 +9,8 @@
     "build": "yarn clean && yarn lc_build --create-entrypoints --pre --tree-shaking",
     "prepublish": "yarn run build",
     "format": "prettier --write src",
-    "lint": "prettier --check src && tsc --noEmit"
+    "lint": "prettier --check src && tsc --noEmit",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest"
```
jacoblee93, line 13: Would add ` --testPathIgnorePatterns=\\.int\\.test.ts`

libs/sdk-js/src/client.ts
jacoblee93, line 1: Are those the only places we use `fetch`?

libs/sdk-js/src/singletons/fetch.ts
```diff
@@ -0,0 +1,29 @@
...
+  "ls:fetch_implementation",
+);
+
+/**
+ * Overrides the fetch implementation used for LangSmith calls.
+ * You should use this if you need to use an implementation of fetch
+ * other than the default global (e.g. for dealing with proxies).
+ * @param fetch The new fetch functino to use.
+ */
+export const overrideFetchImplementation = (fetch: (...args: any[]) => any) => {
```
jacoblee93, line 17: You need to re-export this from the main entrypoint

libs/sdk-js/src/singletons/fetch.ts
```diff
@@ -0,0 +1,29 @@
+// Wrap the default fetch call due to issues with illegal invocations
+// in some environments:
+// https://stackoverflow.com/questions/69876859/why-does-bind-fix-failed-to-execute-fetch-on-window-illegal-invocation-err
+// @ts-expect-error Broad typing to support a range of fetch implementations
+const DEFAULT_FETCH_IMPLEMENTATION = (...args: any[]) => fetch(...args);
+
+const LANGSMITH_FETCH_IMPLEMENTATION_KEY = Symbol.for(
+  "ls:fetch_implementation",
```
jacoblee93, line 8: Make it `lg:fetch_implementation`

PR #3007, opened 2025-01-13
langgraph: add stateful entrypoint

eyurtsev:
* Just experimenting with the API  -- if we like this `@g` would be renamed and moved to appropriate place etc.


libs/langgraph/langgraph/func/__init__.py
```diff
@@ -47,6 +50,14 @@ def call(
     return fut
 
 
+def get_store() -> BaseStore:
```
vbarda, line 53: that's for accessing the store from inside the tasks?

PR #3009, opened 2025-01-13
docs: Add note about LangGraph Platform UI not available for self-hosted deployments

andrewnguonly:
Example screenshot:
image

### DIFF
--- a/docs/docs/concepts/deployment_options.md
+++ b/docs/docs/concepts/deployment_options.md
@@ -28,6 +28,10 @@ The guide below will explain the differences between the deployment options.
 
     The Self-Hosted Enterprise version is only available for the **Enterprise** plan.
 
+!!! warning "Note"
+
+    The LangGraph Platform UI (within LangSmith) is not available for Self-Hosted Enterprise deployments.
+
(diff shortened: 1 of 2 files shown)


docs/docs/concepts/deployment_options.md
```diff
@@ -28,6 +28,10 @@ The guide below will explain the differences between the deployment options.
 
     The Self-Hosted Enterprise version is only available for the **Enterprise** plan.
 
+!!! warning "Note"
+
+    The LangGraph Platform UI (within LangSmith) is not available for Self-Hosted Enterprise deployments.
```
nhuang-lc, line 33: Kind of a nit but had to read this a few times. It's slightly tricky because LangSmith also has self-hosted so I got confused at first.

Wdyt of
`The LangGraph Platform UI is not available within LangSmith for Self-Hosted Enterprise LangGraph Platform deployments.`

Or is that too verbose?

andrewnguonly: I updated the descriptions to be more verbose. Let me know if that helps.

nhuang-lc: Yea that's a lot clearer than what I wrote! Thank you!

nhuang-lc (approved):
Lgtm thank you for doing this!

PR #3025, opened 2025-01-14
functional api: Add ability to request previous output

eyurtsev:
1. The inputs into foo do not affect any state behavior
2. `previous` always reflects the previous return value from the function
3. Anything can be returned and that will be the new state for the function on the next iteration
4. This API is not meant to support reducers in the inputs/state

```python
  from langgraph.func import entrypoint

  states = []

  # In this version reducers do not work
  @entrypoint(checkpointer=MemorySaver())
  def foo(inputs, *, previous: Any) -> Any:
      states.append(previous)
      return {"previous": previous, "current": inputs}

  foo.invoke({"a": "1"}, config)
  foo.invoke({"a": "2"}, config)
  foo.invoke({"a": "3"}, config)
  assert states == [
      None,
      {"current": {"a": "1"}, "previous": None},
      {"current": {"a": "2"}, "previous": {"current": {"a": "1"}, "previous": None}},
  ]
```


libs/langgraph/langgraph/func/__init__.py
```diff
@@ -112,27 +111,8 @@ def entrypoint(
     store: Optional[BaseStore] = None,
 ) -> Callable[[types.FunctionType], Pregel]:
     def _imp(func: types.FunctionType) -> Pregel:
-        if inspect.isgeneratorfunction(func):
```
nfcampos, line 115: why was this removed?

eyurtsev: re-adding w/ tests to make sure that all behavior is well defined

eyurtsev: ah wasn't obvious to me how to implement a state update if it was a generator. users already have a way to stream by requesting `StreamWriter`

nfcampos:
Please don't remove generator support

eyurtsev:
How do you want to deal with the return value In the generator case? Why should we support it?

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -113,22 +113,96 @@ def entrypoint(
...
+        Returns:
+            A Pregel graph.
+        """
+        # wrap generators in a function that writes to StreamWriter
+        if inspect.isgeneratorfunction(func):
+            original_sig = inspect.signature(func)
+            # Check if original signature has a writer argument with a matching type.
+            # If not, we'll inject it into the decorator, but not pass it
+            # to the wrapped function.
+            if "writer" in original_sig.parameters:
```
nfcampos, line 131: why is this changing?

PR #3033, opened 2025-01-14
Remove send v2

nfcampos:


eyurtsev:
Is this removing the ENV variable to make Send eager in anticipation instead of tasks being used?

PR #3036, opened 2025-01-15
Fix tracing hierarchy for imperative api

nfcampos:


hinthornw:
Pretty simple fix. Wish I could forget the `Callbacks` constant though :P

hinthornw (approved):

libs/langgraph/langgraph/pregel/runner.py
```diff
@@ -148,9 +150,12 @@ def call(
             input: Any,
             *,
             retry: Optional[RetryPolicy] = None,
+            callbacks: Optional[Callbacks] = None,
```
eyurtsev, line 153: nit: Callbacks is already wrapped in an Optional in langchain_core

libs/langgraph/langgraph/pregel/executor.py
```diff
@@ -63,10 +63,14 @@ def submit(  # type: ignore[valid-type]
         __next_tick__: bool = False,
         **kwargs: P.kwargs,
     ) -> concurrent.futures.Future[T]:
+        ctx = copy_context()
         if __next_tick__:
-            task = self.executor.submit(next_tick, fn, *args, **kwargs)
+            task = cast(
+                concurrent.futures.Future[T],
+                self.executor.submit(next_tick, ctx.run, fn, *args, **kwargs),  # type: ignore[arg-type]
```
eyurtsev, line 70: nit: `next_tick` is a bit confusing on the sync side -- sleep(0) isn't guaranteed  to context switch

eyurtsev (approved):

libs/langgraph/langgraph/pregel/executor.py
nfcampos: It is the closest equivalent. And yes it does context switch https://stackoverflow.com/a/7273727

nfcampos: Also not something being changed in this PR?

Issue #2890, opened 2024-12-28
Tags added to the graph are not propagated in v0.2.x (worked in v0.1.x)


zoltan-fedor:
### Checked other resources

- [X] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [X] I added a clear and detailed title that summarizes the issue.
- [X] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [X] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from langgraph.graph import StateGraph

workflow = StateGraph(AgentState)
...
workflow.compile().with_config(tags=["mygraph"])
```

### Description

In v0.1.x I could add a tag to a graph with either of the following:

```
from langgraph.graph import StateGraph

workflow = StateGraph(AgentState)
...
workflow.compile().with_config({"run_name": "my_run", "tags": ["mygraph"]})
```

Or with:

```
workflow = StateGraph(AgentState)
...
workflow.compile().with_config(run_name="my_run", tags=["mygraph"])
```

In v0.1.x the added tag (`mygraph`) shows up in the metadata of the events:
```
{
  tags: [
    0: "seq:step:3",
    1: "mygraph"    <=== TAG IS HERE
  ]
  langgraph_step: "1"
  langgraph_node: "agent"
  langgraph_triggers: [
    0: "start:agent"
  ]
}
```

In v0.2.x the added tag is not in the metadata:
```
{
  tags: [
    0: "seq:step:3"   <=== NO 'mygraph' TAG!!!
  ]
  langgraph_step: "1"
  langgraph_node: "agent"
  langgraph_triggers: [
    0: "start:agent"
  ]
}
```

Is it possible that the new tags added to the v0.2.x codebase like `TAG_HIDDEN` overwrite the user-supplied tags somewhere instead of appending to it and so the user-supplied tags gets lost?

See `TAG_HIDDEN`:
https://github.com/langchain-ai/langgraph/blob/effddca49417e2b526ca5affabbbd6d53ed6aa0f/libs/langgraph/langgraph/constants.py#L19

Ps,: 
- The `run_name` can be successfully propagated from the `with_config()` method, only the `tags` are not propagated in v0.2.x
[...]

PR #3039, opened 2025-01-15
langgraph: fix non-empty value check in ensure_config

vbarda:
Fixes https://github.com/langchain-ai/langgraph/issues/2890

### DIFF
--- a/libs/langgraph/langgraph/utils/config.py
+++ b/libs/langgraph/langgraph/utils/config.py
@@ -250,6 +250,13 @@ def get_async_callback_manager_for_config(
         )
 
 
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/utils/config.py
```diff
@@ -250,6 +250,13 @@ def get_async_callback_manager_for_config(
         )
 
 
+def _is_not_empty(value: Any) -> bool:
```
eyurtsev, line 253: Could you add a unit test?

nfcampos (approved):

PR #3046, opened 2025-01-15
Remove duckdb checkpointer and store

nfcampos:
- duckdb is too buggy to be able to provide reliable checkpointer and store


sneerin:
curious on your feedback , is it duckdb itself or the implementation has lots of bugs?

PR #3049, opened 2025-01-15
Add optional task_path arg for put_writes()

nfcampos:
- Will be used for sorting pending_sends when available

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -66,6 +66,7 @@ MIGRATIONS = [
     """
     CREATE INDEX CONCURRENTLY IF NOT EXISTS checkpoint_writes_thread_id_idx ON checkpoint_writes(thread_id);
     """,
+    """ALTER TABLE checkpoint_writes ADD COLUMN task_path TEXT NOT NULL DEFAULT '';""",
 ]
 
 SELECT_SQL = f"""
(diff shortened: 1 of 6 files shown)


libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -66,6 +66,7 @@
     """
     CREATE INDEX CONCURRENTLY IF NOT EXISTS checkpoint_writes_thread_id_idx ON checkpoint_writes(thread_id);
     """,
+    """ALTER TABLE checkpoint_writes ADD COLUMN task_path TEXT NOT NULL DEFAULT '';""",
```
headyin, line 69: This breaks 'postgres_saver()'. 

When call it more than once, it will throw exception:

DuplicateColumn: column "task_path" of relation "checkpoints_writes" already exists

headyin (commented):
Hi

This breaks my project, please see inline comment for the details.

Thank you!

libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
vbarda: Please update to the latest version of `langgraph-checkpoint-postgres`

headyin: Do I need to call setup just once or once very new version update? How to detect if there is a migration needed and need to call setup?

vbarda: there is no way to detect that other than by reading release notes to see if there were any schema changes. there was also a bug in the previous version of the library that could have caused the issues you were seeing

technically there is no harm from calling `.setup()` multiple times, as long as you're not modifying the list of migrations manually yourself (and you shouldn't be running individual migrations manually)

headyin: I'm using `langgraph-checkpoint-postgres` 2.0.13 the latest version. This problem is still there.

vbarda: can you open an issue and provide a full description of the problem and a minimal reproducible example of the issue ? https://stackoverflow.com/help/minimal-reproducible-example

PR #3060, opened 2025-01-16
tests: use send twice in a row?

eyurtsev:


libs/langgraph/tests/test_pregel.py
```diff
@@ -5491,3 +5491,241 @@ def invoke_sub_agent(state: AgentState):
...
+        max_score = 0
+        best_joke = ""
+        for joke in state["jokes_with_grades"]:
+            if joke["grade"] > max_score:
+                max_score = joke["grade"]
+                best_joke = joke["joke"]
+        return {"best_selected_joke": best_joke}
+
+    builder = StateGraph(OverallState)
+    builder.add_conditional_edges("__start__", generate_jokes, ["generate_joke"])
```
eyurtsev, line 5553: fails compilation:

>               raise ValueError(f"Found edge starting at unknown node '{source}'")
E               ValueError: Found edge starting at unknown node 'generate_joke'


1) This should probably work
2) We should've called these "add_routing_node" in retrospect

libs/langgraph/tests/test_pregel.py
```diff
@@ -5491,3 +5491,241 @@ def invoke_sub_agent(state: AgentState):
...
+    # subjects to in order to generate a joke
+    class JokeState(TypedDict):
+        subject: str
+        joke: NotRequired[str]
+        joke_grade: NotRequired[int]
+
+    def generate_jokes(state: OverallState):
+        # Distribute the subjects to the generate_joke node
+        sends = [Send("generate_joke", {"subject": s}) for s in state["subjects"]]
+        return Command(goto=sends)
```
eyurtsev, line 5677: This one fails as well -- looks like a bug

libs/langgraph/tests/test_pregel.py
nfcampos: But your graph doesn't have a "generate_joke" node? What am I missing?

eyurtsev: If i add a node it requires the state to be overall state (and I think forces the reduce step)

PR #3076, opened 2025-01-16
Implement get_graph for imperative api

nfcampos:


libs/langgraph/langgraph/func/__init__.py
```diff
@@ -174,6 +178,84 @@ async def agen_wrapper(
...
+        RunnableCallable,
+        RunnableLambda,
+        RunnableSeq,
+        RunnableSequence,
+    )
+
+    candidates: list[Runnable] = [candidate]
+
+    for c in candidates:
+        print(c, type(c))
```
dqbd, line 202: remove print

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -174,6 +178,84 @@ async def agen_wrapper(
...
+                    for nl in get_function_nonlocals(c.func)
+                )
+            elif c.afunc is not None:
+                candidates.extend(
+                    nl.__self__ if hasattr(nl, "__self__") else nl
+                    for nl in get_function_nonlocals(c.afunc)
+                )
+
+
+def _entrypoint_graph(entrypoint: Runnable, xray: int = 0) -> Graph:
```
dqbd, line 224: _question_: worth doing this on-demand rather than during initialization?

nfcampos: yea probably, will change

PR #3083, opened 2025-01-17
functional api: first pass at api reference

eyurtsev:
First pass at the API reference for the functional API


docs/mkdocs.yml
```diff
@@ -369,6 +369,7 @@ nav:
       - Errors: reference/errors.md
       - Types: reference/types.md
       - Constants: reference/constants.md
+      - Functional API: reference/func.md
```
eyurtsev, line 372:
```suggestion
```

PR #3107, opened 2025-01-20
Fix tracing of args for @task decorated functions

nfcampos:
- now using same logic as in langsmith sdk, treating as single args dict, based on function signature

### DIFF
--- a/libs/langgraph/langgraph/pregel/call.py
+++ b/libs/langgraph/langgraph/pregel/call.py
@@ -1,12 +1,25 @@
 """Utility to convert a user provided function into a Runnable with a ChannelWrite."""
 
+import asyncio
+import concurrent.futures
+import functools
+import inspect
 import sys
 import types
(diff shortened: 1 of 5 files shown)


libs/langgraph/langgraph/pregel/call.py
```diff
@@ -107,18 +120,96 @@ def _lookup_module_and_qualname(
...
+            trace_inputs=functools.partial(
+                _explode_args_trace_inputs, inspect.signature(func)
+            ),
+        )
+        if not _lookup_module_and_qualname(func):
+            return seq
+        return CACHE.setdefault(key, seq)
+
+
+CACHE: dict[tuple[Callable[..., Any], bool], RunnableSeq] = {}
```
hinthornw, line 198: `bool` means explode_args here?

hinthornw (approved):
Nice I like the improvements - should probably include a test b4 landing?

PR #3108, opened 2025-01-20
Re-enable support for running sync tasks from async entrypoints

nfcampos:
- When using an async entrypoint you can now freely mix and match sync and async tasks with a uniform api (ie all tasks return a sync or async future depending on context)
- Fix issues with scheduling deeply nested tasks (use threadsafe methods to schedule coroutines and create futures)

### DIFF
--- a/libs/langgraph/langgraph/pregel/runner.py
+++ b/libs/langgraph/langgraph/pregel/runner.py
@@ -213,9 +213,7 @@ class PregelRunner:
             assert fut is not None, "writer did not return a future for call"
             # return a chained future to ensure commit() callback is called
             # before the returned future is resolved, to ensure stream order etc
-            sfut: concurrent.futures.Future[Any] = concurrent.futures.Future()
-            chain_future(fut, sfut)
-            return sfut
+            return chain_future(fut, concurrent.futures.Future())
 
         tasks = tuple(tasks)
         futures = FuturesDict(
@@ -400,10 +398,6 @@ class PregelRunner:
             retry: Optional[RetryPolicy] = None,
             callbacks: Callbacks = None,
         ) -> Union[asyncio.Future[Any], concurrent.futures.Future[Any]]:
-            if not asyncio.iscoroutinefunction(func):
-                raise RuntimeError(
-                    "In an async context use func.to_thread(...) to invoke tasks"
-                )
             (fut,) = writer(
                 task,
                 [(PUSH, None)],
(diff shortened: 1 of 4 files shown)


libs/langgraph/langgraph/utils/future.py
```diff
@@ -125,3 +134,70 @@ def chain_future(source: AnyFuture, destination: AnyFuture) -> None:
...
+    called_wrap_awaitable = False
+    if not asyncio.iscoroutine(coro_or_future):
+        if inspect.isawaitable(coro_or_future):
+            coro_or_future = cast(
+                Coroutine[None, None, T], _wrap_awaitable(coro_or_future)
+            )
+            called_wrap_awaitable = True
+        else:
+            raise TypeError(
+                "An asyncio.Future, a coroutine or an awaitable is required"
```
hinthornw, line 155:
```suggestion
"An asyncio.Future, a coroutine or an awaitable is required."
f" Got {type(future).__name__} instead."
```

libs/langgraph/langgraph/pregel/runner.py
```diff
@@ -412,9 +406,27 @@ def call(
             assert fut is not None, "writer did not return a future for call"
             # return a chained future to ensure commit() callback is called
             # before the returned future is resolved, to ensure stream order etc
-            sfut: asyncio.Future[Any] = asyncio.Future(loop=loop)
-            chain_future(fut, sfut)
-            return sfut
+            try:
+                asyncio.current_task()
```
hinthornw, line 410: ooc: why this vs. `get_running_loop()`? 

And would this ever suceed but return None?

libs/langgraph/langgraph/pregel/runner.py
```diff
@@ -412,9 +406,27 @@ def call(
...
+                in_async = True
+            except RuntimeError:
+                in_async = False
+            # if in async context return an async future
+            # otherwise return a chained sync future
+            if in_async:
+                if isinstance(fut, asyncio.Task):
+                    sfut: Union[asyncio.Future[Any], concurrent.futures.Future[Any]] = (
+                        asyncio.Future(loop=loop)
```
hinthornw, line 419: Took me a moment to find where loop is defined in the outer context after the fn definition. Would have been easier for me to read if it were passed in as an arg fwiw

hinthornw (approved):

PR #3124, opened 2025-01-21
langgraph[patch]: allow ToolNode to ignore ToolCalls for unknown tools

ccurme:
Alternative to https://github.com/langchain-ai/langgraph/pull/3126

Screenshot 2025-01-21 at 12 55 07 PM


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -762,14 +763,30 @@ def should_continue(state: AgentState) -> str:
...
-    workflow.add_node("tools", tool_node)
+    if version == "v1":
+        workflow.add_node("tools", tool_node)
+        tool_node_names = ["tools"]
+    elif version == "v2":
+        from collections.abc import Sequence
+        assert isinstance(tools, Sequence)  # sequence of BaseTool
+        tool_node_names = []
+        for tool in tools:
+            tool_node = ToolNode(tools=[tool], ignore_unknown_tool_calls=True)
```
vbarda, line 784: why do we want `ignore_unknown_tool_calls=True` by default?

also, we should add some validation in case someone tries to pass a `ToolNode` and also use "v2". we should probably have some of this conditional logic be done earlier as well around line 595 so we don't re-define `ToolNode` several times

ccurme: This loop is creating ToolNodes, and if you have multiple ToolNodes, you likely want `ignore_unknown_tool_calls=True`.

ToolNode as its implemented now is assuming that each `"messages"` key is handled by one ToolNode (I think). When ToolNode plucks the last AIMessage with tool calls, it will generate one ToolMessage per tool call -- if there are tools it isn't initialized with, it will generate a placeholder ToolMessage.

---

Roger on the early validation

vbarda: i wonder if it's better to rewrap this into separate, new AI messages (or handle tool calls directly in the tool node), otherwise it's gonna be weird in tracing i think

PR #3126, opened 2025-01-21
langgraph[patch]: allow ToolNode to accept ToolCalls

ccurme:
Alternative to https://github.com/langchain-ai/langgraph/pull/3124

Currently if a tool interrupts, the entire tool node executes again after resuming. So tools can get executed twice if parallel tool calls are generated. Here we allow ToolNode to accept tool calls, so we can use the `Send` API to distribute the tool calls to multiple instances of the tool node.

```python
from langchain_anthropic import ChatAnthropic
from langgraph.types import Command, Send, interrupt


@tool
def human_assistance(query: str) -> str:
    """Request assistance from a human."""
    human_response = interrupt({"query": query})
    return human_response["data"]


@tool
def get_weather(location: str) -> str:
    """Use this tool to get the weather."""
    return "It's sunny!"


tools = [get_weather, human_assistance]
llm = ChatAnthropic(model="claude-3-5-sonnet-20240620")

agent = create_react_agent(
    llm,
    tools,
    checkpointer=MemorySaver(),
    tool_call_parallelism="parallel_tool_nodes",
)


user_input = (
    "Could you please (1) request assistance for building an AI agent "
    "from a human, and (2) search for the weather in Boston, MA? "
    "Generate two tool calls at once."
)

for event in agent.stream(
    {"messages": [{"role": "user", "content": user_input}]},
    config,
    stream_mode="values",
):
    event["messages"][-1].pretty_print()
```
```
...
```
```python
human_response = "You should check out LangGraph to build your agent."
human_command = Command(resume={"data": human_response})

for event in agent.stream(human_command, config, stream_mode="values"):
    event["messages"][-1].pretty_print()
```


ccurme:
Example trace:
- After interrupt: https://smith.langchain.com/public/6ab97e42-467f-4789-941a-40fcb08bb7a8/r
- After resuming: https://smith.langchain.com/public/349320b8-729b-4daf-9c56-a97e173c57e3/r

vbarda:
This looks good to me -- let's add some tests!

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -405,10 +405,15 @@ def _parse_input(
...
         store: BaseStore,
-    ) -> Tuple[list[ToolCall], Literal["list", "dict"]]:
+    ) -> Tuple[list[ToolCall], Literal["list", "dict", "tool_calls"]]:
         if isinstance(input, list):
-            input_type = "list"
-            message: AnyMessage = input[-1]
+            if isinstance(input[-1], dict) and input[-1].get("type") == "tool_call":
+                input_type = "tool_calls"
+                tool_calls = input
+                return tool_calls, input_type
```
vbarda, line 413: need to inject the tool args for this as well

ccurme: updated this to inject store. I'm unclear if there's anything else to do regarding injecting state args because the state is a tool call in this case.

vbarda: that makes sense, but what if the tools _do_ need to access graph state?

ccurme: Updated:

- ToolNode will always try to inject store
- create_react_agent will inject both state and store.

vbarda: let's simplify to just do this in react agent

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -1069,7 +1142,14 @@ def handle(e: NodeInterrupt):
                 },
             ],
         ),
+        _AnyIdToolMessage(content="normal", name="tool_normal", tool_call_id="2"),
     ]
+    if tool_call_parallelism == "single_tool_node":
+        # Interrupt blocks second tool result
+        assert result["messages"] == expected_messages[:2]
```
vbarda, line 1149: nit for readability
```suggestion
        assert result["messages"] == expected_messages[:-1]
```

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -1081,7 +1161,8 @@ def handle(e: NodeInterrupt):
     not IS_LANGCHAIN_CORE_030_OR_GREATER,
     reason="Langchain core 0.3.0 or greater is required",
 )
-async def test_tool_node_command():
+@pytest.mark.parametrize("input_type", ["dict", "tool_calls"])
+async def test_tool_node_command(input_type: str):
```
vbarda, line 1165: what happens in this case if we do parallel tool nodes?

vbarda (approved):

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -1886,6 +2046,49 @@ def tool4(
     assert tool_message.content == "hi?"
 
 
+@pytest.mark.parametrize("tool_call_parallelism", REACT_TOOL_CALL_PARALLELISM)
+def test_create_react_agent_inject_vars(tool_call_parallelism: str) -> None:
```
ccurme, line 2050: Traces from this:
- [Default mode for react agent](https://smith.langchain.com/public/1168de08-9d90-4503-863c-161be63d02ea/r/17b00296-265b-44a1-83b3-49c219a911fc)
- [New mode (sending tool calls)](https://smith.langchain.com/public/24960fb7-c67a-41bc-b39d-7214aab47613/r/f77205e7-07c3-44a5-8a4e-decac41f5168)

libs/langgraph/tests/test_prebuilt.py
```diff
@@ -1698,6 +1777,87 @@ def prompt(state: State):
     assert tool_message.name == "get_user_name"
 
 
+@pytest.mark.skipif(
+    not IS_LANGCHAIN_CORE_030_OR_GREATER,
+    reason="Langchain core 0.3.0 or greater is required",
+)
+@pytest.mark.parametrize("tool_call_parallelism", REACT_TOOL_CALL_PARALLELISM)
+def test_react_agent_parallel_tool_calls(tool_call_parallelism: str):
```
ccurme, line 1785: Traces from this:
- Default mode for react agent:
  - [After interrupt](https://smith.langchain.com/public/b11706f1-7286-4d6b-ad4e-31a2469a23c1/r)
  - [After resume](https://smith.langchain.com/public/370dc990-26c4-4547-a8a7-537655d24bd4/r)
- New mode (sending tool calls):
  - [After interrupt](https://smith.langchain.com/public/13bc9b37-1838-41fa-80b9-a7d6cd2b63fa/r)
  - [After resume](https://smith.langchain.com/public/136fdf55-4a22-477e-a66e-5a7d521258f5/r)

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -248,6 +248,9 @@ def create_react_agent(
     interrupt_before: Optional[list[str]] = None,
     interrupt_after: Optional[list[str]] = None,
     debug: bool = False,
+    tool_call_parallelism: Literal[
```
vbarda, line 251: can we document this in the docstring? i think important to mention that in v2 each tool result will be streamed /traced as a separate update

ccurme: Yeah good catch.

libs/langgraph/langgraph/prebuilt/tool_node.py
```diff
@@ -168,8 +170,23 @@ def tool_node(state: dict):
...
+        tool_calls = [
+            tool_node.inject_tool_args(call, state, store)
+            for call in last_message.tool_calls
+        ]
+        return [Send("tools", [tool_call]) for tool_call in tool_calls]
+    ```
+
     Important:
-        - The state MUST contain a list of messages.
+        - The input state MUST contain either a list of messages or a list of tool calls.
```
vbarda, line 189: nice update -- i think we can go a bit further and explain that inputs can be one of 3 things:
- dict w/ messages_key -> list of messages
- list of messages
- list of tool calls

libs/langgraph/tests/test_prebuilt.py
ccurme: Spoke offline, ran through agent handoffs notebook with the new mode for create_react_agent. results look reasonable (see trace [here](https://smith.langchain.com/public/a4cda233-c3bf-4777-aa54-39c3d2101a03/r)).

if we have a regular tool call and a handoff tool call in parallel, it looks like both methods currently raise the same error: when they get handed off to agent 2, we have two tool calls generated but only one "successful transfer" ToolMessage.
- single tool node: https://smith.langchain.com/public/264d4239-d77f-4c4d-a0e9-1a52b25cf9c9/r
- Send + tool calls: https://smith.langchain.com/public/ae3da0d9-201c-40a5-a925-1cc30824c32f/r

this is unrelated to the changes here.

Sourav-Nandy-ai:
Hi ccurme,

I ran a real before-and-after test on this rather than just reading the diff, and I found something worth flagging directly. I tested both the default ToolNode topology (one node, multiple tool_calls, plain edge) and the opt-in Send-per-tool-call pattern shown in this PR's description against real installed versions of langgraph==0.2.69 (pre-fix) and 0.2.70 (post-fix).

The Results
- Opt-in Send-per-tool-call pattern: This genuinely fixes double-execution on resume. I observed 0 out of 20 clean runs pre-fix (it was not even possible, as ToolNode rejected bare ToolCall input) versus 20 out of 20 clean runs post-fix.
- Default topology: This topology (which is almost certainly what most existing callers are still using) double-executes identically on both versions. I saw 0 out of 20 clean runs on both 0.2.69 and 0.2.70.

Therefore, this fix closes the gap only for callers who adopt the new Send-per-call pattern. Anyone still on the old topology remains exposed.

I maintain agent-eval (agent_regress), an OSS regression harness. You can find my full scripts and raw JSON output here: https://github.com/RudrenduPaul/agent-eval.

I wanted to flag this distinction since it seems worth calling out more prominently in the PR description or documentation, if it is not already. I am happy to be corrected if I am missing a migration note somewhere.

Best,
Sourav

PR #3129, opened 2025-01-21
docs: add react agent guide using functional API

ccurme:


nfcampos (commented):
- should we make each tool a `@task`?
- I'd probably suggest using concurrent.futures.as_completed when waiting for the results of the tool calls, but not necessary

docs/docs/how-tos/react-agent-from-scratch-functional.ipynb
```diff
@@ -0,0 +1,273 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from langchain_openai import ChatOpenAI\n",
+    "from langchain_core.tools import tool\n",
+    "\n",
+    "model = ChatOpenAI(model=\"gpt-4o-mini\")\n",
+    "\n",
+    "\n",
+    "@tool(response_format=\"content_and_artifact\")\n",
```
vbarda, line 99: why `@tool(response_format="content_and_artifact")`?

ccurme: this isn't needed, will remove

ccurme:
> * should we make each tool a `@task`?

@nfcampos is there a good way to associate the name of a tool (e.g., from a tool call) to a task?

One option is something like
```python
def _make_task_from_tool(tool):
    tool_task = task(tool.invoke)
    tool_task.__name__ = tool.name
    return tool_task
```

nfcampos:
We can add a name arg to task decorator

docs/docs/how-tos/wait-user-input-functional.ipynb
```diff
@@ -0,0 +1,569 @@
+{
```
vbarda, line 1: high level this looks good, but i wonder if people would find it a bit confusing if they try to compare to the original notebook (cause doesn't fully mirror)

docs/docs/how-tos/wait-user-input-functional.ipynb
```diff
@@ -0,0 +1,569 @@
...
+    "def call_model(messages):\n",
+    "    \"\"\"Call model with a sequence of messages.\"\"\"\n",
+    "    response = model.bind_tools(tools).invoke(messages)\n",
+    "    return response\n",
+    "\n",
+    "\n",
+    "@task\n",
+    "def call_tool(tool_call):\n",
+    "    tool = tools_by_name[tool_call[\"name\"]]\n",
+    "    observation = tool.invoke(tool_call[\"args\"])\n",
```
vbarda, line 332: should we just pass `tool_call` here?

docs/docs/how-tos/wait-user-input-functional.ipynb
```diff
@@ -0,0 +1,569 @@
...
+    "        return \"It's rainy!\"\n",
+    "    else:\n",
+    "        return f\"I am not sure what the weather is in {location}\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Human assistance tool\n",
```
vbarda, line 272: probably no need for a separate header here

docs/docs/how-tos/wait-user-input-functional.ipynb
```diff
@@ -0,0 +1,569 @@
...
+       "(Interrupt(value={'query': 'What should I feed my cat?'}, resumable=True, ns=['agent:5e01f3a3-a097-3cb2-7557-b36fa4c5e427', 'call_tool:154fa07a-05dc-b025-2a4d-e7faa29c643a'], when='during'),)"
+      ]
+     },
+     "execution_count": 9,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "next_task = agent.get_state(config).tasks[0]\n",
```
vbarda, line 487: (nit) nothing wrong with showing interrupts this way, but you can also just simply stream updates instead, and interrupt will be streamed as well

docs/docs/how-tos/wait-user-input-functional.ipynb
```diff
@@ -0,0 +1,569 @@
...
+   "metadata": {},
+   "source": [
+    "Above, when we resume we provide the final tool message, allowing the model to generate its response. Check out the LangSmith traces to see a full breakdown of the runs:\n",
+    "\n",
+    "1. [Trace from initial query](https://smith.langchain.com/public/c3d8879d-4d01-41be-807e-6d9eed15df99/r)\n",
+    "2. [Trace after resuming](https://smith.langchain.com/public/97c05ef9-8b4c-428e-8826-3fd417c8c75f/r)"
+   ]
+  },
+  {
+   "cell_type": "code",
```
vbarda, line 541: nit blank cell

docs/docs/how-tos/review-tool-calls-functional.ipynb
```diff
@@ -0,0 +1,585 @@
...
+    "# highlight-next-line\n",
+    "human_input = Command(resume={\"action\": \"continue\"})\n",
+    "\n",
+    "for step in agent.stream(human_input, config):\n",
+    "    _print_step(step)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
```
vbarda, line 558: nit empty cell

docs/docs/how-tos/review-tool-calls-functional.ipynb
```diff
@@ -0,0 +1,585 @@
+{
```
vbarda, line 1: should we lift a short example of the tool review flow to the top? same as in the non-functional how to?

vbarda (approved):

docs/docs/how-tos/wait-user-input-functional.ipynb
ccurme: This will work if the integration adds `"type": "tool_call"` to the tool call dict. I'm not sure that all integrations do this. We could add that in (I believe ToolNode does that). Will keep this for now for simplicity.

docs/docs/how-tos/wait-user-input-functional.ipynb
ccurme: Will come back to this, I kind of like this implementation better (may update the other guide).

PR #3133, opened 2025-01-21
docs: add more entrypoint defintions

eyurtsev:
Show more entrypoint definitions


eyurtsev:
need to document previous as well

PR #3135, opened 2025-01-21
Add entrypoint.final to decouple return value from save value

eyurtsev:
* Introduce `entrypoint.final` that allows decoupling what's returned from the state update.
* moving decorator to class object w/ call to support defining `final` as a property on it -- this should play nicely w/ IDE tooling / type checking.

```python
    previous_ = None

    @entrypoint(checkpointer=MemorySaver())
    def foo(msg: str, *, previous: Any) -> entrypoint.final[int, list[str]]:
        nonlocal previous_
        previous_ = previous
        previous = previous or []
        return entrypoint.final(value=len(previous), save=previous + [msg])

    assert foo.get_output_schema().model_json_schema() == {
        "title": "LangGraphOutput",
        "type": "integer",
    }

    assert foo.invoke("hello", config) == 0
    assert previous_ is None
    assert foo.invoke("goodbye", config) == 1
    assert previous_ == ["hello"]
    assert foo.invoke("definitely", config) == 2
    assert previous_ == ["hello", "goodbye"]

```


libs/langgraph/langgraph/func/__init__.py
```diff
@@ -286,22 +288,62 @@ def _imp(func: types.FunctionType) -> Pregel:
...
+                                    "Yielding a value after a ReturnAndSave "
+                                    "object is not allowed."
+                                )
+                            output_chunk = chunk
                         writer(chunk)
-                        chunks.append(chunk)
-                    return chunks
+                        chunks.append(output_chunk)
+                    if return_and_save:
+                        return ReturnAndSave(chunks, return_and_save.save)
```
nfcampos, line 313: shouldn't this be `ReturnAndSave(return_and_save.value, return_and_save.save)`

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -286,22 +288,62 @@ def _imp(func: types.FunctionType) -> Pregel:
...
+                            if return_and_save is not None:
+                                raise RuntimeError(
+                                    "Yielding a value after a ReturnAndSave "
+                                    "object is not allowed."
+                                )
+                            output_chunk = chunk
+                        writer(output_chunk)
+                        chunks.append(output_chunk)
+                    if return_and_save:
+                        return ReturnAndSave(chunks, return_and_save.save)
```
nfcampos, line 344: same as above

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -266,7 +279,35 @@ def my_workflow(input_data: str, previous: Optional[str] = None) -> str:
...
+
+        value: R
+        """Value to return. A value will always be returned even if it is None."""
+        save: S
+        """The value for the state for the next checkpoint. 
+
+        A value will always be saved even if it is None.
+        """
+
+        def __init__(self, value: R, save: S) -> None:
```
eyurtsev, line 306: @nfcampos thoughts on making these named arguments only given that `entrypoint.final` is not as self-explanatory? it's a lot to type `entrypoint.final(value='hello', save='goodbye')`

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -286,22 +327,57 @@ def _imp(func: types.FunctionType) -> Pregel:
...
+                                final_ = chunk
+                                output_chunk = chunk.value
+                        else:
+                            if final_ is not None:
+                                raise RuntimeError(
+                                    "Yielding a value after a entrypoint.final "
+                                    "object is not allowed."
+                                )
+                            output_chunk = chunk
                         writer(chunk)
```
eyurtsev, line 349: ah one more thing to update

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -178,6 +186,11 @@ def entrypoint(
         config_schema: Specifies the schema for the configuration object that will be
             passed to the workflow.
 
+    Attributes:
+        final: A special attribute that can be used to return a value and save a value
+            to the checkpointer. This is useful when you want to return a value from
+            the function and save a value to the check
```
nfcampos, line 191: I think this should say something like "save a value to the checkpointer distinct from the one returned"

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -266,7 +279,35 @@ def my_workflow(input_data: str, previous: Optional[str] = None) -> str:
...
+        checkpointer: Optional[BaseCheckpointSaver] = None,
+        store: Optional[BaseStore] = None,
+        config_schema: Optional[type[Any]] = None,
+    ) -> None:
+        """Initialize the entrypoint decorator."""
+        self.checkpointer = checkpointer
+        self.store = store
+        self.config_schema = config_schema
+
+    class final(Generic[R, S]):
```
nfcampos, line 292: This can be made a data class again no?

libs/langgraph/langgraph/func/__init__.py
nfcampos: Yea I agree with making them kwargs

libs/langgraph/langgraph/func/__init__.py
```diff
@@ -376,37 +476,69 @@ async def agen_wrapper(
...
+
+        def _pluck_save_value(value: Any) -> Any:
+            """Extract the save value from the entrypoint.final object or pass through."""
+            return value.save if isinstance(value, entrypoint.final) else value
+
+        output_type, save_type = Any, Any
+        if sig.return_annotation is not inspect.Signature.empty:
+            # User does not parameterize entrypoint.final properly
+            if (
+                sig.return_annotation is entrypoint.final
```
nfcampos, line 491: Does signature() resolve string annotations? If not this can also be the string entrypoint.final.

libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -508,8 +509,8 @@ def prepare_single_task(
                             pending_writes,
                             task_id,
                         ),
-                        CONFIG_KEY_END: checkpoint["channel_values"].get(
-                            "__end__", None
+                        CONFIG_KEY_PREVIOUS: checkpoint["channel_values"].get(
```
nfcampos, line 512: This branch is used only for `@task` so this should probably be removed

libs/langgraph/langgraph/func/__init__.py
eyurtsev: Good catch I'll check tomorrow

eyurtsev:
I'll look info forward references in a separate PR -- going to merge right now to unblock downstream work for docs

PR #3139, opened 2025-01-22
docs: Add note about how to share Postgres instance for self-hosted deployments

andrewnguonly:
### Screenshot
image

### DIFF
--- a/docs/docs/how-tos/deploy-self-hosted.md
+++ b/docs/docs/how-tos/deploy-self-hosted.md
@@ -31,7 +31,7 @@ You will eventually need to pass in the following environment variables to the L
     !!! Note "Shared Postgres Instance"
         Multiple self-hosted deployments can share the same Postgres instance. For example, for `Deployment A`, `DATABASE_URI` can be set to `postgres://<user>:<password>@/<database_name_1>?host=<hostname_1>` and for `Deployment B`, `DATABASE_URI` can be set to `postgres://<user>:<password>@/<database_name_2>?host=<hostname_1>`.
 
-        `<database_name_1>` and `database_name_2` are different, but `<hostname_1>` is shared.
+        `<database_name_1>` and `database_name_2` are different databases within the same instance, but `<hostname_1>` is shared. **The same database cannot be used for separate deployments**.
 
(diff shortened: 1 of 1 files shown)


docs/docs/how-tos/deploy-self-hosted.md
```diff
@@ -26,7 +26,13 @@ If you would like to deploy LangGraph Cloud on Kubernetes, you can use this [Hel
...
 
 - `REDIS_URI`: Connection details to a Redis instance. Redis will be used as a pub-sub broker to enable streaming real time output from background runs.
-- `DATABASE_URI`: Postgres connection details. Postgres will be used to store assistants, threads, runs, persist thread state and long term memory, and to manage the state of the background task queue with 'exactly once' semantics.
+- `DATABASE_URI`: Postgres connection details. Postgres will be used to store assistants, threads, runs, persist thread state and long term memory, and to manage the state of the background task queue with 'exactly once' semantics. The value of `DATABASE_URI` must be a valid [Postgres connection URI](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS).
+
+    !!! Note "Shared Postgres Instance"
+        Multiple self-hosted deployments can share the same Postgres instance. For example, for `Deployment A`, `DATABASE_URI` can be set to `postgres://<user>:<password>@/<database_name_1>?host=<hostname_1>` and for `Deployment B`, `DATABASE_URI` can be set to `postgres://<user>:<password>@/<database_name_2>?host=<hostname_1>`.
```
langchain-infra, line 32: should we also make it try to create the db if it doesn't exist?

andrewnguonly: Not sure if that's feasible... The connection string wouldn't include the database initially and then would need to be set after on redeploy.

Regardless, out-of-scope for this documentation update.

langchain-infra (approved):
Fair. Could we also point out they can use Redis with a different database as well? I think we should also make it clear that you cant point multiple deploymnents at the **same** logical database as well as this can cause one deployment to handle something meant for another?

andrewnguonly:
> Fair. Could we also point out they can use Redis with a different database as well? I think we should also make it clear that you cant point multiple deploymnents at the same logical database as well as this can cause one deployment to handle something meant for another?

Will update...

andrewnguonly:
> Could we also point out they can use Redis with a different database as well?

Just looked at the LangGraph API implementation and I don't think this is possible or potentially not safe to do.

PR #3158, opened 2025-01-23
Make scratchpad counters thread-safe

nfcampos:
- Same solution as used in python stdlib to name threads and asyncio tasks

### DIFF
--- a/libs/langgraph/langgraph/pregel/runner.py
+++ b/libs/langgraph/langgraph/pregel/runner.py
@@ -143,9 +143,9 @@ class PregelRunner:
                     continue
                 # schedule the next task, if the callback returns one
                 wcall = calls[idx] if calls else None
-                cnt = scratchpad.call_counter
-                scratchpad.call_counter += 1
-                if next_task := self.schedule_task(task, cnt, wcall):
+                if next_task := self.schedule_task(
+                    task, scratchpad.call_counter(), wcall
+                ):
                     if fut := next(
                         (
                             f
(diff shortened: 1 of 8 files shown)


libs/langgraph/tests/test_pregel_async.py
```diff
@@ -1612,7 +1612,11 @@ async def test_invoke_two_processes_in_out(mocker: MockerFixture) -> None:
 
 
 async def test_invoke_two_processes_in_dict_out(mocker: MockerFixture) -> None:
-    add_one = mocker.Mock(side_effect=lambda x: x + 1)
+    async def add_one_impl(x: int) -> int:
```
nfcampos, line 1615: unrelated, but this test was flaky before

libs/langgraph/langgraph/pregel/runner.py
```diff
@@ -331,9 +331,9 @@ def writer(
                     continue
                 # schedule the next task, if the callback returns one
                 wcall = calls[idx] if calls is not None else None
-                cnt = scratchpad.call_counter
-                scratchpad.call_counter += 1
-                if next_task := self.schedule_task(task, cnt, wcall):
+                if next_task := self.schedule_task(
+                    task, scratchpad.call_counter(), wcall
```
benjamincburns, line 335: nit: make call_counter a property?

benjamincburns (approved):
looks reasonable

libs/langgraph/langgraph/pregel/runner.py
benjamincburns: ah ignore me - I misread it -it's get and increment, so there's a side effect

nfcampos: it is a property, whose value is a function that returns the next value. I think auto incrementing on property getter might be a little too hard to read?

benjamincburns: Yeah, I came back and realized the sytanx I suggested was poor because I realized I hadn't seen where you were incrementing the counter. Revised nit might be that naming it something like `get_and_inc_call_counter()` would be more readable (albeit verbose).

PR #3172, opened 2025-01-23
update concepts

eyurtsev:


docs/docs/concepts/functional_api.md
```diff
@@ -1,15 +1,15 @@
 # Functional API
 
 !!! warning "Beta"
-    The Functional API is currently in **beta** and is subject to change. Please [report any issues](https://github.com/langchain-ai/langgraph/issues) or feedback to the LangGraph team. Use the label `#functional-api` when creating issues in Github.
+    The Functional API is currently in **beta** and is subject to change. Please [report any issues](https://github.com/langchain-ai/langgraph/issues) or feedback to the LangGraph team.
 
 ## Overview
 
-LangGraph's **Functional API** allows using standard control flow constructs (e.g., loops, conditionals) to define complex workflows without having to define a graph.
+The Functional API is an alternative to [State Graph](low_level.md#stategraph) for development in LangGraph.
```
vbarda, line 8: should this be "Graph API" for consistency? or we should use "State Graph" everywhere

docs/docs/concepts/functional_api.md
```diff
@@ -155,117 +153,147 @@ You will usually want to pass a **checkpointer** to the `@entrypoint` decorator
...
     ```
 
-
-!!! important "Serialization"
-
-    The **inputs** and **outputs** of entrypoints must be JSON-serializable to support checkpointing. Please see the [serialization](#serialization) section for more details.
-
 ### Executing
 
 Using the [`@entrypoint`](#entrypoint) yields a [EntrypointPregel][langgraph.func.EntrypointPregel] object that can be executed using the `invoke`, `ainvoke`, `stream`, and `astream` methods.
```
vbarda, line 203: should `EntrypointPregel` removed?

docs/docs/concepts/functional_api.md
```diff
@@ -155,117 +153,147 @@ You will usually want to pass a **checkpointer** to the `@entrypoint` decorator
...
+    }
+
+    async for chunk in my_workflow.astream(some_input, config):
         print(chunk)
     ```
 
 ### Resuming
 
-Execution can be **resumed** using the [Command][langgraph.types.Command] primitive.
+Resuming an execution after an [interrupt][langgraph.types.interrupt] can be done by passing a **resume** value to the [Command][langgraph.types.Command] primitive.
```
vbarda, line 255: should we mention that people should be using non-null resume value?

vbarda (approved):

docs/docs/concepts/functional_api.md
eyurtsev: in the API reference

eyurtsev: too much detail for the conceptual docs I think -- even what we have here is arguably too much detail

PR #3173, opened 2025-01-23
langgraph: use 'prompt' param for model input preprocessing in create_react_agent

vbarda:


libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -244,6 +240,7 @@ def create_react_agent(
     interrupt_before: Optional[list[str]] = None,
     interrupt_after: Optional[list[str]] = None,
     debug: bool = False,
+    **kwargs: dict[str, Any],
```
hinthornw, line 243: should we call this something like "deprecated_kwargs" so people don't think they need to search here for real stuff?

vbarda: sure

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -244,6 +240,7 @@ def create_react_agent(
     interrupt_before: Optional[list[str]] = None,
     interrupt_after: Optional[list[str]] = None,
     debug: bool = False,
+    **deprecated_kwargs: dict[str, Any],
```
baskaryan, line 243:
```suggestion
    **deprecated_kwargs: Any,
```
believe the typing for a `**...` field automatically represents the type of the values (the keys have to be strings and the whole thing has to be a dict)

vbarda: nice will update

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -229,13 +235,13 @@ def _validate_chat_history(
 
 
 @deprecated_parameter("messages_modifier", "0.1.9", "state_modifier", removal="0.3.0")
```
nfcampos, line 237: lets remove this one, no?

eyurtsev: if it's not getting in the way for users (not affecting adding confusing things to API reference or the doc string seeing in an interactive env), we should consider removing the deprecation warning.

okay for the internal code to be backwards compatible -- we just care that things that are user facing aren't confusing

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -254,26 +260,16 @@ def create_react_agent(
...
-
-            - SystemMessage: this is added to the beginning of the list of messages in state["messages"].
             - str: This is converted to a SystemMessage and added to the beginning of the list of messages in state["messages"].
+            - SystemMessage: this is added to the beginning of the list of messages in state["messages"].
             - Callable: This function should take in full graph state and the output is then passed to the language model.
             - Runnable: This runnable should take in full graph state and the output is then passed to the language model.
+
+            !!! Note
+                Prior to `v0.2.68`, this functionality was available through `state_modifier` / `messages_modifier` parameters.
+                The `prompt` parameter is now the recommended parameter.
```
eyurtsev, line 272: say to remove this sentence

```suggestion
```

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -254,26 +260,16 @@ def create_react_agent(
...
-            Can take a few different forms:
-
-            - SystemMessage: this is added to the beginning of the list of messages in state["messages"].
             - str: This is converted to a SystemMessage and added to the beginning of the list of messages in state["messages"].
+            - SystemMessage: this is added to the beginning of the list of messages in state["messages"].
             - Callable: This function should take in full graph state and the output is then passed to the language model.
             - Runnable: This runnable should take in full graph state and the output is then passed to the language model.
+
+            !!! Note
+                Prior to `v0.2.68`, this functionality was available through `state_modifier` / `messages_modifier` parameters.
```
eyurtsev, line 271:
```suggestion
                Prior to `v0.2.68`, the prompt was set using `state_modifier` / `messages_modifier` parameters.
```

libs/langgraph/langgraph/prebuilt/chat_agent_executor.py
vbarda: sure, we can, but we also should remove it in 0.3 and simplify the backwards compatibility code as a result

dpublic:
These types of breaking changes should be highlighted as BREAKING to stand out in the "Changes since" list.

vbarda:
> These` types of breaking changes should be highlighted as BREAKING to stand out in the "Changes since" list.
This is not a breaking change as the previous parameter (`state_modifier`) is fully supported in a backwards-compatible manner. That being said, we will highlight more prominently in the release notes

dpublic:
I checked the file changes and state_modifier is removed from the create_react_agent() argument list.

vbarda:
> I checked the file changes and state_modifier is removed from the create_react_agent() parameter list.

see this https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/prebuilt/chat_agent_executor.py#L134

PR #3185, opened 2025-01-24
Do not inject args in RunnableCallable if arg already exists

nfcampos:


libs/langgraph/langgraph/utils/runnable.py
```diff
@@ -78,9 +77,19 @@ class StrEnum(str, enum.Enum):
...
+            # Best effort to catch some forward references.
+            # This will not work for cases like `"Union[None, BaseStore]"`,
+            # we'll need to re-write logic to use get_type_hints()
+            # to resolve forward references.
+            "Optional[BaseStore]",
+            inspect.Parameter.empty,
+        ),
         CONFIG_KEY_STORE,
-        inspect.Parameter.empty,
+        None,
```
eyurtsev, line 92: Actually this will introduce a new bug since `None` shouldn't be assigned to BaseStore -- i'll split this into two cases prior to merging

libs/langgraph/langgraph/utils/runnable.py
```diff
@@ -77,11 +79,32 @@ class StrEnum(str, enum.Enum):
...
+        (
+            Optional[BaseStore],
+            # Best effort to catch some forward references.
+            # This will not work for cases like `"Union[None, BaseStore]"`,
+            # we'll need to re-write logic to use get_type_hints()
+            # to resolve forward references.
+            "Optional[BaseStore]",
+        ),
+        CONFIG_KEY_STORE,
+        # Default value is None, so we don't need to pass it here.
```
eyurtsev, line 105:
```suggestion
```

PR #3186, opened 2025-01-24
Improve typings for task, it now returns a mixed sync/async future protocol

nfcampos:
- note this type is never instantiated, it is only used for typing (we cannot make it a protocol as it inherits from concurrent.futures.Future)


hinthornw (approved):

eyurtsev (approved):
thank you

PR #3194, opened 2025-01-24
Update type annotations in errors.py and types.py in langgraph

iamarunbrahma:
Update type annotations in `errors.py` for __slots__  and in `types.py` for Error Codes.


nfcampos:
hi, why is this needed?

eyurtsev:
This does not seem necessary. If there's an issue with any type checking linters let us know. Closing for now.

PR #3202, opened 2025-01-24
langgraph: add names for tasks

vbarda:


libs/langgraph/tests/test_pregel.py
```diff
@@ -5570,10 +5570,10 @@ def graph(state: dict) -> dict:
 
 
 @pytest.mark.parametrize("checkpointer_name", ALL_CHECKPOINTERS_SYNC)
-def test_multiple_interrupts_imperative(
+def test_multiple_interrupts_functional(
     request: pytest.FixtureRequest, checkpointer_name: str, snapshot: SnapshotAssertion
 ):
-    """Test multiple interrupts with an imperative API."""
+    """Test multiple interrupts with an functional API."""
```
nfcampos, line 5576: a functional

nfcampos: or even better, just with functional api

vbarda: oops was doing find/replace, ty

PR #3217, opened 2025-01-27
docs: use named tasks for react agents

vbarda:


docs/docs/how-tos/wait-user-input-functional.ipynb
```diff
@@ -300,35 +300,28 @@
...
     "\n",
     "@task\n",
     "def call_model(messages):\n",
     "    \"\"\"Call model with a sequence of messages.\"\"\"\n",
     "    response = model.bind_tools(tools).invoke(messages)\n",
-    "    return response\n",
-    "\n",
-    "\n",
-    "@task\n",
-    "def call_tool(tool_call):\n",
```
eyurtsev, line 328: why can we not use call tool?

vbarda: we can! this was just an alternative approach

PR #3220, opened 2025-01-27
functional api: remove generator support

eyurtsev:
Remove generator support in entrypoint.

### DIFF
--- a/libs/langgraph/tests/test_pregel.py
+++ b/libs/langgraph/tests/test_pregel.py
@@ -5824,50 +5824,13 @@ def test_entrypoint_from_sync_generator() -> None:
     """@entrypoint does not support sync generators."""
     previous_return_values = []
 
-    @entrypoint(checkpointer=MemorySaver())
-    def foo(inputs, previous=None) -> Any:
-        previous_return_values.append(previous)
-        yield "a"
-        yield "b"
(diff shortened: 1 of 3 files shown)


libs/langgraph/tests/test_pregel.py
```diff
@@ -5824,50 +5824,13 @@ def test_entrypoint_from_sync_generator() -> None:
...
-
-
-def test_entrypoint_request_stream_writer() -> None:
-    """Test using a stream writer with an entrypoint."""
-
-    @entrypoint(checkpointer=MemorySaver())
-    def foo(inputs, writer: StreamWriter) -> Any:
-        writer("a")
-        yield "b"
+    with pytest.raises(NotImplementedError):
```
benjamincburns, line 5827: nit - maybe mark the test as skipped or just drop it entirely?

benjamincburns (approved):
LGTM

libs/langgraph/tests/test_pregel.py
eyurtsev: what behavior do you want if someone attempts to decorator a generator? this test verifies that we raise a not implemented error?

benjamincburns: ah, I misread - makes sense

PR #3221, opened 2025-01-27
langgraph: more unittests for functional API

vbarda:


libs/langgraph/tests/test_pregel.py
```diff
@@ -6053,7 +6053,76 @@ def parent_call_multiple_subgraphs(inputs):
...
+        a: int
+        b: int
+
+    class Output(TypedDict):
+        result: int
+
+    # Define addition subgraph
+    @entrypoint()
+    def add(inputs):
+        a, b = inputs
```
eyurtsev, line 6071: Perhaps use a dict here instead of a 2-tuple to make the test code easier to read?

libs/langgraph/tests/test_pregel.py
```diff
@@ -6053,7 +6053,76 @@ def parent_call_multiple_subgraphs(inputs):
 
 
 @pytest.mark.parametrize("checkpointer_name", ALL_CHECKPOINTERS_SYNC)
-def test_multiple_subgraphs_mixed_checkpointer(
+def test_multiple_subgraphs_mixed_state_graph(
```
eyurtsev, line 6056: Maybe do a quick pass with cursor to just inline comments about what's being tested?

eyurtsev (approved):

libs/langgraph/tests/test_pregel.py
vbarda: i'll add a type annotation but i think useful to test non-dict inputs

PR #3228, opened 2025-01-28
docs: add graph API basics section

ccurme:
Screenshot 2025-01-30 at 3 53 05 PM


docs/docs/how-tos/state-reducers.ipynb
```diff
@@ -0,0 +1,421 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# How to define and update graph state\n",
+    "\n",
+    "This guide demonstrates how to define and update [state](../../concepts/low_level/#state) in LangGraph. We will demonstrate:\n",
+    "\n",
+    "1. How do use state to define a graph's [schema](../../concepts/low_level/#schema)\n",
```
vbarda, line 11:
```suggestion
    "1. How to use state to define a graph's [schema](../../concepts/low_level/#schema)\n",
```

docs/docs/how-tos/state-reducers.ipynb
```diff
@@ -0,0 +1,421 @@
...
+    "2. How to use [reducers](../../concepts/low_level/#reducers) to control how state updates are processed.\n",
+    "\n",
+    "We will use [messages](../../concepts/low_level/#working-with-messages-in-graph-state) in our examples. This represents a versatile formulation of state for many LLM applications. See our [concepts page](../../concepts/low_level/#working-with-messages-in-graph-state) for more detail.\n",
```
vbarda, line 14: should messages link to `MessagesState`?

docs/docs/how-tos/state-reducers.ipynb
```diff
@@ -0,0 +1,421 @@
...
+    " </div>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Defining state\n",
+    "\n",
+    "[State](../../concepts/low_level/#state) in LangGraph can be either a `TypedDict` or a `Pydantic` model. Below we will use `TypedDict`. See [this guide](../../how-tos/state-model) for detail on using Pydantic.\n",
```
vbarda, line 49: technically can also be a dataclass

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "\n",
+    "graph_builder = StateGraph(State)\n",
+    "\n",
+    "# Add nodes\n",
+    "graph_builder.add_node(step_1)\n",
+    "graph_builder.add_node(step_2)\n",
+    "graph_builder.add_node(step_3)\n",
+    "\n",
+    "# Add edges\n",
+    "graph_builder.set_entry_point(\"step_1\")\n",
```
vbarda, line 156: same comment about `set_entry_point` -- let's switch to `add_edge`

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "graph.invoke({\"value_1\": \"c\"})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that:\n",
+    "\n",
+    "- We kicked off invocation by updating a single key of the state.\n",
```
vbarda, line 260:
```suggestion
    "- We kicked off invocation by providing a value for a single state key.\n",
```

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+   ],
+   "source": [
+    "graph.invoke({\"value_1\": \"c\"})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that:\n",
```
vbarda, line 258: worth adding a note that you always need to provide a value for at least one of the keys?

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "\n",
+    "1. `.add_edge` takes the names of nodes, which for functions defaults to `node.__name__`.\n",
+    "2. We must specify the entry point of the graph. For this we used `.set_entry_point`, but we can also control where our graph starts and ends by adding edges with the [START node](../../concepts/low_level/#start-node) and [END node](../../concepts/low_level/#end-node).\n",
```
vbarda, line 181: let's change this to just recommend `.add_edge()` instead of entrypoint
also, let's remove the part about "controlling where the graph ends w/ END" cause it confuses people. in fact, it's good to clarify that here as well for new folks, e.g. add something like "the graph halts when there are no more tasks to execute"

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "    return {\"value_2\": 10}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that when issuing updates to the state, each node can just specify the value of the key it wishes to update.\n",
+    "\n",
+    "By default, this will **overwrite** the value of the corresponding key. You can also use [reducers](../../concepts/low_level/#reducers) to control how updates are processed-- for example, you can append successive updates to a key instead. See [this guide](../../how-tos/state-reducers) for more detail.\n",
```
vbarda, line 131: super nit -- instead of `--` could use emdash `— `

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
+{
```
vbarda, line 1: could we lift a small example up top to the header?

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "# How to create a sequence of steps\n",
+    "\n",
+    "!!! info \"Prerequisites\"\n",
+    "    This guide assumes familiarity with the following:\n",
+    "\n",
+    "    - [How to define and update graph state](../../how-tos/state-reducers)\n",
+    "\n",
+    "This guide demonstrates how to construct a simple sequence of steps. We will demonstrate:\n",
+    "\n",
+    "1. How to build the graph\n",
```
vbarda, line 16:
```suggestion
    "1. How to build a sequential graph\n",
```

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "2. Built-in short-hand for constructing similar graphs.\n",
+    "\n",
+    "\n",
+    "<details>\n",
+    "<summary>Why split application steps into a sequence with LangGraph?</summary>\n",
+    "\n",
+    "The primary benefit of using LangGraph for this case is access to the underlying persistence layer. LangGraph\n",
```
vbarda, line 23: other benefits are streaming and tracing (e.g. observability)

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "\n",
+    "def step_3(state: State):\n",
+    "    return {\"value_2\": 10}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that when issuing updates to the state, each node can just specify the value of the key it wishes to update.\n",
```
vbarda, line 129: this is actually a super important point and we should highlight it with a callout. there are a lot of people who have an antipattern of returning `{**state, "updated_key": "updated_value}`

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+       "<IPython.core.display.Image object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from IPython.display import Image, display\n",
+    "from langchain_core.runnables.graph import CurveStyle, MermaidDrawMethod, NodeStyles\n",
```
vbarda, line 220: not needed

docs/docs/how-tos/sequence.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "         Sign up for LangSmith to quickly spot issues and improve the performance of your LangGraph projects. LangSmith lets you use trace data to debug, test, and monitor your LLM aps built with LangGraph — read more about how to get started in the <a href=\"https://docs.smith.langchain.com\">docs</a>. \n",
+    "     </p>\n",
+    " </div>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## How to build the graph\n",
```
vbarda, line 68: consider changing to just "Build the graph" -- i find it a bit confusing to see "how to" in subheaders + breaks symmetry w/ other headers

docs/docs/how-tos/branching.ipynb
```diff
@@ -82,14 +82,12 @@
...
-    "In this example, we fan out from `Node A` to `B and C` and then fan in to `D`. With our state, [we specify the reducer add operation](https://langchain-ai.github.io/langgraph/concepts/low_level/#reducers). This will combine or accumulate values for the specific key in the State, rather than simply overwriting the existing value. For lists, this means concatenating the new list with the existing list. \n",
-    "\n",
-    "Note that LangGraph uses `Annotated` type to specify reducer functions for specific keys in the State: it maintains the original type (`list`) for type checking, but allows attaching the reducer function (`add`) to the type without changing the type itself."
+    "In this example, we fan out from `Node A` to `B and C` and then fan in to `D`. With our state, [we specify the reducer add operation](https://langchain-ai.github.io/langgraph/concepts/low_level/#reducers). This will combine or accumulate values for the specific key in the State, rather than simply overwriting the existing value. For lists, this means concatenating the new list with the existing list. See [this guide](../../how-tos/state-reducers) for more detail on updating state with reducers."
```
vbarda, line 85: can we retitle to "How to run graph nodes in parallel"

docs/docs/how-tos/branching.ipynb
```diff
@@ -192,6 +201,16 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "ea5495cf-9564-40c6-bc2d-0b2a8f72a5df",
+   "metadata": {},
+   "source": [
+    "!!! note\n",
+    "\n",
+    "    In the above example, nodes `\"b\"` and `\"c\"` are executed concurrently in the same [superstep](../../concepts/low_level/#graphs). Because they are in the same step, node `\"d\"` executes after both `\"b\"` and `\"c\"` are finished."
+   ]
+  },
```
vbarda, line 213: the callout below needs to be clarified actually: while the whole superstep errors, we DO actually save the results from successful nodes (when used w/ a checkpointer) so that when we resume, we don't re-execute those nodes

docs/docs/how-tos/branching.ipynb
```diff
@@ -317,6 +320,18 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "id": "00d33cb0-5a47-4057-bc55-0fc14c6034fc",
+   "metadata": {},
+   "source": [
+    "!!! note\n",
+    "\n",
+    "    <p>In the above example, nodes `\"b\"` and `\"c\"` are executed concurrently in the same [superstep](../../concepts/low_level/#graphs). What happens in the next step?</p>\n",
+    "    <p>We use `add_edge([\"b_2\", \"c\"], \"d\")` here to force node `\"d\"` to only run when nodes `\"b_2\"` and `\"c\"` have finished execution. If we added two separate edges,\n",
```
vbarda, line 331:
```suggestion
    "    <p>We use `add_edge([\"b_2\", \"c\"], \"d\")` here to force node `\"d\"` to only run when both nodes `\"b_2\"` and `\"c\"` have finished execution. If we added two separate edges,\n",
```

docs/docs/how-tos/branching.ipynb
```diff
@@ -317,6 +320,18 @@
...
+    "    node `\"d\"` would run twice.</p>"
```
vbarda, line 332: not sure if this is helpful, but might be worth clarifying

```suggestion
    "    node `\"d\"` would run twice: after node `b2` finishes and once again after node `c` (in whichever order those nodes finish).</p>"
```

docs/docs/how-tos/branching.ipynb
```diff
@@ -341,31 +356,48 @@
     "\n",
     "from typing_extensions import TypedDict\n",
     "\n",
-    "from langgraph.graph import END, START, StateGraph\n",
+    "from langgraph.graph import StateGraph, START, END\n",
```
vbarda, line 359: we need to remove or rewrite the line above

```
"If you have a known \"sink\" node that the conditional branches will route to afterwards, you can provide `then=<final-node-name>` when creating the conditional edges."
```

we are not actually using `then` here

docs/docs/how-tos/branching.ipynb
```diff
@@ -480,11 +511,11 @@
    "source": [
     "## Stable Sorting\n",
```
vbarda, line 512: is this example actually helpful? should we remove?

docs/docs/how-tos/index.md
```diff
@@ -9,16 +9,24 @@ Here you’ll find answers to “How do I...?” types of questions. These guide
...
 ## LangGraph
 
-### Controllability
+### Graph API Basics
 
-LangGraph offers a high level of control over the execution of your graph.
+- [How to define and update graph state](state-reducers.ipynb)
+- [How to create a sequence of steps](sequence.ipynb)
+- [How to create branches for parallel execution](branching.ipynb)
+- [How to create and control loops with recursion limits](recursion-limit.ipynb)
```
vbarda, line 17: we have different names for this one -- i kind of like the shorter version `How to create and control loops`

docs/docs/how-tos/recursion-limit.ipynb
```diff
@@ -31,9 +31,26 @@
     "</div> \n",
     "\n",
     "\n",
-    "You can set the graph recursion limit when invoking or streaming the graph. The recursion limit sets the number of **supersteps** that the graph is allowed to execute before it raises an error. Read more about the concept of recursion limits [here](https://langchain-ai.github.io/langgraph/concepts/low_level/#recursion-limit). Let's see an example of this in a simple graph with parallel branches to better understand exactly how the recursion limit works.\n",
+    "You can set the graph recursion limit when invoking or streaming the graph. The recursion limit sets the number of [supersteps](../../concepts/low_level/#graphs) that the graph is allowed to execute before it raises an error. Read more about the concept of recursion limits [here](../../concepts/low_level/#recursion-limit). \n",
```
vbarda, line 34: should we actually add a sentence and a code snippet to demonstrate how to add a loop?
otherwise, the title says "how to create a loop" but we immediately jump to recursion limit

docs/docs/how-tos/recursion-limit.ipynb
```diff
@@ -88,44 +107,44 @@
...
+    "# Define nodes\n",
+    "builder = StateGraph(State)\n",
+    "builder.add_node(a)\n",
+    "builder.add_node(b)\n",
     "\n",
     "\n",
-    "def node_d(state):\n",
-    "    return {\"aggregate\": [\"I'm A\"]}\n",
+    "# Define edges\n",
+    "def route(state: State) -> Literal[\"b\", END]:\n",
```
vbarda, line 127: i wonder if we could restructure this a bit to present this as another way to control a loop, e.g.:
- control via a conditional edge w/ some termination condition (could also link to react agent docs for this as an example)
- control via recursion limit

docs/docs/how-tos/recursion-limit.ipynb
```diff
@@ -175,7 +240,16 @@
...
    "source": [
-    "Success! The graph raised an error as expected - now let's test setting the recursion limit to 4 and ensure that the graph succeeds in this case:"
+    "Note that this time we terminate after the fourth step."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Additional example\n",
```
vbarda, line 250: is this example necessary? perhaps we can simplify? is the idea here to show how supersteps are counted towards recursion limit?

docs/docs/how-tos/state-reducers.ipynb
```diff
@@ -0,0 +1,421 @@
+{
```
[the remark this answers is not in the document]
ccurme: Did (1) and (3). I left (2) because you can see more or less the same thing pretty close to the top as-is.

docs/docs/how-tos/state-reducers.ipynb
ccurme: any docs on this?

docs/docs/how-tos/state-reducers.ipynb
```diff
@@ -0,0 +1,430 @@
...
+   "source": [
+    "%%capture --no-stderr\n",
+    "%pip install -U langgraph"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"admonition tip\">\n",
```
eyurtsev, line 35: Use !!! tip when possible -- we'll have to re-write all of these at some point (not blocking)

ccurme: where these appear it's because I copied them from elsewhere

PR #3248, opened 2025-01-30
docs: concepts update

eyurtsev:


docs/docs/concepts/functional_api.md
```diff
@@ -5,11 +5,21 @@
...
+The Functional API uses two key building blocks:  
+
+- **`@entrypoint`** – Marks a function as the starting point of a workflow, encapsulating logic and managing execution flow, including handling long-running tasks and interrupts.  
+- **`@task`** – Represents a discrete unit of work, such as an API call or data processing step, that can be executed asynchronously within an entrypoint. Tasks return a future-like object that can be awaited or resolved synchronously.  
+
+This provides a minimal abstraction for building workflows with state management and streaming.
+
+!!! tip 
+
+    For users who prefer a more declarative approach, LangGraph's [Graph API](./low_level.md) allows you to define workflows using a Graph DSL. Both APIs share the same underlying runtime, so you can use them together in the same application.
```
vbarda, line 21: nit - should we really call it DSL?

nfcampos: I'd call it a paradigm

nfcampos (approved):

docs/docs/concepts/functional_api.md
eyurtsev, line 21:
```suggestion
    For users who prefer a more declarative approach, LangGraph's [Graph API](./low_level.md) allows you to define workflows using the Graph paradigm. Both APIs share the same underlying runtime, so you can use them together in the same application.
```

PR #3250, opened 2025-01-30
langgraph: inject stream writer into tools

vbarda:


vbarda:
Closing in favor of #3251

PR #3251, opened 2025-01-30
langgraph: add get_stream_writer()

vbarda:
Alternative to #3250


nfcampos (approved):

libs/langgraph/langgraph/config.py
```diff
@@ -0,0 +1,169 @@
...
+        3
+        ```
+    """
+    config = get_config()
+    return config[CONF][CONFIG_KEY_STORE]
+
+
+def get_stream_writer() -> StreamWriter:
+    """Access LangGraph [StreamWriter][langgraph.types.StreamWriter] from inside a graph node or entrypoint task at runtime.
+
```
eyurtsev, line 112: add warning for async 3.10 and below?

PR #3255, opened 2025-01-31
Guard and cache calls to find_subgraph_pregel

nfcampos:
- The result of these doesnt change once a node is created, and it's fairly expensive to run, so great thing to cache
- There's a variety of errors that can come from inspecting the source code of a function (part of what this does) so adding a catch-all try-except block as this should be best-effort, not crash your graph

### DIFF
--- a/libs/langgraph/langgraph/pregel/read.py
+++ b/libs/langgraph/langgraph/pregel/read.py
@@ -22,7 +22,9 @@ from langchain_core.runnables.base import Input, Other, coerce_to_runnable
 from langchain_core.runnables.utils import ConfigurableFieldSpec
 
 from langgraph.constants import CONF, CONFIG_KEY_READ
+from langgraph.pregel.protocol import PregelProtocol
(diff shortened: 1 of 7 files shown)


libs/langgraph/langgraph/pregel/read.py
```diff
@@ -165,9 +170,21 @@ def __init__(
...
+            if subgraph:
+                self.subgraphs = [subgraph]
+            else:
+                self.subgraphs = []
+        else:
+            self.subgraphs = []
 
     def copy(self, update: dict[str, Any]) -> PregelNode:
         attrs = {**self.__dict__, **update}
+        attrs.pop("subgraphs")
```
hinthornw, line 187: I'm guessing no need for a `None` here

hinthornw (approved):

libs/langgraph/langgraph/pregel/read.py
nfcampos: I think we actually always assign to subgraphs?

PR #3278, opened 2025-02-01
Update application_structure.md

laelhalawani:
moved requirements.txt to root of the project in the recommended structure


vbarda (approved):
Thanks for catching

PR #3355, opened 2025-02-07
docs: concepts durable execution

eyurtsev:
Conceptual page for durable execution


docs/docs/concepts/durable_execution.md
```diff
@@ -0,0 +1,50 @@
...
+
+As a result, when you're writing a workflow for durable execution, you should wrap any non-deterministic operations (e.g., random number generation) and any operations with side effects (e.g., file writes, API calls) inside [tasks](./functional_api.md#task).
+
+- **Avoid Repeating Work**  Place operations that produce side effects (such as logging, file writes, or network calls) after an interrupt in Graph API nodes or encapsulate them within [tasks](./functional_api.md#task). This prevents their unintended repetition when the workflow is resumed. LangGraph will look up information stored in the persistence layer about any **tasks** that were previously executed for the specific run and will swap in the results of those tasks instead of re-executing them.
```
benjamincburns, line 27: > encapsulate them within tasks

Or nodes/runnables?

docs/docs/concepts/durable_execution.md
```diff
@@ -0,0 +1,50 @@
...
+- **Encapsulate Non-Deterministic Operations:**  Wrap any code that might yield non-deterministic results (e.g., random number generation) inside **tasks**. This ensures that, upon resumption, the workflow follows the exact recorded sequence of steps with the same outcomes.
+
+For some examples of pitfalls to avoid, see the [Common Pitfalls](./functional_api.md#common-pitfalls) section in the functional API, which shows
+how to structure your code using **tasks** to avoid these issues. The same principles apply to the Graph API (you can use **tasks** from inside [nodes](./low_level.md#nodes) as well).
+
+## Idempotency
+
+Idempotency ensures that running the same operation multiple times produces the same result. This helps prevent duplicate API calls and redundant processing if a step is rerun due to a failure. Always place API calls inside **tasks** functions for checkpointing, and design them to be idempotent in case of re-execution. Re-execution can occur if a **task** starts, but does not complete successfully. Then, if the workflow is resumed, the **task** will run again. Use idempotency keys or verify existing results to avoid duplication.
```
benjamincburns, line 35: I think it would be clearer to mention idempotency in the section above. Making side-effect-producing logic idempotent should be the first suggestion, IMO.

benjamincburns (approved):
Looks good overall, and I think it'd be fine to merge as-is. That said, I do think it could be improved slightly. See comments.

docs/docs/concepts/durable_execution.md
eyurtsev: Nodes yes, runnables no?

I can add a clarification on nodes. I chose to describe `tasks` only with the intention to avoid confusing people with two ways of doing the same thing. `tasks` are lighter weight since you don't need to declare a node (which implies declaring a schema). Imagine that having to accommodate a lot of different side-effects each with its own payload that's not immediately derivable from the shared graph state.

benjamincburns: I generally agree with your remarks about `tasks` and why they're preferable over nodes. But since this is a concept doc, I figured it's best to paint a complete picture, else someone might've read this and thought that functional API is a requirement for durability.

Also re: runnables, I'm fine with leaving those off. I was just being a bit pedantic, as any langchain runnable can be used as a node

docs/docs/concepts/durable_execution.md
eyurtsev: Agreed, merged into the previous paragraph. It's the third bullet point rather than the first, since the main objective of the conceptual doc is to the happy path of durable execution (i.e., what it does and how to set it up).

PR #3356, opened 2025-02-08
docs: fix grammatical error for navigating node to parent section

DangerousPotential:
Add within to better explain moving from one subgraph to another


vbarda (approved):
thank you!

PR #3384, opened 2025-02-11
langgraph: allow passing destinations to .add_node

vbarda:


nfcampos (approved):

libs/langgraph/tests/test_pregel.py
```diff
@@ -6416,3 +6417,61 @@ def test_tags_stream_mode_messages() -> None:
...
+
+        return Command(
+            update={"foo": value},
+            goto=goto,
+            graph=Command.PARENT,
+        )
+
+    subgraph = StateGraph(State).add_node(node_a).add_edge(START, "node_a").compile()
+
+    def node_b(state: State):
```
eyurtsev, line 6441: unrelated to this PR -- should we not have mypy fail this code due to missing type annotations?

libs/langgraph/tests/test_pregel.py
```diff
@@ -6416,3 +6417,61 @@ def test_tags_stream_mode_messages() -> None:
...
+        return {"foo": "b"}
+
+    def node_c(state: State):
+        return {"foo": "c"}
+
+    # destinations w/ tuples
+    builder = StateGraph(State)
+    builder.add_edge(START, "child")
+    builder.add_node("child", subgraph, destinations=("node_b", "node_c"))
```
eyurtsev, line 6450: do we need a test if subgraph is a regular node rather than a subgraph?

libs/langgraph/tests/test_pregel.py
vbarda: we're not running mypy on tests

PR #3411, opened 2025-02-13
langgraph: fix None handling for pydantic state updates

vbarda:


libs/langgraph/tests/test_pregel.py
```diff
@@ -6482,3 +6482,16 @@ def node_c(state: State):
             Edge(source="child", target="node_b", data="foo", conditional=True),
             Edge(source="child", target="node_c", data="bar", conditional=True),
         ] == graph.edges
+
+
+def test_pydantic_none_state_update() -> None:
+    from pydantic import BaseModel
+
+    class State(BaseModel):
+        foo: str | None
```
eyurtsev, line 6491: this works already in the none pydantic case?

eyurtsev:
:rocket:

libs/langgraph/tests/test_pregel.py
vbarda: correct

PR #3440, opened 2025-02-14
docs: how-to batch conversion

eyurtsev:
- update
- x
- x
- update
- x
- x


docs/docs/how-tos/branching.md
```diff
@@ -98,7 +98,7 @@ graph = builder.compile()
 ```
 
 
-```python exec="on" source="above" session="1" result="ansi"
+```python
```
eyurtsev, line 101: hmm i should restore this actually, we want to execute, but without result

PR #3445, opened 2025-02-14
Fix busy loop in AsyncBatchedBaseStore

nfcampos:
- while loop w asyncio.sleep(0) takes up cpu


libs/checkpoint/langgraph/store/base/batch.py
```diff
@@ -55,9 +56,11 @@ def __init__(self) -> None:
         super().__init__()
         self._loop = asyncio.get_running_loop()
         self._aqueue: dict[asyncio.Future, Op] = {}
+        self._aqueue_ = asyncio.Queue()
         self._task = self._loop.create_task(_run(self._aqueue, weakref.ref(self)))
 
     def __del__(self) -> None:
+        print("cancelling task")
```
hinthornw, line 63: nit

hinthornw (approved):

PR #3469, opened 2025-02-17
fix(docs): update outdated docs about [Using the async MongoDB client]

XenoAmess:
as title.

### DIFF
--- a/docs/docs/how-tos/persistence_mongodb.ipynb
+++ b/docs/docs/how-tos/persistence_mongodb.ipynb
@@ -65,7 +65,7 @@
    "outputs": [],
    "source": [
     "%%capture --no-stderr\n",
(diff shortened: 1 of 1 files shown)


vbarda:
The documentation is actually correct -- please use the latest version of `pymongo`

XenoAmess:
> The documentation is actually correct -- please use the latest version of `pymongo`

@vbarda 
image

cannot  
```
use the latest version of `pymongo`
```

XenoAmess:
@vbarda see this. https://github.com/langchain-ai/langchain-mongodb/commit/1dbfee3354bd39246af35909288b0f9d4d98f2da

sigh.

so before this commit, the real 0.1.0 in central actually can only use pymongo 4.9.x

I really suggest you use a snapshot version after you release one version & before next release. but this is python so as python is just always dirty...sigh

PR #3471, opened 2025-02-17
Add custom router support in langgraph-api

hinthornw:
Add support for a custom Router type to specify
- Custom routes
- Custom middleware
- Custom lifespans

Why custom type vs. permitting Starlette Router/App directly?
tl;dr - maintaining deps. Starlette Router() builds the middleware stack without retaining pointers to the original middleware. Starlette() app itself has other things that we may not want to support. 

Custom type lets us limit the surface area. It's just a struct.

### DIFF
--- a/libs/sdk-py/langgraph_sdk/routing/__init__.py
+++ b/libs/sdk-py/langgraph_sdk/routing/__init__.py
@@ -1,0 +1,111 @@
+import typing
+
+from langgraph_sdk.routing import types
+from langgraph_sdk.routing.types import Middleware
+
+
+@typing.final
+class Router:
+    """Add routes, middleware, and manage application lifecycle.
+
+    Define custom routes, apply middleware globally, and handle application startup/shutdown.
+    Middleware runs on all routes (including default LangGraph endpoints like /runs/, /assistants/, etc).
+    Custom routes take precedence over default ones, so you can override default behavior if needed.
+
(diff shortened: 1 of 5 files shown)


libs/sdk-py/langgraph_sdk/routing/__init__.py
```diff
@@ -0,0 +1,111 @@
+import typing
+
+from langgraph_sdk.routing import types
+from langgraph_sdk.routing.types import Middleware
+
+
+@typing.final
+class Router:
```
eyurtsev, line 8: This looks like a container for 3 independent things b/c there's no api router concept in starlette. 

It may be helpful to restructure the first documentation paragraph visually, so it's easier to see that the Middleware has global scope. Folks may have the similar confusion about the lifespan (although less likely to be an issue). -- content is good, i think only visual layout (e.g., markdown bullet point)

hinthornw: (had somem updates so we're goingt o directly accept a starlette app)

hinthornw: There is a Starlette router but it constructs the middleware at init time so we wouldn't be able to extract the relevant bits and put it in our router or apply the middleware to the other routes (which imo is expected by the user) without it gettting a tad fragile

PR #3488, opened 2025-02-18
langgraph: fallback on no-op writer in get_stream_writer

vbarda:


libs/langgraph/langgraph/config.py
```diff
@@ -177,4 +177,4 @@ def workflow(value: int):
         ```
     """
     config = get_config()
-    return config[CONF][CONFIG_KEY_STREAM_WRITER]
+    return config[CONF].get(CONFIG_KEY_STREAM_WRITER, lambda c: None)
```
nfcampos, line 180: lets define the fallback function once at top-level

nfcampos (approved):

PR #3505, opened 2025-02-19
Support http config

hinthornw:
Right now requires that the file be in one of the local packages/dependencies.

### DIFF
--- a/libs/cli/examples/my_app.py
+++ b/libs/cli/examples/my_app.py
@@ -1,0 +1,62 @@
+from contextlib import asynccontextmanager
+from contextvars import ContextVar
+from typing import Any
+
+from starlette.applications import Starlette
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.responses import JSONResponse
+from starlette.routing import Route
+
+my_context_var: ContextVar[str] = ContextVar("my_context_var", default="")
+LIFESPAN_VAL = ""
+other_context_var = ContextVar("other_context_var", default="")
+
+
+@asynccontextmanager
+async def my_lifespan(app):
+    global LIFESPAN_VAL
+    LIFESPAN_VAL = "foobar-lifespan"
+    yield
+    assert LIFESPAN_VAL == "foobar-lifespan"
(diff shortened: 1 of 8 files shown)


libs/cli/langgraph_cli/config.py
```diff
@@ -86,6 +86,33 @@ class AuthConfig(TypedDict, total=False):
...
+    allow_origins: list[str]
+    allow_methods: list[str]
+    allow_headers: list[str]
+    allow_credentials: bool
+    allow_origin_regex: str
+    expose_headers: list[str]
+    max_age: int
+
+
+class HttpConfig(TypedDict, total=False):
```
eyurtsev, line 99: is `httpconfig` the common name?

hinthornw: haha i was debating this with nuno. Unsure. It has:
- Custom app (with custom routes, middleware, and lifespan)
- route toggling
- CORS
could add port configuraiton, etc.

libs/cli/examples/my_app.py
```diff
@@ -0,0 +1,62 @@
+from contextlib import asynccontextmanager
```
eyurtsev, line 1: Why is this file here?

libs/cli/langgraph_cli/config.py
eyurtsev: maybe something in the style of:

AppConfig
ServiceConfig


---

I don't see the custom routes, middleware or lifespan right -- plan is to add to this very same container?

hinthornw: that's all defined in the custom app (via the app path)

libs/cli/examples/my_app.py
hinthornw: An example custom app/routes /etc

libs/cli/langgraph_cli/config.py
nfcampos: App doesn't mean anything. Service will be ambiguous when/if we add non http transports. I'd stick with http

PR #3534, opened 2025-02-20
langgraph: allow passing config_schema to create_react_agent

vbarda:


hinthornw (approved):
Should update docstring

PR #3559, opened 2025-02-23
prebuilt: allow pydantic model as state schema in create_react_agent

ianchi:
Inherited attributes where not considered.
Pydantic model can inherit from other pydantic models. In those cases, inherited attributes where not considered in the check and the code fails.


benjamincburns:
Thanks for the contribution @ianchi! Would you mind adding a test for this, please?

ianchi:
Hi @benjamincburns,

I updated the PR and added the test, but I've run into a problem.
It works ok for `v1` but it fails for `v2`.

It seems that the state that the conditional edge receives is not the full state (but the one in the source node?), so it cannot inject it correctly on the tool.

I don't know if this can be "fixed" somehow.
Any ideas/suggestions??

ianchi:
Hi @benjamincburns,

I found a "fix" for the v2 problem. Anyhow is more of a hack than a good fix.

Although I couldn't find it clearly in the documentation it seems that currently in Langgraph:
- conditional edges recieve the same state defined in the origin node (either by input, type annotation, or implicit). Type annotation in the conditional function are completely ignored.
- for tools, the type in the `InjectedState` annotation is ignored. The available state is injected as is, with no convertion (as is the case in nodes)

I couldn't find any other more elegant solution, given these conditions.

vbarda:
@ianchi thanks for the contribution - i pushed a bunch of updates -- let me know if you have any comments, otherwise will merge shortly

libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -57,13 +57,27 @@ class AgentState(TypedDict):
...
+
+
 class AgentStateWithStructuredResponse(AgentState):
     """The state of the agent with a structured response."""
 
     structured_response: StructuredResponse
 
 
-StateSchema = TypeVar("StateSchema", bound=AgentState)
+class AgentStateWithStructuredResponsePydantic(BaseModel):
```
vbarda, line 74: should be inheriting from `AgentStatePydantic`

libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -595,7 +617,16 @@ class Agent,Tools otherClass
         if response_format is not None:
             required_keys.add("structured_response")
 
-        if missing_keys := required_keys - set(state_schema.__annotations__):
+        # Pydantic v2
```
nfcampos, line 620: shouldn't we just be using get_type_hints for all types? That supports inherited fields I believe

vbarda: yup you're right, and it was in the original change too! will bring it back

vbarda (approved):

PR #3581, opened 2025-02-25
langgraph: add a test for Command.PARENT updates from subgraphs invoked in a function

vbarda:


libs/langgraph/tests/test_pregel.py
```diff
@@ -6321,6 +6321,71 @@ def node_3(state: State, store):
...
+    main_builder = StateGraph(State)
+    main_builder.add_node("node_1", node_1)
+    main_builder.add_node("node_2", call_subgraph)
+    main_builder.add_node("node_3", node_3)
+    main_builder.add_edge(START, "node_1")
+    main_builder.add_edge("node_2", "node_3")
+    main_graph = main_builder.compile()
+
+    assert main_graph.invoke({"foo": []}) == {
+        "foo": ["foo", "baz", "qux"],
```
eyurtsev, line 6385: Yes, this is consistent with what i'd expect teh behavior to be for a subgraph invoked as a function due to scoping

PR #3582, opened 2025-02-25
feat(react): support interrupt_before/after

dqbd:


libs/sdk-js/src/schema.ts
```diff
@@ -142,9 +142,9 @@ export interface AssistantGraph {
  * An interrupt thrown inside a thread.
  */
 export interface Interrupt<TValue = unknown> {
-  value: TValue;
-  when: "during";
-  resumable: boolean;
+  value?: TValue;
+  when: "during" | (string & {});
```
dqbd, line 146: Typing softly here as `breakpoint` is React-specific thing

PR #3584, opened 2025-02-25
StreamMode in Join [sdk]

hinthornw:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -1829,7 +1829,12 @@ async def join(self, thread_id: str, run_id: str) -> dict:
         return await self.http.get(f"/threads/{thread_id}/runs/{run_id}/join")
 
     def join_stream(
-        self, thread_id: str, run_id: str, *, cancel_on_disconnect: bool = False
+        self,
+        thread_id: str,
+        run_id: str,
+        *,
+        cancel_on_disconnect: bool = False,
+        stream_mode: Union[StreamMode, Sequence[StreamMode]] = "values",
```
nfcampos, line 1837: I dont think we should default to anything, as if you dont pass it in you get whatever stream modes you picked when creating the run

nfcampos: Also the docs should make clear that this list here can only ever be a subset of the one passed to create run

nfcampos (approved):

PR #3593, opened 2025-02-26
fix(langgraph): send update events on nodes that end in `Command.PARENT`

benjamincburns:
ported from langchain-ai/langgraphjs#877


libs/langgraph/tests/test_pregel.py
```diff
@@ -6238,26 +6238,31 @@ def node_3(state: State, store):
...
+
+    assert len(stream_results) == 5
+    assert stream_results[0] == ((), {"node_1": {"bar": ["node_1"]}})
+    assert stream_results[1] == (
+        (AnyStr("node_2:"),),
+        {"subgraph_node_1": {"foo": "foo", "bar": ["subgraph_node_1"]}},
+    )
+    assert stream_results[2] == (
+        (AnyStr("node_2:"),),
+        {"subgraph_node_2": None},
```
vbarda, line 6253: shouldn't this be `{"bar": ["subgraph_node_2"]}`?

nfcampos: No, that node didn't return any value

nfcampos (approved):

PR #3641, opened 2025-02-28
Pregel.config_schema should use config_type directly when present

nfcampos:
- the previous behavior of re-creating model through config_specs would lose custom annotations on config_type


eyurtsev (approved):

eyurtsev (changes requested):
Blocking so we don't merge until I finish testing something

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -609,6 +610,36 @@ def config_specs(self) -> list[ConfigurableFieldSpec]:
...
         ]
 
+    def config_schema(
+        self, *, include: Optional[Sequence[str]] = None
+    ) -> Type[BaseModel]:
+        # If the config type is not set explicitly OR
+        # if the checkpointer is turned on, then we'll want to infer the config.
+        # When the checkpointer is turned on, additional fields are injected
+        # into the configurable object which correspond to the checkpointing parameters.
+        if self.config_type is None or self.checkpointer is not None:
```
eyurtsev, line 620: My understanding is that when the config schema is provided, we want to ignore any auto behavior w/ the exception of the checkpointer fields? (i.e., we do want to expose thread_id as part of the configurable schema?)

eyurtsev (approved):

PR #3668, opened 2025-03-04
When retrying a previously attempted run, resume from previous checkpoint

nfcampos:
- Ignore input if being passed in when a checkpoint already exists for that run_id


libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -939,6 +943,8 @@ def __enter__(self) -> Self:
         self.prev_checkpoint_config = saved.parent_config
         self.checkpoint = saved.checkpoint
         self.checkpoint_metadata = saved.metadata
+        if self.config.get("metadata").get("run_id") == saved.metadata.get("run_id"):
```
hinthornw, line 946: oops

hinthornw (approved):

PR #3670, opened 2025-03-04
chore(langgraph): fix typing of task decorator

JP-Ellis:
An async function of the form `def foo(P) -> T` has type `Callable[[P], Awaitable[T]]`. The old type annotations then converted the function into a `Callable[[P], SyncAsyncFuture[Awaitable[T]]]` which is incorrect.

The change introduced in this commit updates the type annotations to ensure the `Awaitable[T]` is correctly unwrapped.

I've tested it locally and confirmed it work on:

```python
@task 
def sync_fn(a: int) -> int: ...

@task 
def async_fn(a: int) -> int: ...
```

Let me know if you want me to add tests, just let me know how you test type annotations.

### DIFF
--- a/libs/langgraph/langgraph/func/__init__.py
+++ b/libs/langgraph/langgraph/func/__init__.py
@@ -36,23 +36,31 @@ from langgraph.types import _DC_KWARGS, RetryPolicy, StreamMode
 
 @overload
 def task(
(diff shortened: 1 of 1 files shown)


nfcampos (approved):

nfcampos:
Thanks!

PR #3699, opened 2025-03-05
docs: Third Party packages update with ads4gpts

blackbak:


hinthornw:
Does this actually export a LangGraph compiled graph?

blackbak:
> Does this actually export a LangGraph compiled graph?

Yes through
from ads4gpts_langgraph_agent import make_ads4gpts_langgraph_agent

PR #3702, opened 2025-03-05
feat: update assistant version class

lc-arjun:


nfcampos (approved):

libs/sdk-js/src/index.ts
```diff
@@ -1,6 +1,7 @@
 export { Client } from "./client.js";
 
 export type {
+  AssistantBase,
```
lc-arjun, line 4: @dqbd any reason against exporting this? I would like to use this in Studio :)

PR #3704, opened 2025-03-05
Add TTL option for store items

hinthornw:


libs/checkpoint/langgraph/store/base/__init__.py
```diff
@@ -463,6 +463,13 @@ class PutOp(NamedTuple):
         ]
         ```
     """
+    ttl: Optional[float] = None
```
hinthornw, line 466: Could expand to a dict {"refresh_on": "rw", "after_minutes": foo} if we want it to be configurable.

nfcampos (approved):

PR #3734, opened 2025-03-07
Refactor deprecated copy method

hgs-study:
When I installed langgraph 0.3.5, it was depending on pydantic 2.10.6. 
The BaseModel class in pydantic.main has a copy() method, but it is currently deprecated and recommends using the model_copy() method. I've changed the code that uses BaseModel accordingly. 
Please check it out. Thank you. 

image


vbarda:
@hgs-study thanks for the contribution! we actually want to support both v1 and v2 - could you just add a check like this in those 2 places?

```python
 if hasattr(message, "model_copy"):
    message_ = message.model_copy()
else:
    message_ = message.copy()
```

hgs-study:
@vbarda 
Ah, I see, I thought we were only supporting v2 statically. 
Like you said, if we wanted to support both v1 and v2, i'd just have to keep the code as it is. 
I changed the code and got it working again.

PR #3737, opened 2025-03-07
feat(pregel): add bulk update state method

dqbd:
This method is useful for recreating a thread from a list of checkpoint writes. A new method is needed to clone a checkpoint that has been created from multiple writes (functional API, map-reduce)

Port of https://github.com/langchain-ai/langgraphjs/pull/969 and https://github.com/langchain-ai/langgraphjs/pull/1007


libs/langgraph/langgraph/types.py
```diff
@@ -133,6 +133,11 @@ class Interrupt:
     when: Literal["during"] = "during"
 
 
+class BulkUpdate(NamedTuple):
```
vbarda, line 136: nit - it's a single update so should probably just be called `Update` or `StateUpdate`. also, should this just be a typeddict? (don't have a super strong preference)

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1151,22 +1152,22 @@ async def aget_state_history(
...
         config: RunnableConfig,
-        values: Optional[Union[dict[str, Any], Any]],
-        as_node: Optional[str] = None,
+        updates: list[BulkUpdate],
     ) -> RunnableConfig:
-        """Update the state of the graph with the given values, as if they came from
-        node `as_node`. If `as_node` is not provided, it will be set to the last node
-        that updated the state, if not ambiguous.
-        """
+        """Apply updates to the graph state in bulk. Requires a checkpointer to be set."""
```
vbarda, line 1160: could we document the API (all of the params) - i think it's useful to know what the `updates` are just from reading the method docstring

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1420,23 +1467,28 @@ def update_state(
                     **checkpoint_metadata,
                     "source": "update",
                     "step": step + 1,
-                    "writes": {as_node: values},
+                    "writes": {k: v for k, v in valid_updates},
```
vbarda, line 1470: could we use better names for unpacking here
```suggestion
                    "writes": {as_node: values for as_node, values in valid_updates},
```

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -1339,79 +1356,109 @@ def update_state(
...
-            # channel writes are saved to current checkpoint
-            # push writes are saved to next checkpoint
-            channel_writes, push_writes = (
-                [w for w in task.writes if w[0] != PUSH],
-                [w for w in task.writes if w[0] == PUSH],
-            )
-            if saved and channel_writes:
-                checkpointer.put_writes(checkpoint_config, channel_writes, task_id)
+            for task in run_tasks:
+                channel_writes = [w for w in task.writes if w[0] != PUSH]
```
vbarda, line 1449: can we collect all channel writes and push writes in a single loop here? and avoid another loop below?

vbarda (commented):
lgtm overall!

libs/langgraph/langgraph/types.py
dqbd: Renamed to `StateUpdate`, leaving TypedDict vs NamedTuple up for grabs.

libs/langgraph/langgraph/pregel/__init__.py
dqbd: Unless I'm mistaken (cc @nfcampos) we need to put push writes after creating a new checkpoint? So we'd iterate anyway.

Sourav-Nandy-ai:
Hi David,

I ran a real existence-and-functional check on this rather than just reading the changelog. I installed langgraph==0.3.16 (the release right before your merge commit) and 0.3.17 (confirmed via the GitHub API comparison to contain it) in two isolated venvs.

The Results
- Pre-fix: hasattr(graph, 'bulk_update_state') returns False, and trying the call raises a real ImportError. Even StateUpdate does not exist yet.
- Post-fix: The method exists and the call succeeds. I confirmed via get_state() afterward that the patched value genuinely landed in the checkpointer instead of just being a silent no-op. This verified real state mutation end-to-end.

I build agent-eval (agent_regress), an OSS regression harness. The langgraph_runner(operation=...) function (and now the async variant) routes through bulk_update_state directly so this class of checkpoint-recreation change is testable through the documented integration.

Given you wrote this method, I am curious whether a get_state()-based verification like this matches how you would want checkpoint-recreation correctness checked, or if there is a subtler edge (such as partial supersteps or concurrent writers) worth testing next.

You can find the repository here: https://github.com/RudrenduPaul/agent-eval

Best,
Sourav

PR #3766, opened 2025-03-10
docs: Fix version badge by linking it to PyPi instead of shield

yasonk:
Currently the version badge showing langgraph version as PyPi shield image is linking to the shield image. It would be more intuitive to link it to PyPi.


vbarda:
Thank you!

vbarda (approved):

PR #3774, opened 2025-03-11
feat(sdk-js): add docs for generative UI

dqbd:
- **Add docs**
- **Update docs**
- **Add missing links**


docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,255 @@
...
+    Currently only LangGraph.js supports Generative UI. Support for Python is coming soon.
+
+## Quickstart
+
+### 1. Define and configure UI components
+
+First, create your first UI component. For each component you need to provide an unique identifier that will be used to reference the component in your graph code.
+
+```tsx
+// src/agent/ui.tsx
```
bracesproul, line 23: do we want to suggest always defining the structure as:
```
component-name
-- index.css
-- index.tsx
```

If not, we should have a callout/link to/collapsable box with docs on how to ensure your components have access to the same default styling via tailwind if you're using, for example shadcn (adding the tailwind styles to `index.css` and `import "./index.css"` inside the `.tsx` file

docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,255 @@
...
+/>;
+```
+
+### Access the thread state from the UI component
+
+You can access the thread state from the UI component by using the `useStreamContext` hook.
+
+```tsx
+const WeatherComponent = (props: { city: string }) => {
+  const { thread, submit } = useStreamContext();
```
bracesproul, line 176: nit, include `useStreamContext` import statement 
```suggestion
```tsx
import { useStreamContext } from "@langchain/langgraph-sdk/react-ui";

const WeatherComponent = (props: { city: string }) => {
  const { thread, submit } = useStreamContext();
```

bracesproul (commented):
looks good with a few minor nit comments

docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,255 @@
...
+  "graphs": {
+    "agent": "./src/agent/index.ts:graph"
+  },
+  "ui": {
+    "agent": "./src/agent/ui.tsx"
+  }
+}
+```
+
+The `ui` section points to the UI components that will be used by the respective graph. LangGraph Platform supports React 18.x and above, as well as Tailwind 4.x. Dependencies will be bundled and served from the LangGraph Platform.
```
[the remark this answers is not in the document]
dqbd: The key inside `ui` doesn't have to match the agent key, but currently the default is the assistant ID / graph ID. Will add a method to configure the namespace for UI.

docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,256 @@
+# How to implement Generative User Interfaces with LangGraph
+
+!!! info "Prerequisites"
+
+    - [LangGraph Platform](../../concepts/langgraph_platform.md)
+    - [LangGraph Server](../../concepts/langgraph_server.md)
+    - [`useStream()` React Hook](./use_stream_react.md)
+
+Generative user interfaces (Generative UI) allows agents to go beyond text and generate rich user interfaces. This enables creating more interactive and context-aware applications where the UI adapts based on the conversation flow and AI responses.
```
jacoblee93, line 9: Should we add a screenshot/diagram here?

docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,256 @@
...
+  "graphs": {
+    "agent": "./src/agent/index.ts:graph"
+  },
+  "ui": {
+    "agent": "./src/agent/ui.tsx"
+  }
+}
+```
+
+The `ui` section points to the UI components that will be used by the respective graph. LangGraph Platform supports React 18.x and above, as well as Tailwind 4.x. Dependencies will be bundled and served from the LangGraph Platform.
```
jacoblee93, line 47: What about other deps?

Just install as normal?

docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,256 @@
...
+
+### 2. Send the UI components in your graph
+
+Use the `typedUi` utility to emit UI elements from your agent nodes:
+
+```typescript
+// src/agent/index.ts
+import type ComponentMap from "./ui.js";
```
jacoblee93, line 55: nit: put below non-relative paths

docs/docs/cloud/how-tos/generative_ui_react.md
```diff
@@ -0,0 +1,256 @@
...
+
+export const graph = new StateGraph(AgentState)
+  .addNode("weather", async (state, config) => {
+    // Provide the type of the component map to ensure
+    // type safety of `ui.push()` calls.
+    const ui = typedUi<typeof ComponentMap>(config);
+
+    const weather = await new ChatOpenAI({ model: "gpt-4o-mini" })
+      .withStructuredOutput(z.object({ city: z.string() }))
+      .withConfig({ tags: ["langsmith:nostream"] })
```
jacoblee93, line 85: nit: just use `nostream`? Don't think you need `langsmith:nostream` and don't want people to get the wrong idea

jacoblee93 (commented):
Super slick

However I think this reads as more of a tutorial than a how-to

We should have a tutorial and then some how-to guides on the specific configuration topics below

PR #3778, opened 2025-03-11
Update dynamic_barrier_value.py

NinoRisteski:
i think it clearer


eyurtsev (changes requested):
should be re-written more substantially for this to be worth a PR?

PR #3791, opened 2025-03-12
Update recursion-limit.ipynb

El-Invierno:
There was a typo of returning 'a' when the fn def clearly states that: def route(state: State) -> Literal["b", END]:


vbarda:
@El-Invierno i think the typo is actually in the return annotation

PR #3801, opened 2025-03-12
docs: fix state type in Persistence documentation

Mathieuka:
This PR fixes the type of `foo` in the `State` class in Persistence documentation. The type was previously defined as `int`, but the code uses it as a `str`.

Updated the type of `foo` in the documentation to `str` to match its actual usage in the code.

No changes to the functionality or codebase, only a documentation fix.


vbarda (approved):
thanks!

PR #3806, opened 2025-03-12
Coerce nested pydantic

hinthornw:


libs/langgraph/tests/test_pregel.py
```diff
@@ -3027,6 +3031,97 @@ def decider(data: State) -> str:
...
+    class NestedModel(BaseModel):
+        value: int
+        name: str
+
+    class State(BaseModel):
+        top_level: str
+        nested: NestedModel
+        optional_nested: Optional[NestedModel] = None
+        dict_nested: dict[str, NestedModel]
+        list_nested: Annotated[
```
nfcampos, line 3053: oh is this one expected to work?

hinthornw: as a user i'd expect it to i think

libs/langgraph/langgraph/graph/state.py
```diff
@@ -940,23 +943,156 @@ def _pick_mapper(
...
+        if issubclass(schema, dict):
+            return None
+        if issubclass(schema, (BaseModel, BaseModelV1)):
+            return _SchemaCoercionMapper(schema)
     return partial(_coerce_state, schema)
 
 
-def _coerce_state_pydantic(schema: Type[Any], input: dict[str, Any]) -> dict[str, Any]:
-    return schema.model_construct(**input)
+class _SchemaCoercionMapper:
```
nfcampos, line 954: should we move this to diff file?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -940,23 +943,156 @@ def _pick_mapper(
 ) -> Optional[Callable[[Any], Any]]:
     if state_keys == ["__root__"]:
         return None
-    if issubclass(schema, dict):
-        return None
-    if issubclass(schema, BaseModel):
-        return partial(_coerce_state_pydantic, schema)
-    if issubclass(schema, BaseModelV1):
-        return partial(_coerce_state_pydantic_v1, schema)
+    if isclass(schema):
```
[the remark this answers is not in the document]
hinthornw: Don't want to get an error in issubclass, but it _is_ a bit redundant because in python 3.9 it still will pass this/instance check of type but then fail if it's a generic or a few other types that aren't classes

libs/langgraph/langgraph/graph/state.py
hinthornw: Ya can do

PR #3807, opened 2025-03-12
Coerce nested pydantic models up to some depth

nfcampos:


libs/langgraph/langgraph/graph/state.py
```diff
@@ -949,13 +949,43 @@ def _pick_mapper(
...
-def _coerce_state_pydantic(schema: Type[Any], input: dict[str, Any]) -> dict[str, Any]:
+def _coerce_state_pydantic(
+    schema: Type[BaseModel], input: dict[str, Any], *, __depth__: int = 3
+) -> dict[str, Any]:
+    if __depth__ > 0:
+        for key in input:
+            field = schema.model_fields.get(key)
+            if field is None:
+                continue
+            if (
```
hinthornw, line 960: what about list[BaseModel] or the like?

hinthornw: Probably better than mine but:
 https://github.com/langchain-ai/langgraph/pull/3806/files

PR #3812, opened 2025-03-12
Make default recursion_limit configurable by env var

nfcampos:


hinthornw (approved):
We use the full LANGGRAPH_ prefix elsewhere - not sure if we already use LG also

PR #3839, opened 2025-03-14
Update handling of updates/inputs passed in as pydantic models

nfcampos:
- remove usage of require_at_least_one_of, we shouldn't be enforcing presence of keys in inputs/updates, an empty dict is a valid input/update
- ensure that values that were explcitly set/assigned in pydantic model are saved even if equal to default value

### DIFF
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -766,7 +766,7 @@ class CompiledStateGraph(CompiledGraph):
                 # Pydantic v2
                 if isinstance(input, BaseModel):
                     keep: Optional[set[str]] = input.model_fields_set
-                    defaults = {k: v.default for k, v in t.model_fields.items()}
+                    defaults = {k: v.default for k, v in input.model_fields.items()}
                 # Pydantic v1
                 elif isinstance(input, BaseModelV1):
                     keep = input.__fields_set__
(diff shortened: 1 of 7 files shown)


libs/langgraph/tests/test_pregel.py
```diff
@@ -2625,6 +2625,9 @@ class StateUpdate(BaseModel):
         answer: Optional[str] = None
         docs: Optional[list[str]] = None
 
+    class UpdateDocs34(BaseModel):
+        docs: list[str] = ["doc3", "doc4"]
```
hinthornw, line 2629: Should we add another field and set it to "None" explicitly to avoid regressions on that or is that captured well by snapshot tests?

libs/langgraph/tests/test_pregel.py
```diff
@@ -2808,7 +2814,7 @@ def retriever_one(data: State) -> State:
 
     def retriever_two(data: State) -> State:
         time.sleep(0.1)
-        return {"docs": ["doc3", "doc4"]}
+        return UpdateDocs34()
```
hinthornw, line 2817: should we actually assert that the result is doc3 and doc4 after this?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -762,23 +762,32 @@ def _get_updates(
...
-                if hasattr(input, "model_fields"):
+                if isinstance(input, BaseModel):
+                    keep: Optional[set[str]] = input.model_fields_set
                     defaults = {k: v.default for k, v in input.model_fields.items()}
                 # Pydantic v1
-                elif hasattr(input, "__fields__"):
-                    defaults = {k: v.default for k, v in input.__fields__.items()}
+                elif isinstance(input, BaseModelV1):
+                    keep = input.__fields_set__
+                    defaults = {k: v.default for k, v in t.__fields__.items()}
```
hinthornw, line 773: can `v.default` be MISSING ? Should we exclude that?

hinthornw (approved):

PR #3852, opened 2025-03-14
Add optional encryption of checkpointer payloads

nfcampos:
- no dependency on any particular encryption lib (there is no py stdlib encryption lib)
- works with any modern checkpointer, ie. those which use dumps_typed and loads_typed methods to serialize data
- uses the default msg pack serializer, but also works with any custom serializer
- backwards compatible with unencrypted data in same storage (will just be read unencrypted)
- providing easy constructor to use AES encryption through pycriptodome library, one single line of code to add it in
- other encryption libraries or algorithms (even assymetric ones) can be used by implementing the two-method CipherProtocol interface
- cipher name (eg. aes) is stored with encrypted payload for forwards compatibility

```py
import sqlite3

from langgraph.checkpoint.serde.encrypted import EncryptedSerializer

# will read AES key from env var LANGGRAPH_AES_KEY
serde = EncryptedSerializer.from_pycryptodome_aes()
# works with any other checkpointer, including custom ones
checkpointer = SqliteSaver(sqlite3.connect('...'), serde=serde)
```


hinthornw:
wow you made the change i was queuing to submit

nfcampos:
I read minds

libs/checkpoint/langgraph/checkpoint/serde/encrypted.py
```diff
@@ -0,0 +1,90 @@
...
+        except ImportError:
+            raise ImportError(
+                "Pycryptodome is not installed. Please install it with `pip install pycryptodome`."
+            ) from None
+
+        # check if AES key is provided
+        if "key" in kwargs:
+            key = kwargs.pop("key")
+        else:
+            key = os.getenvb(b"LANGGRAPH_AES_KEY")
```
hinthornw, line 64: Seems like this is not necessarily available on windows?

libs/checkpoint/langgraph/checkpoint/serde/encrypted.py
```diff
@@ -0,0 +1,90 @@
...
+        return f"{typ}+{self.cipher.name}", encrypted_data
+
+    def loads_typed(self, data: tuple[str, bytes]) -> Any:
+        enc_typ, enc_data = data
+        # unencrypted data
+        if "+" not in enc_typ:
+            return self.serde.loads_typed(data)
+        # verify cipher name
+        typ, cipher_name = enc_typ.split("+", 1)
+        if cipher_name != self.cipher.name:
```
hinthornw, line 39: Is there a world where we should let you provide fallback ciphers for decrypting?

hinthornw (approved):

libs/checkpoint/langgraph/checkpoint/serde/encrypted.py
```diff
@@ -0,0 +1,90 @@
...
+            if key is None:
+                raise ValueError("LANGGRAPH_AES_KEY environment variable is not set.")
+            if len(key) not in (16, 24, 32):
+                raise ValueError("LANGGRAPH_AES_KEY must be 16, 24, or 32 bytes long.")
+
+        # set default mode to EAX if not provided
+        if kwargs.get("mode") is None:
+            kwargs["mode"] = AES.MODE_EAX
+
+        class PycryptodomeAesCipher(CipherProtocol):
```
akira, line 74: not sure if makes sense to split this out from inner class if want to reuse this

akira (approved):

libs/checkpoint/langgraph/checkpoint/serde/encrypted.py
nfcampos: Will make the conditional imports a little messier, so I think will leave like this

RisaKirisu:
Wow, this came in just as I'm thinking of ways to enable encryption in our prod application. Thank you!

Is it possible to use different encryption keys based on config? i.e. different keys for different `thread_id`

PR #3857, opened 2025-03-15
Improve documentation string for SqliteSaver

ArrayPD:
* Document that `check_same_thread` as an option when creating sqlite connection.
* Document why it's OK to do that.


libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -56,7 +56,7 @@ class SqliteSaver(BaseCheckpointSaver[str]):
         >>> builder.add_node("add_one", lambda x: x + 1)
         >>> builder.set_entry_point("add_one")
         >>> builder.set_finish_point("add_one")
-        >>> conn = sqlite3.connect("checkpoints.sqlite")
+        >>> conn = sqlite3.connect("checkpoints.sqlite", check_same_thread=False)
```
eyurtsev, line 59: Would you mind adding an inline explanation explaining why this is OK to do in the context of langggraph. This assumes that writes are sequential and are not in parallel

eyurtsev (approved):

PR #3861, opened 2025-03-15
(wip) prebuilt: add utils for long context management

vbarda:


libs/prebuilt/langgraph/prebuilt/memory_handler.py
```diff
@@ -0,0 +1,307 @@
...
+from langgraph.store.base import BaseStore
+
+SUMMARIES_NS = ("summaries",)
+
+TokenCounter = Callable[[list[BaseMessage]], int]
+
+
+# TODO: does this need to be a Runnable?
+class BaseMemoryHandler(Runnable):
+    def invoke(  # type: ignore
```
eyurtsev, line 20: If we do `Runnables:

1) We'll want an async variant.
2) The interface for implementation should probably not be `invoke` but something like` _invoke` or `_process` so that `invoke` is reserved for instrumentation purposes


The reason to make this or not make this into a Runnable is if we want dedicated tracing for this in langsmith. But it's not clear to me that we do

libs/prebuilt/langgraph/prebuilt/memory_handler.py
```diff
@@ -0,0 +1,307 @@
...
+        self,
+        messages: list[BaseMessage],
+        config: RunnableConfig,
```
eyurtsev, line 23:
```suggestion
        config: RunnableConfig | None,
```

libs/prebuilt/langgraph/prebuilt/memory_handler.py
```diff
@@ -0,0 +1,307 @@
...
+
+from langgraph.store.base import BaseStore
+
+SUMMARIES_NS = ("summaries",)
+
+TokenCounter = Callable[[list[BaseMessage]], int]
+
+
+# TODO: does this need to be a Runnable?
+class BaseMemoryHandler(Runnable):
```
eyurtsev, line 19: Packaging in a class may make it a bit easier to plug in an async variant?

vbarda:
Closing since will be moved to a different lib

PR #3889, opened 2025-03-18
When using global resume value, ensure subgraphs consume it

nfcampos:
- Previously the global resume value was passed to subgraphs without being consumed
- This would result in two parallel subgraph calls being able to use the same resume value
- Note this behavior can't be implemented over the wire, that will be fixed in future PR

Closes #3398

### DIFF
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -130,7 +130,7 @@ class Interrupt:
     value: Any
     resumable: bool = False
     ns: Optional[Sequence[str]] = None
-    when: Literal["during"] = "during"
+    when: Literal["during"] = dataclasses.field(default="during", repr=False)
 
 
 class PregelTask(NamedTuple):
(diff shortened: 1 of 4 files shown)


libs/langgraph/langgraph/types.py
```diff
@@ -130,7 +130,7 @@ class Interrupt:
     value: Any
     resumable: bool = False
     ns: Optional[Sequence[str]] = None
-    when: Literal["during"] = "during"
+    when: Literal["during"] = dataclasses.field(default="during", repr=False)
```
hinthornw, line 133: ooc: Why are we hiding from the repr?

hinthornw (approved):

libs/langgraph/langgraph/types.py
nfcampos: Because this field shouldn't exist, it's a constant, I just can't remove it bc backwards compat

PR #3916, opened 2025-03-19
langgraph: incorporate information about previously updated channels to identify which tasks to execute next

eyurtsev:
Leverage information about which channels were updated in the previous step to determine which tasks should be triggered. This can result in significant speed up in prepare_next_tasks in some situations.


libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -480,6 +484,15 @@ def tick(
             self.status = "out_of_steps"
             return False
 
+        # If updated channels is available, we project only a subset of the nodes.
+        # since only those nodes have been updated?
+        triggered_nodes: set[str] = set()
```
nfcampos, line 489: I think prepare_next_tasks should receive updated_nodes, and do this work inside?

eyurtsev:
i love to lint

nfcampos (approved):

PR #3923, opened 2025-03-19
feat(sdk): add bulk_update_state in SDK

dqbd:


libs/sdk-js/src/client.ts
```diff
@@ -482,15 +482,47 @@ export class ThreadsClient<
...
      */
     metadata?: Metadata;
+    /**
+     * ID of the thread to create.
+     *
+     * If not provided, a random UUID will be generated.
+     */
     threadId?: string;
+    /**
+     * How to handle dplicate creation.
```
hinthornw, line 492: nit

PR #3926, opened 2025-03-19
DOCS: Update state-model.ipynb to use "AnyMessage"

hesreallyhim:
## Description
The documentation for working with Pydantic and graph State recommends to use `AnyMessage` when working with LangChain types, but the code example uses `BaseMessage`.


vbarda (approved):

hesreallyhim:
@vbarda is this meant to be a general recommendation? i notice some other places in the documentation where `BaseMessage` is used where perhaps `AnyMessage` is preferred(?) - example: [Tutorial - Agentic RAG](https://langchain-ai.github.io/langgraph/tutorials/rag/langgraph_agentic_rag/#agent-state)

Is this too trivial of a distinction to merit a fix? I'm happy to submit a fix for the dozen or so docs (quick estimate from glance with VS-Code find-all) that break this convention.

maver1ck:
I found this PR and I think there is general problem with having both BaseMessage and AnyMessage.
For example:
Langchain BaseChatModel is using BaseMessage
https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/language_models/chat_models.py#L165

Langgraph MessagesState is using AnyMessage
https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/graph/message.py#L274

The problem is that those definitions are not compatible with each other which results in multiple Mypy errors :(

hesreallyhim:
@maver1ck can you give an example (like a LOC that shows a MyPy error)? I'm still trying to grok the distinction, I know BaseMessage is an ABC, and AnyMessage is a union of subclasses (AIMessage, HumanMessage, etc.), so it has something to do with ABC's not being serializable, and maybe whether there are any methods that are defined in the BaseMessage but not declared in the AnyMessage(?) but yeah would love to know more.

maver1ck:
@hesreallyhim 
Yes.
This is original code (not mine but I'm trying to use this template in my project)
https://github.com/JoshuaC215/agent-service-toolkit/blob/main/src/agents/bg_task_agent/bg_task_agent.py

Original code is giving following mypy error:
```
src/agents/bg_task_agent/bg_task_agent.py:26: error: Incompatible return value type (got "RunnableSerializable[AgentState, BaseMessage]", expected "RunnableSerializable[AgentState, AIMessage]")  [return-value]
```
So naturally I'd like to change line 21 to:
```
def wrap_model(model: BaseChatModel) -> RunnableSerializable[AgentState, BaseMessage]:
```

Then I'm getting this which is impossible to fix when using Langchain / Langgraph built-in mechanisms. 
```
src/agents/bg_task_agent/bg_task_agent.py:35: error: List item 0 has incompatible type "BaseMessage"; expected "AIMessage | HumanMessage | ChatMessage | SystemMessage | FunctionMessage | ToolMessage"  [list-item]
```

hesreallyhim:
@maver1ck sorry I've only skimmed the code so far, and I'd still love someone with deeper context to chime in, but have you tried using `AnyMessage` at the end of 21 instead of `BaseMessage`? I'm not sure why the types are "incompatible" but `AnyMessage` I think is literally an alias for `"AIMessage | HumanMessage | ChatMessage | SystemMessage | FunctionMessage | ToolMessage"`

maver1ck:
@hesreallyhim 
Not working
```
src/agents/bg_task_agent/bg_task_agent.py:27: error: Incompatible return value type (got "RunnableSerializable[AgentState, BaseMessage]", expected "RunnableSerializable[AgentState, AIMessage | HumanMessage | ChatMessage | SystemMessage | FunctionMessage | <7 more items>]")  [return-value]
```

Probably because we're using BaseChatModel from Langchain which is defined using BaseMessage
```
class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
```

PR #3929, opened 2025-03-19
GC test all

hinthornw:


hinthornw:
TODO: get this working with gc enable/disable

PR #3931, opened 2025-03-20
Use fast path for prepare_next_tasks on input

nfcampos:
- When there are no values in checkpoint no need to run through all the PULL candidates
- When there are input writes save updated_channels to use on the next call to prepare_next_tasks

### DIFF
--- a/libs/langgraph/langgraph/pregel/algo.py
+++ b/libs/langgraph/langgraph/pregel/algo.py
@@ -347,7 +347,7 @@ def prepare_next_tasks(
     store: Literal[None] = None,
     checkpointer: Literal[None] = None,
     manager: Literal[None] = None,
-    trigger_to_nodes: Optional[Mapping[str, list[str]]] = None,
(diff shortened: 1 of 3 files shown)


libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -452,6 +452,8 @@ def prepare_next_tasks(
                 triggered_nodes.update(node_ids)
         # Sort the nodes to ensure deterministic order
         candidate_nodes: Iterable[str] = sorted(triggered_nodes)
+    elif not checkpoint["channel_versions"]:
```
eyurtsev, line 455: an inline doc would be helpful here -- it's not obvious what this means semantically -- requires knowledge of how / when channel_versions in a checkpoint are set

I can update after the fact fwiw

eyurtsev (approved):

eyurtsev:
Can look in the AM at failing tests

PR #3953, opened 2025-03-20
Switch serialization lib to ormsgpack

nfcampos:
- faster for large states
- supports tuples as dict keys


libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -192,8 +192,10 @@ def dumps_typed(self, obj: Any) -> tuple[str, bytes]:
         else:
             try:
                 return "msgpack", _msgpack_enc(obj)
-            except UnicodeEncodeError:
-                return "json", self.dumps(obj)
+            except ormsgpack.MsgpackEncodeError as exc:
+                if "valid UTF-8" in str(exc):
```
hinthornw, line 196: oof

hinthornw: Maybe something we can upstream as a specific exception type

hinthornw (commented):
How did you get the enum support working?

PR #3958, opened 2025-03-20
Bugfix: proper link to multi-agent-collaboration.ipynb

nikita-rvachev:
There is a mistake in agent_supervisor.ipynb.
Current "previous example" link leads to 404 - page not found
This is fix for that.


nikita-rvachev:
@eyurtsev please let me know if anything needed to PR to be approved.

PR #3976, opened 2025-03-21
prebuilt: pass last message to structured response model in create_react_agent

vbarda:


hinthornw (approved):
Thanks for checking compatibility with Google/oai/anthropic

PR #3988, opened 2025-03-23
docs: add version admonitions for `Interrupt` and `RetryPolicy`

YassinNouh21:
This pull request includes changes to add version admonitions to the documentation and update the styling for these admonitions. The most important changes include the addition of version information to the documentation, updates to the CSS for version admonitions, and modifications to the `mkdocs.yml` configuration file to include the new stylesheets.
this should solve this #3991


docs/cassettes/test_versions.md
```diff
@@ -0,0 +1,44 @@
+# Testing Version Admonitions
```
eyurtsev, line 1: Let's remove this file.

Have you tried serving docs locally to check if ti works for the API reference?

`make serve-docs`

YassinNouh21: I've removed the test_versions.md file and updated the mkdocs.yml configuration to remove its reference. The documentation should now build without the test versions page.

I've also tried serving the docs locally with `make serve-docs` to verify the changes. The documentation server is running and you can check the API reference at http://localhost:8000.

libs/langgraph/langgraph/types.py
```diff
@@ -101,7 +101,10 @@ def default_retry_on(exc: Exception) -> bool:
 
 
 class RetryPolicy(NamedTuple):
-    """Configuration for retrying nodes."""
+    """
+    !!! version-added "Added in v0.2.24"
```
eyurtsev, line 105:
```suggestion
    !!! version-added "0.2.24"
```

libs/langgraph/langgraph/types.py
```diff
@@ -101,7 +101,10 @@ def default_retry_on(exc: Exception) -> bool:
 
 
 class RetryPolicy(NamedTuple):
-    """Configuration for retrying nodes."""
+    """
```
eyurtsev, line 104: Let's restore the object descriptions.

```suggestion
    """Configuration for retrying nodes.
```

libs/langgraph/langgraph/types.py
```diff
@@ -101,7 +101,10 @@ def default_retry_on(exc: Exception) -> bool:
...
+    !!! version-added "Added in v0.2.24"
+        The RetryPolicy class was introduced in version 0.2.24 (September 2024).
```
eyurtsev, line 106: This doesn't provide any additional information. We just want a minimal version added.

```suggestion
```

---

A description could be useful if we start using ..version-modified

eyurtsev:
Cool it looks like this is I make it minimal.

Is there a way to add a title saying "Added in version {version}." by default?

Here's how it looks like:

image


Here's a default from rst / sphinx:

image

libs/langgraph/langgraph/types.py
```diff
@@ -120,13 +123,18 @@ class RetryPolicy(NamedTuple):
 
 
 class CachePolicy(NamedTuple):
-    """Configuration for caching nodes."""
+    """
```
eyurtsev, line 126:
```suggestion
    """Configuration for caching nodes.
```

libs/langgraph/langgraph/types.py
```diff
@@ -101,7 +101,10 @@ def default_retry_on(exc: Exception) -> bool:
 
 
 class RetryPolicy(NamedTuple):
-    """Configuration for retrying nodes."""
+    """Configuration for retrying nodes.
+    
+    !!! version-added "Added in version 0.2.24."
```
eyurtsev, line 106: Is there anyway to have "Added in version" be prepended automatically, so that we only put the version?

eyurtsev:
@YassinNouh21 I'm going to merge this change. If you can figure out how to make the prepending automatic, please follow up with a PR for that! Would be appreciated. 

We want the syntax to just be ` !!! version-added "0.2.24" inside the code, and anything else to be done by mkdocs

eyurtsev (approved):

YassinNouh21:
@eyurtsev I solved the issue in the ci for python linter 3.12 
now I think it is ready for merge

YassinNouh21:
@eyurtsev it is ok can u merge

libs/langgraph/langgraph/types.py
eyurtsev: Could you restore the title? we shouldn't delete the doc-strings for types when adding versions

libs/langgraph/langgraph/types.py
```diff
@@ -417,7 +428,7 @@ def node(state: State):
                 # as part of the interrupt information.
                 \"what is your age?\"
             )
-            print(f\"> Received an input from the interrupt: {answer}\")
+            print(f"> Received an input from the interrupt: {answer}\")
```
eyurtsev, line 431: I believe the `\` is necessary in this case

```suggestion
            print(f"> Received an input from the interrupt: {answer}\")
```

YassinNouh21: this got me an issue at the ci but I pushed the old version anyways

YassinNouh21:
Hey @eyurtsev ,

I've found configuration in the `mkdocs.yml` that allows us to automatically prepend "Added in version" to the version admonitions. Specifically, we've set up the `markdown_version_annotations` plugin with this configuration:

```yaml
- markdown_version_annotations:
    version_added_title: "Added in version \\1"
```

This means when we use `+++ 0.2.24` or `!!! version-added "0.2.24"`, it will automatically generate an admonition that says "Added in version 0.2.24".

Let me know what you think!

eyurtsev (approved):

eyurtsev:
@YassinNouh21 good find! I think it's a good solution, but I would like to vendor it in rather than rely on an external dependency. 

It's under an MIT license: https://github.com/glennmatthews/markdown-version-annotations/blob/main/pyproject.toml#L6

So we can copy paste the code with appropriate attribution and hook it up. If you're interested in doing that, you can create a vendored version under _scripts and figure out how to hook it up (could potentially do it via the hook system rather than as a plugin)

PR #4000, opened 2025-03-24
sdk: allow specifying run time headers in http client

eyurtsev:
This PR only allows this in the HTTP client. 

I can follow up with a PR to allow throughout the entire API.

The use case is to allow instantiating the client once (w/ a single connection pool), but allowing changing api keys and any other headers at run time


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -222,13 +228,23 @@ async def get(self, path: str, *, params: Optional[QueryParamTypes] = None) -> A
...
         """Send a POST request."""
         if json is not None:
-            headers, content = await aencode_json(json)
+            content_headers, content = await aencode_json(json)
         else:
-            headers, content = {}, b""
-        r = await self.client.post(path, headers=headers, content=content)
+            content_headers, content = {}, b""
+        # Merge headers, with runtime headers taking precedence
+        request_headers = {**content_headers}
```
hinthornw, line 244: ooc: why do we need to spread here?

eyurtsev: we don't -- i'll fix up

PR #4023, opened 2025-03-25
docs(pregel): One-line markdown formatting quick-fix

hesreallyhim:
## Description

I noticed a very minor issue in the formatting of the Concepts > Pregel doc: (https://langchain-ai.github.io/langgraph/concepts/pregel/#high-level-api) (https://github.com/langchain-ai/langgraph/blob/main/docs/docs/concepts/pregel.md)

You can see in the image below that there is a python codeblock, then a pycon block, and inside that block there is an extra triple-backtick/code fence, and then text at the bottom, which it appears like it is supposed to be a separate python block, like the one above it. I.e., clearly:

```
```python
print(graph.channels)
```

is intended to be:

```python
print(graph.channels)
```

I'm pretty sure this is due to an extra whitespace character before the preceding closing code fence, which is throwing off the formatting.

image

My VS-Code/extensions can't really render the Markdown the way it appears on the Website, I think because of the tabs (Graph API vs. Functional API), but I noticed that if I remove the extra whitespace, the highlighting on the python codeblock is corrected:

BEFORE:
Screenshot 2025-03-25 at 5 02 41 PM

AFTER:
Screenshot 2025-03-25 at 5 02 50 PM

## Fix
* Remove one whitespace character that was throwing off markdown rendering

### DIFF
--- a/docs/docs/concepts/pregel.md
+++ b/docs/docs/concepts/pregel.md
@@ -284,7 +284,7 @@ LangGraph provides two high-level APIs for creating a Pregel application: the [S
     {'__start__': <langgraph.pregel.read.PregelNode at 0x7d05e3ba1810>,
      'write_essay': <langgraph.pregel.read.PregelNode at 0x7d05e3ba14d0>,
      'score_essay': <langgraph.pregel.read.PregelNode at 0x7d05e3ba1710>}
-     ```
+    ```
 
     ```python
     print(graph.channels)


docs/docs/concepts/pregel.md
```diff
@@ -344,4 +344,4 @@ LangGraph provides two high-level APIs for creating a Pregel application: the [S
     {'write_essay': <langgraph.pregel.read.PregelNode object at 0x7d05e2f9aad0>}
     Channels: 
     {'__start__': <langgraph.channels.ephemeral_value.EphemeralValue object at 0x7d05e2c906c0>, '__end__': <langgraph.channels.last_value.LastValue object at 0x7d05e2c90c40>, '__previous__': <langgraph.channels.last_value.LastValue object at 0x7d05e1007280>}
-    ```
\ No newline at end of file
+    ```
```
hesreallyhim, line 347: i actually can't tell what the diff here is :(

eyurtsev: hm let's try it out -- some spacing difference maybe tab vs. regular space?

eyurtsev (approved):

docs/docs/concepts/pregel.md
hesreallyhim: i made the edit directly in github since it was so tiny, i thought it was a newline issue, but i didn't touch that line as far as i recall, anyway lmk if you want me to run the lint/fix locally, thanks!

PR #4043, opened 2025-03-26
langgraph: allow using RemoteGraph from a graph on the same server

vbarda:


nfcampos:
Closing, this was solved in langgraph-supervisor instead

PR #4044, opened 2025-03-26
Set store & stream writer

hinthornw:


nfcampos (commented):
Doesn't this potentially set things globally?

hinthornw:
Ya it would, which I suppose could create memory leaks.

Context for this was that some people wanted to do `graph.nodes["my_node"].invoke()` with store. Could also just continue recommending:

```
from langgraph.constants import CONF, CONFIG_KEY_STORE

config = {CONF: {CONFIG_KEY_STORE: my_store, thread_id: str(uuid.uuid4())}}
```

PR #4049, opened 2025-03-27
prebuilt: allow prompt to return state updates in create_react_agent

vbarda:


vbarda:
Closing in favor of #4059

PR #4055, opened 2025-03-27
In Python 3.12 or above, use asyncio eager task factory

nfcampos:
- This is a performance improvement when calling async functions that do not use await, as they are run immediately and never scheduled in the loop


libs/langgraph/langgraph/utils/future.py
```diff
@@ -180,25 +186,32 @@ def _wrap_awaitable(awaitable: Awaitable[T]) -> Generator[None, None, T]:
...
-            )
-        except (SystemExit, KeyboardInterrupt):
-            raise
-        except BaseException as exc:
-            future.set_exception(exc)
-            raise
-
-    loop.call_soon_threadsafe(callback, context=context)
-    return future
+    if asyncio._get_running_loop() is loop:
```
nfcampos, line 199: note this is function is documented "This is a low-level function intended to be used by event loops", so not private api

PR #4065, opened 2025-03-28
 docs: Add Version 0.2.24 Documentation to for PregelTask, PregelExecutableTask, StateSnapshot, Send

YassinNouh21:
# Add Version 0.2.24 Documentation to Types

## Changes
Added version documentation to classes and functions introduced in version 0.2.24:
- `PregelTask` class
- `PregelExecutableTask` class
- `StateSnapshot` class
- `Send` class

## Why
Improves code maintainability by clearly indicating when these types were introduced in the codebase.

### DIFF
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -149,6 +149,9 @@ class StateUpdate(NamedTuple):
 
 
 class PregelTask(NamedTuple):
+    """
+    !!! version-added "Added in version 0.2.24."
+    """
     id: str
     name: str
     path: tuple[Union[str, int, tuple], ...]
@@ -166,6 +169,9 @@ else:
 
 @dataclasses.dataclass(**_T_DC_KWARGS)
 class PregelExecutableTask:
+    """
+    !!! version-added "Added in version 0.2.24."
+    """
     name: str
     input: Any
     proc: Runnable
@@ -182,7 +188,9 @@ class PregelExecutableTask:
 
 
 class StateSnapshot(NamedTuple):
-    """Snapshot of the state of the graph at the beginning of a step."""
+    """Snapshot of the state of the graph at the beginning of a step.
+    !!! version-added "Added in version 0.2.24."
+    """
 
     values: Union[dict[str, Any], Any]
     """Current values of channels"""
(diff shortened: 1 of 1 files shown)


YassinNouh21:
@eyurtsev caan u check this

eyurtsev:
> Improves code maintainability by clearly indicating when these types were introduced in the codebase.

I'd probably say that this is much more useful for our end users than for internal development. End users may need to know sometimes that a feature is not available for their current version

YassinNouh21:
> > Improves code maintainability by clearly indicating when these types were introduced in the codebase.
> 
> I'd probably say that this is much more useful for our end users than for internal development. End users may need to know sometimes that a feature is not available for their current version

Could u suggest some to work on

libs/langgraph/langgraph/types.py
```diff
@@ -149,6 +149,9 @@ class StateUpdate(NamedTuple):
 
 
 class PregelTask(NamedTuple):
+    """
+    !!! version-added "Added in version 0.2.24."
```
eyurtsev, line 153: Looks good, but I think the versions are incorrect.

Try using the published version tags

```
git tag -l
```

```
git checkout 0.1.10
```

```
git grep StateSnapshot
```

You'll already get a hit on state snapshot.

---

You can use the steps above ^ To create a simple script for automatically identifying when a particular type was introduced

YassinNouh21:
@eyurtsev   I checked the history of types.py, and it looks like at some point, the entire file was either moved or significantly changed—including PregelTask, which you mentioned.

I followed your suggested steps and created a release based on that version, but let me know if I misunderstood anything or if there’s a better way to approach it. I just want to make sure the GitHub diff actions correctly reflect when PregelTask was added to the class.

Here are the relevant links for reference:
[GitHub Link 1](https://github.com/langchain-ai/langgraph/blob/d9854b36ebb6690a5106ad73a3099ef3ee386259/libs/langgraph/langgraph/types.py)
[GitHub Link 2](https://github.com/langchain-ai/langgraph/blob/d9854b36ebb6690a5106ad73a3099ef3ee386259/libs/langgraph/langgraph/types.py)

Let me know what you think!

eyurtsev:
Use version added for when the object / type was introduced into langgraph.

version-modified can be used to capture modifications (e.g., changing namespaces for the type etc.)

YassinNouh21:
> Use version-added for when the object / type was introduced into langgraph.
> 
> version-modified can be used to capture modifications (e.g., changing namespaces for the type etc.)

Ok thanks for the clarification. I will remove it

PR #4100, opened 2025-03-31
feat(cli): pass ui and ui config to inmem server, handle Docker setup for UI

dqbd:


libs/cli/langgraph_cli/cli.py
```diff
@@ -658,6 +658,8 @@ def dev(
         wait_for_client=wait_for_client,
         auth=config_json.get("auth"),
         http=config_json.get("http"),
+        ui=config_json.get("ui"),
```
hinthornw, line 661: Without reading into it, i would have expected these two to be a single config

hinthornw (approved):

libs/cli/langgraph_cli/cli.py
hinthornw: I don't think` ui_config` is in the schema either - what's its purpose?

PR #4101, opened 2025-03-31
Lazily create atomic counters in pregel scratchpad

nfcampos:
- many times these aren't actually used, so makes sense to delay creation until needed

### DIFF
--- a/libs/langgraph/langgraph/pregel/algo.py
+++ b/libs/langgraph/langgraph/pregel/algo.py
@@ -1,6 +1,7 @@
 import binascii
 import itertools
 import sys
+import threading
 from collections import defaultdict, deque
 from functools import partial
 from hashlib import sha1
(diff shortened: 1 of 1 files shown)


libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -990,3 +991,22 @@ def task_path_str(tup: Union[str, int, tuple]) -> str:
...
     )
+
+
+LAZY_ATOMIC_COUNTER_LOCK = threading.Lock()
+
+
+class LazyAtomicCounter:
+    __slots__ = ("_counter",)
+
+    _counter: Optional[Callable[[], int]]
```
hinthornw, line 1002: ooc: why not just type this as an instance attribute?

hinthornw (approved):

PR #4107, opened 2025-04-01
fix: update py document for add_messages

warjiang:
update python document

### DIFF
--- a/libs/langgraph/langgraph/graph/message.py
+++ b/libs/langgraph/langgraph/graph/message.py
@@ -78,7 +78,7 @@ def add_messages(
         message from `right` will replace the message from `left`.
 
     Examples:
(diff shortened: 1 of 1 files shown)


warjiang:
@nfcampos cloud you help me review that, just a simple chnage

hinthornw:
pycon means "python console syntax" ([pycon](https://github.com/squidfunk/mkdocs-material/issues/3647))

warjiang:
> pycon means "python console syntax" ([pycon](https://github.com/squidfunk/mkdocs-material/issues/3647))
> 
> 
> 
> But i do think we do this copying bit wrong so we should check our tooling
> 
> 

thanks for pointing that， i got that～

PR #4114, opened 2025-04-01
langgraph: allow overwriting messages in add_messages

vbarda:
This allows nodes to completely replace message history, if needed, by returning

```python
def my_node(state):
    ...
    return {"messages": ("overwrite", brand_new_messages)}
```


vbarda:
Closing in favor of #4117

PR #4117, opened 2025-04-01
langgraph: support removing all messages with RemoveMessage

vbarda:


libs/langgraph/langgraph/graph/message.py
```diff
@@ -179,6 +182,12 @@ def add_messages(
     for m in right:
         if m.id is None:
             m.id = str(uuid.uuid4())
+        if isinstance(m, RemoveMessage) and m.id == REMOVE_ALL_MESSAGES:
```
hinthornw, line 185: maybe silly but shoudl this ALSO clear any messages in `right` preceding the remove all command?

vbarda: yea, i see your point. yea, i can add and if it causes any headache we can relax

hinthornw (approved):

PR #4157, opened 2025-04-03
feat(langgraph): Add UI messages API

dqbd:
Sample usage:

```python
from typing import Annotated, Sequence, TypedDict

from langchain_core.messages import BaseMessage
from langchain_openai import ChatOpenAI

from langgraph.graph.ui import AnyUIMessage, ui_message_reducer, push_ui_message


class AgentState(TypedDict):  # noqa: D101
    messages: Annotated[Sequence[BaseMessage], add_messages]
    ui: Annotated[Sequence[AnyUIMessage], ui_message_reducer]


async def agent(state: AgentState):  # noqa: D103
    message = await ChatOpenAI(model="gpt-4o-mini", temperature=0).ainvoke(
        state["messages"]
    )

    # Also directly writes the message to `ui`
    push_ui_message("simple", {"hello": "123"}, message=message, state_key="ui")

    return {"messages": [message]}

```

### DIFF
--- a/libs/langgraph/langgraph/graph/ui.py
+++ b/libs/langgraph/langgraph/graph/ui.py
@@ -115,7 +115,7 @@ def push_ui_message(
     return evt
 
 
-def remove_ui_message(id: str, *, state_key: str = "ui") -> RemoveUIMessage:
+def delete_ui_message(id: str, *, state_key: str = "ui") -> RemoveUIMessage:
     """Delete a UI message by ID from the UI state.
 
     This function creates and sends a message to remove a UI component from the current state.
@@ -132,7 +132,7 @@ def remove_ui_message(id: str, *, state_key: str = "ui") -> RemoveUIMessage:
 
     .. code-block:: python
 
-        remove_ui_message("message-123")
+        delete_ui_message("message-123")
 
     """
     writer = get_stream_writer()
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/graph/ui.py
```diff
@@ -0,0 +1,206 @@
...
+
+        if (existing_idx := merged_by_id.get(msg_id)) is not None:
+            if msg.get("type") == "remove-ui":
+                ids_to_remove.add(msg_id)
+            else:
+                ids_to_remove.discard(msg_id)
+                merged[existing_idx] = msg
+        else:
+            if msg.get("type") == "remove-ui":
+                raise ValueError(
```
hinthornw, line 198: If I have two parallel nodes that both try to remove the same message, would this raise an error?

Should we just `pass` in that case/

libs/langgraph/langgraph/graph/ui.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+
+    type: Literal["ui"]
+    id: str
+    name: str
+    props: dict[str, Any]
+    metadata: dict[str, Any]
+
+
+class RemoveUIMessage(TypedDict):
```
hinthornw, line 32: Should we just re-use the RemoveMessage here?

PR #4169, opened 2025-04-03
Add checkpoint_during arg

nfcampos:
- This provides a new mode of execution where only the last checkpoint is saved
- We save the last checkpoint no matter how the agent run is terminated (success, error, interrupt, etc)
- This cuts down on cpu time spent on checkpointing, while not losing any resilience benefits, given individual task writes are still saved
- If an error occurs and the run is retried, any tasks that completed successfully before will be skipped (as currently)
- checkpoint_during=True is useful when you want to time-travel to inner steps of a run
- The default value will remain the current behavior, ie. checkpoint_during=True

### DIFF
--- a/libs/langgraph/langgraph/pregel/loop.py
+++ b/libs/langgraph/langgraph/pregel/loop.py
@@ -316,7 +316,7 @@ class PregelLoop(LoopProtocol):
                 self.checkpoint_pending_writes[idx] = (task_id, c, v)
             else:
                 self.checkpoint_pending_writes.append((task_id, c, v))
-        if self.checkpointer_put_writes is not None:
+        if self.checkpoint_during and self.checkpointer_put_writes is not None:
             config = patch_configurable(
                 self.checkpoint_config,
                 {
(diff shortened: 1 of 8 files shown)


libs/langgraph/tests/test_interruption.py
```diff
@@ -78,11 +79,11 @@ async def noop(_state):
         initial_input = {"input": "hello world"}
         thread = {"configurable": {"thread_id": "1"}}
 
-        await graph.ainvoke(initial_input, thread, debug=True)
+        await graph.ainvoke(initial_input, thread, checkpoint_during=checkpoint_during)
         assert (await graph.aget_state(thread)).next == ("step_2",)
 
-        await graph.ainvoke(None, thread, debug=True)
```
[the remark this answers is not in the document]
nfcampos: Because that's just a holdover from someone debugging this test and forgetting to remove it before

libs/langgraph/tests/test_interruption.py
```diff
@@ -78,11 +79,11 @@ async def noop(_state):
...
+        await graph.ainvoke(None, thread, checkpoint_during=checkpoint_during)
         assert (await graph.aget_state(thread)).next == ("step_3",)
 
-        await graph.ainvoke(None, thread, debug=True)
+        await graph.ainvoke(None, thread, checkpoint_during=checkpoint_during)
```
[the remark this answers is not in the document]
nfcampos: Oh yes there's lots of tests missing. But there will be as many checkpoints as there are calls to invoke

libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -294,29 +296,17 @@ def put_writes(self, task_id: str, writes: Sequence[tuple[str, Any]]) -> None:
         """Put writes for a task, to be read by the next tick."""
         if not writes:
             return
+        checkpoint_during = self.checkpoint_during or any(w[0] == TASKS for w in writes)
```
hinthornw, line 299: Could we have a comment *why* we're checkpointing if there are tasks? I'm assuming this is to handle the fact that tasks are only saved to the writes() in the parent checkpoint?

libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -294,29 +296,17 @@ def put_writes(self, task_id: str, writes: Sequence[tuple[str, Any]]) -> None:
...
         # deduplicate writes to special channels, last write wins
         if all(w[0] in WRITES_IDX_MAP for w in writes):
             writes = list({w[0]: w for w in writes}.values())
+        # remove existing writes for this task
+        self.checkpoint_pending_writes = [
```
hinthornw, line 304: Dumb q but this is diffrent than before in the case where there's multiple writes for a single WRITES_IDX_MAP channel - did we add an invarian where we know it won't be duplicated here?

libs/langgraph/langgraph/pregel/loop.py
```diff
@@ -747,6 +789,7 @@ def _put_checkpoint(self, metadata: CheckpointMetadata) -> None:
                 **self.checkpoint_config,
                 CONF: {
                     **self.checkpoint_config[CONF],
+                    CONFIG_KEY_CHECKPOINT_ID: self.checkpoint_id_prev,
```
hinthornw, line 792: Wasn't obvious to me initially that we could guarantee that self.checkpoint_id_prev had been set by the point this branch was run

hinthornw (commented):
Do we have a good test that's run for `checkpoint_during=False` for the folllowing?

- for resuming from an interrupt & resume within a subgraph? 
- Or that where you Send to node that is interrupted?

libs/langgraph/langgraph/pregel/loop.py
nfcampos: I think that's covered by line 302 above

libs/langgraph/tests/test_pregel_async.py
```diff
@@ -5348,6 +5433,131 @@ async def side(state: State):
     assert times_called == 1
 
 
+@pytest.mark.parametrize("checkpoint_during", [True, False])
```
nfcampos, line 5436: these tests existed only for sync case before, figured was worth adding here

hinthornw (approved):

Issue #4184, opened 2025-04-06
Pydantic graph state no longer deserializes UUIDs correctly


optimalstrategy:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from typing import Any, Dict
from uuid import UUID

from pydantic import BaseModel
from langgraph.graph import StateGraph


class State(BaseModel):
    id: UUID


def my_node(state: State) -> Dict[str, Any]:
    assert isinstance(state.id, UUID), type(state.id)  # assertion error
    print(state)

    return {"id": state.id}


workflow = StateGraph(State)
workflow.add_node("my_node", my_node)
workflow.add_edge("__start__", "my_node")
graph = workflow.compile()

graph.invoke({"id": "fe096781-5601-53d2-b2f6-0d3403f7e9ca"})
```

### Error Message and Stack Trace (if applicable)

```shell
/main.py:463: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `uuid` - serialized value may not be as expected [input_value='fe096781-5601-53d2-b2f6-0d3403f7e9ca', input_type=str])
  return self.__pydantic_serializer__.to_python(
```

### Description

* I'm storing UUIDs in my Pydantic graph state
* I expect UUIDs to be deserialized into `UUID` objects (as per normal Pydantic behavior and like they used to in previous LangGraph versions)
* The UUIDs are actually incorrectly deserialized into `str` objects

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #1 SMP PREEMPT_DYNAMIC Mon, 10 Mar 2025 01:49:31 +0000
> Python Version:  3.12.8 (main, Dec 19 2024, 14:33:20) [Clang 18.1.8 ]

Package Information
-------------------
> langchain_core: 0.3.51
> langsmith: 0.3.24
> langgraph_api: 0.0.46
> langgraph_cli: 0.1.89
> langgraph_license: Installed. No version info available.
> langgraph_sdk: 0.1.61
> langgraph_storage: Installed. No version info available.

Optional packages not installed
-------------------------------
> langserve

[...]

PR #4200, opened 2025-04-08
Validate other types in model_construct

hinthornw:
Resolves:
https://github.com/langchain-ai/langgraph/issues/4184
https://github.com/langchain-ai/langgraph/issues/4198 <- tested on python 3.9 and 3.10


libs/langgraph/langgraph/graph/schema_utils.py
```diff
@@ -204,7 +216,97 @@ def union_coercer(v: Any, d: Any) -> Any:
...
+    str,
+    bool,
+    bytes,
+    bytearray,
+    complex,
+    memoryview,
+    type(None),
+)
+
+try:
```
hinthornw, line 243: Grotesque.

libs/langgraph/langgraph/graph/schema_utils.py
```diff
@@ -25,54 +28,57 @@
...
+    ) -> None:
+        if hasattr(self, "_initialised"):
             return
-        self._inited = True
+        self._initialised = True
+
         self.schema = schema
-        self.type_hints = (
-            type_hints
-            if type_hints is not None
```
nfcampos, line 55: Why change the comparison to truthy?

PR #4242, opened 2025-04-11
Avoid validating node input more than once per superstep

nfcampos:


libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -926,34 +935,32 @@ def _proc_input(
...
-                        continue
-                else:
-                    val[k] = managed[k]()
-        except EmptyChannelError:
-            return
+        val: dict[str, Any] = {}
+        for k, chan in proc.channels.items():
+            if chan in channels:
+                if channels[chan].is_available():
+                    val[k] = channels[chan].get()
```
hinthornw, line 951: So skip the channel if it's not available? When wold it not be avialble?

libs/langgraph/langgraph/pregel/algo.py
```diff
@@ -926,34 +935,32 @@ def _proc_input(
...
         for chan in proc.channels:
-            try:
-                val = read_channel(channels, chan, catch=False)
-                break
-            except EmptyChannelError:
-                pass
+            if chan in channels:
+                if channels[chan].is_available():
+                    val = channels[chan].get()
+                    break
```
hinthornw, line 959: Why do we break here but not in the dict case?

hinthornw (approved):

hinthornw:
curious on benchmark results

libs/langgraph/langgraph/pregel/algo.py
nfcampos: Empty, ie not written to yet

libs/langgraph/langgraph/pregel/algo.py
nfcampos: Because for list we get the first channel that contains something. For dict we get all channels that contain something

eyurtsev:
it's getting faster and faster

image

hinthornw:
> it's getting faster and faster
> 


Just wait until we use LangGraph's time travel capabilities to let the agent respond before you ask the question

PR #4250, opened 2025-04-11
langgraph: support streaming messages from Command.update

vbarda:


libs/langgraph/langgraph/pregel/messages.py
```diff
@@ -153,32 +153,35 @@ def on_chain_end(
         **kwargs: Any,
     ) -> Any:
         if meta := self.metadata.pop(run_id, None):
-            if isinstance(response, BaseMessage):
```
nfcampos, line 156: Do we really need to make this recursive? Commands, when used, are always top level return values no?

vbarda: they can be returned inside lists - tool node returns a list of commands, for example

nfcampos (approved):

PR #4255, opened 2025-04-11
langgraph: handle pydantic updates consistently in Command

### LINKED ISSUE #3950: Command update overrides States fields with None
### Checked other resources

### Example Code

```python

class State(BaseModel):
    foo: str | None = None
    bar: str | None = None

def node_a(state: State):
    return Command(goto="node_b", update=State(foo='foo'))

def node_b(state: State):
    return Command(goto=END, update=State(bar='bar'))

builder = StateGraph(State)
builder.add_node(node_a)
builder.add_node(node_b)
builder.add_edge(START, "node_a")
graph = builder.compile()

print(graph.invoke(State()))
```

```shell

```

### Description

Output:
{'foo': None, 'bar': 'bar'}
Expected:
{'foo': 'foo', 'bar': 'bar'}

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000
> Python Version:  3.11.9 (main, Apr  2 2024, 08:25:04) [Clang 15.0.0 (clang-1500.3.9.4)]

Package Information
-------------------
> langchain_core: 0.3.46
> langchain: 0.3.21
> langchain_community: 0.3.20
> langsmith: 0.3.18
> langchain_openai: 0.3.9
> langchain_text_splitters: 0.3.7
> langgraph_sdk: 0.1.58

vbarda:
Fixes https://github.com/langchain-ai/langgraph/issues/3950

### DIFF
--- a/libs/langgraph/langgraph/utils/pydantic.py
+++ b/libs/langgraph/langgraph/utils/pydantic.py
@@ -1,15 +1,11 @@
 import sys
 import typing
 from dataclasses import is_dataclass
-from typing import Any, Dict, Optional, Sequence, Union
+from typing import Any, Dict, Optional, Union
 
 import typing_extensions
 from pydantic import BaseModel
 from pydantic.v1 import BaseModel as BaseModelV1
-
-# NOTE: this is redefined here separately from langgraph.constants
-# to avoid a circular import
-MISSING = object()
 
 
 def create_model(
@@ -43,36 +39,6 @@ def create_model(
             v1_kwargs["__root__"] = root
 
         return create_model(model_name, **v1_kwargs, **(field_definitions or {}))
-
-
-def get_update_as_tuples(input: Any, keys: Sequence[str]) -> list[tuple[str, Any]]:
-    """Get Pydantic state update as a list of (key, value) tuples."""
-    # Pydantic v1
-    if isinstance(input, BaseModelV1):
-        keep: Optional[set[str]] = input.__fields_set__
-        defaults = {k: v.default for k, v in input.__fields__.items()}
-    # Pydantic v2
-    elif isinstance(input, BaseModel):
-        keep = input.model_fields_set
-        defaults = {k: v.default for k, v in input.model_fields.items()}
-    else:
-        keep = None
-        defaults = {}
-
-    # NOTE: This behavior for Pydantic is somewhat inelegant,
-    # but we keep around for backwards compatibility
-    # if input is a Pydantic model, only update values
-    # that are different from the default values or in the keep set
-    return [
-        (k, value)
-        for k in keys
-        if (value := getattr(input, k, MISSING)) is not MISSING
-        and (
-            value is not None
-            or defaults.get(k, MISSING) is not None
-            or (keep is not None and k in keep)
-        )
-    ]
 
 
 def is_supported_by_pydantic(type_: Any) -> bool:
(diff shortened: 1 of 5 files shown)


libs/langgraph/langgraph/utils/pydantic.py
```diff
@@ -1,12 +1,16 @@
...
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Sequence, Union
 
 import typing_extensions
 from pydantic import BaseModel
 from pydantic.v1 import BaseModel as BaseModelV1
 
+# NOTE: this is redefined here separately from langgraph.constants
+# to avoid a circular import
+MISSING = object()
```
nfcampos, line 12: How could there be a circular import w constants? Does constants import any file?

libs/langgraph/langgraph/utils/pydantic.py
```diff
@@ -41,6 +45,36 @@ def create_model(
         return create_model(model_name, **v1_kwargs, **(field_definitions or {}))
 
 
+def get_update_as_tuples(input: Any, keys: Sequence[str]) -> list[tuple[str, Any]]:
```
nfcampos, line 48: Note this logic isn't for Pydantic only, it covers all non-dict cases

libs/langgraph/langgraph/utils/pydantic.py
vbarda: yea, imports `langgraph.types` for backwards compatibility

libs/langgraph/langgraph/utils/pydantic.py
vbarda: agreed, i think might need to move to a different file to make that clear

libs/langgraph/tests/test_pregel.py
```diff
@@ -7246,6 +7246,39 @@ def node_a(state: State) -> State:
...
+
+    class State(BaseModel):
+        foo: Optional[str] = None
+        bar: Optional[str] = None
+
+    def node_a(state: State):
+        return State(foo="foo")
+
+    def node_b(state: State):
+        return Command(update=State(bar="bar"))
```
chenweisomebody126, line 7269: Thanks for this example

Sourav-Nandy-ai:
Hi Vadym,

I ran an actual before-and-after test on this rather than trusting the description. I installed langgraph==0.3.29 (the release right before your merge) and 0.3.30 (released approximately 2 hours after). I then built a graph where node 1 sets field_b, and node 2 returns Command(update=MyState(field_a="new_a")). This explicitly sets only field_a while leaving field_b unset on the model instance.

The Results
- Pre-fix: field_b gets silently reset to None. Node 2's update wipes a sibling field it never touched.
- Post-fix: field_b correctly survives with node 1's value.

I confirmed this at the source level as well. The model_fields_set-aware handling is genuinely absent in 0.3.29 and present in 0.3.30, exactly matching your fix.

I build agent-eval (agent_regress), an OSS harness for exactly this kind of cross-version behavioral comparison. A new subprocess_runner() lets compare() diff two installed langgraph builds directly. This is exactly what made this test possible without hand-rolling process isolation.

Given you are deep in this exact surface, I am curious whether a model_fields_set-based repro like this is the shape you would want for regression-testing future Command-update changes.

You can find the full script and raw output here: https://github.com/RudrenduPaul/agent-eval

Best,
Sourav

PR #4286, opened 2025-04-15
docs(agents): minor fixes

eyurtsev:


docs/docs/agents/agents.md
```diff
@@ -55,63 +55,66 @@ agent = create_react_agent(
 )
 ```
 
+See the [models](./models.md) page for more information on how to configure LLMs.
+
 ## Custom Prompts
 
 Prompts instruct the LLM how to behave. They can be:
 
 * **Static**: A fixed string or list of [messages](https://python.langchain.com/docs/concepts/messages/)
```
vbarda, line 64: actually this one is misleading, because you can't currently pass a list of messages as a prompt to `create_react_agent`, so it might confuse people. we probably need to rephrase

eyurtsev: Got it -- will add to do list

eyurtsev: updated

PR #4296, opened 2025-04-16
docs(agents): add memory to the agents page + run agents page

vbarda:


docs/docs/agents/agents.md
```diff
@@ -112,6 +112,59 @@ agent.invoke(
...
+
+# Run the agent
+# highlight-next-line
+config = {"configurable": {"thread_id": "1"}}
+sf_response = agent.invoke(
+    {"messages": "what is the weather in sf"},
+    # highlight-next-line
+    config
+)
+ny_response = agent.invoke(
```
eyurtsev, line 141: It's not obvious why this shows multi-turn interaction. There needs to be an explanation that it responds about the weather even even though it wasn't included in the query.

One could argue that it would do that by default b/c it only has access to the get weather tool (though I'm not sure)

docs/docs/agents/agents.md
```diff
@@ -112,6 +112,59 @@ agent.invoke(
...
+    {"messages": "what about new york?"},
+    # highlight-next-line
+    config
+)
+```
+
```
eyurtsev, line 147: Add a link to the memory.md page from here

"For more details about how to work with memory please see..."

docs/docs/agents/agents.md
```diff
@@ -112,6 +112,59 @@ agent.invoke(
...
+agent = create_react_agent(
+    model="anthropic:claude-3-7-sonnet-latest",
+    tools=[get_weather],
+    # highlight-next-line
+    checkpointer=checkpointer
+)
+
+# Run the agent
+# highlight-next-line
+config = {"configurable": {"thread_id": "1"}}
```
eyurtsev, line 135: need explanation about what a thread id is

docs/docs/agents/agents.md
```diff
@@ -112,6 +112,59 @@ agent.invoke(
...
+ny_response = agent.invoke(
+    {"messages": "what about new york?"},
+    # highlight-next-line
+    config
+)
+```
+
+## Inputs and outputs
+
+Agent operates on a list of messages via `messages` key in its [state](https://langchain-ai.github.io/langgraph/concepts/low_level/#working-with-messages-in-graph-state).
```
eyurtsev, line 150: nit: there might be a way to paraphrase this to be more precise.

e.g., the agent uses an llm that expects messages as an input and outputs a message. The inputs (or prompt / chat history) are stored in the 'messages' key in the agent state.

^ This a bit more precise, but too long

eyurtsev (approved):

Issue #4343, opened 2025-04-18
LangGraph PostgresStore bug: index `fields` parameter ignored due to incorrect naming


VMinB12:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
import os
from langgraph.store.postgres import PostgresStore
from langgraph.store.postgres.base import PostgresIndexConfig
from langgraph.store.base import PutOp
from langchain_core.embeddings import FakeEmbeddings

size = 1024
embeddings = FakeEmbeddings(size=size)

db_string = os.getenv("DB_CONN_STRING")
assert db_string is not None, "DB_CONN_STRING environment variable is not set"

index: PostgresIndexConfig = {
    "dims": size,
    "embed": embeddings,
    "fields": ["text"], # <--- Ignored!
    #"text_fields": ["text"], # <--- Correct behaviour, but mismatch with `PostgresIndexConfig` type.
}


with PostgresStore.from_conn_string(
    db_string,
    index=index,
) as store:
    print(store.index_config)

    preps = store._prepare_batch_PUT_queries(
        [
            (
                1,
                PutOp(
                    namespace=("documents", "user123"),
                    key="report1",
                    value={
                        "text": "hello world",
                        "metadata": {"id": "1"},
                        "user_ids": ["1"],
                    },
                ),
            )
        ]
    )
    for _prep in preps:
        for prep in _prep:
            print("=" * 20)
            print(prep)
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

I am using the `langgraph.store.postgres.base.PostgresStore` implementation and noticed that all fields were embedded, not just the fields I specified in the IndexConfig.

After digging into the code, I found this:
```
def _ensure_index_config(
    index_config: PostgresIndexConfig,
) -> tuple[Optional["Embeddings"], PostgresIndexConfig]:
    index_config = index_config.copy()
    tokenized: list[tuple[str, Union[Literal["$"], list[str]]]] = []
    tot = 0
    text_fields = index_config.get("text_fields") or ["$"]
```
[...]

PR #4344, opened 2025-04-18
bug: use "fields" instead of "text_fields" in _ensure_index_config

VMinB12:
Fix for #4343


hinthornw:
Thanks for detecting and following through on a fix! Great work

hinthornw (approved):

PR #4354, opened 2025-04-21
Rewrite graph drawing logic

nfcampos:
- It now executes the same pregel algo as when the graph is executed (without running any user code in nodes or conditional edges) to discover all the edges
- This means we now support drawing the graph for all Pregel instances, not just StateGraph
- This is done in preparation for new edge/node type coming in separate PR
- Known changes
  - custom labels on conditional edges to END are no longer displayed


libs/langgraph/tests/test_pregel_async.py
```diff
@@ -4980,7 +4982,7 @@ async def decider(data: State) -> str:
 
     app = workflow.compile()
 
-    if SHOULD_CHECK_SNAPSHOTS:
+    if SHOULD_CHECK_SNAPSHOTS and checkpointer_name == "memory":
```
hinthornw, line 4985: Why does this have to change?

libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -562,14 +563,87 @@ def __init__(
...
+                            )
+                            for p in subpregels.values()
+                        )
+                    ),
+                )
+            }
+        else:
+            subgraphs = {}
+
+        return draw_graph(
```
hinthornw, line 629: do we wanna do asyncio.to_thread here?

libs/langgraph/tests/test_pregel_async.py
nfcampos: Because it's pointless to snapshot the same thing 10 times for different checkpointers

libs/langgraph/langgraph/pregel/__init__.py
nfcampos: no, this doesn't do IO so nothing to gain?

libs/langgraph/langgraph/channels/dynamic_barrier_value.py
```diff
@@ -8,7 +9,7 @@
 
 
 class WaitForNames(NamedTuple):
-    names: set[Any]
+    names: Set[Any]
```
nfcampos, line 12: now accepts mutable or frozen set

libs/langgraph/langgraph/graph/branch.py
```diff
@@ -138,7 +142,15 @@ def run(
                 name=None,
                 trace=False,
                 func_accepts_config=True,
+            ),
+            list(
+                zip_longest(
+                    writer([e for e in self.ends.values() if e != END]),
```
hinthornw, line 148: Why would the writer's output sequence length differ frmo the input sequence length?

hinthornw (approved):

libs/langgraph/langgraph/graph/branch.py
nfcampos: When you use `then` there is an extra write

PR #4359, opened 2025-04-21
INTPYTHON-433 LangGraph long-term memory for MongoDB

caseyclements:
Basic draft of `langgraph.store.mongodb.MongoDBStore`

Async, and Vector Search aren't shown.

Questions

- What is being done in postgres' "migrations"?
- Is there a particular behaviour done on concurrent writes? 
  - I am getting a DuplicateKeyError in the test in `test_pregel.test_store_injected`
  - [test_async_store::test_large_batches](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/tests/test_async_store.py#L128-L174) highlights the situation.

### DIFF
--- a/libs/checkpoint-mongodb/CHANGELOG.md
+++ b/libs/checkpoint-mongodb/CHANGELOG.md
@@ -1,0 +1,19 @@
+# Changelog
+
+---
+
+## Changes in version 0.1.3 (2025/04/01)
+
+- Add compatibility with `pymongo.AsyncMongoClient`.
+
+## Changes in version 0.1.2 (2025/03/26)
+
+- Add compatibility with `langgraph-checkpoint` 2.0.23.
+
+## Changes in version 0.1.1 (2025/02/26)
+
+- Remove dependency on `langgraph`.
+
+## Changes in version 0.1 (2024/12/13)
+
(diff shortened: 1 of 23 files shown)


hinthornw:
Hello!

> What is being done in postgres' "migrations"?
These are run when you call `setup()` to create the tables & indices expected by the store. The migrations contain the actual SQL migration code.

> Is there a particular behaviour done on concurrent writes?
> I am getting a DuplicateKeyError in the test in test_pregel.test_store_injected
[test_async_store::test_large_batches](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/tests/test_async_store.py#L128-L174) highlights the situation.

For concurrent writes, we assume the "last write wins", so if you have a batch containing put ops to create, update , then delete an item, the query would end with the item being deleted. In postgres (and the batch api), we currently deduplicate in python before even creating the query

caseyclements:
> For concurrent writes, we assume the "last write wins", so if you have a batch containing put ops to create, update , then delete an item, the query would end with the item being deleted. In postgres (and the batch api), we currently deduplicate in python before even creating the query

Thanks for the prompt reply! Ok - I will deduplicate and see if the test in test_pregel passes.

PR #4361, opened 2025-04-21
fix(langgraph): Fix deserialization issue for AnyMessage objects, and…

hinthornw:
…… (#4317)

Fix deserialization issue for AnyMessage objects, and add support for deserializing Pydantic generic and polymorphic models.

bug detail:
https://github.com/langchain-ai/langgraph/issues/4316

Thanks to [ittlebai3618](https://github.com/langchain-ai/langgraph/commits?author=littlebai3618) for raising and proposing the fix. Going to double check + add tests before landing.

### DIFF
--- a/libs/langgraph/langgraph/graph/schema_utils.py
+++ b/libs/langgraph/langgraph/graph/schema_utils.py
@@ -5,13 +5,14 @@ from inspect import isclass
 from typing import (
     Any,
     Callable,
+    Hashable,
     Optional,
     Type,
-    get_args,
     TypeVar,
+    Union,
+    get_args,
+    get_origin,
     get_type_hints,
-    Hashable,
-    get_origin, Union,
 )
 
 from pydantic import BaseModel, Discriminator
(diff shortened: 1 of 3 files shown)


hinthornw:
(the suggested fix doesn't work - working on this after some other fixes)

littlebai3618:
What are the failing cases? Can you describe them?
Right now, I have to use version 0.3.27 and manually call model_validate as a temporary workaround for the deserialization errors.
I’ll have some time in the next few days to look into this issue.

crenus:
Does 3.27 work? with that I no longer get errors whe nI start from langgraph studio, but when it gets to a subgraph expecting AnyMEssage, that still doesnt work. Previously worked on 0.3.24 for me

littlebai3618:
> Does 3.27 work? with that I no longer get errors whe nI start from langgraph studio, but when it gets to a subgraph expecting AnyMEssage, that still doesnt work. Previously worked on 0.3.24 for me



In version 0.3.27, polymorphic types are stored as raw dictionaries. Therefore, it’s necessary to manually deserialize them into the correct types using BaseModel.model_validate() when handling state objects.

like:
```python
def node(state):
      state.polymorphic = polymorphic_type.model_validate(state.polymorphic.value)
```

PR #4484, opened 2025-04-30
feat: assistants sorting sdk spec

lc-arjun:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -767,6 +767,10 @@ async def search(
         graph_id: Optional[str] = None,
         limit: int = 10,
         offset: int = 0,
+        sort_by: Optional[
+            Literal["assistant_id", "graph_id", "name", "created_at", "updated_at"]
+        ] = None,
+        sort_order: Optional[Literal["asc", "desc"]] = None,
```
vbarda, line 773: (nit) could we just put this as a constant in the `schema.py`?

vbarda (approved):

PR #4520, opened 2025-05-02
docs: update persistence & memory how-tos

vbarda:


docs/docs/how-tos/memory.ipynb
```diff
@@ -0,0 +1,1089 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "15ed7413-c876-4d38-b080-79c71133549b",
+   "metadata": {},
+   "source": [
+    "# Manage Conversation History\n",
```
lnhsingh, line 9: Maybe "Optimize memory" instead?

We also probably want to cross link with the Persistence how-tos in the intro here too.

docs/docs/how-tos/memory.ipynb
```diff
@@ -0,0 +1,1089 @@
...
+    "    print(event)\n",
+    "    print(\"\\n\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ae65737a-dcf9-4987-9de8-948afd6a820d",
+   "metadata": {},
+   "source": [
+    "## Strategies"
```
lnhsingh, line 474: You can probably get rid of this H2 and promote the H3s under it up

docs/docs/how-tos/memory.ipynb
```diff
@@ -0,0 +1,1089 @@
...
+    "\n",
+    "_set_env(\"ANTHROPIC_API_KEY\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2c338d33-c17f-40a5-8167-a9ee4aaeeea7",
+   "metadata": {},
+   "source": [
+    "## Process Model Input"
```
lnhsingh, line 66:
```suggestion
    "## Process model input"
```

lnhsingh: Might be good to add a sentence or two explaining why you would need to do this

lnhsingh (commented):
Quick feedback added - can add more once you've worked in Eugene's feedback

PR #4539, opened 2025-05-05
docs: Fix broken navigation for `Local deploy` page

andrewnguonly:
### Summary
The `Local deploy` page's left-hand navigation looks like this:
image

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -330,11 +330,10 @@ nav:
       - Environment variables: cloud/reference/env_var.md
 
   - Examples:
-    - tutorials/langgraph-platform/local-server.md
     - tutorials/workflows/index.md
-    - tutorials/rag/langgraph_agentic_rag.ipynb
-    - tutorials/multi_agent/agent_supervisor.ipynb
-    - tutorials/customer-support/customer-support.ipynb
+    - Agentic RAG: tutorials/rag/langgraph_agentic_rag.ipynb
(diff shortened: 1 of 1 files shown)


docs/mkdocs.yml
```diff
@@ -330,11 +330,10 @@ nav:
       - Environment variables: cloud/reference/env_var.md
 
   - Examples:
-    - tutorials/langgraph-platform/local-server.md
```
andrewnguonly, line 333: For some reason, including the page here causes the issue. I suspect `mkdocs` is just taking the location of the last reference to the page (`Examples`) and using the navigation from there.

In other words, you can't reference a page twice in `mkdocs.yml`?

docs/mkdocs.yml
```diff
@@ -330,11 +330,10 @@ nav:
...
     - tutorials/workflows/index.md
-    - tutorials/rag/langgraph_agentic_rag.ipynb
-    - tutorials/multi_agent/agent_supervisor.ipynb
-    - tutorials/customer-support/customer-support.ipynb
+    - Agentic RAG: tutorials/rag/langgraph_agentic_rag.ipynb
```
andrewnguonly, line 334: Adding these navigation titles. We can change them later if we want.

docs/mkdocs.yml
andrewnguonly: Confirmed. It's not recommended to reference a page twice in `mkdocs.yml`.

PR #4589, opened 2025-05-08
langgraph: use tuples for streamed message events in RemoteGraph

vbarda:


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -687,6 +687,10 @@ def stream(
             # filter for what was actually requested
             if mode not in requested:
                 continue
+
+            if chunk.event.startswith("messages"):
```
eyurtsev, line 691: why `startswith`?

nfcampos (approved):

PR #4590, opened 2025-05-08
Fix remote streaming of subgraphs

hinthornw:
Because we were propagating the task ID config key, the stream mode was always overridden as "values", meaning the token, etc. callback handlers were never added within the remote graphs.


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -50,6 +51,7 @@
         CONFIG_KEY_CHECKPOINT_MAP,
         CONFIG_KEY_CHECKPOINT_ID,
         CONFIG_KEY_CHECKPOINT_NS,
+        CONFIG_KEY_TASK_ID,
```
hinthornw, line 54: This is the only actual change. The rest is just to setup the testing harness

libs/langgraph/integration_tests/example_app/example_graph.py
```diff
@@ -0,0 +1,90 @@
+from typing import Annotated
```
hinthornw, line 1: Copy/pasted from unit test folder

PR #4606, opened 2025-05-09
docs: Update application structure pages

andrewnguonly:
### Summary
Small wording updates to the application structure documentation pages.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -186,9 +186,8 @@ nav:
             - cloud/deployment/setup.md
             - cloud/deployment/setup_pyproject.md
             - cloud/deployment/setup_javascript.md
-            - cloud/deployment/semantic_search.md
             - cloud/deployment/custom_docker.md
-            - how-tos/autogen-langgraph-platform.ipynb
+            
         - LangGraph CLI: concepts/langgraph_cli.md
         - LangGraph Studio:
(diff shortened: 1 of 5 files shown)


docs/mkdocs.yml
```diff
@@ -202,6 +201,7 @@ nav:
           - how-tos/local-studio.md
         - LangGraph SDK: concepts/sdk.md
       - Data management:
+        - cloud/deployment/semantic_search.md
```
andrewnguonly, line 204: This feels more appropriate here.

docs/mkdocs.yml
```diff
@@ -308,6 +308,7 @@ nav:
         - tutorials/auth/add_auth_server.md
       - Rebuild graph at runtime: cloud/deployment/graph_rebuild.md
       - how-tos/use-remote-graph.md
+      - how-tos/autogen-langgraph-platform.ipynb
```
andrewnguonly, line 311: This page is unrelated to application (file) structure. It feels more appropriate as an example.

docs/mkdocs.yml
hinthornw: hm fine with it ya! maybe should ahve a whole section on the long-term memory store

andrewnguonly: I was actually looking for a top-level `Memory` section; I was kind of expecting one...

PR #4607, opened 2025-05-09
docs: Small changes to Server Customization section

andrewnguonly:
### Summary
Small changes to Server Customization section.


docs/mkdocs.yml
```diff
@@ -245,7 +245,7 @@ nav:
       - Cron Jobs:
         - Overview: cloud/concepts/cron_jobs.md
         - cloud/how-tos/cron_jobs.md
-      - Modifying the API:
+      - Server Customization:
```
andrewnguonly, line 248: Opinion: I think this is way better than `Modifying the API`.

PR #4609, opened 2025-05-09
docs: Updates to deployment options pages

andrewnguonly:


docs/docs/concepts/platform_architecture.md
andrewnguonly, line 1: This page is kind of random. I moved it's content under `LangGraph Data Plane`.

docs/docs/concepts/scalability_and_resilience.md
```diff
@@ -25,16 +25,16 @@ When a graceful shutdown request is received (SIGINT) an instance enters shutdow
 - gives any in-progress runs a limited number of seconds to finish (if not finished it will be put back in the queue)
 - stops the instance from picking up more runs from the queue
 
-If a hard shutdown occurs, eg. due to a server crash, or an infra failure, any runs that were in progress will be picked up by a periodic sweeper task that looks for in-progress runs that have breached their heartbeat window, which will put them back in the queue for another instance to pick them up.
+If a hard shutdown occurs due to a server crash or an infrastructure failure, any runs that were in progress will be picked up by a periodic sweeper task that looks for in-progress runs that have breached their heartbeat window, which will put them back in the queue for another instance to pick them up.
 
 ## Postgres resilience
 
-For deployment modalities where we manage the Postgres database we have periodic backups, continuously replicated standby replicas for automatic failover. Optionally, on request, we can also setup read replicas as well as other advanced failover capabilities.
```
andrewnguonly, line 32: @langchain-infra, please double check these changes.

docs/mkdocs.yml
```diff
@@ -251,23 +251,21 @@ nav:
         - how-tos/http/custom_routes.md
       - Deployment options:
         - Overview: concepts/deployment_options.md
-        - concepts/platform_architecture.md
-        - Scalability & resilience: concepts/scalability_and_resilience.md
+        - LangGraph Data Plane: concepts/langgraph_data_plane.md
```
andrewnguonly, line 254: Clarifying: The `LangGraph Data Plane` and `LangGraph Control Plane` are components of all deployment options.

docs/docs/concepts/scalability_and_resilience.md
```diff
@@ -25,16 +25,16 @@ When a graceful shutdown request is received (SIGINT) an instance enters shutdow
 - gives any in-progress runs a limited number of seconds to finish (if not finished it will be put back in the queue)
 - stops the instance from picking up more runs from the queue
 
-If a hard shutdown occurs, eg. due to a server crash, or an infra failure, any runs that were in progress will be picked up by a periodic sweeper task that looks for in-progress runs that have breached their heartbeat window, which will put them back in the queue for another instance to pick them up.
+If a hard shutdown occurs due to a server crash or an infrastructure failure, any runs that were in progress will be picked up by a periodic sweeper task that looks for in-progress runs that have breached their heartbeat window, which will put them back in the queue for another instance to pick them up.
```
lnhsingh, line 28:
```suggestion
If a hard shutdown occurs due to a server crash or an infrastructure failure, any runs that were in progress will be picked up by a periodic sweeper task that looks for in-progress runs that have breached their heartbeat window. The sweeper will put the runs back in the queue for another instance to pick them up.
```
Can we provide any more details here like how often the sweeper runs or if it's configurable / how you do it?

docs/mkdocs.yml
lnhsingh: I don't think `data plane` and `control plane` should be proper nouns

Also, I think this section should be structured a little differently since not all of the content is related to picking a deployment option-- it's about deployment in general. Maybe something like:

```
- Deployments:
    - Overview:
    - Data plane
    - Control plane
    - Deployment options:
        - Cloud Saas:
            - Overview:
            - Deploy Cloud SaaS:
        - SH Control Plane:
        - etc...
```

docs/mkdocs.yml
```diff
@@ -251,23 +251,21 @@ nav:
...
-          - LangGraph Data Plane: concepts/langgraph_data_plane.md
           - Deploy Self-Hosted Data Plane: cloud/deployment/self_hosted_data_plane.md
         - Self-Hosted Control Plane:
           - Overview: concepts/langgraph_self_hosted_control_plane.md
-          - LangGraph Control Plane: concepts/langgraph_control_plane.md
           - Deploy Self-Hosted Control Plane: cloud/deployment/self_hosted_control_plane.md
         - Standalone Container:
           - Overview: concepts/langgraph_standalone_container.md
           - Deploy Standalone Container: cloud/deployment/standalone_container.md
-          # ADD SH LITE?
```
lnhsingh, line 270: I think we're missing information about Self-Hosted Lite, wdyt? If it doesn't warrant its own page, I think there should be a heading for it on the Standalone Container overview page

docs/docs/concepts/scalability_and_resilience.md
andrewnguonly: Accepting your changes and adding a detail about the sweeper running every 2 minutes. It's not configurable. Do we need to explicitly say it's "not configurable"? I changed the adjective `periodic sweeper` to `internal sweeper` to try to convey that it's something that isn't exposed to the users and can't be touched/configured/etc.

docs/mkdocs.yml
andrewnguonly: 4-levels of nesting :+1:. I'm fine with this, but just calling it out.

andrewnguonly: Going to uncapitalize all instances of `control plane` and `data plane` besides headings and references to the actual deployment option name (e.g. Self-Hosted Data Plane).

docs/mkdocs.yml
andrewnguonly: I'll add a heading for it in the `Standalone Container` page. Will clarify offline.

docs/docs/concepts/scalability_and_resilience.md
```diff
@@ -25,16 +25,16 @@ When a graceful shutdown request is received (SIGINT) an instance enters shutdow
...
 
 ## Postgres resilience
```
langchain-infra, line 30: I think true for prod type deployments? Maybe can say SaaS only.

docs/mkdocs.yml
andrewnguonly: Actually, I'm going to do this in a separate, smaller PR.

docs/docs/concepts/scalability_and_resilience.md
lnhsingh: SGTM

lnhsingh:
LGTM

PR #4662, opened 2025-05-12
feat(docs): threads

lc-arjun:


docs/docs/cloud/how-tos/create_threads.md
```diff
@@ -0,0 +1,234 @@
+# How to create a thread
```
lnhsingh, line 1: Can we change this to something more generic like "Use threads" and then the headings would be
```
## Create a thread
### Empty thread
### Prepopulated thread
## Copy a thread
## List threads
...
## Inspect threads
...
```

docs/docs/cloud/how-tos/view_threads.md
```diff
@@ -0,0 +1,266 @@
+# How to view and inspect Threads
+
+!!! info "Prerequisites"
+
+    - [Threads Overview](../concepts/threads.md)
+    - [How to create threads](./create_threads.md)
+
+This guide shows how to view threads in a LangGraph Platform application and inspect their state.
+
+## List Threads
```
lnhsingh, line 10:
```suggestion
## List threads
```

docs/docs/cloud/how-tos/create_threads.md
lc-arjun: sounds good

PR #4690, opened 2025-05-14
docs: document tuples in streaming guides

vbarda:


libs/langgraph/langgraph/pregel/__init__.py
```diff
@@ -2386,13 +2394,64 @@ def node_a(state: State):
...
+            # Events are emitted as 2-tuples (LLM token, metadata).
             # (AIMessageChunk(content='The', additional_kwargs={}, response_metadata={}, id='...'), {'langgraph_step': 1, 'langgraph_node': 'a', 'langgraph_triggers': ['start:a'], 'langgraph_path': ('__pregel_pull', 'a'), 'langgraph_checkpoint_ns': '...', 'checkpoint_ns': '...', 'ls_provider': 'openai', 'ls_model_name': 'gpt-4o-mini', 'ls_model_type': 'chat', 'ls_temperature': 0.7})
             # (AIMessageChunk(content=' capital', additional_kwargs={}, response_metadata={}, id='...'), {'langgraph_step': 1, 'langgraph_node': 'a', 'langgraph_triggers': ['start:a'], ...})
             # (AIMessageChunk(content=' of', additional_kwargs={}, response_metadata={}, id='...'), {...})
             # (AIMessageChunk(content=' France', additional_kwargs={}, response_metadata={}, id='...'), {...})
             # (AIMessageChunk(content=' is', additional_kwargs={}, response_metadata={}, id='...'), {...})
             # (AIMessageChunk(content=' Paris', additional_kwargs={}, response_metadata={}, id='...'), {...})
             ```
+
+        Example: Using subgraphs=True:
```
eyurtsev, line 2406: IMO maybe we consider removing some of the examples in the reference? The reference becomes a bit unmangeable

eyurtsev (approved):

libs/langgraph/langgraph/pregel/__init__.py
vbarda: good point - i will cross link instead

PR #4695, opened 2025-05-15
langgraph: fix drawing graph with __root__ channel

vbarda:


nfcampos (approved):

libs/langgraph/langgraph/pregel/draw.py
```diff
@@ -120,16 +125,16 @@ def draw_graph(
                                     edges.add((task.name, t[0], True, t[2]))
                             writes = [t for t in writes if t[0] != END]
                             conditionals.update(
-                                {(task.name, *t[:2]): t[2] for t in writes}
+                                {(task.name, t[0], t[1] or None): t[2] for t in writes}
```
nfcampos, line 128: Why is this change needed?

libs/langgraph/langgraph/pregel/draw.py
```diff
@@ -120,16 +125,16 @@ def draw_graph(
...
                             )
                             task.config[CONF][CONFIG_KEY_SEND]([t[:2] for t in writes])
             # collect sources
             step_sources = {
                 task.name: {
                     (
                         w[0],
-                        (task.name, *w) in conditionals,
-                        conditionals.get((task.name, *w)),
+                        (task.name, w[0], w[1] or None) in conditionals,
```
nfcampos, line 136: Same as above

libs/langgraph/langgraph/pregel/draw.py
vbarda: because the empty writes values in case of list/dict are unhashable

vbarda: (needed for the line below, and this one is just for consistency)

PR #4751, opened 2025-05-19
Remove local_write utility

nfcampos:
- The validation isn't worth the cost of having to pass list of nodes to task config


sydney-runkle (approved):
Nice simplification

PR #4761, opened 2025-05-20
docs: typo in npx command to install LangGraph CLI

leeroybrun:
Docs have the wrong npx command to install LangGraph CLI:

`npx @langchain/langgraph-cl` instead of `npx @langchain/langgraph-cli`


sydney-runkle (approved):
Thanks!

PR #4765, opened 2025-05-20
feat(sdk-py): resumable streams

dqbd:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -344,6 +390,8 @@ async def stream(
         async with self.client.stream(
             method, path, headers=request_headers, content=content, params=params
         ) as res:
+            if on_response:
+                on_response(res)
```
hinthornw, line 394: should we say this is async? (would someone ever want to decode the response here?)

libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -116,6 +118,26 @@ def _orjson_default(obj: Any) -> Any:
         raise TypeError(f"Object of type {type(obj)} is not JSON serializable")
 
 
+def _get_run_metadata_from_response(
+    response: httpx.Response,
+) -> Optional[RunCreateMetadata]:
+    """Extract run metadata from the response headers."""
+    import re
```
hinthornw, line 125: maybe import and compile this outside the function?

hinthornw (approved):

PR #4784, opened 2025-05-21
Prebuild `rag agent`

leo-gan:
Update:
- Add a prebuilt `RAG agent` as the `create_rag_agent` function.

REASON:
- `CrewAI` uses the **default** RAG component a lot, in many workflows. I like it. It speeds up the experimentations, decreases the cognitive load. It is the same as the `ReAct agent`.

CONS:
- Compared with `ReAct agent`, the `RAG agent` has a heavyweight component, a vector store, which requires a separate initialization.

Discussion:
@eyurtsev  Eugene, I did this PR only as a discussion. If you want to move with it, let me know, and I will  proceed with these next steps:
- add a notebook with examples;
- ? What else?


leo-gan:
Closing it because of the #4889

PR #4813, opened 2025-05-24
Remove postgres shallow checkpointer

nfcampos:
- This was deprecated, and superseded by checkpoint_during=False, which is available for all checkpointers


hinthornw (approved):

sydney-runkle (approved):
Makes sense, given that this was deprecated!

satyaprakash1729:
> * This was deprecated, and superseded by checkpoint_during=False, which is available for all checkpointers

@nfcampos @sydney-runkle Thank you much for all the work you're doing and running this ship.
I've a question regarding this - what we're noticing with checkpointer is our storage is growing steadily (pace is increasing infact). Shallow checkpointer seems like a very good option rather than storing all the old checkpointers. This is a very good feature for optimizing storage. Is there any replacement available for this? Please point us at any document if available. Lot of people are complaining about the storage actually. Please check the discussion here. https://github.com/langchain-ai/langgraph/discussions/5166

Thanks in advance!

PR #4814, opened 2025-05-24
Remove old checkpoint test fixtures

nfcampos:
- Now all tests fully migrated to more recent sync_checkpointer and async_checkpointer fixtures for parametrising on checkpointer
- Use sync/async_store fixtures where tests used only in memory store
- Remove unused "should snapshot" check for older versions of langchain core no longer tested against


nfcampos:
We have stopped testing it long ago, I'm just cleaning up code someone forgot to remove at the time

PR #4816, opened 2025-05-24
Remove dict subclasses used for values/updates stream chunks

nfcampos:
- This is only used by astream_events, which is a langchain core code path we no longer recommend
- No documented langgraph code relies on this, so prefer to remove


hinthornw (approved):

sydney-runkle (commented):
Is this supposed to be merged into your rm test fixtures branch, or v1, or main?

PR #4820, opened 2025-05-25
Remove Checkpoint.pending_sends

nfcampos:
- Instead store sends in a Topic channel, removing the need to fetch sends as writes against the parent checkpoint
- Remove deprecated/unused functions in langgraph-checkpoint (will require bumping min range for langgraph-checkpoint in langgraph lib)
- Implement migration of old pending sends in langgraph-checkpoint-postgres
- Ensure parent config of `checkpoint_during=False` checkpoints always points to checkpoints that were also saved


libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -130,7 +131,37 @@ async def alist(
         # if we change this to use .stream() we need to make sure to close the cursor
         async with self._cursor() as cur:
             await cur.execute(query, args, binary=True)
-            async for value in cur:
+            values = await cur.fetchall()
+            if not values:
+                return
+            print(values)
```
hinthornw, line 137: Nit

sydney-runkle (approved):

PR #4857, opened 2025-05-28
Remove Context channel / managed value, Remove SharedValue

nfcampos:
- Both have been deprecated long ago, have not been present in docs for quite a while
- Removing these lets us delete some code paths that were dedicated to this, easing work on distributed runner


rhlarora84:
Can you suggest what is the alternative to this? Using configurable?

helioLJ:
> Can you suggest what is the alternative to this? Using configurable?

Yep, in this documentation page they provide 3 ways to supply context:

https://langchain-ai.github.io/langgraph/agents/context/

PR #4886, opened 2025-05-30
docs: Update `1-build-basic-chatbot.md` to add a section about `END` node

LostInCode404:
The current [Build a basic chatbot](https://langchain-ai.github.io/langgraph/tutorials/get-started/1-build-basic-chatbot/) tutorial shows a graph diagram that includes both a `START` and an `END` node. However, the tutorial does not define or mention the END node at all, which may cause confusion.

Running the code as-is from the current version of this tutorial does not produce a graph matching the image shown in the tutorial.

This PR adds documentation and code for adding the `END` node as well. The alternative was to update the image, but it might be more valuable to show users how to add an exit point explicitly.

### DIFF
--- a/docs/docs/tutorials/get-started/1-build-basic-chatbot.md
+++ b/docs/docs/tutorials/get-started/1-build-basic-chatbot.md
@@ -32,7 +32,7 @@ from typing import Annotated
 
 from typing_extensions import TypedDict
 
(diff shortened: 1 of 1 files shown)


sydney-runkle (approved):
Thanks!

Issue #4923, opened 2025-06-02
ClientConfig in langgraph-js headers need to be case insensitive for auth


kingsotn-twelve:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
// reproduce:
// npm install @langchain/langgraph @langchain/core
// vim headers_case_issue.mjs && node headers_case_issue.mjs

import { ThreadsClient } from "@langchain/langgraph-sdk/client";

const state = { apiKey: "api-key-from-state" };
const currentHeaders = { "x-api-key": "from-headers" }; // lowercase on purpose

const threadsClient = new ThreadsClient({
  apiUrl: "http://localhost:8123",
  apiKey: state.apiKey,
  // @ts-ignore: force override
  defaultHeaders: {
    ...currentHeaders,
    "X-Api-Key": state.apiKey,
  },
});

// @ts-ignore: inspect protected fetch options
const [, init] = threadsClient.prepareFetchOptions("/debug");
console.log("Merged headers:", init.headers);
```

### Error Message and Stack Trace (if applicable)

```shell
Merged headers: { 'x-api-key': 'from-headers', 'X-Api-Key': 'api-key-from-state' }
```

### Description

Using `defaultHeaders` in `ClientConfig` can result in duplicate headers due to case-sensitive merging:

```ts
new ThreadsClient({
  apiUrl: "http://localhost:8123",
  apiKey: "sampleKey",
  defaultHeaders: {
    "x-api-key": "sampleKey",
    "X-Api-Key": "sampleKey",
  },
});
```

results in

```bash
{
  'x-api-key': 'sampleKey',
  'X-Api-Key': 'sampleKey'
}
```

more specifically this [line](https://github.com/langchain-ai/langgraph/blob/7828003958547f0f9fc96fee97925a9752ddfe2b/libs/sdk-js/src/client.ts#L150) breaks expected HTTP behavior (header keys are case-insensitive) and can cause auth issues. `defaultHeaders` should normalize keys (e.g. using Headers or lowercased map).

```js
// package.json 
...
"@langchain/core": "^0.3.57",
[...]

PR #4926, opened 2025-06-03
fix: header merging

hinthornw:
need a better test. Current function is ridiculous will change b4 landing, but resolves https://github.com/langchain-ai/langgraph/issues/4923

### DIFF
--- a/libs/sdk-js/src/client.ts
+++ b/libs/sdk-js/src/client.ts
@@ -88,7 +88,7 @@ function* iterateHeaders(
 function mergeHeaders(
   ...headerObjects: (
     | HeadersInit
-    | Record<string, string | null | undefined>
+    | Record<string, HeaderValue>
     | undefined
     | null
   )[]
@@ -165,7 +165,7 @@ export interface ClientConfig {
   apiKey?: string;
   callerOptions?: AsyncCallerParams;
   timeoutMs?: number;
-  defaultHeaders?: Record<string, string | null | undefined>;
+  defaultHeaders?: Record<string, HeaderValue>;
   onRequest?: RequestHook;
 }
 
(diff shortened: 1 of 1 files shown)


libs/sdk-js/src/client.ts
```diff
@@ -39,6 +39,75 @@ import { getEnvironmentVariable } from "./utils/env.js";
...
+      if (value === undefined) continue;
+
+      // New object keys should always overwrite older headers
+      // Yield a null to clear the header in the headers object
+      // before adding the new value
+      if (shouldClear && !didClear) {
+        didClear = true;
+        yield [name, null];
+      }
+      yield [name, value];
```
hntrl, line 83: this covers cases where we don't want to retain headers from two objects with conflicting naming (including casing)
```
{ defaultHeaders: { 'x-api-key': 'test123' } }
{ headers: { 'X-Api-Key': 'test456' } }

// 😡
{ headers: { 'x-api-key': 'test123, test456' } }
// 🥳
// { headers: { 'x-api-key': 'test456' } }
```

libs/sdk-js/src/client.ts
```diff
@@ -39,6 +39,75 @@ import { getEnvironmentVariable } from "./utils/env.js";
...
+    | undefined
+    | null
+  )[]
+) {
+  const outputHeaders = new Headers();
+  for (const headers of headerObjects) {
+    if (!headers) continue;
+    for (const [name, value] of iterateHeaders(headers)) {
+      if (value === null) outputHeaders.delete(name);
+      else outputHeaders.append(name, value);
```
hntrl, line 101: `headers.append` has casing considerations built in (we <3 web standards)

hinthornw: oh fun!

PR #4953, opened 2025-06-04
docs: enhance PostgresSaver connection requirements explanation

### LINKED ISSUE #4937: DOC: Langgraph checkpoint postgres setup is incomplete
### Issue with current documentation:

I have recently started using the langgraph postgres checkpointer. 

When reviewing the documentation, I've noticed the following:
https://github.com/langchain-ai/langgraph/blob/3a74514bc3aaa4fc186553f05901d280effc3c39/libs/checkpoint-postgres/README.md?plain=1#L15

I understand why the `autocommit=True` parameter is currently requested from a pre-configured connection provided for the PostgresSaver class as it comes handy and avoid problems with the `.setup()` method not committing the checkpoint tables to the DB. However, `row_factory=dict_row` is also "necessary" according to the README, but there should be more detail into why this is necessary.   
I've set up my current project with the default [`tuple_row()`](https://www.psycopg.org/psycopg3/docs/api/connections.html) and it's working properly, but changing this would not be backwards-compatible right now. Hence, I'd need more justification to see why I'd need to switch to this setting. I think it would generally benefit other developers. 

### Idea or request for content:

Detail in the PostgresSaver class or elsewhere in the repo why do we need to use `autocommit=True` and `row_factory=dict_row`

YassinNouh21:
## Description
Enhances the PostgreSQL checkpointer documentation to address issue #4937 by providing detailed explanations of why `autocommit=True` and `row_factory=dict_row` parameters are required when manually creating database connections.

## Problem
The current documentation states that these parameters are required but doesn't explain **why**, leading to developer confusion. Users were questioning the necessity of `row_factory=dict_row` and reporting that their code "works fine" with the default `tuple_row`, not realizing they hadn't hit the problematic code paths yet.

## Changes Made
- ✨ Added detailed explanation of why `autocommit=True` is required
  - Required for the `.setup()` method to properly commit checkpoint tables to the database
  - Without this, table creation may not be persisted
  
- ✨ Added detailed explanation of why `row_factory=dict_row` is required
  - PostgresSaver implementation uses dictionary-style row access (`row["column_name"]`)
  - Default `tuple_row` factory only supports index-based access (`row[0]`)
  - Missing this causes `TypeError: tuple indices must be integers or slices, not str`
  
- ✨ Added example of incorrect usage showing the specific error that occurs
- ✨ Improved developer experience by providing clear technical justification

## Technical Background
The PostgresSaver source code has multiple instances of dictionary-style row access:
- `row["v"]` in the `setup()` method (line 89)
- `value["thread_id"]`, `value["checkpoint_id"]` in the `list()` method
- `value["checkpoint"]`, `value["metadata"]` in the `get_tuple()` method

This pattern **only works** with `dict_row` factory, not the default `tuple_row`.

## Testing
- [x] Verified documentation renders correctly in Markdown
- [x] Confirmed code examples are syntactically correct
- [x] Validated technical explanations against source code

## Fixes
Closes #4937

### DIFF
--- a/libs/checkpoint-postgres/README.md
+++ b/libs/checkpoint-postgres/README.md
@@ -13,6 +13,20 @@ By default `langgraph-checkpoint-postgres` installs `psycopg` (Psycopg 3) withou
 
 > [!IMPORTANT]
 > When manually creating Postgres connections and passing them to `PostgresSaver` or `AsyncPostgresSaver`, make sure to include `autocommit=True` and `row_factory=dict_row` (`from psycopg.rows import dict_row`). See a full example in this [how-to guide](https://langchain-ai.github.io/langgraph/how-tos/persistence_postgres/).
+>
+> **Why these parameters are required:**
+> - `autocommit=True`: Required for the `.setup()` method to properly commit the checkpoint tables to the database. Without this, table creation may not be persisted.
+> - `row_factory=dict_row`: Required because the PostgresSaver implementation accesses database rows using dictionary-style syntax (e.g., `row["column_name"]`). The default `tuple_row` factory returns tuples that only support index-based access (e.g., `row[0]`), which will cause `TypeError` exceptions when the checkpointer tries to access columns by name.
+>
+> **Example of incorrect usage:**
+> ```python
+> # ❌ This will fail with TypeError during checkpointer operations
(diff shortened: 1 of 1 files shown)


sydney-runkle (approved):
Thanks!

PR #5041, opened 2025-06-10
Disable values update on use stream in interrupt events

OfirTeneJuno:
Since the change of the interrupt, that send the interrupt event as values event instead of update,
The messages update has issues after interrupt.
The interrupt data does not include the graph state, and the structure of the interrupt values is nested in the interrupt data.
Which causes the getMessages function to return empty array after each interrupt event.
https://github.com/langchain-ai/langgraphjs/pull/1244 the pr changing the interrupt event type.

related issue:
https://github.com/langchain-ai/langgraphjs/issues/1273


dqbd:
Thank you!

PR #5045, opened 2025-06-10
docs: fixed a wrong import in persistence docs

SimonFrank14:
Hi, I think I found an error in the docs on the persistence page [Click here for reference in docs](https://langchain-ai.github.io/langgraph/how-tos/persistence/#use-in-production_1) and I fixed it


sydney-runkle (approved):
Thanks!

PR #5103, opened 2025-06-13
Restore compatibility with custom checkpointer classes created in prior versions

nfcampos:
- Ensure existing custom checkpointer classes are compatible with new langgraph-checkpoint release


libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -116,8 +118,19 @@ class BaseCheckpointSaver(Generic[V]):
...
 
+    _get_next_version_legacy: ClassVar[bool] = False
+    """Flag indicating if get_next_version method is legacy (takes two parameters)."""
+
     serde: SerializerProtocol = JsonPlusSerializer()
 
+    def __init_subclass__(cls) -> None:
+        cls._get_next_version_legacy = (
+            len(signature(cls.get_next_version).parameters) > 2  # self + current
+            if hasattr(cls, "get_next_version")
```
hinthornw, line 129: When would nit not have `get_next_version` at all?

nfcampos: Ah I guess always inherits?

PR #5105, opened 2025-06-13
Restore shallow checkpointer

nfcampos:
- This should definitely be removed soon, but let's give people more time to update


satyaprakash1729:
Hi @nfcampos, Thanks for maintaining the postgres checkpointer library. It's super useful. 
I've a question regarding this PR - without shallow checkpointer we would unnecessarily store lot of unwanted/duplicate information right? For example, last/recent checkpointer state is good enough since it contains all the information right?
What's the rationale for removing this? (I understand that this PR is temporarily enabling, but eventually going to be removed)
We're noticing that the Postgres is filling up quite fast. Thanks in advance!

PR #5106, opened 2025-06-14
Add migration for pending_sends

nfcampos:
- Checkpoints saved on older versions of langgraph will be compatible with langgraph 0.5 and 1.0


libs/langgraph/tests/memory_assert.py
```diff
@@ -24,6 +26,28 @@ def dumps_typed(self, obj: Any) -> tuple[str, bytes]:
         return "type", obj
 
 
+class MemorySaverNeedsPendingSendsMigration(BaseCheckpointSaver):
+    def __init__(self) -> None:
+        self.saver = InMemorySaver()
+
+    def __getattribute__(self, name):
```
nfcampos, line 33: I am very bad

hinthornw: ha nice. it's tests anyhow

hinthornw (approved):

PR #5111, opened 2025-06-14
feat(react): enhance useStream with initialValues, newThreadId, and onStop callback for improved UX

MauritsBrinkman:
## Summary
Adds three new options to the `useStream` React hook to improve user experience in common chat application scenarios:

- **`initialValues`**: Display cached thread data immediately while history loads
- **`newThreadId`**: Enable optimistic thread creation with predetermined IDs  
- **`onStop`**: Customize UI behavior when streams are stopped/cancelled

### `initialValues` Option
- Shows cached thread data instantly when navigating to existing threads
- Improves perceived performance by eliminating loading states for cached content
- Values priority: `streamValues > initialValues > historyValues`
- Backward compatible

### `newThreadId` Option  
- Enables optimistic UI patterns for thread creation
- Prevents 404 errors when automatically fetching thread history when navigating to routes before thread exists
- Allows predetermined thread IDs for better URL management in practice (e.g. when navigating to `/[threadId]` routes when submitting a message)
- Works with `threadId: null` for new thread creation

### `onStop` Callback
- Enables custom behavior when streams are stopped or cancelled
- Provides immediate local state updates via `mutate` function
- Particularly useful for handling loading states in generative UI components
- Prevents loading components from remaining in infinite loading state after cancellation
- Supports optional server thread state updates

```typescript
const stream = useStream({
  assistantId: "my-assistant",
  onStop: async ({ mutate }) => {
    // Immediate UI update - stop loading components
    mutate((prev) => ({
      ...prev,
      ui: prev.ui?.map(component =>
        component.props?.isLoading
          ? {
              ...component,
              props: {
                ...component.props,
                isLoading: false,
                isStopped: true
              }
            }
          : component
      )
    }));

    // Optional server thread state update
    if (stream.threadId) {
      await stream.client.threads.updateState(stream.threadId, {
        values: { ui: prev.ui }
      });
    }
  }
});
```

## Documentation & Testing
- Added comprehensive documentation with examples for all features
- Test coverage for `initialValues`, `newThreadId`, and `onStop` callback
- Updated React integration guide
- Clean test implementation with proper cleanup

## Use Cases
- **Cached display**: Show cached thread data while server loads
- **Optimistic navigation**: Navigate to `/threads/[id]` before thread creation
- **Better UX**: Eliminate loading states and navigation delays
- **Graceful cancellation**: Handle stopped streams with custom UI updates
- **Generative UI**: Properly manage loading states in dynamic components

## Breaking Changes
None - all three options are optional and fully backward compatible.

### DIFF
--- a/libs/sdk-js/src/react/stream.tsx
+++ b/libs/sdk-js/src/react/stream.tsx
@@ -512,6 +512,31 @@ export interface UseStreamOptions<
   onDebugEvent?: (data: DebugStreamEvent["data"]) => void;
 
   /**
+   * Callback that is called when the stream is stopped by the user.
+   * Provides a mutate function to update the stream state immediately
+   * without requiring a server roundtrip.
+   * 
+   * @example
+   * ```typescript
+   * onStop: ({ mutate }) => {
+   *   mutate((prev) => ({
+   *     ...prev,
+   *     ui: prev.ui?.map(component => 
+   *       component.props.isLoading
+   *         ? { ...component, props: { ...component.props, stopped: true, isLoading: false }}
+   *         : component
+   *     )
+   *   }));
+   * }
+   * ```
+   */
+  onStop?: (options: {
+    mutate: (
+      update: Partial<StateType> | ((prev: StateType) => Partial<StateType>),
+    ) => void;
+  }) => void;
+
+  /**
    * The ID of the thread to fetch history and current values from.
    */
   threadId?: string | null;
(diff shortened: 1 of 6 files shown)


libs/sdk-js/src/react/stream.tsx
```diff
@@ -997,23 +1053,25 @@ export function useStream<
       // Assumption: we're setting the initial value
       // Used for instant feedback
       setStreamValues(() => {
-        const values = { ...historyValues };
+        const baseValues = options.initialValues ? { ...historyValues, ...options.initialValues } : historyValues;
```
dqbd: My take is that we should clear out the `initialValues` when we receive new `values` rather than always keeping those values around.

libs/sdk-js/src/react/stream.tsx
```diff
@@ -997,23 +1053,25 @@ export function useStream<
...
 
-        return values;
+        return baseValues;
       });
 
       let usableThreadId = threadId;
       if (!usableThreadId) {
-        const thread = await client.threads.create();
+        const thread = await client.threads.create({
+          threadId: options.newThreadId,
```
dqbd: I think we want to have this as an srgument for `submit` instead?

dqbd:
Thank you so much for the PR @MauritsBrinkman! Will rebase and work on top of this PR 🙂

PR #5128, opened 2025-06-17
docs: Add conditional js/python rendering

eyurtsev:
# Overview

Adding conditional rendering logic to co-locate js and python documentation.

* `:::` conditional syntax can be used to switch between python only or js only content.
* Contains simple unit tests for `:::`
* PR adds set up for a way to implement a context switch between languages, but it will not be enabled until JS content is merged in.
* Contains a script that can add javascript documentation 

Implementation of: https://github.com/langchain-ai/langgraph/pull/5118

## Example

Example of conditional rendering / compilation.

```markdown

### Config (static context)

Config is for immutable data like user metadata or API keys. Use
when you have values that don't change mid-run.

Specify configuration using a key called **"configurable"** which is reserved
for this purpose:


:::python

This content will only be rendered for the python site.
:::

:::js
this content will only be rendered for the js / ts site.
:::

```

### DIFF
--- a/docs/_scripts/generate_llms_text.py
+++ b/docs/_scripts/generate_llms_text.py
@@ -3,19 +3,21 @@
 import asyncio
 import glob
 import os
-from typing import TypedDict, List, Optional
-import pydantic
 import re
-from pydantic import BaseModel, Field
-from langchain_core.rate_limiters import InMemoryRateLimiter
+from typing import TypedDict, List, Optional
 
 import yaml
 from langchain.chat_models import init_chat_model
+from langchain_core.rate_limiters import InMemoryRateLimiter
 from mkdocs.structure.files import File
 from mkdocs.structure.pages import Page
+from pydantic import BaseModel, Field
 from yaml import SafeLoader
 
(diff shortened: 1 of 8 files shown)


docs/_scripts/generate_llms_text.py
```diff
@@ -3,19 +3,21 @@
 import asyncio
```
eyurtsev, line 3: formatting with ruff -- but unrelated change

eyurtsev:
cc @hntrl this is supporting code. It will result in a no-op on existing documentation (with exception of injection of unused js, the html won't be included yet).

docs/_scripts/link_map.py
hntrl, line 1: Maybe we can have a route that transpiles typedoc refs into this format and have that accessible on the js refs site? then either retroactively apply the links in the browser (probably hard to do) or at build time (can fall out of sync).

eyurtsev: I have a suspicion that we won't need automation here -- there's probably no more than ~20-30 links that are referenced from docs manually.

Issue #5137, opened 2025-06-18
Incorrect Blockbuster configuration prevents Trustcall extractors from being invoked async


mobiware:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
import os
import asyncio
import dotenv
from blockbuster import blockbuster_ctx

from typing import List, Optional

from pydantic import BaseModel
from langchain_openai import ChatOpenAI
from trustcall import create_extractor

dotenv.load_dotenv()


class OutputFormat(BaseModel):
    preference: str
    sentence_preference_revealed: str


class TelegramPreferences(BaseModel):
    preferred_encoding: Optional[List[OutputFormat]] = None
    favorite_telegram_operators: Optional[List[OutputFormat]] = None
    preferred_telegram_paper: Optional[List[OutputFormat]] = None


class MorseCode(BaseModel):
    preferred_key_type: Optional[List[OutputFormat]] = None
    favorite_morse_abbreviations: Optional[List[OutputFormat]] = None


class Semaphore(BaseModel):
    preferred_flag_color: Optional[List[OutputFormat]] = None
    semaphore_skill_level: Optional[List[OutputFormat]] = None


class TrustFallPreferences(BaseModel):
    preferred_fall_height: Optional[List[OutputFormat]] = None
    trust_level: Optional[List[OutputFormat]] = None
    preferred_catching_technique: Optional[List[OutputFormat]] = None


class CommunicationPreferences(BaseModel):
    telegram: TelegramPreferences
    morse_code: MorseCode
    semaphore: Semaphore


class UserPreferences(BaseModel):
    communication_preferences: CommunicationPreferences
    trust_fall_preferences: TrustFallPreferences


class TelegramAndTrustFallPreferences(BaseModel):
    pertinent_user_preferences: UserPreferences


conversation = """Operator: How may I assist with your telegram, sir?
Customer: I need to send a message about our trust fall exercise.
Operator: Certainly. Morse code or standard encoding?
Customer: Morse, please. I love using a straight key.
Operator: Excellent. What's your message?
[...]

PR #5139, opened 2025-06-18
Fix: Correct Blockbuster config for lock.acquire

shaktiman101:
This commit fixes issue #5137.

The Blockbuster configuration in `langgraph_runtime_inmem/queue.py` was using "lock.acquire" to disable blocking call detection for thread locks. The correct identifier for this is "threading.Lock.acquire".

This change updates the configuration to use the correct identifier, preventing erroneous `BlockingError` exceptions when using features like Trustcall extractors asynchronously.


hinthornw:
Oh this is a good fix! I'll replicate it  to the package. Thank you for your contribution!

hinthornw:
Should be live in langgraph-runtime-inmem 0.3.2!
https://pypi.org/project/langgraph-runtime-inmem/

I'll close this PR since the change was made to that repo - If you have a twitter handle or something, I'd be happy to give you a shoutout for the contribution (in lieu of credit on merging this PR, since the target repo isn't open source)

PR #5158, opened 2025-06-22
feat: post hook - create_react_agent support identical tool call ids

### LINKED ISSUE #5157: create_react_agent with `post_model_hook` - early stop on non unique tool call id
### Checked other resources

### Example Code

```python
# New test added to test suite

def test_post_model_hook_identical_tool_call_ids() -> None:
    def get_weather():
        """Get the weather"""
        return "The weather is sunny and 75°F."

    # two identical calls
    tool_calls = [[{"args": {}, "id": "1", "name": "get_weather"}]] * 2

    model = FakeToolCallingModel(
        tool_calls=tool_calls,
        # Only two tool calls, then, done.
        max_generations_with_tools=2,
    )

    class State(AgentState):
        called_count: Annotated[int, add]

    def post_model_hook(state: State) -> Union[dict[str, bool], Command]:
        # Track post model hook calls
        return {"called_count": 1}

    agent = create_react_agent(
        model,
        [get_weather],
        post_model_hook=post_model_hook,
        state_schema=State,
    )

    assert "post_model_hook" in agent.nodes

    response = agent.invoke(
        {"messages": [HumanMessage("What's the weather?")], "called_count": 0}
    )
    assert response["called_count"] == 3
```

```shell
# None - the graph ends prematurely with no error
```

### Description

@sydney-runkle 

I have stumbled upon an edge case where Gemini Flash 2 would use the same tool call ids on multiple consecutive AI messages within a thread (not within a parallel tool call).

The `post_model_hook` router would not handle this case properly - it would "detect" the first tool call response as if the last AI message with the same tool call id was already fulfilled - and will end the Graph.

Although I assumed this situation should not happen, [Gemini docs](https://ai.google.dev/api/caching#FunctionCall) state the id is optional, and do not guarantee uniqueness within the same thread.

I have implemented a simple test to cover this case, and implemented a reasonable fix for your review on a separate PR.

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000
> Python Version:  3.11.9 (main, Aug 14 2024, 04:17:21) [Clang 18.1.8 ]

Package Information
-------------------
> langchain_core: 0.3.60
> langsmith: 0.3.42
> langgraph_sdk: 0.1.70

> langserve

Other Dependencies
------------------
> httpx: 0.28.1
> httpx>=0.25.2: Installed. No version info available.
> jsonpatch<2.0,>=1.33: Installed. No version info available.
> langsmith-pyo3: Installed. No version info available.
> langsmith<0.4,>=0.1.126: Installed. No version info available.
> openai-agents: Installed. No version info available.
> opentelemetry-api: Installed. No version info available.
> opentelemetry-exporter-otlp-proto-http: Installed. No version info available.
> opentelemetry-sdk: Installed. No version info available.
> orjson: 3.10.18
> orjson>=3.10.1: Installed. No version info available.
> packaging: 24.2
> packaging<25,>=23.2: Installed. No version info available.
> pydantic: 2.11.4
> pydantic>=2.7.4: Installed. No version info available.
> pytest: 8.3.5
> PyYAML>=5.3: Installed. No version info available.
> requests: 2.32.3
> requests-toolbelt: 1.0.0
> rich: Installed. No version info available.
> tenacity!=8.4.0,<10.0.0,>=8.1.0: Installed. No version info available.
> typing-extensions>=4.7: Installed. No version info available.
> zstandard: 0.23.0

yoadsn:
@sydney-runkle 
This fixes #5157.

While working on the fix/tests I have observed that `FakeToolCallingModel` had no way to "stop producing tool calls" which was working only because the bug existed. (Think about it this way, it would keep producing the same tool call id - over and over, and the router would always assume the tool call is completed, and thus would end the execution).

but I needed to reproduce the above case, without hitting the max graph recursion limit - so I added an optional arg to allow max generations with tool calls.

After fixing the bug, I observed the test `test_post_model_hook_with_structured_output` fails on max graph recursion due to the above reason, so I took the liberty of fixing that test, resetting the agent/model state such that the two tested modes (invoke vs stream) are independent. I suggest splitting them to two tests though.

As an aside - I do wonder - why even test for tool call completions in the `post_model_hook` router?
And why does the logic assumed tools->post_model_hook is a possible path?

This probably complicated the logic and is not expected to happen - right?

Without this assumption it could be as simple as the logic in `should_continue` - if there are any tool_calls route to `tools`.
Wdyt?

### DIFF
--- a/libs/prebuilt/tests/model.py
+++ b/libs/prebuilt/tests/model.py
@@ -41,7 +41,10 @@ class FakeToolCallingModel(BaseChatModel):
     ) -> ChatResult:
         """Top Level call"""
         messages_string = "-".join([m.content for m in messages])
-        allow_tool_calls = self.max_generations_with_tools is None or self.index < self.max_generations_with_tools
+        allow_tool_calls = (
+            self.max_generations_with_tools is None
+            or self.index < self.max_generations_with_tools
+        )
         tool_calls = (
             self.tool_calls[self.index % len(self.tool_calls)]
             if self.tool_calls and allow_tool_calls
--- a/libs/prebuilt/tests/test_react_agent.py
+++ b/libs/prebuilt/tests/test_react_agent.py
@@ -2,6 +2,7 @@ import dataclasses
 import inspect
 import json
 from functools import partial
+from operator import add
 from typing import (
     Annotated,
     List,
@@ -10,7 +11,6 @@ from typing import (
     TypeVar,
     Union,
 )
-from operator import add
 
 import pytest
 from langchain_core.messages import (
@@ -1432,7 +1432,8 @@ def test_post_model_hook_with_structured_output() -> None:
 
     expected_structured_response = WeatherResponse(temperature=75)
     model = FakeToolCallingModel(
-        tool_calls=tool_calls, structured_response=expected_structured_response,
+        tool_calls=tool_calls,
+        structured_response=expected_structured_response,
         max_generations_with_tools=1,
     )
 
@@ -1463,7 +1464,8 @@ def test_post_model_hook_with_structured_output() -> None:
     # Test with stream - reset state
 
     model = FakeToolCallingModel(
-        tool_calls=tool_calls, structured_response=expected_structured_response,
+        tool_calls=tool_calls,
+        structured_response=expected_structured_response,
         max_generations_with_tools=1,
     )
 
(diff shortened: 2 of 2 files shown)


casparb:
Hey @yoadsn, thank you for your contribution. We'll soon be releasing our new prebuilt agent [`create_agent()`](https://docs.langchain.com/oss/python/langchain/agents) with the upcoming v1.0 LangChain, which will depricate `create_react_agen()t`. Pre-/post-model hooks will be replaced by our new [middleware abstractions](https://docs.langchain.com/oss/python/langchain/middleware), so we aren’t taking new features on that surface. I'm sorry we didn't get to you earlier!

PR #5188, opened 2025-06-24
docs: cli data storage

lc-arjun:


docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
...
+
+This document describes how [LangGraph CLI](../../concepts/langgraph_cli.md) handles data storage, collection, and privacy. It covers what data is stored locally versus remotely, security measures, and privacy controls.
+
+## Overview
+
+LangGraph CLI is designed with privacy-first principles:
+
+- **Local-first storage**: Your data stays on your machine by default
+- **Minimal telemetry**: Optional analytics that can be disabled
+- **Encryption support**: Optional encryption for sensitive data
```
hinthornw: Does this just mean the telemetry uses Tls? Seems distracting

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
+# CLI Data Storage and Privacy
+
+This document describes how [LangGraph CLI](../../concepts/langgraph_cli.md) handles data storage, collection, and privacy. It covers what data is stored locally versus remotely, security measures, and privacy controls.
+
+## Overview
+
+LangGraph CLI is designed with privacy-first principles:
+
+- **Local-first storage**: Your data stays on your machine by default
+- **Minimal telemetry**: Optional analytics that can be disabled
```
hinthornw: Redundant (optional and can be disabled) - and seems odd to put this in a "design" - it's not really deserving of a bullet point for CLI design imo

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Data Storage Locations
+
+### Local Storage (Default)
+
+All application data is stored locally unless explicitly configured otherwise:
+
+- **Checkpoints**: Graph state snapshots for debugging and development
+- **Thread data**: Conversation and execution history
+- **Graph state**: Node outputs and intermediate results
```
hinthornw: Graph state IS checkpoints

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
...
+- **Local-first storage**: Your data stays on your machine by default
+- **Minimal telemetry**: Optional analytics that can be disabled
+- **Encryption support**: Optional encryption for sensitive data
+- **Clear opt-in controls**: External services require explicit configuration
+
+## Data Storage Locations
+
+### Local Storage (Default)
+
+All application data is stored locally unless explicitly configured otherwise:
```
hinthornw: This only applies to dev and up . If it's deployed then the data resides on database the user specifies (or which is provisioned for them in saas)

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
...
+
+- **Checkpoints**: Graph state snapshots for debugging and development
+- **Thread data**: Conversation and execution history
+- **Graph state**: Node outputs and intermediate results
+- **Memory/Store data**: Information that persists across threads
```
jdrogers940: I feel like this detail is unnecessary. Would just link application data to relevant docs if needed.

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
...
+
+**Storage locations:**
+
+- **Development mode**: Local directory (configurable)
+- **Docker deployment**: Local Docker volumes
+- **LangGraph Platform**: Managed database infrastructure
+
+### Remote Storage (Optional)
+
+Remote storage only occurs when explicitly configured:
```
jdrogers940: This is unclear to me - do we require opt-out for our own telemetry? If so, I think we can't say this.

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
...
+- Feature usage statistics
+- Browser type and version
+- Screen resolution and viewport size
+
+**Not collected:**
+
+- Application data or code
+- Sensitive configuration details
+
+**To disable:** Use browser privacy settings or ad blockers to prevent analytics collection
```
jdrogers940: Do we want to give some examples?

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,121 @@
+# CLI Data Storage and Privacy
+
+This document describes how [LangGraph CLI](../../concepts/langgraph_cli.md) handles data storage, collection, and privacy. It covers what data is stored locally versus remotely, security measures, and privacy controls.
```
hinthornw: We need to be clear exactly what this is covering. In this doc, you also sort of talk about the in-mem server, the local single-container server (`up`) and then sort of mention others.

Need to be up front about what this does and does not cover.

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,54 @@
+# Data Storage and Privacy
+
+This document describes how data is processed in the LangGraph CLI and the LangGraph Server for both the in-memory server (`langgraph dev`) and the local docker server (`langgraph up`). It also describes what data is tracked when interacting with the hosted LangGraph Studio frontend.
+
+## CLI
```
lc-arjun, line 5: For each section probably worth adding a link to "learn more" and linking to the overview page?

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,54 @@
...
+
+By default, calls to most CLI commands log a single analytics event to help improve the CLI experience. Each telemetry event contains the calling process's OS, OS version, Python version, the CLI version, the command name (`dev`, `up`, `run`, etc.), and booleans representing whether a flag was passed to the command. You can see the full analytics logic [here](https://github.com/langchain-ai/langgraph/blob/main/libs/cli/langgraph_cli/analytics.py). 
+
+You can disable all CLI telemetry by setting `LANGGRAPH_CLI_NO_ANALYTICS=1`.
+
+## LangGraph Server (in-memory & docker)
+
+The LangGraph server provides a durable execution runtime that relies on persisting checkpoints of your application state, long-term memories, thread metadata, assistants, and similar resources to a database or file system. Unless you have deliberately customized the storage location, this information is either stored in a postgres database (for `langgraph up` and in deployments) or written to local disk (for `langgraph dev`).
```
lc-arjun: nit: header order is in-memory then docker but this sentence order is docker then in-memory.

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,54 @@
...
+
+### In-memory development server (`langgraph dev`)
+
+`langgraph dev` runs an in-memory variant of the server as a single Python process. It saves all checkpointing & memory data to disk within a `.langgraph` directory in the current working directory. Apart from the telemetry data described in [CLI only](#cli-only), no data leaves the machine unless you have enabled tracing or your graph code explicitly contacts an external service.
```
lc-arjun: `.langgraph_api`

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,54 @@
...
+
+LangSmith tracing is disabled by default, but you can opt-in to tracing by setting `LANGSMITH_TRACING=true` and providing a valid `LANGSMITH_API_KEY` either in your terminal or in the `.env` file indicated by your `langgraph.json` configuration file.
```
lc-arjun: technically isn't it the case that tracing is enabled if `langsmith_api_key` is defined and `langsmith_tracing` is not `false`? You don't need to explicitly set  it to `true`

lc-arjun: assuming this is true, i think we should move the tracing info up a level as it applies to both cases of the server and can be consolidated into one section

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,54 @@
...
+
+Additional API calls are made to confirm that the server has a valid license and to track the number of executed runs and tasks. Periodically, the API server validates the provided license key (or API key).
+
+In this configuration, LangSmith tracing is enabled by default. To opt-out, set `LANGSMITH_TRACING=false` in your environment.
+
+If you've disabled tracing, no user data is persisted externally unless your graph code explicitly contacts an external service.
+
+## Studio
+
+LangGraph Studio is a browser client for interacting with your LangGraph server. Though it caches responses in your browser's local storage for performance, it does not trace or persist your private data.
```
lc-arjun: curious what you mean by this part -- we don't cache model responses in local storage. we do store some assistant ids so we can load things faster, and we store custom headers, but no application data

docs/docs/cloud/concepts/data_storage_and_privacy.md
```diff
@@ -0,0 +1,54 @@
...
+
+In summary, you can opt-out of server-side telemetry by turning off CLI analytics and disabling tracing.
+
+| Variable                       | Purpose                   | Default                          |
+| ------------------------------ | ------------------------- | -------------------------------- |
+| `LANGGRAPH_CLI_NO_ANALYTICS=1` | Disable CLI analytics     | Analytics enabled                |
+| `LANGSMITH_API_KEY`            | Enable LangSmith tracing  | Tracing disabled                 |
+| `LANGSMITH_TRACING=false`      | Disable LangSmith tracing | Depends on environment           |
+
+You can opt-out of browser analytics by setting the `DoNotTrack` header in your browser.
```
lc-arjun: just checked and this doesn't work. i know arc browser disables these requests by default, although i'm not entirely sure what the mechanism is behind this (i assumed it was their built in ad blocker but i just tried with a chrome ad blocker extension and that didn't work so not sure exactly what actionable advice to give user here)

docs/docs/cloud/concepts/data_storage_and_privacy.md
hinthornw: Thanks for clarifying! I had been assuming some more caching of the actual assistant metadata, etc. but seems I was mistaken. Will remove

docs/docs/cloud/concepts/data_storage_and_privacy.md
hinthornw: I'll just remove

docs/docs/cloud/concepts/data_storage_and_privacy.md
hinthornw: I like the suggestion of moving it up! I'm just gonna make our messaging simpler and say "LangSmith tracing may be enabled to facilitate faster development and observability in production. To disable, set LANGSMITH_TRACING=false"

docs/docs/cloud/concepts/data_storage_and_privacy.md
hinthornw: oops!

docs/docs/cloud/concepts/data_storage_and_privacy.md
lc-arjun: nice, clean

PR #5195, opened 2025-06-25
Remove unused branch in update_state

nfcampos:
- update_state(None, as_node=None) was not tested/used, and was too similar in behavior to __copy__ / __end__, so removing it


sydney-runkle (approved):
Fantastic!

PR #5197, opened 2025-06-25
feat: crons sorting sdk

lc-arjun:


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -2532,6 +2533,8 @@ async def search(
         thread_id: str | None = None,
         limit: int = 10,
         offset: int = 0,
+        sort_by: CronSortBy | None = None,
```
hinthornw, line 2536: what about the SyncClient?

lc-arjun: ah good callout. realizing i never did this for assistants or threads search either oops

lc-arjun: will add for all 3

Issue #4596, opened 2025-05-08
Authorization header missing on /threads/{thread_id}/history request when using useStream with defaultHeaders


zyoung1212:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
const thread = useStream({
  apiUrl: 'http://localhost:2024',
  assistantId: "app",
  threadId: threadId,
  onThreadId: setThreadId,
  messagesKey: "messages",
  defaultHeaders: {
    Authorization: `Bearer ${token}`,
  }
});
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

Hi LangGraph team,

I'm currently developing a LangGraph application using the officially recommended React frontend from the guide here:
https://langchain-ai.github.io/langgraph/cloud/how-tos/use_stream_react

I have implemented authentication and am passing a token through the Authorization header.

The token is successfully included when sending messages (e.g., to /stream、/threads). However, I noticed that when a request is sent to /threads/{thread_id}/history, the token is not included in the headers, resulting in a 403 error from my backend.

### System Info

    "@langchain/core": "^0.3.49",
    "@langchain/langgraph-sdk": "^0.0.74",

> langgraph_api: 0.1.17
> langchain_core: 0.3.56
> langchain: 0.3.24
> langgraph_sdk: 0.1.63

zyoung1212:
<img width="1105" alt="Image" src="https://github.com/user-attachments/assets/dbbc77bd-84c6-42d5-845d-25efcf07ceb7" />

<img width="1010" alt="Image" src="https://github.com/user-attachments/assets/aa8a140a-ebfa-4f38-92d2-ff26a47a5426" />

zyoung1212:
sorry, it's my bug.
Fixed code example:
`export default function Home() {
  // 立刻同步读 localStorage，确保首帧就有 Token
  const initialToken =
    typeof window !== "undefined" ? localStorage.getItem("token") : null;
  const [token, setToken] = useState(initialToken);

  // 如果还没拿到 token，先别渲染 useStream
  if (!token) return <p>Loading...</p>;

  const thread = useStream({
[...]

PR #5208, opened 2025-06-26
Allow React useStream client changes for thread/{id}/history requests

Kreezag:
# Update the useThreadHistory logic to respond to stream client  change

## Problem definition

The useStream is a React hook that should react to component rerenders (ex, state changes).

React useStream encapsulates multiple calls, like `/threads`, `/threads/{id}/runs/stream`, `/threads/{id}/runs/history`.

First and second endpoints respond to options changes (defaultHeaders and apiKey in my case), but `/threads/{id}/runs/history` is not. Even if threadId is changed, new option changes are unavailable at the history level.
I hope this fix will solve this behavior

This fix is working on my local machine, but I'm not sure if the stable history client is a necessary logic. If it is true, I'd like to see a reason description in the code or the docs. The actual [example section in documentation](https://langchain-ai.github.io/langgraph/cloud/how-tos/use_stream_react/#example) doesn't respond to my question.

--- 

## References

I'm not sure that it is a bug, so I prepared [discussion](https://github.com/langchain-ai/langgraph/discussions/5202). It includes more description about my case.

I found a similar issue to this behavior: https://github.com/langchain-ai/langgraph/issues/4596

### DIFF
--- a/libs/sdk-js/src/react/stream.tsx
+++ b/libs/sdk-js/src/react/stream.tsx
@@ -324,6 +324,7 @@ function useThreadHistory<StateType extends Record<string, unknown>>(
   const fetcher = useCallback(
     (
       threadId: string | undefined | null,
+      client: Client,
     ): Promise<ThreadState<StateType>[]> => {
       if (threadId != null) {
         return fetchHistory<StateType>(client, threadId).then((history) => {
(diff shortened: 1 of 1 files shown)


dqbd:
Hello! Thank you for the PR, although I think here in particular we don't want to pass the `client` directly to `useEffect`, since it is possible to accidentally pass in a new instance of a client with every render, causing infinite fetching of `/history`.

Closing this in favour of #5240

PR #5217, opened 2025-06-26
Add test for reducer exceptions

nfcampos:


libs/langgraph/tests/test_pregel_async.py
```diff
@@ -106,13 +107,29 @@ class FaultyVersionCheckpointer(InMemorySaver):
         def get_next_version(self, current: Optional[int], channel: None) -> int:
             raise ValueError("Faulty get_next_version")
 
+    class FaultySerializer(JsonPlusSerializer):
+        def dumps_typed(self, obj: Any) -> tuple[str, bytes]:
```
eyurtsev, line 111: I remember the traceback message on failed serialization in the past was difficult to act on since it was not clear what failed during serialization or what action to take to fix. If still an issue, could be nice to add a more informative error message on failure

PR #5224, opened 2025-06-27
Draft : a filesystem checkpointer

eric-burel:
Hi folks,

I am experimenting with a new filesystem checkpointer, which I don't think exists in the codebase and didn't found in the community.

If there is already such a checkpointer, feel free to dismiss.

I am particularly interested into such a checkpointer because it can be coupled with Google Colab capability to mount Google Drive folders, so it would allow to save the agent state in a Google Drive folder.

This is draft pull-request.

**Most of the changes at this point are AI-generated!** (+1 hour of proof reading and tweaking things to understand what happens) 

I am an experienced developer with some Python background but I don't practice this language, so I've used this PR as a practice and I had to figure new concepts I wasn't aware about (`defaultdict` for handling threads that do not exist yet, stacking exits, the `PersistentDict` implementation etc.).

- At this point, it doesn't seem to actually create the file. I've pinpointed the issue to the fact that dumping files only happens in a `PersistentDict` close event. But I fail to understand when this happens, and whether I should dump the file more frequently eg on each "put" operation.
- I think the stack exit calls should be reenabled in __enter, __exit and so on (they are present in the MemorySaver), but I am not 100% sure what they do so I'd be eager to get some confirmation (perhaps this explains why the file is not dumped)
- The PR definitely lacks a unit test, I've just tested the code in a notebook by copy-pasting so far.
- PersistentDict could be factored with `InMemorySaver` if we happen to reuse the exact same structure

So I hope this could help as a basis for a file system checkpointer, and I'd be eager to keep iterating on it with light directions from experienced contributors.

Also let me know if you prefer using a discussion rather than a pull request.

Update: after more exploration, I feel like `PersistentDict` is meant to be used as a `InMemorySaver` "factory" parameter, to allow dumping the data on save? So could `InMemorySaver` + `PersistentDict` act as a `FileSystemSaver`?

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/fs/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/fs/__init__.py
@@ -1,0 +1,605 @@
+from __future__ import annotations
+
+import logging
+import os
+import pickle
+import random
+import shutil
+from collections import defaultdict
+from collections.abc import AsyncIterator, Iterator, Sequence
+from contextlib import AbstractAsyncContextManager, AbstractContextManager, ExitStack
+from types import TracebackType
+from typing import Any
+
+from langchain_core.runnables import RunnableConfig
+
+from langgraph.checkpoint.base import (
+    WRITES_IDX_MAP,
+    BaseCheckpointSaver,
+    ChannelVersions,
+    Checkpoint,
+    CheckpointMetadata,
+    CheckpointTuple,
+    SerializerProtocol,
+    get_checkpoint_id,
+    get_checkpoint_metadata,
+)
+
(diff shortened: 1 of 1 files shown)


eric-burel:
Closing the PR as I figured an alternative: using `SqliteSaver` works as a filesystem saver.
Demoes here:
- https://colab.research.google.com/drive/16V9TpOsLMOuU655ryerni7Hc7mni6hcU?usp=sharing
- https://gist.github.com/eric-burel/97156eea3f865104f0815c1aa143af19
Exploring `InMemorySaver` implementation was very insightful though!

PR #5236, opened 2025-06-27
docs: Nav reorg

lnhsingh:


docs/mkdocs.yml
```diff
@@ -92,38 +92,75 @@ nav:
...
+          -  4. Add human-in-the-loop: tutorials/get-started/4-human-in-the-loop.md
+          -  5. Customize state: tutorials/get-started/5-customize-state.md
+          -  6. Time travel: tutorials/get-started/6-time-travel.md
+      - Run a local server: tutorials/langgraph-platform/local-server.md
+      - Basic deployment: agents/deployment.md
+    - Agent development:
+      - Prebuilt components: agents/overview.md
+      - Run an agent: agents/run_agents.md
+      - Agent architectures: concepts/agentic_concepts.md
+    - Workflow orchestration:
```
ccurme: I'd consider calling this `Concepts` or similar. My main concern is that it's hard to find the meat of the docs; you need to navigate to Guides in the top horizontal bar. If on the landing page I see a link to Graph API, I'm liable to go there looking for details.

I'd also suggest adding a link to the Guides section near the top of the landing page (e.g., under the "Trusted by companies" text).

These are just initial reactions, feel free to incorporate or not :)

docs/mkdocs.yml
```diff
@@ -92,38 +92,75 @@ nav:
...
-        - Runtime: concepts/pregel.md
-        - Functional API: concepts/functional_api.md
+          -  1. Build a basic chatbot: tutorials/get-started/1-build-basic-chatbot.md
+          -  2. Add tools: tutorials/get-started/2-add-tools.md
+          -  3. Add memory: tutorials/get-started/3-add-memory.md
+          -  4. Add human-in-the-loop: tutorials/get-started/4-human-in-the-loop.md
+          -  5. Customize state: tutorials/get-started/5-customize-state.md
+          -  6. Time travel: tutorials/get-started/6-time-travel.md
+      - Run a local server: tutorials/langgraph-platform/local-server.md
+      - Basic deployment: agents/deployment.md
```
eyurtsev: Basic deployment is a lot like the local server, so we can remove it from here.

For deployment as long as users can discover that langgraph platform / deployment exist as an option it'd be great. We have some information about that in the `Home` page it just might not be immediately visible

PR #5237, opened 2025-06-27
Test windows make availability

foie0222:


foie0222:
I created this by mistake

PR #5257, opened 2025-06-30
Updated examples for SummarizationNode to account for serde with persistence layers

sparkida:
This snagged me for a little bit, not sure if there is a reason to not have this in there.

### What's happening
When using persistence layers with JSON serde, the running_summary is a plain dict instead of a dataclass. If you're using a Pydantic state then adding this in will automatically handle proper deserialization.

### Concern
This leaves a bit of hole in the docs, because the examples are using TypedDict classes, maybe there is an additional improvement we can add for that (but maybe it's out of scope). At the very least this provides a type hint of what's missing.


hinthornw (approved):
Thanks for the improvement!

PR #5287, opened 2025-07-01
switch order of MCP methods in API spec

samecrowder:
Placing the implemented MCP method as the first of the three

### DIFF
--- a/docs/docs/cloud/reference/api/openapi.json
+++ b/docs/docs/cloud/reference/api/openapi.json
@@ -2388,17 +2388,6 @@
       }
     },
     "/mcp/": {
-      "get": {
-        "operationId": "get_mcp",
-        "summary": "MCP Get",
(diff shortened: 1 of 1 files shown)


docs/docs/cloud/reference/api/openapi.json
```diff
@@ -2457,6 +2446,17 @@
         },
         "tags": ["MCP"]
       },
+      "get": {
```
andrewnguonly, line 2449: This might be a tricky thing to maintain since this file is copied directly from the source in `openapi.json` in `langgraph-api`. Can you update it there as well?

andrewnguonly (approved):

docs/docs/cloud/reference/api/openapi.json
samecrowder: https://github.com/langchain-ai/langgraph-api/pull/984

PR #5318, opened 2025-07-02
ci: set explicit workflow permissions to read (should be a no-op)

eyurtsev:
* We're using restricted GITHUB_TOKENS by default.
* This is expected to be a no-op operation to appease codeql.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,9 @@ on:
         type: string
         default: "libs/langgraph"
 
+permissions:
+  contents: read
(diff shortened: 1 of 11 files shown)


.github/workflows/release.yml
```diff
@@ -8,6 +8,9 @@ on:
         type: string
         default: "libs/langgraph"
 
+permissions:
```
eyurtsev, line 11: I'll double check this file

sydney-runkle (approved):

PR #5328, opened 2025-07-03
Fix: Updating mcp_tools_node fn & referencing runtime graph rebuild docs

j-broekhuizen:
Incorporating comments from @hinthornw and @eyurtsev

### DIFF
--- a/docs/docs/concepts/server-mcp.md
+++ b/docs/docs/concepts/server-mcp.md
@@ -209,7 +209,7 @@ To make user-scoped tools available to your LangGraph Platform deployment, start
 ```python
 from langchain_mcp_adapters.client import MultiServerMCPClient
 
-def get_mcp_tools_node(state, config):
+def mcp_tools_node(state, config):
     user = config["configurable"].get("langgraph_auth_user")
 		 # e.g., user["github_token"], user["email"], etc.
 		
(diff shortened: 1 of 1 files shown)


docs/docs/concepts/server-mcp.md
```diff
@@ -223,14 +223,19 @@ def get_mcp_tools_node(state, config):
...
+    
+    tool_messages = ...
+    return {"messages": tool_messages}
 ```
 
 1. MCP only supports adding headers to requests made to `streamable_http` and `sse` `transport` servers.
 2. Your MCP server URL.
 3. Get available tools from your MCP server.
 
+_This can also be done by [rebuilding your graph at runtime](https://langchain-ai.github.io/langgraph/cloud/deployment/graph_rebuild/) to have a different configuration for a new run_
```
hinthornw, line 237: @eyurtsev wdyt of having this shown as the default way? then could put in a TOolNode and go

eyurtsev: (1) Rebuilding at run time is indeed better as the default solution. user may have a custom workflow that involves looping behavior (e.g., agent) and we wouldn't want to fetch the tools for every loop iteration.
(2) Whatever we recommend should work nicely with human-in-the-loop by default. When resuming the graph run, the client should be re-initialized as connections could've closed / timed out

eyurtsev (approved):

eyurtsev:
Going to merge as we at least want to make the current change. I agree that rebuilding at run time is a better thing to show by default

PR #5399, opened 2025-07-08
fix(doc): remove incorrect navigation title overrides for mobile

nlimpid:
This is a small fix to resolve inconsistent navigation titles between desktop and mobile views. 

The `navigation_title_ovverides.css` file was incorrectly overriding navigation titles on mobile devices, changing "Get started" and "Guides" to "Home", while desktop navigation displayed the correct titles. This created a confusing user experience.


eyurtsev:
Confirmed that this messed up mobile view with double "Home" categories

PR #5404, opened 2025-07-08
feat: [CLI] Add arg to retain build deps (setuptools, pip, wheel)

hinthornw:
By default, we remove these at the end, since these are deps installed in the base image but which are typically unnecessary at runtime.

Some users' dependencies rely on these being there.

This PR adds support for retaining these:
```js
{
   ...
   "retain_build_tools": true
}
```

or
```js
{
   ...
   "retain_build_tools": ["wheel", "setuptools"] // can also be "wheel"
}
```

So you can opt-in to retaining these deps.


libs/cli/langgraph_cli/config.py
```diff
@@ -471,21 +471,63 @@ class Config(TypedDict, total=False):
     """Optional. Named definitions of UI components emitted by the agent, each pointing to a JS/TS file.
     """
 
+    retain_build_tools: Optional[Union[bool, list[str]]]
+    """Optional. Control whether to retain Python build tools (`pip`, `setuptools`, `wheel`) in the final image.
```
jdrogers940: nit: Change to make options explicit. Control whether to retain Python build tools in the final image. Allowed tools are:`pip`, `setuptools`, `wheel`. You can also set to true to include all build tools.

PR #5419, opened 2025-07-09
feat: add copy page button functionality and fix llms-text output

EugeneJinXin:
- Add copy page button with CSS and JS implementation
- Implement copy page hooks for MkDocs integration
- Add necessary styling and package.json dependencies


docs/_scripts/copy_page_hooks.py
```diff
@@ -0,0 +1,195 @@
...
+        # Skip completely empty lines for now
+        if line:
+            cleaned_lines.append(line)
+    
+    # Join lines and then clean up spacing
+    content = '\n'.join(cleaned_lines)
+    
+    # Add proper paragraph breaks by looking for markdown patterns
+    # Add double newline before headers
+    content = re.sub(r'\n(#{1,6}\s)', r'\n\n\1', content)
```
eyurtsev: Why is this necessary?

docs/_scripts/copy_page_hooks.py
```diff
@@ -0,0 +1,195 @@
...
+    content = re.sub(r'<script[^>]*>.*?</script\s*>', '', content, flags=re.DOTALL | re.IGNORECASE)
+    
+    # Remove style tags (handle spaces in closing tags)
+    content = re.sub(r'<style[^>]*>.*?</style\s*>', '', content, flags=re.DOTALL | re.IGNORECASE)
+    
+    # Remove HTML comments
+    content = re.sub(r'<!--.*?-->', '', content, flags=re.DOTALL)
+    
+    # Remove all HTML tags (more aggressive cleaning)
+    content = re.sub(r'<[^>]+>', '', content)
```
eyurtsev: Why is this needed?

eyurtsev:
Looks like whitespaces get stripped for markdown

docs/mkdocs.yml
```diff
@@ -362,6 +362,7 @@ markdown_extensions:
   - github-callouts
 hooks:
   - _scripts/notebook_hooks.py
+  - _scripts/copy_page_hooks.py
```
eyurtsev, line 365: We could try to combine this with the notebook_hooks -- notebooks_hooks are just representing the entrypoint (bad name)

docs/_scripts/copy_page_hooks.py
```diff
@@ -0,0 +1,195 @@
...
+    
+    # Remove markdown image references that might be logos
+    content = re.sub(r'!\[[^\]]*\]\([^)]*logo[^)]*\)', '', content, flags=re.IGNORECASE)
+    
+    # Clean up line by line
+    lines = content.split('\n')
+    cleaned_lines = []
+    
+    for line in lines:
+        line = line.strip()
```
eyurtsev: This .strip() is likely messing up formatting inside of code blocks and will likely cause issues in other places like tabs etc

docs/_scripts/copy_page_hooks.py
EugeneJinXin: not useful, was having issue with API reference and forgot to remove this, now removed most of these.

docs/_scripts/copy_page_hooks.py
EugeneJinXin: removed

docs/mkdocs.yml
EugeneJinXin: sounds like refactor, is this absolute necessary since we move to mintlify ?

docs/_scripts/copy_page_hooks.py
EugeneJinXin: exactly, fixed, seems to be working now

eyurtsev:
@EugeneJinXin  I'm deploying for now and trying a refactor to leverage the other plugins that we have. I want to handle the refactor now since we still haven't committed to migrating to mintlify and the timeline for that is uncertain even if it happens

PR #5453, opened 2025-07-11
fix: correct indentation for parallel tool calls handling

jitokim:
fix correct indentation for parallel tool calls handling


nfcampos (dismissed):

nfcampos:
The code is actually correct as is, closing PR

PR #5483, opened 2025-07-14
fix(docs): broken URL in _AIO_ERROR_MSG for AsyncSqliteSaver

jitokim:
remove unreachable `yield` from unimplemented async methods

### DIFF
--- a/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
+++ b/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
@@ -521,7 +521,6 @@ class SqliteSaver(BaseCheckpointSaver[str]):
         raise NotImplementedError(_AIO_ERROR_MSG)
         yield
 
-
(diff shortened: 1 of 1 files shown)


sydney-runkle (approved):

jitokim:
@sydney-runkle 
Thanks for the review. 
I’ve merged the main branch.

libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
```diff
@@ -519,7 +519,7 @@ async def alist(
             Use list() instead, or consider using [AsyncSqliteSaver][langgraph.checkpoint.sqlite.aio.AsyncSqliteSaver].
         """
         raise NotImplementedError(_AIO_ERROR_MSG)
-        yield
```
nfcampos: No this is here to ensure the method is a generator instead of a function, it cannot be removed

PR #5489, opened 2025-07-14
fix(docs): jokes needs reducer in graph-api.md (#5489)

sakshi1989:
The jokes key in the OverallState requires a reducer function because it is designed to aggregate results during the REDUCE phase of the graph execution. In this workflow, jokes are generated for each subject in the subjects list during the MAP phase, and these individual jokes need to be combined into a single list in the jokes key during the REDUCE phase. Without a reducer function, the graph cannot handle the aggregation of multiple values into the jokes key, resulting in an error.


sydney-runkle (approved):

Maxi91f:
Aligned to my proposal. Thanks!

https://github.com/langchain-ai/langgraph/pull/5521

PR #5505, opened 2025-07-14
feat(sdk-py): Show is_studio_user

hinthornw:


docs/docs/how-tos/auth/custom_auth.md
```diff
@@ -133,15 +133,44 @@ To allow an agent to perform authenticated actions on behalf of the user, access
...
     ...
 ```
 
 !!! note
-    Fetch user credentials from a secure secret store. Storing secrets in graph state is not recommended.
+Fetch user credentials from a secure secret store. Storing secrets in graph state is not recommended.
+
+### Authorizing a Studio user
+
+By default, if you add custom authorization on your resources, this will also apply to interactions made from the Studio. If you want, you can handle logged-in Studio users in a special way with [is_studio_user()](../../reference/functions/sdk_auth.isStudioUser.html). 2Code has comments. Press enter to view.
```
dqbd: Seems like there's an extra sentence?

hinthornw: ha copy paste

PR #5512, opened 2025-07-15
Fix Examples link in README to correct URL

Faisal-Alsrheed:
Update the Examples link in the README to point to the correct URL for guided examples on getting started with LangGraph.


eyurtsev:
There's another PR that does this already. Thank you for the contribution!

PR #5584, opened 2025-07-19
feat(cli): add support for api-version

phvash:
**Description:** 
Adds support for specifying the version of the api-server to use for docker base images.


libs/cli/langgraph_cli/cli.py
```diff
@@ -739,11 +762,12 @@ def prepare_args_and_stdin(
     debugger_port: Optional[int] = None,
     debugger_base_url: Optional[str] = None,
     postgres_uri: Optional[str] = None,
+    api_version: Optional[str] = None,
     # Like "my-tag" (if you already built it locally)
     image: Optional[str] = None,
     # Like "langchain/langgraphjs-api" or "langchain/langgraph-api
     base_image: Optional[str] = None,
-) -> tuple[list[str], str]:
```
andrewnguonly: Can we restore this type annotation?

PR #5600, opened 2025-07-21
feat(langgraph): Handle ParentCommand in RemoteGraph

nfcampos:
- when receiving a "command" stream event raise ParentCommand exception for caller graph to handle


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -672,6 +672,9 @@ def stream(
                 ns = tuple(ns_.split(NS_SEP))
             else:
                 mode, ns = chunk.event, ()
+            # raise ParentCommand exception for command events
+            if mode == "command":
```
hinthornw: could we be extra cautious about forward compatibility by also checking for "parent" in the goto of the payload(?)

sydney-runkle (approved):

PR #5609, opened 2025-07-21
docs: clarify draw_mermaid_png only works in jupyter

EugeneJinXin:
Nit, without comments I thought image would somehow show in terminal, but that's not true you'd only get image to show if in jupyter notebook. 
Don't have to merge, i'm just seeing this particular as not a pleasure DevX. 

Screenshot 2025-07-21 at 12 26 04 PM


docs/docs/agents/overview.md
```diff
@@ -159,7 +159,7 @@ function generateCodeSnippet({ tools, pre, post, response }) {
   if (post) lines.push("    post_model_hook=post_model_hook,");
   if (response) lines.push("    response_format=ResponseFormat,");
 
-  lines.push(")", "", "agent.get_graph().draw_mermaid_png()");
+  lines.push(")", "", "# Only works in Jupyter or GUI environments", "agent.get_graph().draw_mermaid_png()");
```
hinthornw: Could show draw_ascii as an alternative?

EugeneJinXin: Good point! Adding

EugeneJinXin: let's offer 3 ways: output image , terminal ascii, and jupyter? 

FYI  ascii doesn't seem to be super useful to me,  due to the confusing  * vs `  , but I'd still keep it as i can image people find it useful in logger print etc. I also 

Screenshot 2025-07-21 at 1 35 16 PM


Screenshot 2025-07-21 at 1 44 58 PM

PR #5619, opened 2025-07-22
chore (langgraph): Support sending distributed tracing headers

hinthornw:


nfcampos (approved):

samecrowder:
@hinthornw why is `self.distributed_tracing` not actually used anywhere?

PR #5720, opened 2025-07-30
chore(docs): Fix pages on get-started tab

eyurtsev:
Fix pages on get started guide


docs/docs/tutorials/get-started/2-add-tools.md
```diff
@@ -208,7 +208,7 @@ We can now incorporate it into a `StateGraph`:
 
 :::python
 
-```python hl_lines="15"
+```python
```
eyurtsev, line 211: No need to do this. We use special `highlight-next-line` syntax.

`hl_lines` is too brittle as it breaks upon any code refactor -- we will likely enforce linting rules to forbid it if we keep the pipeline around

hntrl (approved):

PR #5723, opened 2025-07-30
fix(docs): squash js docs build errors

hntrl:


docs/_scripts/link_map.py
```diff
@@ -92,7 +93,7 @@
     "entrypoint.final": "reference/functions/langgraph_func.entrypoint.final.html",
     "entrypoint": "reference/functions/langgraph_func.entrypoint.html",
     "from_pycryptodome_aes": "reference/functions/langgraph_checkpoint_serde_encrypted.EncryptedSerializer.fromPycryptodomeAes.html",
-    # "getContextVariable": "<insert-ref>",
+    "getContextVariable": "https://v03.api.js.langchain.com/functions/_langchain_core.context.getContextVariable.html",
```
eyurtsev, line 96: All the other references are prefixed with a different HOST I think is that correct?

docs/docs/agents/context.md
```diff
@@ -51,38 +51,12 @@ graph.invoke( # (1)!
 )
 ```
 
-:::
```
eyurtsev, line 54: why this change?

docs/docs/tutorials/workflows.md
```diff
@@ -1274,7 +1274,7 @@ With orchestrator-worker, an orchestrator breaks down a task and delegates each
...
+    Because orchestrator-worker workflows are common, LangGraph **has the `Send` API to support this**. It lets you dynamically create worker nodes and send each one a specific input. Each worker has its own state, and all worker outputs are written to a *shared state key* that is accessible to the orchestrator graph. This gives the orchestrator access to all worker output and allows it to synthesize them into a final output. As you can see below, we iterate over a list of sections and `Send` each to a worker node. See further documentation [here](https://langchain-ai.github.io/langgraph/how-tos/map-reduce/) and [here](https://langchain-ai.github.io/langgraph/concepts/low_level/#send).
```
eyurtsev, line 1277: why the change here? could we point to the relative files?

docs/_scripts/link_map.py
hntrl: yup! `getContextVariable` has to do with core logic, so it's pointing to langchain core refs instead of langgraph refs (two separate sites)

docs/docs/agents/context.md
hntrl: The thing this section is covering (runtime context) is conceptually different between the two languages currently (standard config vs. runtime context). Rather than trying to intermingle the code blocks together I thought it best to just leave it separated until we have an equivalent in JS. The deleted section here has been moved + condensed down lower

docs/docs/tutorials/workflows.md
hntrl: From what I can gather this has to do with some special ipynb/redirect handling. They aren't dead links, but they don't have a 1:1 doc file in the tree either. This is mirroring the current python block which lets the build pass:

https://github.com/langchain-ai/langgraph/blob/2180e0f80fcf783684ab64abd23562f6bc8a692e/docs/docs/tutorials/workflows.md?plain=1#L1137

Issue #5739, opened 2025-07-30
update type annotations in create_react_agents


eyurtsev:
### Privileged issue

- [x] I am a LangGraph maintainer, or was asked directly by a LangGraph maintainer to create an issue here.

### Issue Content

bind_tools return Runnable[LanguageModelInput, BaseMessage], not BaseChatModel and this is actually a bug in type annotations.

Please update the type annotations in create_react_agent to allow one to provide a callable for the model that uses bind_tools and returns a Runnable[LanguageModelInput, BaseMessage].

i.e., we want to widen the accepted types

<details>
<summary>Agent Context</summary>

<open-swe-do-not-edit-task-plan>
{
  "tasks": [
    {
      "id": "9679a1a1-5ed2-4d57-81c5-7e0167d92d05",
      "taskIndex": 0,
      "request": "[original issue]\n**update type annotations in create_react_agents**\n### Privileged issue\n\n- [x] I am a LangGraph maintainer, or was asked directly by a LangGraph maintainer to create an issue here.\n\n### Issue Content\n\nbind_tools return Runnable[LanguageModelInput, BaseMessage], not BaseChatModel and this is actually a bug in type annotations.\n\nPlease update the type annotations in create_react_agent to allow one to provide a callable for the model that uses bind_tools and returns a Runnable[LanguageModelInput, BaseMessage].\n\ni.e., we want to widen the accepted types\n\n",
      "title": "Update type annotations in create_react_agent to allow bind_tools return type",
      "createdAt": 1753900472096,
      "completed": true,
      "planRevisions": [
        {
          "revisionIndex": 0,
          "plans": [
            {
              "index": 0,
              "plan": "**Update the type annotations for dynamic model callables in `create_react_agent`** in `/home/daytona/langgraph/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py` at lines 255-256. Change `Callable[[StateSchema, Runtime[ContextT]], BaseChatModel]` to `Callable[[StateSchema, Runtime[ContextT]], LanguageModelLike]` and `Callable[[StateSchema, Runtime[ContextT]], Awaitable[BaseChatModel]]` to `Callable[[StateSchema, Runtime[ContextT]], Awaitable[LanguageModelLike]]`. This allows dynamic model functions to return the result of `bind_tools()` which is `Runnable[LanguageModelInput, BaseMessage]` (aliased as `LanguageModelLike`).",
              "completed": true,
[...]

PR #5748, opened 2025-07-30
chore(prebuilt): resolve flag parsing error

keenborder786:
- **Description:** Update the type annotations in create_react_agent to allow one to provide a callable for the model that uses bind_tools and returns a Runnable[LanguageModelInput, BaseMessage]
  - **Issue:** #5739


keenborder786:
@eyurtsev

libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -254,6 +254,9 @@ def create_react_agent(
         LanguageModelLike,
         Callable[[StateSchema, Runtime[ContextT]], BaseChatModel],
         Callable[[StateSchema, Runtime[ContextT]], Awaitable[BaseChatModel]],
+        Callable[
+            [StateSchema, Runtime[ContextT]], Runnable[LanguageModelInput, BaseMessage]
```
eyurtsev, line 258: hi @keenborder786,

We also need to support the async version of this.

But if we want a proper fix we likely need to specify a Protocol which supports `invoke`, `stream`, `ainvoke`, `astream` with the appropriate type signatures (based on whatever the implementation in create react agent uses.

keenborder786: okay

PR #5767, opened 2025-07-31
docs: redirects for LGP mintlify

lnhsingh:
- Added redirects for all of the LGP docs we're moving to Mintlify
- Exclude files not listed in nav from search
- Update banner


katmayb (approved):
LGTM, once live, I'll also give a test

PR #5791, opened 2025-08-01
chore(docs): fix admonitions in graph api page

eyurtsev:
fix many admonitions in the graph API page


hntrl (approved):
yay

ty for cleaning up my mess, am not sure where these weird indentation errors came from

PR #5811, opened 2025-08-02
fix(langgraph): deterministic caching for Pydantic models

### LINKED ISSUE #5733: LangGraph Caching Issue with Pydantic Models
### Checked other resources

```shell

```

### Description

# LangGraph Caching Issue with Pydantic Models

## Summary
LangGraph caching does not work consistently with Pydantic models as state objects due to non-deterministic serialization, despite documentation claiming support for both Pydantic models and TypedDict.

## Environment
- LangGraph version: [latest]
- Python version: 3.11+
- Pydantic version: 2.x

## Problem Description
When using Pydantic models as state objects in LangGraph workflows with caching enabled (`CachePolicy()`), the same logical state produces different cache keys on subsequent runs, leading to cache misses and redundant expensive operations.

## Root Cause
The issue stems from how Python's `pickle` module (used internally by LangGraph for state serialization) handles Pydantic models:

1. **Memory addresses in objects**: Pydantic models contain internal references that include memory addresses
2. **Non-deterministic internal metadata**: Pydantic stores internal metadata that varies between instantiations
3. **Optional field handling**: Different serialization behavior for `None` vs unset optional fields

### Pydantic Model (Inconsistent Hashing)
```python
from pydantic import BaseModel, Field
import hashlib
import pickle

class MyState(BaseModel):
    drug_name: str = Field(..., description="Name of drug")
    model: str
    result: str | None = None

# Same logical data, different hashes
state1 = MyState(drug_name="aspirin", model="gpt-4", result=None)
state2 = MyState(drug_name="aspirin", model="gpt-4", result=None)

hash1 = hashlib.sha256(pickle.dumps(state1)).hexdigest()
hash2 = hashlib.sha256(pickle.dumps(state2)).hexdigest()

print(f"Hash 1: {hash1}")
print(f"Hash 2: {hash2}")
print(f"Equal: {hash1 == hash2}")  # Often False!
```

### TypedDict (Consistent Hashing)
```python
from typing import TypedDict, NotRequired
import hashlib
import json

class MyState(TypedDict):
    drug_name: str
    model: str
    result: NotRequired[str | None]

# Same logical data, same hashes
state1: MyState = {"drug_name": "aspirin", "model": "gpt-4", "result": None}
state2: MyState = {"drug_name": "aspirin", "model": "gpt-4", "result": None}

hash1 = hashlib.sha256(json.dumps(state1, sort_keys=True, default=str).encode()).hexdigest()
hash2 = hashlib.sha256(json.dumps(state2, sort_keys=True, default=str).encode()).hexdigest()

print(f"Hash 1: {hash1}")
print(f"Hash 2: {hash2}")
print(f"Equal: {hash1 == hash2}")  # Always True
```

## Impact
- Cache misses lead to redundant expensive LLM calls
- Inconsistent performance in production
- Wasted computational resources
- Misleading cache hit rates

## Workaround
Convert Pydantic models to TypedDict for state objects:

```python
# Before (Pydantic - inconsistent caching)
class WorkflowState(BaseModel):
    drug_request: DrugRequest
    model: str
    result: MyResult | None = None

# After (TypedDict - consistent caching)  
class WorkflowState(TypedDict):
    drug_request: DrugRequest
    model: str
    result: NotRequired[MyResult | None]
```

### Option 1: Documentation Update
Update documentation to clarify that while Pydantic models are *supported* as state objects, TypedDict is *recommended* for optimal caching performance.

### Option 2: Improve Caching Logic
Implement a more deterministic serialization strategy for Pydantic models in the caching layer:
- Use `model.model_dump()` with `sort_keys=True` instead of pickle
- Add custom serialization logic for Pydantic models

### Option 3: Cache Key Generation
Allow users to provide custom cache key generation functions for state objects.

## Expected Behavior
Identical state objects (same logical data) should produce identical cache keys regardless of whether they're Pydantic models or TypedDict objects.

## Current Behavior  
Pydantic models with identical logical data produce different cache keys, leading to cache misses.

## Additional Context
This issue affects any workflow that:
- Uses expensive operations (LLM calls, API requests)
- Relies on caching for performance
- Uses Pydantic models for type safety and validation

The workaround (TypedDict) works perfectly but requires sacrificing some of Pydantic's validation and serialization benefits.

chitralputhran:
**Description:** Implement deterministic caching for Pydantic models to fix issue #5733 where identical Pydantic state objects would produce different cache keys, causing cache misses and redundant expensive operations.

The fix modifies the `_freeze()` function in `langgraph/_internal/_cache.py` to handle Pydantic BaseModel instances using `model_dump(mode='python')` for consistent serialization. This addresses the root cause where Python's `pickle` module handles Pydantic models non-deterministically due to memory addresses and internal metadata.

This PR also adds test functions that validate deterministic `_freeze()` behavior for Pydantic models, cache key consistency across identical instances, nested model handling, field order independence, and backward compatibility with non-Pydantic objects. The tests include a specific reproduction of the exact problem reported in the GitHub issue using `pickle.dumps()` and `hashlib.sha256()`. Closes #5733

**Issue:** #5733

**Dependencies:** None - uses existing pytest framework and Pydantic library

### DIFF
--- a/libs/langgraph/langgraph/_internal/_cache.py
+++ b/libs/langgraph/langgraph/_internal/_cache.py
@@ -3,11 +3,21 @@ from __future__ import annotations
 from collections.abc import Hashable, Mapping, Sequence
 from typing import Any
 
+try:
+    from pydantic import BaseModel
+except ImportError:
+    BaseModel = None
+
 
 def _freeze(obj: Any, depth: int = 10) -> Hashable:
     if isinstance(obj, Hashable) or depth <= 0:
         # already hashable, no need to freeze
         return obj
+    elif BaseModel is not None and isinstance(obj, BaseModel):
+        # Handle Pydantic models deterministically
+        # Convert to Python dict, then freeze recursively to sort keys
+        dumped = obj.model_dump(mode='python')
+        return (type(obj).__name__, _freeze(dumped, depth - 1))
     elif isinstance(obj, Mapping):
         # sort keys so {"a":1,"b":2} == {"b":2,"a":1}
         return tuple(sorted((k, _freeze(v, depth - 1)) for k, v in obj.items()))
--- a/libs/langgraph/tests/test_pydantic_caching_fix.py
+++ b/libs/langgraph/tests/test_pydantic_caching_fix.py
@@ -1,0 +1,307 @@
+"""Tests for the Pydantic caching consistency fix (issue #5733)."""
+
+import hashlib
+import pickle
+from typing import Optional
+
+from pydantic import BaseModel, Field
+from typing_extensions import NotRequired, TypedDict
+
+from langgraph._internal._cache import _freeze, default_cache_key
+
+
+class UserRequest(BaseModel):
+    """Example Pydantic model for testing."""
+    name: str = Field(..., description="User name")
+    email: Optional[str] = None
+    user_id: int
+
+
+class ProcessResult(BaseModel):
+    """Example result model."""
+    success: bool
+    message: str
+    data: Optional[dict] = None
+
+
+class WorkflowState(BaseModel):
+    """Complex Pydantic state object like in the issue."""
+    user_request: UserRequest
+    model: str
+    result: Optional[ProcessResult] = None
+    metadata: dict = Field(default_factory=dict)
+
+
+class WorkflowStateTypedDict(TypedDict):
+    """Equivalent TypedDict for comparison."""
+    user_request: UserRequest
+    model: str
+    result: NotRequired[Optional[ProcessResult]]
+    metadata: NotRequired[dict]
+
+
+def test_pydantic_model_freeze_deterministic():
+    """Test that _freeze produces deterministic results for Pydantic models."""
+    # Test with regular values
+    user1 = UserRequest(name="john_doe", email="john@example.com", user_id=123)
+    user2 = UserRequest(name="john_doe", email="john@example.com", user_id=123)
+    
+    # These should be different objects but have same content
+    assert user1 is not user2
+    assert user1 == user2
+    
+    # _freeze should produce identical results
+    frozen1 = _freeze(user1)
+    frozen2 = _freeze(user2)
+    
+    assert frozen1 == frozen2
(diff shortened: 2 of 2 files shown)


docs/docs/how-tos/pydantic-state-caching.md
```diff
@@ -0,0 +1,143 @@
+# Use Pydantic models as state with reliable caching
+
+This guide explains how Pydantic models work as state objects with LangGraph's caching system after the fixes implemented in [issue #5733](https://github.com/langchain-ai/langgraph/issues/5733).
+
+
+## The Problem (Fixed)
+
+Previously, identical Pydantic models would produce different cache keys due to non-deterministic pickle serialization:
```
pascalwhoop: IMO a post mortem / fix report doesn't belong in the happy path of docs docs. So I'd just describe the caching system and that it works for both vanilla dicts and pydantic objects but that devs are responsible for ensuring the hashed object is deterministic

chitralputhran: Good point! You are right, this reads more like a bug report than user docs. I will rewrite it to focus on how to use caching with Pydantic models rather than explaining what was broken.

Will update it shortly 👍

docs/docs/how-tos/pydantic-state-caching.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. **Use immutable data**: Prefer immutable field values for predictable caching
+2. **Consistent instantiation**: Create models with identical data for cache hits
+3. **Monitor cache performance**: Use cache hit rates to validate effectiveness
+
+## Related
+
+- [Node caching documentation](graph-api.md#add-node-caching)
+- [Pydantic state models guide](../concepts/low_level.md#state)
+- [Cache types reference](../reference/cache.md)
```
pascalwhoop, line 121: IMO this is AI generated noise

pascalwhoop: the frameworks should "just work" with pydantic without making a deal out if it. No need to document a fixed bug

casparb:
Closing as I cannot repro #5733

PR #5825, opened 2025-08-04
chore(docs): add remaining js translations

hntrl:
Related Linear ticket: https://linear.app/langchain/issue/DOC-51/add-js-translations-for-remaining-pages

### DIFF
--- a/docs/docs/how-tos/graph-api.md
+++ b/docs/docs/how-tos/graph-api.md
@@ -674,9 +674,10 @@ Output of graph invocation: {"a":"set by node3"}
 ```
 :::
 
+:::python
+
 ### Use Pydantic models for graph state
 
-:::python
 A [StateGraph](https://langchain-ai.github.io/langgraph/reference/graphs.md#langgraph.graph.StateGraph) accepts a `state_schema` argument on initialization that specifies the "shape" of the state that the nodes in the graph can access and update.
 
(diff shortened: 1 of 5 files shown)


docs/docs/how-tos/graph-api.md
```diff
@@ -2485,9 +2485,10 @@ Recursion Error
 ```
 :::
 
-??? example "Extended example: return state on hitting recursion limit"
 
     :::python
```
beklapko: I moved these two start tags above the callouts to remove them from the JS versions. I don't know if we're supposed to have equivalent callouts or not for the JS version, but I made this change initially to just hide the dropdowns from the JS version.

beklapko:
@hntrl this is ready for you to review. You'll probably want to look at the individual commits to see where I moved the Python tags to remove headers and things from the JS versions of the docs. I wasn't sure if there should be equivalent JS sections for those but I just removed them from now by moving the start tags for Python.

I also wasn't sure what kind of validation (if any) needed to happen for the code examples.

The only other thing I noticed was that the Python docs have this API Reference links but the JS docs don't. Not sure if that's expected or not.
jsversion
pythonversion

beklapko (approved):

beklapko:
@hntrl I think you can merge whenever you're ready.

beklapko:
Found one more thing https://github.com/langchain-ai/langgraph/pull/5825/commits/771f0d8eac77180672078c847632c5330c5387b0

Issue #5722, opened 2025-07-30
The LLM invoked the parameterless tool and received a 400 error


lzj09:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from langchain_core.tools import tool
from langchain_openai import ChatOpenAI
from langgraph.graph import StateGraph, START, END, MessagesState
from langgraph.prebuilt import ToolNode, tools_condition


llm = ChatOpenAI(model="xx", api_key="xx", base_url="xxx", streaming=True)

@tool
def current_user_info() -> str:
    """get current user info"""

    return '{"name": "test", "age": 18}'

tools = [current_user_info]


# Define the two nodes we will cycle between
async def call_model(state: MessagesState):
    """Call the LLM powering our "agent"
    """

    model = llm.bind_tools(tools)
    # Get the model's response
    response = await model.ainvoke([*state['messages']])
    return {"messages": [response]}



builder = StateGraph(MessagesState)

builder.add_node(call_model)
builder.add_node("tools", ToolNode(tools))
builder.add_edge(START, "call_model")
builder.add_conditional_edges(
    "call_model",
    tools_condition,
    {"tools": "tools", END: END}
)
builder.add_edge("tools", "call_model")

graph = builder.compile(name="Test Agent")
```

### Error Message and Stack Trace (if applicable)

```shell
openai.BadRequestError: Error code: 400 - {'object': 'error', 'message': '1 validation error for ValidatorIterator\n0.function.arguments\n  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]\n    For further information visit https://errors.pydantic.dev/2.11/v/string_type', 'type': 'BadRequestError', 'param': None, 'code': 400}
```

### Description

Run the program through "langgraph dev", and then in LangGraph Studio, ask "get current user info", finally, a 400 error will be reported.

**Is it required that the "tool" method must have parameters?**

<img width="1062" height="591" alt="Image" src="https://github.com/user-attachments/assets/ad098b87-ebbc-44a1-8e20-186bdbdb0955" />

### System Info

[...]

PR #5833, opened 2025-08-05
fix(prebuilt): Allow functions with no parameters to be used as tools in ToolNode

jitokim:
Description:
Allows functions with no parameters to be used as tools in ToolNode.
Previously, if the args field in a ToolCall was missing or None, tool execution would fail.
This change ensures that args defaults to {} when not provided, enabling support for no-arg tools.

Issue:
Closes https://github.com/langchain-ai/langgraph/issues/5722

Dependencies:
None

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/tool_node.py
+++ b/libs/prebuilt/langgraph/prebuilt/tool_node.py
@@ -445,6 +445,8 @@ class ToolNode(RunnableCallable):
         if invalid_tool_message := self._validate_tool_call(call):
             return invalid_tool_message
         try:
+            if call.get("args") is None:
+                call["args"] = {}
             call_args = {**call, **{"type": "tool_call"}}
             response = self.tools_by_name[call["name"]].invoke(call_args, config)
 
@@ -502,6 +504,8 @@ class ToolNode(RunnableCallable):
             return invalid_tool_message
 
         try:
+            if call.get("args") is None:
+                call["args"] = {}
             input = {**call, **{"type": "tool_call"}}
             response = await self.tools_by_name[call["name"]].ainvoke(input, config)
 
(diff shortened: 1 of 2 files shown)


eyurtsev (changes requested):
Hi @jitokim,

Take a look at the type signature of ToolCall. args are required and can be an empty dict to support a parameterless function. 

We cannot/should not update the code here to support missing args given that it's not consistent w/ the tool call type.

eyurtsev (changes requested):
Hi @jitokim,

Take a look at the type signature of ToolCall. args are required and can be an empty dict to support a parameterless function. 

We cannot/should not update the code here to support missing args given that it's not consistent w/ the tool call type.

jitokim:
> Hi @jitokim,
> 
> Take a look at the type signature of ToolCall. args are required and can be an empty dict to support a parameterless function.
> 
> We cannot/should not update the code here to support missing args given that it's not consistent w/ the tool call type.

Hi @eyurtsev  , Thanks for the clarification.

You're absolutely right that the ToolCall type requires args, and that a missing field is technically invalid.
Since LangGraph (e.g., in [#5722](https://github.com/langchain-ai/langgraph/issues/5722)) may emit tool calls without args when a tool has no parameters, this results in unintended validation errors.

To align with the ToolCall spec and avoid relaxing LangGraph's schema, I’ve opened a PR in LangChain to patch this at the source by defaulting args to {} if missing:

https://github.com/langchain-ai/langchain/pull/32408

Given that, I think it's best to close this PR on the LangGraph side. Thanks again!

eyurtsev:
> Since LangGraph (e.g., in https://github.com/langchain-ai/langgraph/issues/5722) may emit tool calls without args when a tool has no parameters, this results in unintended validation errors.

We need to fix the root cause instead of adjusting the schema in langchain. The langchain schema is correct.

Could we determine whether the issue is in LangGraph or ChatOpenAI?

Based on the issue that you've referenced, it appears that the issue is actually that someone is using a NON open ai model via the ChatOpenAI interface and that model API is actually complying with the openai completions API.

If so, we'll need to make the adjustment for this in ChatOpenAI for the logic that generates the AIMessage that has tool calls

jitokim:
The issue has been fixed in LangChain.

PR #5899, opened 2025-08-13
feat(prebuilt): Add structured output tools to ToolNode

eyurtsev:
* Add structured output tools to ToolNode
* Fix default tool node name to match the actual default ('tools')
* Update doc-strings to explain what inputs/outputs are for the ToolNode.
* Mark internal attributes as private (potentially breaking -- although hopefully users aren't accessing these)

## Decisions points

* OK with two properties? Done since users may be relying on `tools_by_name` and expanding the return type will break user code.

### Marked as public

* Make `tools_by_name` an official public property
* Make `structured_output_tools` a public property

### Marked as private

There should be no reason why users are accessing these attributes

```python
_tool_to_state_args
_tool_to_store_arg
_handle_tool_errors
_messages_key
```

### Usage

```python

    class OutputSchema(BaseModel):
        name: str
        age: int
        location: str

    tool_node = ToolNode([OutputSchema])

    # Test that the structured output tool is registered correctly
    assert "OutputSchema" in tool_node.structured_output_tools

    # Create a tool call that matches the schema
    tool_call = {
        "name": "OutputSchema",
        "args": {"name": "Alice", "age": 30, "location": "NYC"},
        "id": "call_123",
        "type": "tool_call",
    }

    # Test sync execution
    result = tool_node.invoke(
        {"messages": [AIMessage(content="", tool_calls=[tool_call])]}
    )

    # Should return a Command with structured response
    assert isinstance(result, list)
    assert len(result) == 1
    command = result[0]
    assert isinstance(command, Command)

    # Check the update structure
    assert "messages" in command.update
    assert "structured_response" in command.update

    # Check the tool message
    tool_message = command.update["messages"][0]
    assert isinstance(tool_message, ToolMessage)
    assert tool_message.name == "OutputSchema"
    assert tool_message.tool_call_id == "call_123"

    # Check the structured response
    structured_response = command.update["structured_response"]
    assert isinstance(structured_response, OutputSchema)
    assert structured_response.name == "Alice"
    assert structured_response.age == 30
    assert structured_response.location == "NYC"
```

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/tool_node.py
+++ b/libs/prebuilt/langgraph/prebuilt/tool_node.py
@@ -340,14 +340,22 @@ class ToolNode(RunnableCallable):
         self.tools_by_name: dict[str, BaseTool] = {}
         self.tool_to_state_args: dict[str, dict[str, Optional[str]]] = {}
         self.tool_to_store_arg: dict[str, Optional[str]] = {}
+        self.structured_output_tools: list[str] = []
         self.handle_tool_errors = handle_tool_errors
         self.messages_key = messages_key
         for tool_ in tools:
-            if not isinstance(tool_, BaseTool):
-                tool_ = create_tool(tool_)
-            self.tools_by_name[tool_.name] = tool_
-            self.tool_to_state_args[tool_.name] = _get_state_args(tool_)
-            self.tool_to_store_arg[tool_.name] = _get_store_arg(tool_)
+            if inspect.isclass(tool_) and issubclass(tool_, BaseModel):
+                # Handle Pydantic model classes as structured output tools
+                self.tools_by_name[tool_.__name__] = tool_
+                self.tool_to_state_args[tool_.__name__] = {}
+                self.tool_to_store_arg[tool_.__name__] = None
+                self.structured_output_tools.append(tool_.__name__)
+            else:
+                if not isinstance(tool_, BaseTool):
+                    tool_ = create_tool(tool_)
+                self.tools_by_name[tool_.name] = tool_
+                self.tool_to_state_args[tool_.name] = _get_state_args(tool_)
+                self.tool_to_store_arg[tool_.name] = _get_store_arg(tool_)
 
     def _func(
         self,
@@ -390,7 +398,7 @@ class ToolNode(RunnableCallable):
 
     def _combine_tool_outputs(
         self,
-        outputs: list[ToolMessage],
+        outputs: list[Union[ToolMessage, Command]],
         input_type: Literal["list", "dict", "tool_calls"],
     ) -> list[Union[Command, list[ToolMessage], dict[str, list[ToolMessage]]]]:
         # preserve existing behavior for non-command tool outputs for backwards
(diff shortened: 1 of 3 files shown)


libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -283,42 +321,35 @@ def calculator(a: int, b: int) -> int:
...
-
-        For advanced use cases involving state injection or store access, tools
-        can be annotated with InjectedState or InjectedStore to receive graph
-        context automatically.
+        tool_node = ToolNode([my_tool], handle_tool_errors=handle_errors)
+        ```
     """
 
-    name: str = "ToolNode"
+    name: str = "tools"
```
eyurtsev, line 348: change has no effect since the correct default is "tools" (see __init__)

libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -283,42 +321,35 @@ def calculator(a: int, b: int) -> int:
...
 
     def __init__(
         self,
-        tools: Sequence[Union[BaseTool, Callable]],
+        tools: Sequence[Union[BaseTool, BaseModel, Callable]],
```
eyurtsev, line 352: * Accepting BaseModel in `tools` for convenience (used for structured output tools).
* Internally, we separate structured tools for output from regular tools.
* Inspection (via the two properties) also separates them.

libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -337,17 +368,36 @@ def __init__(
...
+                if not isinstance(tool, BaseTool):
+                    tool_ = create_tool(cast(Type[BaseTool], tool))
+                else:
+                    tool_ = tool
+                self._tools_by_name[tool_.name] = tool_
+                self._tool_to_state_args[tool_.name] = _get_state_args(tool_)
+                self._tool_to_store_arg[tool_.name] = _get_store_arg(tool_)
+
+    @property
+    def tools_by_name(self) -> dict[str, BaseTool]:
```
eyurtsev, line 393: Users are likely using `tools_by_name` already

libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -237,39 +239,75 @@ def _infer_handled_types(handler: Callable[..., str]) -> tuple[type[Exception],
...
+           - List of messages with tool calls in the last AIMessage
+
+        3. **Direct Tool Calls**: ``[{"name": "tool", "args": {...}, "id": "1", "type": "tool_call"}]``
+           - Bypasses message parsing for direct tool execution
+           - For programmatic tool invocation and testing
+
+    Tool Types:
+        1. **Regular tools**: Functions or BaseTool instances that return values or
+            Commands.
+        2. **Structured output tools**: Pydantic model classes for schema-validated
```
eyurtsev, line 263: We'll need to add support for typed-dict. Potential complication is that either here or upstream we'll need to discriminate between typed-dict vs. built-in tools.

LLM Built-in tools are also described as a typed dict at the moment.

PR #5931, opened 2025-08-18
feat(sdk-py): define aclose method to LangGraphClient

wakita181009:
This PR adds an aclose method to the LangGraphClient.

When using the client in a FastAPI application, it's common to share a single instance across the application's lifespan. The absence of an aclose method makes it difficult to gracefully close the underlying HTTP session on application shutdown. This change enables proper resource management by allowing the client to be closed cleanly.


wakita181009:
@hinthornw I fixed a test case.

PR #6016, opened 2025-08-26
feat(checkpoint-mcp): adds MCP based memory store support

kk-src:
Introduce support for MCP-based memory store, enabling integration with any backend exposing specific MCP tools:
- store_put
- store_get
- store_delete
- store_search
- store_list_namespaces

With this feature, the memory store can be backed by any data source (e.g., a PostgreSQL-based MCP server) or service implementing the required tools. Initial implementation supports async operations and streaming http.

This unlocks new scenarios, such as enabling Langgraph agents using LangMem to persist long-term memory and perform semantic search against MCP server as long as store_search tool is implemented.

Additional notes:
- Implements MCP store functionality: put, get, delete, search, list_namespaces
- Adds in-memory test MCP server for development and validation.
- Includes batch operation tests to simulate LangMem usage.
- Provides end-to-end tests for MCP store workflows.
- Adds sample LangMem documentation demonstrating MCP store integration.

### DIFF
--- a/libs/checkpoint-mcp/Makefile
+++ b/libs/checkpoint-mcp/Makefile
@@ -1,0 +1,38 @@
+
+.PHONY: test test_watch lint format
+
+######################
+# TESTING AND COVERAGE
+######################
+
+TEST ?= .
+
+test:
+	uv run pytest $(TEST)
+
+test_watch:
+	uv run ptw $(TEST)
+
+######################
+# LINTING AND FORMATTING
+######################
+
+# Define a variable for Python and notebook files.
+PYTHON_FILES=.
+MYPY_CACHE=.mypy_cache
+lint format: PYTHON_FILES=.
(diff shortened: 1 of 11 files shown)


eyurtsev:
@kk-src thanks for putting this together -- we're working toward the 1.0 release, so this PR won't be prioritized until after the release. We'll need to discuss internally whether we want this functionality or not.

There are two things:

High level: I don't know whether we want to commit to maintaining this integration yet b/c we just haven't seen anyone creating such MCP servers.

Also some technical issues: The life cycle management w/ MCP scares me a bit; e.g., in the given code the client is holding long lived connections to the MCP server which will very likely cause a bunch of issues in production. I'm suspect the existing implementation will need to be refactored before folks can use it.


You do have the option to roll this out as an independent package under your own github user, and if the package proves popular, we could move it more officially into langgraph.

eyurtsev:
cc @sydney-runkle  / @hinthornw

eyurtsev:
@kk-src we discussed a bit more internally I think the best path forward for you is to release your own package if you're interested in pursuing MCP server implementations for this interface.

PR #6028, opened 2025-08-27
WIP: monorepo support in CLI

isahers1:
This PR introduces the `--build-command` and `--install-command` arguments to `langgraph build`.

`--install-command` is a custom install command. If passed, it will be run from wherever the `langgraph build` call was made, i.e. NOT where the langgraph.json file lives (except if these are the same place). This will override the detected install command that we previously used.

`--build-command` is a custom build command. This will run from wherever the langgraph.json file lives, and will be done after the install has been run.

You don't need to provide both. Just providing one will make the install (detected or supplied) run in the directory from where `langgraph build was called` and then have the build command (if one exists) run in the directory where langgraph.json exists. 

I think we should probably allow configuring the directories from which these commands get run, but I don't think this needs to be part of the MVP.


libs/cli/langgraph_cli/config.py
```diff
@@ -911,10 +911,10 @@ def check_reserved(name: str, ref: str):
                         "Rename the directory to use it as flat-layout package."
                     )
                 check_reserved(resolved.name, local_dep)
-                container_path = f"/deps/__outer_{resolved.name}/{resolved.name}"
```
isahers1, line 916: was getting this error:
```
\ Building...ERROR: invalid context name __outer_common: invalid reference format
ERROR: invalid context name __outer_common: invalid reference format
```
until I changed this and then stuff worked fine

docs/docs/cloud/reference/cli.md
```diff
@@ -483,19 +483,19 @@ The LangGraph CLI requires a JSON configuration file that follows this [schema](
 
     RUN PIP_CONFIG_FILE=/pipconfig.txt PYTHONDONTWRITEBYTECODE=1 pip install --no-cache-dir -c /api/constraints.txt langchain_community langchain_anthropic langchain_openai wikipedia scikit-learn
 
-    ADD ./graphs /deps/__outer_graphs/src
+    ADD ./graphs /deps/outer-graphs/src
```
hinthornw, line 486: ooc: why do we need to remove the dunder?

libs/cli/tests/unit_tests/cli/test_cli.py
```diff
@@ -570,8 +570,9 @@ def test_build_generate_proper_build_context():
                 catch_exceptions=True,
             )
 
-        build_context_pattern = re.compile(r"--build-context\s+(\w+)=([^\s]+)")
+        build_context_pattern = re.compile(r"--build-context\s+([\w-]+)=([^\s]+)")
 
+        print(result.output)
```
hinthornw: nit

.github/workflows/_integration_test.yml
```diff
@@ -87,3 +87,15 @@ jobs:
         working-directory: libs/cli/js-examples
         run: |
           langgraph build -t langgraph-test-e
+
+      - name: Build JS monorepo service
+        if: steps.changed-files.outputs.all
+        working-directory: libs/cli/js-monorepo-example
+        run: |
+          langgraph build -t langgraph-test-f -c apps/agent/langgraph.json --build-command "yarn run turbo build" --install-command "yarn install"
```
hinthornw, line 95: i think we want to add a run_langgraph_cli_test.py call and maybe update that one to actually create a run

hinthornw (commented):
Will revisit after these meetings but high level:
- Is it a requirement that weremove the dunder prefix? I didn't catch why that was needed.
- I think we should actually run the built server in the tests. The hard part is ensuring imports, etc. are all resolved correctly
- Is it possible to handle both monorepo and other cases without the user having to provide additional configuration?

libs/cli/langgraph_cli/cli.py
```diff
@@ -322,22 +324,36 @@ def _build(
         "-t",
         tag,
     ]
+    # determine build context: use current directory for JS projects, config parent for Python
+    is_js_project = config_json.get("node_version") and not config_json.get(
+        "python_version"
+    )
+    if is_js_project and (build_command or install_command):
```
hinthornw, line 333: Could we add a comment here? Not clear why js project with no build/install command should be the second case or why the build / install command in python context wouldn't be the former case.

PR #6085, opened 2025-09-05
feat(cli): support prereleases

isahers1:
We previously errored when a user had prerelease dependencies, this PR passes the `--prereleases=allow` flag to our `uv pip install` call. 

This PR also adds a test to verify that said deployments will build and run as expected.

### DIFF
--- a/libs/cli/examples/graph_prerelease_reqs/requirements.txt
+++ b/libs/cli/examples/graph_prerelease_reqs/requirements.txt
@@ -1,0 +1,5 @@
+requests
+langchain_anthropic
+langchain_openai
+langchain_community
+langchain==1.0.0a2
--- a/libs/cli/examples/graph_prerelease_reqs/requirements.txt
+++ b/libs/cli/examples/graph_prerelease_reqs/requirements.txt
@@ -2,4 +2,5 @@ requests
 langchain_anthropic
 langchain_openai
 langchain_community
-langchain==1.0.0a2
+langchain
+langgraph==1.0.0a2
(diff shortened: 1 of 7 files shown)


libs/cli/examples/graph_prerelease_reqs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+requests
+langchain_anthropic
+langchain_openai
+langchain_community
+langchain==1.0.0a2
```
hinthornw:
```suggestion
langchain==1.0.0a2
langgraph==1.0.0a2
```

This actually passes for me even before your fix, since langchain 1.0.0a2 doesn't depend on an alpha.

hinthornw (approved):

nikita-preql:
I think this behaviour should be disabled by default @hinthornw . 

We had an incident in prod due to an upgrade from pymysql:

  TypeError: AsyncAdapt_aiomysql_connection.ping() missing
     1 required positional argument: 'reconnect'

    Reproduced verbatim:
    - Local poetry install → PyPI pymysql 1.1.2 →
    reconnect=True default → works
    - Local langgraph build → pymysql dev "1.2.0/2.2.8" →
    reconnect=False default → crashes
    - Deployed pod (langgraph cloud) → same pymysql dev
    build → same crash on every pool checkout in
    validate_pre_merge_node, blocking the entire deployment
    pipeline

    Consequences for projects using Dolt as a metadata store
     via SQLAlchemy+aiomysql:
    1. E2E tests pass (CI uses poetry install honoring the
    lock).
    2. langgraph build succeeds locally and on cloud.
    3. Every checkout in production crashes — silently
    surfaces as "Deployment Failed" on the first DB
    operation (in our case, semantic-merger's pre-merge
    validation).
    4. No way to disable --prerelease=allow from
    langgraph.json config — only escape is tightening upper
    bounds in pyproject.toml (e.g. pymysql =
    ">=1.1.2,<1.2").

    Underlying SQLAlchemy bug (not reported upstream yet):
    AsyncAdapt_aiomysql_connection.ping(self, reconnect:
    bool) should be reconnect: bool = False to match how
    MySQLDialect_pymysql.do_ping may call it.

    Asks:
    1. Provide a way to opt out of --prerelease=allow from
    langgraph.json (e.g., a prerelease: false field), or
    revert to opt-in behavior. Hard-coded --prerelease=allow
     is a footgun for any project with caret/range
    constraints on libraries that publish dev builds.
    2. Pin pymysql in /api/constraints.txt to the latest
    PyPI stable (currently 1.1.2) so downstream projects
    depending on Dolt/aiomysql don't get the unreleased
    build.

PR #6095, opened 2025-09-07
fix(langgraph): task result from stream mode debug / tasks should match format from getStateHistory / getState

keenborder786:
- **Issue:** #6073


casparb:
Closing as solved here https://github.com/langchain-ai/langgraph/pull/6233

PR #6107, opened 2025-09-08
fix(checkpoint-postgres): retry & reconnect when DB restarts or SSL connection drops

NDNM1408:
I have a problem using the Postgres checkpointer at my company. Whenever the DB restarts every day, the chatbot, which is equipped with a Postgres checkpointer, raises an error that says, `the connection is closed`

This PR hardens the Postgres checkpointer against transient connection failures seen when the DB restarts or the network/SSL channel is dropped (e.g., `OperationalError: SSL connection has been closed unexpectedly`, `InterfaceError: the connection is closed`). Previously, a restarted database or stale socket would cause operations to fail and bubble up to callers.

**What’s changed (logic preserved):**

- Add lightweight retry/reconnect around all DB ops while keeping the original behavior of each method.

  - `_with_retry(...)`: Exponential backoff (+ jitter) for clearly transient psycopg errors (closed connection, EOF, reset by peer, timeouts, admin termination).
  - `_ensure_connection_open()` / `_reconnect_if_single_conn()`: Proactively replace a dead single-connection client; invalidate old pipeline handles.
- Wrap these methods with the retry layer (no SQL/flow changes):
  - `setup`, `alist`, `aget_tuple` (including the **pending-sends migration** for `v < 4`), `aput`, `aput_writes`, `adelete_thread`.
- Keep all existing data shapes and migration behavior; no API or serialization changes.

**Why**

- In the prior version, if Postgres restarted or an SSL socket closed mid-request, the saver raised and left callers to recover. This PR makes the saver resilient by transparently retrying and, when needed, re-establishing the connection.

**Backward compatibility**

- No breaking changes. Public signatures and return types unchanged.
- Migration and read/write semantics unchanged.

**Performance**

- Negligible impact on the happy path (single attempt).
- Retries trigger only on recognized transient failures and use capped backoff.

**Testing**

- Existing test suite passes against PostgreSQL **15** and **16**.
- Lint (ruff), types (mypy), and formatting all pass.

**Dependencies**

- No new dependencies.


NDNM1408:
@nfcampos @eyurtsev Can you review this PR? When I deploy my agent using Postgres checkpointer, every new day it meet the connection is closed error. When i using this code, the error is resolved. You can check by run a Postgres Docker and restart it then run the agent with the current checkpointer to see the error

PR #6129, opened 2025-09-10
fix(cli): fix CLI integration test

isahers1:
CLI integration tests were failing due to missing env vars, just needed to copy to the places where the build commands were running from


casparb (approved):
could you update PR desc with a short description please

Issue #6137, opened 2025-09-12
Langgraph postgresql checkpointer support old checkpoints


razmikmelikbekyan:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
here:


def _load_checkpoint_tuple(self, value: DictRow) -> CheckpointTuple:
        """
        Convert a database row into a CheckpointTuple object.

        Args:
            value: A row from the database containing checkpoint data.

        Returns:
            CheckpointTuple: A structured representation of the checkpoint,
            including its configuration, metadata, parent checkpoint (if any),
            and pending writes.
        """
        return CheckpointTuple(
            {
                "configurable": {
                    "thread_id": value["thread_id"],
                    "checkpoint_ns": value["checkpoint_ns"],
                    "checkpoint_id": value["checkpoint_id"],
                }
            },
            {
                **value["checkpoint"],
                "channel_values": {
                    **value["checkpoint"].get("channel_values"),
                    **self._load_blobs(value["channel_values"]),
                },
            },
            value["metadata"],
            (
                {
                    "configurable": {
                        "thread_id": value["thread_id"],
                        "checkpoint_ns": value["checkpoint_ns"],
                        "checkpoint_id": value["parent_checkpoint_id"],
                    }
                }
                if value["parent_checkpoint_id"]
                else None
            ),
            self._load_writes(value["pending_writes"]),
        )


this line `**value["checkpoint"].get("channel_values"), should be replaced by this

`**value["checkpoint"].get("channel_values", {}),`

to avoid issue when old checkpoint does not have `channel_values` key.
```

### Error Message and Stack Trace (if applicable)

```shell
File "/Users/rmelikbe/Desktop/projects/diva-backend/venv/lib/python3.11/site-packages/langgraph/checkpoint/base/__init__.py", line 149, in get
    if value := self.get_tuple(config):
                ^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #5677, opened 2025-07-27
AsyncPostgresSaver 2.0.23 fails to read channel values: NoneType object is not a mapping


ManuelAngel99:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
When using: anext(checkpointer.alist({'configurable': {'configurable': {'thread_id': '5df66a44-081d-4060-a858-4cfbdbb1b126'}}, limit=1), None)
```

### Error Message and Stack Trace (if applicable)

```shell
sdk-api-debug-1           |     return await anext(checkpointer.alist(read_config, limit=1), None)
sdk-api-debug-1           |                        │            │     └ {'configurable': {'thread_id': '5df66a44-081d-4060-a858-4cfbdbb1b126'}}
sdk-api-debug-1           |                        │            └ <function AsyncPostgresSaver.alist at 0xffff6684df80>
sdk-api-debug-1           |                        └ <langgraph.checkpoint.postgres.aio.AsyncPostgresSaver object at 0xffff62b7d7f0>
sdk-api-debug-1           | 
sdk-api-debug-1           |   File "/app/lib/python3.13/site-packages/langgraph/checkpoint/postgres/aio.py", line 166, in alist
sdk-api-debug-1           |     yield await self._load_checkpoint_tuple(value)
sdk-api-debug-1           |                 │    │                      └ {'thread_id': '5df66a44-081d-4060-a858-4cfbdbb1b126', 'checkpoint': {'v': 3, 'id': '1f0687d6-2d05-6dd0-8086-fcaa43ed2123', 't...
sdk-api-debug-1           |                 │    └ <function AsyncPostgresSaver._load_checkpoint_tuple at 0xffff6684e3e0>
sdk-api-debug-1           |                 └ <langgraph.checkpoint.postgres.aio.AsyncPostgresSaver object at 0xffff62b7d7f0>
sdk-api-debug-1           |   File "/app/lib/python3.13/site-packages/langgraph/checkpoint/postgres/aio.py", line 413, in _load_checkpoint_tuple
[...]

PR #6162, opened 2025-09-17
fix(langgraph): fix PostgresSaver crashing when loading older checkpoints

casparb:
### Description

https://github.com/langchain-ai/langgraph/issues/6137 and https://github.com/langchain-ai/langgraph/issues/5677 reported issues where older checkpoints read by AsyncPostgresSaver/PostgresSaver from `langgraph-checkpoint-postgres==2.0.19` fail to read channel values, throwing `NoneType object is not a mapping`. This was due to a bug in how `channel_values` is assembled:
```python
"channel_values": {
    **value["checkpoint"].get("channel_values"),  # <--- if channel_values doesn't exist (old checkpoint), **None errors
    **self._load_blobs(value["channel_values"]),
},
```
This bug was observed for checkpoints generated by `langgraph-checkpoint-postgres<=2.0.19`.

Fixed by providing a fallback to `value["checkpoint"].get("channel_values")`:
```python
**value["checkpoint"],
"channel_values": {
    **(
        value["checkpoint"].get("channel_values") or {}
    ),  # 'or {}' needed for backwards compat with v3 checkpoints and below, as v4 introduced channel_values key
    **self._load_blobs(value["channel_values"]),
},
```

### Tests
Added test for AsyncPostgresSaver and test for PostgresSaver, using monkeypatch to remove `channel_values` before CheckpointTuple is assembled in `_load_checkpoint_tuple`.

### Solves
https://github.com/langchain-ai/langgraph/issues/6137 and https://github.com/langchain-ai/langgraph/issues/5677

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
@@ -450,7 +450,9 @@ class PostgresSaver(BasePostgresSaver):
             {
                 **value["checkpoint"],
                 "channel_values": {
-                    **value["checkpoint"].get("channel_values"),
+                    **(
+                        value["checkpoint"].get("channel_values") or {}
+                    ),  # 'or {}' needed for backwards compat with v3 checkpoints and below, as v4 introduced channel_values key
                     **self._load_blobs(value["channel_values"]),
                 },
             },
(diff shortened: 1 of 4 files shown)


libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
```diff
@@ -450,7 +450,9 @@ def _load_checkpoint_tuple(self, value: DictRow) -> CheckpointTuple:
             {
                 **value["checkpoint"],
                 "channel_values": {
-                    **value["checkpoint"].get("channel_values"),
+                    **(
+                        value["checkpoint"].get("channel_values") or {}
+                    ),  # 'or {}' needed for backwards compat with checkpoints that do not have a channel_values key
```
hinthornw: comment probably not necessary here (same in aio.py)

PR #6173, opened 2025-09-20
chore(sdk-py): type errors nicely

isahers1:
This PR types errors in a nicer way

### DIFF
--- a/libs/sdk-py/langgraph_sdk/client.py
+++ b/libs/sdk-py/langgraph_sdk/client.py
@@ -24,6 +24,7 @@ from typing import (
     Literal,
     Optional,
     Union,
+    cast,
     overload,
 )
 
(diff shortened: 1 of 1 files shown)


libs/sdk-py/langgraph_sdk/errors.py
```diff
@@ -0,0 +1,218 @@
...
+        super().__init__(message, request, body=None)
+
+
+class APITimeoutError(APIConnectionError):
+    def __init__(self, request: httpx.Request) -> None:
+        super().__init__(message="Request timed out.", request=request)
+
+
+class BadRequestError(APIStatusError):
+    status_code: Literal[400] = 400  # type: ignore[assignment]
```
jdrogers940: Why the ignore[assignment]?

Can we also have some quick tests to validate types/status codes are getting set on responses for a couple of these?

hinthornw:
> We should probably do some sort of larger bump given that folks could be relying on these being HTTPStatusErrors for exception handling. Alternatively, did we consider inheriting from that in the API Error?


Yeah we should inherit from the originating error for these ^

PR #6179, opened 2025-09-22
feat(cli): add configuration for server customization ordering

pjrule:
This adds a configuration option in `HttpConfig` that allows LangGraph Platform users to apply custom authentication hooks before (other) custom middleware. Currently, the order is fixed (custom middleware is always evaluated before custom auth).

### DIFF
--- a/libs/cli/langgraph_cli/config.py
+++ b/libs/cli/langgraph_cli/config.py
@@ -18,6 +18,7 @@ DEFAULT_IMAGE_DISTRO = "debian"
 
 
 Distros = Literal["debian", "wolfi", "bullseye", "bookworm"]
+MiddlewareOrders = Literal["custom_auth_first", "custom_middleware_first"]
 
 
 class TTLConfig(TypedDict, total=False):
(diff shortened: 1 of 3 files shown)


libs/cli/langgraph_cli/config.py
```diff
@@ -18,6 +18,7 @@
 
 
 Distros = Literal["debian", "wolfi", "bullseye", "bookworm"]
+MiddlewareOrders = Literal["custom_auth_first", "custom_middleware_first"]
```
hinthornw: This also applies to default auth, right? (Langsmith auth)

WDYT of naming like "auth_first", "middleware_first"

pjrule: indeed, it does apply to default auth as well—will update

hinthornw (approved):
Great. Could you also update the docs page here: https://github.com/langchain-ai/docs/blob/main/src/langgraph-platform/cli.mdx

And include the minimum server version that includes it.

libs/cli/langgraph_cli/config.py
eyurtsev: Left a comment on the commit itself -- just in case commenting here. idiomatic naming for the type for a Literal is in singular form

Issue #5769, opened 2025-07-31
A JSON serializable problem on PostgresSaver


gs-max:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from langchain.chat_models import init_chat_model
from langgraph.graph import StateGraph, MessagesState, START
from langgraph.checkpoint.postgres import PostgresSaver

model = init_chat_model(model="anthropic:claude-3-5-haiku-latest")

DB_URI = "postgresql://postgres:postgres@localhost:5442/postgres?sslmode=disable"
with PostgresSaver.from_conn_string(DB_URI) as checkpointer:
    # checkpointer.setup()

    def call_model(state: MessagesState):
        response = model.invoke(state["messages"])
        return {"messages": response}

    builder = StateGraph(MessagesState)
    builder.add_node(call_model)
    builder.add_edge(START, "call_model")

    graph = builder.compile(checkpointer=checkpointer)

    config = {
        "configurable": {
            "thread_id": "1"
        }
    }

    for chunk in graph.stream(
        {"messages": [{"role": "user", "content": "hi! I'm bob"}]},
        config,
        stream_mode="values"
    ):
        chunk["messages"][-1].pretty_print()

    for chunk in graph.stream(
        {"messages": [{"role": "user", "content": "what's my name?"}]},
        config,
        stream_mode="values"
    ):
        chunk["messages"][-1].pretty_print()
```

### Error Message and Stack Trace (if applicable)

```shell
Traceback (most recent call last):
  File "/Users/gaoshuo/Downloads/项目2_基于LangGraph实现智能分诊系统/L1-Project-2/apiTest.py", line 31, in <module>
    for chunk in graph.stream(
  File "/Users/gaoshuo/Downloads/miniconda3/envs/L1-Project-2/lib/python3.11/site-packages/langgraph/pregel/__init__.py", line 1723, in stream
    with SyncPregelLoop(
  File "/Users/gaoshuo/Downloads/miniconda3/envs/L1-Project-2/lib/python3.11/site-packages/langgraph/pregel/loop.py", line 959, in __exit__
[...]

PR #6236, opened 2025-10-03
fix(checkpoint): handle metadata.writes when serializing old checkpoints with Jsonb

casparb:
Issue

Support for `Checkpoint.metadata.writes` was dropped in `langgraph` v0.5.x. 

In `langgraph-checkpoint-postgres` v2.0.23, metadata was serialized with `BasePostgresSaver._dump_metadata` -> `JsonPlusSerializer.dumps` which handles `pydantic.BaseModel`.

In v2.0.23, metadata is serialized with `psycopg.types.json.Jsonb`, which raises `TypeError: Object of type AIMessage is not JSON serializable` when trying to serialize `writes`.

Solution

- Add `BaseCheckpointSaver.get_serializable_checkpoint_metadata` which pops the `writes` key. 
- Log deprecation warning when strange version combinations are used 

Solves https://github.com/langchain-ai/langgraph/issues/5769

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -11,9 +11,7 @@ from langgraph.checkpoint.base import (
     WRITES_IDX_MAP,
     BaseCheckpointSaver,
     ChannelVersions,
-    CheckpointMetadata,
     get_checkpoint_id,
-    get_checkpoint_metadata,
 )
 from langgraph.checkpoint.serde.types import TASKS
 from psycopg.types.json import Jsonb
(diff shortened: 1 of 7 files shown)


libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -299,3 +316,12 @@ def _search_where(
             "WHERE " + " AND ".join(wheres) if wheres else "",
             param_values,
         )
+
+    def get_serializable_checkpoint_metadata(
+        self, config: RunnableConfig, metadata: CheckpointMetadata
+    ) -> CheckpointMetadata:
+        """Get checkpoint metadata in a backwards-compatible manner."""
+        checkpoint_metadata = get_checkpoint_metadata(config, metadata)
```
nfcampos: why wasn't this change made in get_checkpoint_metadata directly?

libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -1,21 +1,38 @@
...
 from langgraph.checkpoint.serde.types import TASKS
 from psycopg.types.json import Jsonb
 
 MetadataInput = Optional[dict[str, Any]]
 
+try:
+    major, minor = get_version("langgraph").split(".")[:2]
+    if int(major) == 0 and int(minor) < 5:
+        warnings.warn(
+            "LangGraph versions < 0.5.x may have compatibility issues with this version of checkpoint-postgres. "
```
nfcampos: This should be worded more like "You're using incompatible versions of langgraph and langgraph-checkpoint-postgres. ..."

PR #6269, opened 2025-10-10
chore: Restrict "json" type deserialization

hinthornw:
- Rm untyped loads/dumps
- Restrict to an allow list


eyurtsev (approved):

10ca1h0st:
checkpoint\langgraph\checkpoint\memory\__init__.py
line 152: checkpoint_: Checkpoint = self.serde.loads_typed(checkpoint)

there exists the same problem. Should it be fixed?

PR #6276, opened 2025-10-13
feat(cli): cache optimized dockerfile

phvash:
# Description

Adds support for multilayer deps installation even when requirements.txt is missing in packages. This eliminates cache invalidation due to code changes. 

## How it works 

```
For each real package without requirements.txt:

├─ Has pyproject.toml?
│  ├─ YES
│  │  ├─ Has uv.lock?
│  │  │  ├─ YES → Copy pyproject.toml + uv.lock
│  │  │  │       → RUN uv export --no-hashes -o requirements.txt
│  │  │  │       (Uses locked dependencies with exact versions)
│  │  │  │
│  │  │  └─ NO  → Copy pyproject.toml
│  │  │          → RUN uv pip compile pyproject.toml -o requirements.txt
│  │  │          (Resolves dependencies fresh)
│  │
│  └─ NO (has setup.py only)
│     └─ Copy setup.py + setup.cfg (if exists)
│        → RUN uv pip compile setup.py -o requirements.txt
│        (setup.py packages can't have uv.lock, so always compile)

Then: Use existing requirements.txt installation logic for ALL packages
```

Summary of Changes to Cache flow:
1. move step to auto generate requirements.txt (when missing) away from ci to  inside the autogenerated dockerfile by cli
  a. away from ci, because we don't have the constraint.txt file available outside base image 
  b. and generating the requirement.txt without constraint.txt can lead to installation failures (different, unreconciliable dep versions in req and const files)

2. make use of only requirements.txt file 
  a. as opposed to maintaining mulitiple code path/logic to install deps 
          i). . i.e requirements file + lock files 
  b. generate requirement file from lock file if it exists, rely on existing support for requirements to handle installation 
  c. drop support for poetry lock files (avoid having to extend cli to support poetry, and update base images)

3. only generate requirements when uv  is the pip installer 
  a. cache is needed for ci, ci uses uv, no need to support other installers atm

### DIFF
--- a/libs/cli/langgraph_cli/config.py
+++ b/libs/cli/langgraph_cli/config.py
@@ -1331,7 +1331,7 @@ def _metadata_files(spec: ReqGenSpec) -> list[str]:
 
 def _get_reqs_gen_cmd(spec: ReqGenSpec) -> str:
     if spec.package_type == "pyproject" and spec.has_uv_lock:
-        return "uv export --no-hashes --no-dev -o 'requirements.txt'"
+        return "uv export --no-hashes --no-dev --no-emit-local -o 'requirements.txt'"
     if spec.package_type == "pyproject":
         return f"uv pip compile pyproject.toml -o 'requirements.txt' --constraint {CONSTRAINTS_PATH}"
     return (


libs/cli/langgraph_cli/config.py
```diff
@@ -923,6 +945,35 @@ def check_reserved(name: str, ref: str):
...
+            if "requirements.txt" not in files:
+                has_pyproject = "pyproject.toml" in files
+                has_uv_lock = "uv.lock" in files
+
+                if has_pyproject:
+                    pkg_type = "pyproject"
+                    has_lock = has_uv_lock
+                else:  # setup.py only
+                    pkg_type = "setup"
+                    has_lock = False
```
isahers1, line 960:
```suggestion
pkg_type = "pyproject" if "pyproject.toml" in files else "setup"
has_lock = "uv.lock" in files and pkg_type == "pyproject"
```

libs/cli/langgraph_cli/config.py
```diff
@@ -1311,22 +1478,59 @@ def python_config_to_docker(
...
+        else:
+            copy_lines.append(
+                f"ADD {reqpath.relative_to(config_path.parent)} {destpath}"
+            )
+        all_req_paths.append(destpath)
+
+    # Generated reqs
+    if (
+        pip_installer == "uv"
+    ):  # we are only generate a requirements.txt if installer is uv (for now)
```
isahers1, line 1518: nit: `we only ...` instead of `we are only ...`

libs/cli/langgraph_cli/config.py
```diff
@@ -1311,22 +1478,59 @@ def python_config_to_docker(
...
+        for spec in local_deps.pkgs_missing_reqs or []:
+            all_req_paths.append(spec.container_req_path)
+
+    pip_reqs_str = ""
+    if all_req_paths:
+        # Stabilize order
+        all_req_paths = sorted(set(all_req_paths))
```
isahers1, line 1525: i _think_ the `set` call is not needed since container names are guaranteed to be unique

libs/cli/langgraph_cli/config.py
```diff
@@ -840,6 +850,15 @@ class LocalDeps(NamedTuple):
         additional_contexts: A list of paths to directories that contain local
             dependencies in parent directories. These directories are added to the
             Docker build context to ensure that the Dockerfile can access them.
+
+        pkgs_missing_reqs: A list of packages that need requirements.txt generated.
```
hinthornw, line 854: I think for these descriptions, ir makes more sense to put them inline with the ReqGenSpec object above in this case (since we do have a typed object for it)

PR #6277, opened 2025-10-14
add overwrite command

hwchase17:
writen by codex: prompt:

right now, the way pregel in langgraph works is that there are reducers for each state channel. that is a custom function. thats nice, but sometimes i just want to bypass the reducer and overwrite the state channel directly. add some method for doing so. it should work for any state channel regardless of reducer. come up with a nice devx for specifying this. happy to brainstorm!


eyurtsev:
This introduces a race condition for map-reduce pattern. What semantics do you want it to have in that context? last one wins?

PR #6322, opened 2025-10-22
fix(sdk-py): remove `body` param (`Auth.authenticate`)

pjrule:
Requests are not guaranteed to contain a body, and a request's body is not guaranteed to be valid JSON. This updates the type signature for authentication handlers to account for these scenarios.


hinthornw (approved):
I think we never pass body through actually (?)

PR #6366, opened 2025-10-31
fix(langgraph): get_subgraph fails when nodes share a common prefix

Joehunk:
# Description
`Pregel.get_subgraphs()` had a subtle bug due to the fact that it matches node names with namespaces using `startswith`. This created the possibility (which I did in fact hit in production) that 2 nodes whose names share a common prefix, and both contain subgraphs, could cause the algorithm to malfunction. It would find the first node (whose name was a prefix of the second), chop off that many characters leaving a partial namespace, and recurse on that.

# Issue
Fixes https://github.com/langchain-ai/langgraph/issues/6924

# Dependencies
None

# Twitter Handle
`@crain36003`

`make lint/format/test` run successfully.


Joehunk:
Fixes https://github.com/langchain-ai/langgraph/issues/6924

PR #6385, opened 2025-11-03
fix(cli): add buildkit syntax directiive, update tests

phvash:
**Description:** Adds the syntax directive to generated dockerfile for langgraph builds if we have additional contexts
**Issue:** fixes issue with python monorepo builds failing
**Dependencies:** N/A


libs/cli/langgraph_cli/config.py
```diff
@@ -926,35 +926,51 @@ def python_config_to_docker(
...
-        installs,
-        "",
-        "# -- Installing all local dependencies --",
-        f"""RUN for dep in /deps/*; do \
+
+    # Prepare docker file contents
+    docker_file_contents = []
+
+    # Add syntax directive if we have additional contexts (requires BuildKit frontend.contexts capability)
+    if local_deps.additional_contexts:
```
andrewnguonly, line 934: Does adding this check mean that builds without additional contexts will break if `# syntax=docker/dockerfile:1.4` is included?

phvash: Nope, they won't break. We just avoid the additional step of having to pull the image when we don't need it.

andrewnguonly: Ah, makes sense. Thanks for clarifying.

Issue #6409, opened 2025-11-07
Returning `Command` with `goto` attribute in a graph with >= 3 total graph depth results in `langgraph.errors.ParentCommand` error


egebilecen:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from langchain_core.messages import BaseMessage
from langchain_core.runnables import RunnableConfig
from langgraph.checkpoint.memory import MemorySaver
from langgraph.graph import StateGraph, START
from langgraph.types import Command
from langgraph.graph.message import add_messages
from typing import Annotated, TypedDict, Sequence

class State(TypedDict):
    messages: Annotated[Sequence[BaseMessage], add_messages]

def sub_sub_child(state):
    return Command(
        graph=Command.PARENT,
        goto="sub_child_3",
        update={"messages": state["messages"] + ["sub_sub_child executed, going to sub_child_3"]},
    )

sub_sub_builder = StateGraph(State)
sub_sub_builder.add_node("sub_sub_child", sub_sub_child)
sub_sub_builder.add_edge(START, "sub_sub_child")
sub_sub_graph = sub_sub_builder.compile(name="sub_sub_graph", checkpointer=True)

def sub_child_1(state):
    return {"messages": state["messages"] + ["sub_child_1 executed"]}

def sub_child_3(state):
    return {"messages": state["messages"] + ["sub_child_3 executed"]}

sub_builder = StateGraph(State)
sub_builder.add_node("sub_child_1", sub_child_1)
sub_builder.add_node("sub_child_2", sub_sub_graph)
sub_builder.add_node("sub_child_3", sub_child_3)
sub_builder.add_edge(START, "sub_child_1")
sub_builder.add_edge("sub_child_1", "sub_child_2")
sub_graph = sub_builder.compile(name="sub_graph", checkpointer=True)

def child_1(state):
    return {"messages": state["messages"] + ["child_1 executed"]}

builder = StateGraph(State)
builder.add_node("child_1", child_1)
builder.add_node("child_2", sub_graph)
builder.add_edge(START, "child_1")
[...]

PR #6411, opened 2025-11-07
fix(langgraph): fix command usage with nested subgraphs

eyurtsev:
Need to re-review prior to merge

Resolve https://github.com/langchain-ai/langgraph/issues/6409

### DIFF
--- a/libs/langgraph/tests/test_pregel.py
+++ b/libs/langgraph/tests/test_pregel.py
@@ -7911,6 +7911,78 @@ def test_parent_command_goto(
     }
 
 
+@pytest.mark.parametrize("subgraph_persist", [True, False])
+def test_parent_command_goto_deeply_nested(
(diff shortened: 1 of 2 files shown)


libs/langgraph/tests/test_pregel.py
```diff
@@ -7911,6 +7911,78 @@ def node_b_parent(state):
...
+    This tests the fix for issue #6409 where Command.PARENT with goto
+    would fail in graphs with 3 or more levels of nesting due to
+    namespace comparison issues.
+    """
+    class State(TypedDict):
+        messages: Annotated[list[str], operator.add]
+
+    # Level 3 (deepest): sub_sub_graph
+    def sub_sub_node(state):
+        """Returns Command.PARENT to jump to grandparent's node."""
```
sydney-runkle, line 7929: This is jumping to parent node, not grandparent node?

PR #6418, opened 2025-11-08
feat(docs): enhance return type documentation for graph run method

mdrxy:
Might also want to do for `stream`?


sydney-runkle (changes requested):
This isn't right :/ it's largely based on the stream mode or combo of them

PR #6448, opened 2025-11-14
fix(checkpoint): InMemorySaver context managers should return self instance

lexi-k:
Hello, I believe that the `__enter__` and `__aenter__` functions of the `InMemorySaver` class should return `self` instances. I tried wrapping checkpointers in a custom function in our library that returns `AbstractAsyncContextManager`s and found out that actually this does not work as IMO intended for the `InMemorySaver`. 

In this PR I am proposing a fix and adding two simple test cases.


lexi-k:
Hi, do I please need to do something to fix the state of the tests or is it just waiting for approval from the maintainers? Thanks! (sorry for tagging you specifically @hinthornw, but I am not sure what would be a better way to make sure that someone got actually pinged with this PR as there are no reviewers assigned and also given the issue with vercel bot, and you seem to be quite active in this repository :pray:)

hinthornw:
oops thanks for your patience!

lexi-k:
Hello @hinthornw , thanks for the branch update. Does this PR need something more from me please? The CI still did not run :/

PR #6472, opened 2025-11-20
chore(langgraph): Add __init__.pyi file for auto import support

nithinjp1997:
Added a __init__.py file to langgraph folder. This helps python recognize langgraph as a package. Additionally this allows pylance or other python LSPs to provide import suggestions on the top level modules mainly `runtime.py` and `types.py`. Without __init__.py pylance code actions wouldn't suggest `Runtime` class or `Send` , `Command` from the top level modules. After addition of __init__.py these are now suggested by code actions auto import thereby improving developer experience.

### DIFF
--- a/libs/sdk-py/langgraph_sdk/__init__.py
+++ b/libs/sdk-py/langgraph_sdk/__init__.py
@@ -3,6 +3,6 @@ from langgraph_sdk.client import get_client, get_sync_client
 from langgraph_sdk.encryption import Encryption
 from langgraph_sdk.encryption.types import EncryptionContext
 
-__version__ = "0.3.1"
+__version__ = "0.3.2"
 
 __all__ = ["Auth", "Encryption", "EncryptionContext", "get_client", "get_sync_client"]
--- a/libs/sdk-py/langgraph_sdk/client.py
+++ b/libs/sdk-py/langgraph_sdk/client.py
@@ -3064,6 +3064,7 @@ class CronClient:
         interrupt_before: All | list[str] | None = None,
         interrupt_after: All | list[str] | None = None,
         webhook: str | None = None,
+        on_run_completed: OnCompletionBehavior | None = None,
         multitask_strategy: str | None = None,
         headers: Mapping[str, str] | None = None,
         params: QueryParamTypes | None = None,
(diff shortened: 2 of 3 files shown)


hinthornw:
Hi Nithin - thanks for the PR!

We deliberately do not do this.  LangGraph uses namespace packaging  for checkpointers and similar packages, following PEP 420 syntax [1]. 

[1] https://peps.python.org/pep-0420/

nithinjp1997:
So I have added the images specifying the developer experience in vscode. As you can see when `__init__.py` file is added auto import finds the correct suggestion, but without the init file the suggestion doesn't find the root level modules. Without init file the suggestions for classes or methods inside the subpackages are found but the root level modules is not found.

image

image

hinthornw:
If it works with an `__init__.pyi` we could consider that

nithinjp1997:
> If it works with an `__init__.pyi` we could consider that

I checked by adding an `__init__.pyi` file instead of an `__init__.py` file and it works too. Hence I have committed the change.

nithinjp1997:
```
"python.analysis.packageIndexDepths": [
    {
      "name": "langgraph",
      "depth": 4
    },
    {
      "name": "langchain",
      "depth": 3
    },
    { "name": "langchain_community", "depth": 3 }
    ]
```
    
The above entry needs to be added to vscode `settings.json` file for auto imports to find the packages. The issue is at depth level 2 for eg. `from langgraph.types ` here `langgraph` is depth level 1 and `types.py` is at depth 2. Without a `__init__.py` or `__init__.pyi` file, any `.py` files i.e modules at depth level 2 are not being indexed for auto imports even if the max depth is greater than 2 (As you can see I have langgraph depth set at 4). 
 
 So if anyone wants to have auto import support for langgraph or langchain they need to first add the above configuration in their vscode `settings.json` (one time setup) and then add an `__init__.pyi` file into the package inside the `.venv` (cumbersome setup/hack). This is not ideal since if we upgrade the langgraph package again one needs to add this `__init__.pyi`. If the PR is approved then the cumbersome part of this setup is avoided.

nithinjp1997:
> Hi Nithin - thanks for the PR!
> 
> We deliberately do not do this. LangGraph uses namespace packaging for checkpointers and similar packages, following PEP 420 syntax [1].
> 
> [1] https://peps.python.org/pep-0420/

From what I observed except `cache` and `checkpoint` every other package uses non namespace packaging standard. Why were these kept different from the rest ? Anyhow the `__init__.pyi` works and it still maintains the namespace packaging standard.

nithinjp1997:
@hwchase17 @hinthornw Guys could you please look into this issue.

nithinjp1997:
So i have a new update. I switched out pylance with ty for my typechecker and lsp in VSCode. Now everything is astronomically faster than before. Anyway ty has an issue for finding classes or methods for namespace packages and it doesn't have the similar setting of providing analysis depth as in pylance. So I tried adding a `__init__.pyi` in the cache, store and checkpoint namespace package folders and then it worked. If you could look into this it would be a great addition.

PR #6482, opened 2025-11-21
feat: custom encryption at rest

cwlbraa:
**Description:** 
This PR adds the Python SDK types necessary for langgraph platform users to inject their own custom encryption-at-rest functions. See [docs PR](https://github.com/langchain-ai/docs/pull/1715) for more details.

note: this PR adds a starlette dev dependency so that custom encryption can access BaseUser information. 

**Issue:**
required for LSD-172 

**Dependencies:**
- [depended upon by associated langgraph-api changes](https://github.com/langchain-ai/langgraph-api/pull/1773)(this PR must merge before that one)
- [docs PR](https://github.com/langchain-ai/docs/pull/1715)

**TODO:**
- [x] move docs to docs repo
- [x] bump package versions before merge


docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
+# Add custom at-rest encryption
```
jdrogers940: This will need to live here I believe: https://github.com/langchain-ai/docs

libs/cli/tests/unit_tests/test_config.py
```diff
@@ -748,6 +750,23 @@ def test_config_to_docker_nodejs():
...
 
+def test_config_to_docker_python_encrypt():
+    # Test that encrypt config is included in validation
+    graphs = {"agent": "./agent.py:graph"}
+    validated = validate_config(
+        {
+            "python_version": "3.11",
+            "graphs": graphs,
+            "dependencies": ["."],
+            "encrypt": {"path": "./encrypt.py:encrypt"},
```
jdrogers940: Worth adding some other cases here? Bad encrypt path and one that will get properly formatted?

libs/sdk-py/langgraph_sdk/encryption/types.py
```diff
@@ -0,0 +1,119 @@
...
+        model: The model type being encrypted (e.g., "assistant", "thread", "run", "checkpoint")
+        field: The specific field being encrypted (e.g., "metadata", "context", "kwargs", "values")
+        metadata: Additional context metadata that can be used for encryption decisions
+    """
+
+    __slots__ = ("field", "metadata", "model")
+
+    def __init__(
+        self,
+        model: str | None = None,
```
jdrogers940, line 34: Any opposition to better typing here with the fields that will definitely be supported? Same for field (though less concerned on that one)

cwlbraa: Yes, typing this would mean you need to change the sdk to add additional model or field types.

libs/cli/tests/unit_tests/test_config.py
cwlbraa: Can do

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: Why? What's the difference?

libs/sdk-py/langgraph_sdk/encryption/types.py
cwlbraa: gonna add some validation of these to the langgraph-api code, though

cwlbraa: https://github.com/langchain-ai/langgraph-api/pull/1773/commits/dff694a655e4ac364bc52f40a58019e4451a4786

docs/docs/how-tos/encryption/custom_encryption.md
hinthornw: The ones in this repo are deprecated and will be deleted. A couple months ago all the how-to's, etc. were all migrated to a docs monorepo.

cwlbraa: cool, will move then.

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+    Returns:
+        Decrypted data dictionary
+    """
+    tenant_id = ctx.metadata.get("tenant_id", "default")
+
+    decrypted_data = {}
+    for key, value in data.items():
+        if key.startswith("my.customer.org/") and isinstance(value, str):
+            # Decrypt the VALUE for customer-specific fields
+            import base64
```
hinthornw: We should keep the imports cleanly at the top of the file

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+            encrypted_data[key] = value
+
+    return encrypted_data
+
+
+@encrypt.decrypt.json
+async def decrypt_json_data(ctx: EncryptionContext, data: dict) -> dict:
+    """Decrypt JSON data fields.
+
+    Inverse of `encrypt_json_data`. Decrypts VALUES for fields with
```
hinthornw: Hmm is this a use case we'd expect? (encrypting only data for particular keys)? 

I may be mis-reading, but I'd assume if the goal is to show how to concditionally encrypt data, it'd be  more likely for a deployment to have a particular key ("messages" maybe) that is encrypted and then fetch the key based on encryption context vs. having the keys be org-coded

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+    "my.customer.org/" prefix, passes through all other fields.
+
+    The encryption context (tenant_id, key_id) is automatically retrieved
+    from storage, so callers don't need to re-pass the X-Encryption-Context header.
+
+    Args:
```
hinthornw: Personal preference, but i think these args/returns docstring parts aren't helpful unless you are actually connecting the dots within them. Right now they're just re-stating what is already communicated

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+
+!!! warning "Python only"
+
+    Custom encryption is currently only supported for Python graphs. JavaScript/TypeScript support is not yet available.
+
+## Overview
+
+Custom at-rest encryption allows you to:
+
+- **Encrypt JSON fields** - Selectively encrypt metadata, values, and other JSON data on assistants, threads, runs, and crons
```
hinthornw: Bullet points 1 and 2 feel redundant. User cares about what they can encrypt, and "checkpoint blobs" is another one in the list of assistants, threads, runs, and crons.

Multi-tenant isolation is also one division it supports but i would also work per user or other entity right? 

In general these bullet points could be tightened

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+- **Use your own encryption service** - Integrate with AWS KMS, Google Cloud KMS, HashiCorp Vault, or any other encryption service
+- **Multi-tenant isolation** - Use different encryption keys per tenant or customer
+
+The encryption system provides a decorator-based API similar to the Auth system, allowing you to define custom encryption and decryption handlers that are executed server-side.
+
+## How it works
+
+1. **Define handlers** - Create encryption and decryption functions decorated with `@encrypt.encrypt.blob`, `@encrypt.decrypt.blob`, `@encrypt.encrypt.json`, and `@encrypt.decrypt.json`
+2. **Configure** - Add the path to your encryption module in `langgraph.json`
+3. **Pass context** - Send encryption context (like tenant ID and key ID) via the `X-Encryption-Context` header
```
hinthornw: Header makes sense for the REST API, but I think it would be good to have a dedicated argument in the relevant SDK methods rather than relying on users to pass via a custom header.

We can chat more about this if you disagree.

libs/sdk-py/langgraph_sdk/encryption/__init__.py
```diff
@@ -0,0 +1,454 @@
...
+
+    @property
+    def json(self) -> _JsonDecryptDecorators:
+        """Access JSON decryption decorators.
+
+        Supports model-specific handlers:
+        - @decrypt.json - default handler for all models
+        - @decrypt.json.thread - handler for thread model only
+        - @decrypt.json.assistant - handler for assistant model only
+        - @decrypt.json("custom") - handler for custom model name
```
hinthornw: ooc: what's the use case for "custom" ?

libs/sdk-py/langgraph_sdk/encryption/__init__.py
```diff
@@ -0,0 +1,454 @@
...
+                self._parent._json_encryptors[model_name] = fn
+                return fn
+
+            return decorator
+        else:
+            fn = fn_or_model
+            self._parent._json_encryptor = fn
+            return fn
+
+    def __getattr__(
```
hinthornw, line 97: ooc: any reason why we shouldn't explicitly add assistants, threads, etc. so you get better type hinting/checking?

docs/docs/how-tos/encryption/custom_encryption.md
hinthornw, line 1: good start. I think we could better motivate the json codeblocks a bit?

and then as a negligent reader, I presume I have to implement both blobs and json methods to get proper encryption, but I am not 100% sure here.

libs/sdk-py/langgraph_sdk/encryption/__init__.py
hinthornw, line 1: Could we raise exceptions if someone tries to add 2 handlers for the same resource?

Any other exceptions we should be raising?

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: i'm kind of inclined to call this YAGNI, but we could ask qualtrics?

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: agreed

libs/sdk-py/langgraph_sdk/encryption/__init__.py
cwlbraa: idk, i just wrote it first before i figured out that the other technique could be dynamically dispatched. will remove.

libs/sdk-py/langgraph_sdk/encryption/__init__.py
cwlbraa: yes, see resolved comment [here](https://github.com/langchain-ai/langgraph/pull/6482#discussion_r2578859878)

tldr rn im validating at server startup so you get an error if you've misspelled, but i don't want to be required to update the SDK for any additional encryption we might want to add to future api objects.

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: it's definitely a use case _I_ expect, or at least how I would want to implement this if I was spinning up a multi-tenant platform for internal customers, but I'm obviously operating with pretty thin context here.

wdym "fetch the key based on encryption context"? like are you thinking about having another data store controlling which keys get encrypted for each context? i was kinda hoping to give an example with the fewest possible # of extra network roundtrips...

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: agreed, i just let claude go wild here bc i don't have existing python style opinions.

libs/sdk-py/langgraph_sdk/encryption/__init__.py
cwlbraa: yeah, i can add this validation. regarding other exceptions, i don't really know what realistic red paths are... do we need to check, for example, that the annotated fns have the correct type signatures?

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: oh gross, will fix.

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: yeah, i'll workshop it a bit, i'm a little close to the metal so they read as self-motivating to me, but good to have some outside perspective.

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+- **Encrypt JSON fields** - Selectively encrypt metadata, values, and other JSON data on assistants, threads, runs, and crons
+- **Encrypt checkpoint blobs** - Encrypt opaque checkpoint data stored during graph execution
+- **Use your own encryption service** - Integrate with AWS KMS, Google Cloud KMS, HashiCorp Vault, or any other encryption service
+- **Multi-tenant isolation** - Use different encryption keys per tenant or customer
+
+The encryption system provides a decorator-based API similar to the Auth system, allowing you to define custom encryption and decryption handlers that are executed server-side.
+
+## How it works
+
+1. **Define handlers** - Create encryption and decryption functions decorated with `@encrypt.encrypt.blob`, `@encrypt.decrypt.blob`, `@encrypt.encrypt.json`, and `@encrypt.decrypt.json`
```
jorgeg-qualtrics: I found it confusing to have the `encrypt` concept as two nested namespaces. What about naming the first namespace `custom_encryption` or just `encryption`? Or even better, use a completely different concept like `persistence` or `storage`.

This way the decorators would become something like: `@persistence.encrypt.blob`, `@persistence.decrypt.blob`, `@persistence.encrypt.json`, and `@persistence.decrypt.json`

And the json configuration would be something like:

```json hl_lines="7-9"
{
  "dependencies": ["."],
  "graphs": {
    "agent": "./agent.py:graph"
  },
  "env": ".env",
  "persistence": {
    "path": "./custom_persistence.py:persistence"
  }
}
```

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+## Overview
+
+Custom at-rest encryption allows you to:
+
+- **Encrypt JSON fields** - Selectively encrypt metadata, values, and other JSON data on assistants, threads, runs, and crons
+- **Encrypt checkpoint blobs** - Encrypt opaque checkpoint data stored during graph execution
+- **Use your own encryption service** - Integrate with AWS KMS, Google Cloud KMS, HashiCorp Vault, or any other encryption service
+- **Multi-tenant isolation** - Use different encryption keys per tenant or customer
+
+The encryption system provides a decorator-based API similar to the Auth system, allowing you to define custom encryption and decryption handlers that are executed server-side.
```
jorgeg-qualtrics: This makes sense. I was expecting something similar to the Auth decorator

docs/docs/how-tos/encryption/custom_encryption.md
jorgeg-qualtrics: Great call outs. Honestly, we don't need to differentiate between the tenant ID and key ID as the value for both will be the same in our case.

We already integrated with the Auth decorator, and we provided a custom JWT header, that we decrypt/decode and extract multiple claims. One of the claims that we passed inside the JWT payload (as a header), is the tenant ID (aka brand ID) and that is the only key that we need to pass to our encryption decryption layer (along  with the data that needs to be encrypted/decrypted.

jorgeg-qualtrics: Something like this:

```
@auth.authenticate
async def authenticate(request: Request) -> Auth.types.MinimalUserDict:
    if request.url.path in NO_AUTH_PATHS:
        return None
    passed_token = request.headers.get("x-jwt", None)
    if passed_token is None:
        raise Auth.exceptions.HTTPException(status_code=401, detail="Missing X-JWT Header")

.... <redacted code> ....

    result.update({
        "identity": identity,
        "display_name": user_id,
        "is_authenticated": True,
        "permissions": [],

        # Qualtrics specific dictionary elements (key and values)
        "issuer": issuer,
        "brandId": brand_id,
        "userId": user_id,
        "identityOverride": identity_override,
        "used_auth_token": used_auth_token
    })
    return result
```

This allows us to later own retrieve the tenant ID which is also our desired key ID as `ctx.user["brandId"]` or `ctx.user.brandId`

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+    Args:
+        ctx: Encryption context with metadata containing:
+             - tenant_id: Tenant identifier for multi-tenant isolation
+             - key_id: KMS key identifier to use
+        blob: The raw checkpoint bytes to encrypt
+
+    Returns:
+        Encrypted checkpoint bytes
+    """
+    # Extract context from X-Encryption-Context header
```
jorgeg-qualtrics: We prefer extracting this from the JWT claims (provided as a X-JWT Header) and obtained in the Auth layer/middleware, as it adds extra security because the claims/ jwt-payload is encrypted at transit and trusted that it was not tampered with

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+    )
+    return response['Plaintext']
+
+
+@encrypt.encrypt.json
+async def encrypt_json_data(ctx: EncryptionContext, data: dict) -> dict:
+    """Encrypt JSON data fields.
+
+    This example demonstrates a practical encryption strategy:
+    - "owner" field: Left unencrypted for search/filtering (indexed field)
```
jorgeg-qualtrics: This is a great idea, because we will need to be able to preserve ownership enforement and searching capabilities:

not all data is necessarily encrypted at rest. Certain data (e.g. Qualtrics-defined data), such as brand IDs or internal keys may not be encrypted, both because it’s not customer data and it may be necessary to facilitate internal operations (e.g. logging, data cleanup, etc…).

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+
+    # Encode as base64
+    encoded_context = base64.b64encode(
+        json.dumps(encryption_context).encode()
+    ).decode()
+
+    # Create client with encryption context header
+    client = get_client(
+        url="http://localhost:2024",
+        headers={"X-Encryption-Context": encoded_context}
```
jorgeg-qualtrics: This client always provides the same value for the encryption context header. In practice, we need to dynamically pass different tenant IDs/key IDs (in our case it is the same) per request. We already created our own SDK clients that provide a dynamic JWT header per request which injects different values for the brand ID (aka tenant ID). This is because we invoke the same graph for different tenants and need to reuse the client in a way that each API can receive different values for those headers (instead of being always the same during initialization)

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+client = get_client(
+    url="http://localhost:2024",
+    headers={"X-Encryption-Context": encoded_context}
+)
+thread = await client.threads.create(metadata={"my.customer.org/secret": "value"})
+
+# Decrypt: no context needed
+client2 = get_client(url="http://localhost:2024")
+retrieved = await client2.threads.get(thread["thread_id"])
+# The secret field is automatically decrypted using the stored context
```
jorgeg-qualtrics: This makes the encryption isolated from Auth and ownership (authN&authZ). Ideally, we would like to tie the encryption to the auth layers, so we can also trust that the client asking for decrypted data is also the owner of such data/resources

docs/docs/how-tos/encryption/custom_encryption.md
```diff
@@ -0,0 +1,451 @@
...
+```
+
+### What gets encrypted
```
jorgeg-qualtrics: Nice, ,this is very clear and I am sure this is what we need 🙏

cwlbraa: truthfully i want to delete this hunk of docs because i fear it will quickly get out of date, but I left it in just for you :)

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: makes sense. even in my assumed understanding of how y'all might wanna implement this, access or forwarding of the tamper-proofed jwt is necessary so it can be passed downstream to the KMS... makes sense that it might as well also be used to override values in the ContextVar we end up setting.

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: i hope to encourage anybody that wants to use this customization layer to not blindly encrypt everything, doing so potentially hobbles a lot of functionality.

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: https://github.com/langchain-ai/langgraph/pull/6482#discussion_r2579720521 this comment clarifies things a lot, thank you... makes me wonder if maybe sdk-fn-level auth injection should also be a thing...

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: the point where this gets ugly and stuttery is actually user-controllable, like in docs this could be:

```
persistence = Encryption()
@persistence.encrypt.json
```

calling it persistence kinda worries me, though, like it's a very wide name for something that is currently very specific. calling it "encryption" is very safe here, though, and less confusing.

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: gonna treat https://github.com/langchain-ai/langgraph/pull/6482#discussion_r2579679436 as the tracking comment for this

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: tentative plan for this is to add the following extension point:

```
@encryption.context
async def foo(user: BaseUser, existing: EncryptionContext) -> dict[str, typing.Any]
  // default returns existing.metadata, returned value becomes existing.metadata
  // note that returned dict is persisted unencrypted to aide in later decryption
```

the idea here being you can choose to merge keys from your authenticated user into existing.metadata, fully discard and replace it with user information, etc etc

does that work for you @jorgeg-qualtrics?

cwlbraa: regarding my previous comment, forwarding the JWT actually remains difficult here because typically we don't want to store auth tokens unencrypted... this means that for searches and bulk reads, the customizer remains responsible for authorizing a given user to whatever the EncryptionContext.metadata grants decrypt to.

libs/sdk-py/langgraph_sdk/encryption/__init__.py
jdrogers940: Yeah that would be nice to add

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: addressed here https://github.com/langchain-ai/langgraph/pull/6482/commits/372abf7739d3a93dc77287de2257444547b468e2

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: addressed here https://github.com/langchain-ai/langgraph/pull/6482/commits/372abf7739d3a93dc77287de2257444547b468e2

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: https://github.com/langchain-ai/langgraph/pull/6482/commits/b43a17f4b9aa5c85bd578f8d415f2d87c001b9b3

docs/docs/how-tos/encryption/custom_encryption.md
cwlbraa: https://github.com/langchain-ai/docs/pull/1715

PR #6500, opened 2025-11-25
feat(sdk-py): add sentinel to skip auto loading api key on sdk client create

jdrogers940:
**Description:** There are times a user might want to create the client, but conditionally set the API key. For example, consider a complex auth situation where the system has user callers using jwts and system callers using API keys. This allows explicitly disabling the auto-loading behavior of API keys in the client today, so no key is set.
**Issue:** N/A
**Dependencies:** None
**Twitter handle:** N/A

### DIFF
--- a/libs/sdk-py/langgraph_sdk/client.py
+++ b/libs/sdk-py/langgraph_sdk/client.py
@@ -81,44 +81,36 @@ logger = logging.getLogger(__name__)
 
 RESERVED_HEADERS = ("x-api-key",)
 
-
-class _SkipAutoLoad:
-    """Sentinel value to skip auto-loading API key from environment variables."""
-
-    def __repr__(self) -> str:
-        return "SKIP_AUTO_LOAD"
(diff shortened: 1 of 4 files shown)


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -82,24 +82,44 @@
 RESERVED_HEADERS = ("x-api-key",)
 
 
-def _get_api_key(api_key: str | None = None) -> str | None:
+class _SkipAutoLoad:
```
hinthornw: Ah so actually what I had in mind was mor elike what I do [here](https://github.com/langchain-ai/langsmith-sdk/blob/5e7abf1d7314eb30a8ac5b8700edbab8e5fbfd3f/python/langsmith/run_trees.py#L54).

Basically, python doesn't distinguish `undefined` and `null` values, but in this case we want an explicit `None` to be accepted, while an undefined value would imply it should check the environment.

PR #6515, opened 2025-11-28
docs: add clarity to use of `thread_id`

mdrxy:


libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -119,6 +119,25 @@ class BaseCheckpointSaver(Generic[V]):
     Checkpointers allow LangGraph agents to persist their state
     within and across multiple interactions.
 
+    When a checkpointer is configured, you **must** pass a `thread_id` in the
```
sydney-runkle: not necessarily must

PR #6526, opened 2025-12-02
feat: Include pagination in assistants search response

hinthornw:


libs/sdk-py/langgraph_sdk/schema.py
```diff
@@ -246,6 +246,15 @@ class Assistant(AssistantBase):
     """The last time the assistant was updated."""
 
 
+class AssistantsSearchResponse(TypedDict):
+    """Paginated response for assistant search results."""
+
+    assistants: list[Assistant]
+    """The assistants returned for the current search page."""
+    next: str | None
+    """Pagination cursor from the ``X-Pagination-Next`` response header."""
```
andrewnguonly, line 255: Might as well include `count` too.
```
count: int
"""Page size from the ``X-Pagination-Total`` response header."""
```

(or call it `size`)

hinthornw: Ya I was considering hat. I think I'd prefer not though, since we actually only return `max(len(assistants), limit + 1)` and just kept the header set for backwards compat reasons. Since this is a net-new API, I'd prefer to avoid adding rn

Open to being overruled ofc

PR #6529, opened 2025-12-02
fix(checkpoint): InMemorySaver context managers should return self in…

hinthornw:
h/t to @lexi-k for openening. merged here to check/fix CI


lexi-k:
Hello @hinthornw, sorry for bothering you with this again, but do you have a timeline please when this could be released to the pypi package? 🙏

PR #6543, opened 2025-12-06
Fix type signature for add_conditional_edges path_map parameter

### LINKED ISSUE #6540: `add_conditional_edges` type signature causes incompatibility with dict[str, str]
### Checked other resources

### Example Code

```python
from typing import TypedDict

from langgraph.graph import END, StateGraph


class State(TypedDict):
    approved: bool


def check_approval(state: State) -> str:
    return "yes" if state["approved"] else "no"


def approved_node(state: State) -> State:
    return state


def rejected_node(state: State) -> State:
    return state


graph = StateGraph(State)
graph.add_node("approved", approved_node)
graph.add_node("rejected", rejected_node)
# This line causes pyright error:
# dict[str, str] is not assignable to dict[Hashable, str]
path_map: dict[str, str] = {"yes": "approved", "no": "rejected"}
graph.add_conditional_edges("__start__", check_approval, path_map)
graph.add_edge("approved", END)
graph.add_edge("rejected", END)
app = graph.compile()
```

```shell
Argument of type "dict[str, str]" cannot be assigned to parameter "path_map" of type "dict[Hashable, str] | list[str] | None" in function "add_conditional_edges"
    Type "dict[str, str]" is not assignable to type "dict[Hashable, str] | list[str] | None"
      "dict[str, str]" is not assignable to "dict[Hashable, str]"
        Type parameter "_KT@dict" is invariant, but "str" is not the same as "Hashable"
      "dict[str, str]" is not assignable to "list[str]"
      "dict[str, str]" is not assignable to "None" (reportArgumentType)
1 error, 0 warnings, 0 informations
```

#### Problem

  The current type signature of `add_conditional_edges` uses `dict[Hashable, str]` for the `path_map` parameter:

  ```python
  def add_conditional_edges(
      self,
      source: str,
      path: Callable[..., str],
      path_map: dict[Hashable, str] | list[str] | None = None,
  ) -> Self:
  ```

  This causes a type error when passing a `dict[str, str]`:

  ```python
  path_map: dict[str, str] = {"yes": "node_a", "no": "node_b"}
  graph.add_conditional_edges("source", my_condition, path_map)
  # pyright error: dict[str, str] is not assignable to dict[Hashable, str]
  ```

  #### Why This Happens

  In Python's type system, `dict` is **invariant** in its key type. Even though `str` is a subtype of `Hashable`, `dict[str, str]` is not a
  subtype of `dict[Hashable, str]`.

  This is by design — if it were allowed, one could insert non-`str` keys (like `int`) through a `dict[Hashable, str]` reference, breaking the
  original type constraint.

  #### Proposed Solution

  Use a generic type parameter to link the return type of `path` with the key type of `path_map`:

  ```python
  from typing import TypeVar
  from collections.abc import Hashable

  K = TypeVar("K", bound=Hashable)

  def add_conditional_edges(
      self,
      source: str,
      path: Callable[..., K],
      path_map: dict[K, str] | list[str] | None = None,
  ) -> Self:
  ```

  This way:

  - When `path` returns `str`, `path_map` expects `dict[str, str]`
  - When `path` returns an `Enum`, `path_map` expects `dict[ThatEnum, str]`
  - Type safety is preserved

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 25.1.0: Mon Oct 20 19:33:00 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6020
> Python Version:  3.12.12 (main, Oct 31 2025, 23:21:56) [Clang 21.1.4 ]

Package Information
-------------------
> langchain_core: 1.1.1
> langsmith: 0.4.55
> langgraph_sdk: 0.2.12

> langserve

Other Dependencies
------------------
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.4
> packaging: 25.0
> pydantic: 2.12.5
> pyyaml: 6.0.3
> requests: 2.32.5
> requests-toolbelt: 1.0.0
> tenacity: 9.1.2
> typing-extensions: 4.15.0
> uuid-utils: 0.12.0
> zstandard: 0.25.0

debu-sinha:
## Summary

Fixes the type signature incompatibility in `add_conditional_edges` where `dict[str, str]` could not be assigned to `dict[Hashable, str]` due to Python's type invariance.

## Problem

The current type signature uses `dict[Hashable, str]` for the `path_map` parameter:

```python
def add_conditional_edges(
    self,
    source: str,
    path: Callable[..., Hashable | Sequence[Hashable]],
    path_map: dict[Hashable, str] | list[str] | None = None,
) -> Self:
```

This causes pyright/mypy errors when passing a `dict[str, str]`:

```python
path_map: dict[str, str] = {"yes": "approved", "no": "rejected"}
graph.add_conditional_edges("__start__", check_approval, path_map)
# Error: dict[str, str] is not assignable to dict[Hashable, str]
```

## Solution

Introduced a `TypeVar` bound to `Hashable` that links the return type of `path` with the key type of `path_map`:

```python
_PathReturnT = TypeVar("_PathReturnT", bound=Hashable)

def add_conditional_edges(
    self,
    source: str,
    path: Callable[..., _PathReturnT | Sequence[_PathReturnT]]
        | Callable[..., Awaitable[_PathReturnT | Sequence[_PathReturnT]]]
        | Runnable[Any, _PathReturnT | Sequence[_PathReturnT]],
    path_map: dict[_PathReturnT, str] | list[str] | None = None,
) -> Self:
```

This allows:
- `dict[str, str]` when `path` returns `str`
- `dict[MyEnum, str]` when `path` returns `MyEnum`
- Type safety is preserved since `_PathReturnT` is bound to `Hashable`

## Testing

Verified the fix works with the example from the issue:

```python
from typing import TypedDict
from langgraph.graph import END, StateGraph

class State(TypedDict):
    approved: bool

def check_approval(state: State) -> str:
    return "yes" if state["approved"] else "no"

graph = StateGraph(State)
graph.add_node("approved", approved_node)
graph.add_node("rejected", rejected_node)

# This now works without type errors!
path_map: dict[str, str] = {"yes": "approved", "no": "rejected"}
graph.add_conditional_edges("__start__", check_approval, path_map)
```

Closes #6540

### DIFF
--- a/libs/langgraph/langgraph/graph/_branch.py
+++ b/libs/langgraph/langgraph/graph/_branch.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Awaitable, Callable, Hashable, Sequence
+from collections.abc import Awaitable, Callable, Hashable, Mapping, Sequence
 from inspect import (
     isfunction,
     ismethod,
@@ -12,6 +12,7 @@ from typing import (
     Any,
     Literal,
     NamedTuple,
+    TypeVar,
     cast,
     get_args,
     get_origin,
@@ -36,6 +37,10 @@ _Writer = Callable[
     [Sequence[str | Send], bool],
     Sequence[ChannelWriteEntry | Send],
 ]
+
+# TypeVar for path_map key type, bound to Hashable for type safety
+# This allows dict[str, str] to be passed where the path function returns str
+_PathReturnT = TypeVar("_PathReturnT", bound=Hashable)
 
 
 def _get_branch_path_input_schema(
@@ -88,15 +93,16 @@ class BranchSpec(NamedTuple):
     @classmethod
     def from_path(
         cls,
-        path: Runnable[Any, Hashable | list[Hashable]],
-        path_map: dict[Hashable, str] | list[str] | None,
+        path: Runnable[Any, _PathReturnT | list[_PathReturnT]],
+        path_map: Mapping[_PathReturnT, str] | list[str] | None,
         infer_schema: bool = False,
     ) -> BranchSpec:
         # coerce path_map to a dictionary
         path_map_: dict[Hashable, str] | None = None
         try:
-            if isinstance(path_map, dict):
-                path_map_ = path_map.copy()
+            if isinstance(path_map, Mapping):
+                # Cast is safe: _PathReturnT is bound to Hashable
+                path_map_ = cast(dict[Hashable, str], dict(path_map))
             elif isinstance(path_map, list):
                 path_map_ = {name: name for name in path_map}
             else:
@@ -117,7 +123,12 @@ class BranchSpec(NamedTuple):
         # infer input schema
         input_schema = _get_branch_path_input_schema(path) if infer_schema else None
         # create branch
-        return cls(path=path, ends=path_map_, input_schema=input_schema)
+        # Cast is safe: _PathReturnT is bound to Hashable
+        return cls(
+            path=cast(Runnable[Any, Hashable | list[Hashable]], path),
+            ends=path_map_,
+            input_schema=input_schema,
+        )
 
     def run(
         self,
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -5,7 +5,7 @@ import logging
 import typing
 import warnings
 from collections import defaultdict
-from collections.abc import Awaitable, Callable, Hashable, Sequence
+from collections.abc import Awaitable, Callable, Hashable, Mapping, Sequence
 from functools import partial
 from inspect import isclass, isfunction, ismethod, signature
 from types import FunctionType
@@ -635,7 +635,7 @@ class StateGraph(Generic[StateT, ContextT, InputT, OutputT]):
         path: Callable[..., _PathReturnT | Sequence[_PathReturnT]]
         | Callable[..., Awaitable[_PathReturnT | Sequence[_PathReturnT]]]
         | Runnable[Any, _PathReturnT | Sequence[_PathReturnT]],
-        path_map: dict[_PathReturnT, str] | list[str] | None = None,
+        path_map: Mapping[_PathReturnT, str] | list[str] | None = None,
     ) -> Self:
         """Add a conditional edge from the starting node to any number of destination nodes.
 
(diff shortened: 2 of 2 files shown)


mgisr:
@debu-sinha Thanks for the fix! The approach of using a TypeVar to link the path return type with the path_map key type makes sense.

However, I noticed this change introduces a new type error at `state.py:672`:

https://github.com/langchain-ai/langgraph/blob/94698c8a349e2f795593ee3a94e03e0a60228240/libs/langgraph/langgraph/graph/state.py#L672

  The issue is that `BranchSpec.from_path` still expects `dict[Hashable, str]`, but now we're passing `dict[_PathReturnT, str]`. Since dict key types are invariant in Python, this causes a type mismatch:

  ```
  Argument of type "dict[_PathReturnT@add_conditional_edges, str] | list[str] | None"
  cannot be assigned to parameter "path_map" of type "dict[Hashable, str] | list[str] | None"
  ```

  To fully fix this, we might need to either:
  1. Update `BranchSpec.from_path` signature to use the same TypeVar pattern, or
  2. Use `Mapping[Hashable, str]` instead of `dict[Hashable, str]` (since `Mapping` is covariant in key type)

  Would you be able to address this? Happy to help if needed!

debu-sinha:
@mgisr Thanks for the thorough review and catching that type error! You're absolutely right.

I've pushed a fix that addresses the `state.py:672` type error. Here's the approach I took:

## Root Cause
Python's type system treats `dict` keys as **invariant**, meaning `dict[str, str]` is not assignable to `dict[Hashable, str]` even though `str` is a subtype of `Hashable`. This applies to both `dict` and `Mapping` (pyright treats `Mapping._KT` as invariant too, despite what the docs might suggest).

## Solution
Rather than trying to make the types covariant (which Python's type system doesn't fully support for mapping keys), I used `cast` at the boundaries where we convert from the generic `_PathReturnT` to concrete `Hashable`:

1. **Added `_PathReturnT` TypeVar to `_branch.py`** - Same pattern as `state.py`, bound to `Hashable`

2. **Changed `BranchSpec.from_path` signature** - Now accepts `Mapping[_PathReturnT, str]` with the TypeVar, allowing it to accept `dict[str, str]` when called from `add_conditional_edges`

3. **Used `cast` at the conversion boundary** - In `from_path`, we cast when storing to the `dict[Hashable, str]` internal type. This is safe because `_PathReturnT` is bound to `Hashable`.

4. **Changed `path_map` in `add_conditional_edges` to `Mapping`** - More flexible for input (accepts dict, OrderedDict, etc.)

The casts are safe because:
- `_PathReturnT` is explicitly bound to `Hashable`
- We're only widening the type at storage boundaries, not narrowing
- Runtime behavior is unchanged

Verified with pyright - no errors in `state.py` or `_branch.py`, and the original use case now type-checks correctly:
```python
path_map: dict[str, str] = {"yes": "approved", "no": "rejected"}
graph.add_conditional_edges("__start__", check_approval, path_map)  # ✓ No type error
```

Let me know if you have any concerns with this approach!

debu-sinha:
Closing to create a fresh PR with clean history.

mgisr:
@debu-sinha Thanks for the detailed explanation! The approach looks solid - using `cast` at storage boundaries is a reasonable workaround for Python's invariant dict keys. LGTM! 👍

PR #6547, opened 2025-12-08
Fix type signature for add_conditional_edges path_map parameter

### LINKED ISSUE #6540: `add_conditional_edges` type signature causes incompatibility with dict[str, str]

debu-sinha:
## Summary

Fixes the type signature incompatibility in `add_conditional_edges` where `dict[str, str]` could not be assigned to `dict[Hashable, str]` due to Python's type invariance.

## Problem

The current type signature uses `dict[Hashable, str]` for the `path_map` parameter, which causes pyright/mypy errors when passing a `dict[str, str]`:

```python
path_map: dict[str, str] = {"yes": "approved", "no": "rejected"}
graph.add_conditional_edges("__start__", check_approval, path_map)
# Error: dict[str, str] is not assignable to dict[Hashable, str]
```

## Solution

- Added `TypeVar` `_PathReturnT` bound to `Hashable` that links the return type of `path` with the key type of `path_map`
- Changed `path_map` to use `Mapping` instead of `dict` (more flexible for input)
- Used `cast` at boundaries to convert from generic TypeVar to concrete Hashable types (safe because `_PathReturnT` is bound to `Hashable`)

This allows:
- `dict[str, str]` when `path` returns `str`
- `dict[MyEnum, str]` when `path` returns `MyEnum`
- Type safety is preserved via TypeVar bounds

## Testing

Verified the fix works with pyright (0 errors) and at runtime:

```python
from typing import TypedDict
from langgraph.graph import END, StateGraph

class State(TypedDict):
    approved: bool

def check_approval(state: State) -> str:
    return "yes" if state["approved"] else "no"

graph = StateGraph(State)
graph.add_node("approved", approved_node)
graph.add_node("rejected", rejected_node)

# This now works without type errors!
path_map: dict[str, str] = {"yes": "approved", "no": "rejected"}
graph.add_conditional_edges("__start__", check_approval, path_map)
```

Closes #6540


debu-sinha:
Thanks @mgisr for catching the type error at `state.py:672` in the previous PR! Your analysis was spot on.

I went with `Mapping` + `cast` at the boundaries since pyright treats both `dict` and `Mapping` key types as invariant. The casts are safe because `_PathReturnT` is bound to `Hashable`.

debu-sinha:
Hi! This is a follow-up to #6543 with a cleaner approach for the type signature fix.

The PR fixes the `add_conditional_edges` path_map parameter to correctly accept both `dict` and `Mapping` types. Uses `Mapping` + `cast` at boundaries since pyright treats both key types as invariant.

Would appreciate a review when you have a moment! @hinthornw @eyurtsev

debu-sinha:
Closing this since I am not pursuing this direction anymore.

PR #6562, opened 2025-12-09
chore: Flip default in BaseCache

hinthornw:


sydney-runkle:
finished via https://github.com/langchain-ai/langgraph/pull/6677

PR #6581, opened 2025-12-12
feat(sdk-py): emit id as part of stream events

jdrogers940:
- **Description:** Provide the id of the event for routes that use SSE streams. This will allow for more custom retry logic when streams disconnect if needed.
  - **Issue:** N/A
  - **Dependencies:** N/A
  - **Twitter handle:** N/A

### DIFF
--- a/libs/langgraph/tests/test_remote_graph.py
+++ b/libs/langgraph/tests/test_remote_graph.py
@@ -567,9 +567,9 @@ def test_stream():
         stream_parts.append(stream_part)
 
     assert stream_parts == [
-        ("updates", {"chunk": "data3"}, "1"),
-        ("updates", {"chunk": "data4"}, "2"),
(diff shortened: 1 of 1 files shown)


libs/sdk-py/langgraph_sdk/sse.py
```diff
@@ -103,6 +103,7 @@ def decode(self, line: bytes) -> StreamPart | None:
             sse = StreamPart(
                 event=self._event,
                 data=orjson.loads(self._data) if self._data else None,  # type: ignore[invalid-argument-type]
+                id=self._last_event_id,
```
hinthornw:
```suggestion
                id=self.last_event_id,
```

Otherwise we'll put an empty string which I think isn't what we intended here?

hinthornw (approved):

PR #6595, opened 2025-12-15
fix(sdk-py): remove model-type specific custom json encryption annotations & document key preservation limitations

cwlbraa:
In langgraph-api, custom-encrypted JSONs need to continue to be SQL-json-mergable after encryption. Previous WIP docs advocated for custom encryption impls where all encrypted kv pairs were shoved into a `__encrypted__: $encrypted_kvs` meta-key. Turns out that pattern causes data loss when running PATCH-style partial updates or in the many places langgraph-api json-SQL-merges across model types.

This PR contains 2 SDK fixes:
1. remove model-type specific custom json encryption annotations - these cause surprising behavior as config and context data propagates across model types, specifically because today we can't guarantee that data encrypted as one model-type will be decrypted as the same model-type because kv pairs move across model-types in pure SQL
2. document limitations and validation around "key preservation" in custom json encryption functions. langgraph-api now validates that custom JSON encryption fns don't change keys. That validation prevents customizers from writing custom encryption functions that cause data loss through patch endpoints and x-model merge propagation.

### DIFF
--- a/libs/sdk-py/langgraph_sdk/encryption/__init__.py
+++ b/libs/sdk-py/langgraph_sdk/encryption/__init__.py
@@ -235,8 +235,9 @@ class Encryption:
     !!! warning "JSON Encryptors Must Preserve Keys"
 
         JSON encryptors **must not add or remove keys** from the input dict.
-        Only values may be transformed. This constraint exists because SQL JSONB
-        merge operations (used for partial updates) work at the key level.
+        Only values may be transformed. This constraint is **enforced at runtime
+        by the server** and exists because SQL JSONB merge operations (used for
+        partial updates) work at the key level.
 
         **Correct (per-key encryption):**
         ```python


libs/sdk-py/langgraph_sdk/encryption/__init__.py
```diff
@@ -336,113 +232,134 @@ class Encryption:
     Then the LangGraph server will load your encryption file and use it to
     encrypt/decrypt data at rest.
 
+    !!! warning "JSON Encryptors Must Preserve Keys"
+
+        JSON encryptors **must not add or remove keys** from the input dict.
```
jdrogers940, line 237: Any way that we can enforce this? Maybe even like a test suite folks can run against their own custom encryptor to validate best practices?

cwlbraa: yep, it's enforced at runtime on the server side. will make that clear, here.

PR #6612, opened 2025-12-19
fix: add msgpack allowlist

casparb:
adds msgpack deserialization allowlist                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                        
- add SAFE_MSGPACK_TYPES constant containing types with safe constructors . these always deserialize without warnings                                                                                                                                                                                                                                                                                                
- add allowed_msgpack_modules to JsonPlusSerializer. default is True right now but will switch to None later
  - when true, warns on unregistered types
  - when None, enter strict mode and those types are blocked                                                                                                                                                                                                                                                 
  - users can set to an explicit allowlist to silence warnings                                                                                                                                                                                                                                                                                                      
                                                                                            

beware this involves a schema change

### DIFF
--- a/libs/cli/schemas/schema.json
+++ b/libs/cli/schemas/schema.json
@@ -658,7 +658,7 @@
     },
     "EncryptionConfig": {
       "title": "EncryptionConfig",
-      "description": "Configuration for custom at-rest encryption logic.\n\n    Allows you to implement custom encryption for sensitive data stored in the database,\n    including metadata fields and checkpoint blobs.",
+      "description": "Configuration for custom at-rest encryption logic.\n\nAllows you to implement custom encryption for sensitive data stored in the database,\nincluding metadata fields and checkpoint blobs.",
       "type": "object",
       "properties": {
         "path": {
--- a/libs/cli/schemas/schema.v0.json
+++ b/libs/cli/schemas/schema.v0.json
@@ -658,7 +658,7 @@
     },
     "EncryptionConfig": {
       "title": "EncryptionConfig",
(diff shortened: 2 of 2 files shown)


libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -37,6 +37,42 @@
 EMPTY_BYTES = b""
 logger = logging.getLogger(__name__)
 
+SAFE_MSGPACK_TYPES: frozenset[tuple[str, ...]] = frozenset(
```
casparb, line 40: we should review these carefully, I'm pretty confident that these are all safe. Maybe Path types are not though

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -54,18 +90,29 @@ def __init__(
...
         self.pickle_fallback = pickle_fallback
-        self._allowed_modules = (
+        # JSON allowlist
+        self._allowed_json_modules: set[tuple[str, ...]] | Literal[True] | None = (
             {mod_and_name for mod_and_name in allowed_json_modules}
             if allowed_json_modules and allowed_json_modules is not True
             else (allowed_json_modules if allowed_json_modules is True else None)
         )
+        # Msgpack allowlist
+        self._allowed_msgpack_modules: set[tuple[str, ...]] | Literal[True] | None = (
```
casparb, line 111: decided to keep this separate from json allowlist because its legacy, but perhaps should consolidate? as old checkpoints still exist

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
casparb: we also need lc core message types

PR #6613, opened 2025-12-20
docs: clarify relationship between LangGraph and LangChain

Adirohansuyal:
This PR clarifies the distinction between LangGraph and LangChain in the README
by explaining their respective roles and how they complement each other.


Adirohansuyal:
Thanks for the review! Let me know if any wording changes are needed

Issue #6626, opened 2025-12-25
`interrupt()` calls in parallel tools generate identical IDs, making multi-interrupt resume impossible


AbhinaavRamesh:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
import asyncio
from typing import Annotated, List
from uuid import uuid4

from langchain_core.messages import AnyMessage, HumanMessage, AIMessage
from langchain_core.tools import tool
from langgraph.checkpoint.memory import MemorySaver
from langgraph.graph import END, START, StateGraph, add_messages
from langgraph.prebuilt import ToolNode
from langgraph.types import interrupt, Command
from typing_extensions import TypedDict


@tool
def create_circle(shape_id: str, color: str) -> str:
    """Create a circle on the canvas."""
    response = interrupt({"tool": "create_circle", "shapeId": shape_id, "color": color})
    return f"Created circle {shape_id}. Response: {response}"


@tool
def create_square(shape_id: str, color: str) -> str:
    """Create a square on the canvas."""
    response = interrupt({"tool": "create_square", "shapeId": shape_id, "color": color})
    return f"Created square {shape_id}. Response: {response}"


@tool
def create_triangle(shape_id: str, color: str) -> str:
    """Create a triangle on the canvas."""
    response = interrupt({"tool": "create_triangle", "shapeId": shape_id, "color": color})
    return f"Created triangle {shape_id}. Response: {response}"


class State(TypedDict):
    messages: Annotated[List[AnyMessage], add_messages]


tools = [create_circle, create_square, create_triangle]


async def agent_node(state: State):
    """Simulates an LLM returning multiple parallel tool calls."""
    return {"messages": [AIMessage(
        content="I'll create three shapes for you.",
        tool_calls=[
            {"id": "call_1", "name": "create_circle", "args": {"shape_id": "c1", "color": "red"}},
            {"id": "call_2", "name": "create_square", "args": {"shape_id": "s1", "color": "blue"}},
            {"id": "call_3", "name": "create_triangle", "args": {"shape_id": "t1", "color": "green"}},
        ],
[...]

Issue #6533, opened 2025-12-03
Interrupt resume values misrouted between tools when using a ToolNode


d4n-a:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
import asyncio
import logging
import math
import random
from typing import Annotated, TypedDict

from langchain_core.messages import AIMessage, HumanMessage
from langchain_core.tools import tool
from langgraph.checkpoint.memory import MemorySaver
from langgraph.graph import END, START, StateGraph
from langgraph.graph.message import add_messages
from langgraph.prebuilt import ToolNode
from langgraph.types import Command, interrupt

logger = logging.getLogger(__name__)


@tool
async def tool_a():
    """The tool raises Interrupt A and expects that exact value in order to continue"""
    expected = "Interrupt A"
    await delay()
    result = interrupt(expected)
    assert expected == result, f"Expected {expected}, got {result}"
    return result


@tool
async def tool_b():
    """The tool raises Interrupt B and expects that exact value in order to continue"""
    expected = "Interrupt B"
    await delay()
    result = interrupt(expected)
    assert expected == result, f"Expected {expected}, got {result}"
    return result


async def parallel_interrupts():
    """Resubmitting interrupt messages back to confirm they are channeled correctly"""
    graph = make_graph()
    config = {"configurable": {"thread_id": "test-thread-1"}}

    # First invocation to trigger both tools in parallel
    first_result = await graph.ainvoke(
        {"messages": [HumanMessage(content="Tool calls")]}, config=config, stream_mode="debug"
    )

    first_interrupt = first_result[-1]["payload"]["interrupts"][0]
    first_interrupt_id = first_interrupt["id"]
    first_interrupt_value = first_interrupt["value"]
    logger.info(f"Resuming first interrupt with id {first_interrupt_id} and value {first_interrupt_value}")

    second_result = await graph.ainvoke(
        Command(resume={first_interrupt_id: first_interrupt_value}), config=config, stream_mode="debug"
    )

    # Resume second interrupt
    second_interrupt = second_result[-1]["payload"]["interrupts"][0]
[...]

PR #6627, opened 2025-12-25
fix(langgraph): generate unique interrupt IDs for parallel tool execution

AbhinaavRamesh:
Fix for #6626 

Enabling proper interrupt handling when multiple tools run in parallel.

## Description

  Updated Interrupt handling mechanism to support multiple tools running in parallel by generating unique interrupt IDs for each concurrent execution.

  When multiple tools execute in parallel with interrupts, the resume mechanism fails because all interrupts share the same ID. This prevents proper handling of `Command(resume={id1: response1, id2: response2, id3: response3})`.

**Example of the bug:**
```python
  # Three parallel tool calls, each with interrupt()
  Interrupts collected: 3
    0: id=1b09852344aeb4aee6776a24d09c3f9e, tool=create_circle
    1: id=1b09852344aeb4aee6776a24d09c3f9e, tool=create_square  # Same ID!
    2: id=1b09852344aeb4aee6776a24d09c3f9e, tool=create_triangle # Same ID!
  # Resume dict collapses due to key collision:
  resume_data = {intr.id: value for intr in interrupts}  # Only 1 entry!
```
### Solution

  Changed the interrupt ID format to include the interrupt index:

  | Before                                 | After                              |
  |----------------------------------------|------------------------------------|
  | a61a86efc1202990625a73c5591eee22       | a61a86efc1202990625a73c5591eee22:0 |
  | (same for all interrupts in namespace) | (unique per interrupt: :0, :1, :2) |

### Changes

  - langgraph/types.py
    - Interrupt.from_ns() now accepts idx parameter
    - ID format: {hash(namespace)}:{idx}
    - interrupt() function passes idx from scratchpad counter
  - langgraph/pregel/_algo.py
    - _scratchpad() updated to match resume values with new ID format
    - Maintains backward compatibility with old-format IDs
    - Iterates through {namespace_hash}:{idx} entries in resume_map
  - langgraph/pregel/_utils.py
    - is_xxh3_128_hexdigest() regex updated to match both formats:
        - Legacy: [0-9a-f]{32}
      - New: [0-9a-f]{32}:\d+

### Backward Compatibility

  - Old checkpoints with legacy interrupt IDs continue to work
  - The _scratchpad() function checks for both old and new formats

## Test Plan

- [x] test_from_ns_unique_ids_for_different_indices - Unit test verifying unique IDs
- [x] test_multiple_interrupts_in_node_have_unique_ids - Sequential interrupts in single node
- [x] test_parallel_tool_interrupts_have_unique_ids - Integration test with parallel Send() dispatches
- [x] Existing tests continue to pass
- [x] Format and lint checks pass

### Manual verification tests passed:

- [x]  Interrupt IDs now have format {hash}:{idx}
- [x] Multiple interrupts in same node get unique IDs
- [x]  Parallel tools via Send() get unique IDs
- [x]  Resume with Command(resume={id: value}) works correctly

##  Breaking Changes
 None. This change is fully backward compatible.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_algo.py
+++ b/libs/langgraph/langgraph/pregel/_algo.py
@@ -1082,9 +1082,24 @@ def _scratchpad(
         del w
 
         # find namespace and task-specific resume value
-        if resume_map and namespace_hash in resume_map:
-            mapped_resume_write = resume_map[namespace_hash]
-            task_resume_write.append(mapped_resume_write)
+        # Interrupt IDs now have format: {namespace_hash}:{idx}
+        # So we need to match resume values by iterating through possible idx values
+        if resume_map:
+            # First check for backward compatibility: old format (namespace_hash only)
+            if namespace_hash in resume_map:
+                mapped_resume_write = resume_map[namespace_hash]
+                task_resume_write.append(mapped_resume_write)
+            else:
+                # New format: look for {namespace_hash}:{idx} entries
+                idx = 0
+                while True:
+                    interrupt_id = f"{namespace_hash}:{idx}"
+                    if interrupt_id in resume_map:
+                        mapped_resume_write = resume_map[interrupt_id]
+                        task_resume_write.append(mapped_resume_write)
+                        idx += 1
+                    else:
+                        break
 
     else:
         null_resume_write = None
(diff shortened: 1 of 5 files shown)


sydney-runkle:
Closing based on https://github.com/langchain-ai/langgraph/issues/6533#issuecomment-3739779037

PR #6631, opened 2025-12-26
feat(langgraph): Custom Injection for ToolRuntime

keenborder786:
# Add RuntimeStateField for Selective State Injection in ToolRuntime

## Overview

This PR introduces `RuntimeStateField`, a new annotation that allows tools using `ToolRuntime` to specify exactly which state field(s) they need, rather than receiving the entire graph state. This makes tool dependencies explicit in function signatures and reduces unnecessary data passing.

## Motivation

Previously, when using `ToolRuntime` to access graph state, tools would always receive the entire state object via `runtime.state`. This had several drawbacks:

- **Unclear dependencies**: Function signatures didn't indicate which parts of state a tool actually used
- **Excessive data passing**: Tools received all state data even if they only needed one or two fields
- **Harder to understand**: Reading tool code required examining the implementation to understand state dependencies

While `InjectedState` could be used for field-specific injection, the shift towards `ToolRuntime` as the recommended pattern meant these benefits were lost.

### New `RuntimeStateField` Annotation

```python
from typing_extensions import Annotated
from langgraph.prebuilt import ToolRuntime, RuntimeStateField


# Inject single field - runtime.state is just that field's value
@tool
def tool_with_messages(
    x: int,
    runtime: Annotated[ToolRuntime, RuntimeStateField("messages")]
) -> str:
    messages = runtime.state  # Just the messages list
    return f"Found {len(messages)} messages"

# Inject multiple fields - runtime.state is a dict with only those fields
@tool
def tool_with_fields(
    x: int,
    runtime: Annotated[ToolRuntime, RuntimeStateField(fields=["messages", "foo"])]
) -> str:
    messages = runtime.state["messages"]
    foo = runtime.state["foo"]
    return f"Processed with {foo}"

# No annotation - runtime.state is full state (backward compatible)
@tool
def tool_with_full_state(x: int, runtime: ToolRuntime) -> str:
    return f"Access all: {runtime.state['messages']}, {runtime.state['foo']}"### Implementation Details
```

1. **`RuntimeStateField` class**: New annotation class supporting both single field (`field="name"`) and multiple fields (`fields=["name1", "name2"]`)

2. **`_InjectedArgs` enhancement**: Added `runtime_state_field` attribute to track field specifications

3. **Detection logic**: `_get_all_injected_args` now detects `RuntimeStateField` annotations during tool initialization

4. **Injection logic**: `_inject_tool_args` creates a new `ToolRuntime` instance with filtered state when field specification is present

5. **State extraction**: New `_extract_runtime_state` method handles:
   - Single field extraction (returns field value directly)
   - Multiple field extraction (returns dict with only specified fields)
   - Dict state, BaseModel state, and list state
   - Helpful error messages for missing fields

### Issue: 
https://github.com/langchain-ai/langchain/issues/33727


keenborder786:
@eyurtsev

PR #6641, opened 2025-12-30
feat: Update types for context in astream_events

hinthornw:
Mentioned in https://github.com/langchain-ai/langgraph/issues/6046


hinthornw:
That's already covered in several other tests.

PR #6646, opened 2026-01-05
perf: remove importlib.metadata for faster version access

pdaxt:
## Summary

Removes usage of `importlib.metadata.version()` which adds ~5ms to import time. Instead, uses hardcoded version strings.

- **libs/langgraph/langgraph/version.py**: Hardcode version directly
- **libs/cli/langgraph_cli/version.py**: Re-export from `__init__.py`  
- **libs/checkpoint-postgres/.../base.py**: Import from `langgraph.version` instead of `importlib.metadata`

## Benchmark Results

| Metric | Before | After |
|--------|--------|-------|
| Import time | ~5ms | ~0.1ms |
| Improvement | - | **50x faster** |

## Test Plan

- [x] `make lint` passes on all modified packages (langgraph, cli, checkpoint-postgres)
- [x] Version imports work correctly
- [x] Benchmark confirms performance improvement

Closes #5040

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import random
 import warnings
 from collections.abc import Sequence
-from importlib.metadata import version as get_version
 from typing import Any, cast
 
 from langchain_core.runnables import RunnableConfig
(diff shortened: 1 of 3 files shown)


hinthornw:
We should update the pyproject.toml to use dynamic versioning to make it easier to have one source of truth

prandata:
Good point on dynamic versioning. Looking at the current setup, there are a few options:

1. **setuptools-scm** - derive version from git tags
2. **hatch-vcs** - similar approach, integrates well with hatch
3. **Keep `__version__` in `__init__.py`** - import it in pyproject.toml via `dynamic = ["version"]`

For this PR, I can:
- Add `[tool.setuptools.dynamic]` to pyproject.toml pointing to a `version.py` or `__init__.py`
- Update the release workflow to ensure tags are available during build

Which approach does the team prefer? Happy to implement whichever aligns with your existing tooling.

hinthornw:
hatch - we use this already in the sdk etc.

aaron-seq (commented):
The changes look good overall. The hardcoded approach is valid until dynamic versioning is implemented.

However, there's a potential issue in the checkpoint-postgres file: the version check comment was updated incorrectly. Line 34 says "skip version check if langgraph not installed or running from source", but the import approach means if langgraph isn't installed, the import will fail before reaching this exception handler.

Consider updating the comment to: "skip version check if running from source or if version cannot be determined"

The hardcoded version "1.0.5" will need manual updates on each release until the team implements hatch dynamic versioning as discussed in the conversation thread.

PR #6656, opened 2026-01-07
fix(langgraph,checkpoint): Pyodide WASM support

mdrxy:
- Add pure Python fallbacks for `xxhash` and `ormsgpack` C extensions to enable LangGraph to run in Pyodide/WASM environments (e.g., Cloudflare Python Workers)
- Make C extensions optional for performance on native Python

`langchain >= 1.0.0` rebuilt create_agent on LangGraph runtime. This pulls in `xxhash` and `ormsgpack` - C extensions with no Pyodide wheels. Users in WASM environments (Cloudflare Workers, JupyterLite) get `ModuleNotFoundError` at import time.


libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -29,7 +29,16 @@
...
-from xxhash import xxh3_128_hexdigest
+
+try:
+    from xxhash import xxh3_128_hexdigest  # type: ignore[import-not-found]
+except ImportError:
+    import hashlib
+
+    def xxh3_128_hexdigest(data: bytes) -> str:
+        """Fallback using MD5 for environments without xxhash (e.g., Pyodide/WASM)."""
+        return hashlib.md5(data, usedforsecurity=False).hexdigest()
```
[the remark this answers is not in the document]
aaron-seq: The CodeQL security check is failing because MD5 is flagged as a weak cryptographic hashing algorithm. While using `usedforsecurity=False` is correct for non-cryptographic purposes, it's not sufficient to pass the security check.

As Copilot Autofix suggests, replacing MD5 with SHA-256 would resolve the security alert:

```python
def xxh3_128_hexdigest(data: bytes) -> str:
    """Fallback using SHA-256 for environments without xxhash (e.g., Pyodide/WASM)."""
    return hashlib.sha256(data).hexdigest()
```

This change maintains the same function signature while using a more secure hashing algorithm. The same fix should be applied to both occurrences in `libs/langgraph/langgraph/pregel/_algo.py` (line 40) and `libs/langgraph/langgraph/types.py` (line 30).

libs/checkpoint/pyproject.toml
```diff
@@ -13,9 +13,13 @@ license = "MIT"
 license-files = ['LICENSE']
 dependencies = [
     "langchain-core>=0.2.38",
-    "ormsgpack>=1.12.0",
+    "msgpack>=1.0.0",
 ]
 
+[project.optional-dependencies]
```
aaron-seq, line 19: Great approach! Making these C extensions optional with a `fast` extra is the right way to maintain WASM compatibility while still allowing users to install the faster C-based packages on native Python. This follows the same pattern as other libraries and provides a clear upgrade path for users who want better performance.

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -26,7 +26,73 @@
...
+    import msgpack  # type: ignore[import-untyped]
+
+    _ORMSGPACK_AVAILABLE = False
+
+    def _pairs_hook(pairs: list[tuple[Any, Any]]) -> dict[Any, Any]:
+        """Convert list keys to tuples during dict construction (for hashability)."""
+        return {(tuple(k) if isinstance(k, list) else k): v for k, v in pairs}
+
+    # Create ormsgpack-compatible interface using msgpack
+    class _OrmsgpackCompat:
```
aaron-seq, line 43: Excellent implementation of the compatibility layer! The `_OrmsgpackCompat` class effectively mimics the ormsgpack API using msgpack as a fallback. The `_pairs_hook` function is particularly clever for handling list-to-tuple conversion during dict construction, which is necessary for hashable keys.

One minor suggestion: Consider adding a test to ensure the compatibility layer properly handles edge cases like nested data structures with list keys, to verify that the fallback behavior matches ormsgpack's behavior in WASM environments.

PR #6661, opened 2026-01-08
feat: add cleanup_old_threads methods for thread retention management

samirpatil2000:
**Description:** Add [cleanup_old_threads](cci:1://file:///Users/samirpatil/Desktop/Dev/langgraph/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py:643:4-667:18) method to PostgreSQL checkpointers for deleting expired checkpoint data based on a retention period.

LangGraph checkpoint tables (`checkpoints`, `checkpoint_blobs`, `checkpoint_writes`) grow indefinitely without cleanup. In production, this can cause database bloat, slower queries, and increased storage costs. This PR adds methods to clean up threads whose latest checkpoint is older than a specified retention period.

### Changes
- `PostgresSaver.cleanup_old_threads(retention_days)` - Sync cleanup method
- `AsyncPostgresSaver.acleanup_old_threads(retention_days)` - Async cleanup method
- `AsyncPostgresSaver.cleanup_old_threads(retention_days)` - Sync wrapper

### Usage
```python
# Delete threads older than 30 days
deleted_count = checkpointer.cleanup_old_threads(retention_days=30)
```


Nakagawa-master (commented):
There is a TOCTOU correctness issue in the current bulk-retention flow that can delete a thread which is no longer expired.

The implementation does:

```text
1. SELECT thread_ids
   GROUP BY thread_id
   HAVING MAX(checkpoint->>'ts') < cutoff

2. later, in a separate cursor:
   DELETE ... WHERE thread_id = ANY(selected_ids)
```

A concurrent writer can add a fresh checkpoint for one of those thread IDs after step 1 but before step 2. The cleanup then deletes **all** checkpoints/blobs/writes for that thread, including the fresh checkpoint that made the thread ineligible for cleanup.

So the eligibility fact is historical by the time destructive execution occurs.

A hostile regression would be:

```text
thread A latest checkpoint = 60 days old
→ cleanup selects A
→ before DELETE, another writer appends a current checkpoint to A
→ cleanup executes
→ current checkpoint must survive / A must no longer be deleted
```

I would make eligibility and deletion one atomic current-state decision: e.g. perform the cutoff re-check inside the deletion transaction/CTE (or lock the relevant thread rows so a new checkpoint cannot appear between qualification and delete). The important invariant is that the destructive action must be authorized by the state **at deletion time**, not only by an earlier snapshot.

There is also a simpler destructive-footgun: `retention_days` accepts any `int`. A negative value moves the cutoff into the future and can make nearly every normal thread eligible. This should reject non-positive/negative values explicitly (depending on the intended `0` semantics) and have a regression test.

For a bulk irreversible API, a dry-run/candidate-list path would also make operator review safer, but the race above is the correctness blocker even if the API remains intentionally destructive.

Source relation: the state-boundary framing is informed by Nakagawa Master's public structural work separating historical state from current execution conditions and preserving future options before irreversible effects. The concrete race and negative-retention finding above are specific to this PR's SQL/control flow. A relevant public return route is:
https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/HISTORICAL_FACT_REUSE_KIT.md

For the broader irreversible future-option framing:
https://github.com/Nakagawa-master/nakagawa-theory-archive/tree/main/derivatives/297

PR #6676, opened 2026-01-12
fix: change default recursion limit

sydney-runkle:
25 is pretty unreasonable for most applications, bumping up to 1000 by default, burden really should be on the user to enforce this based on their application

### DIFF
--- a/libs/langgraph/langgraph/_internal/_config.py
+++ b/libs/langgraph/langgraph/_internal/_config.py
@@ -28,7 +28,7 @@ from langgraph._internal._constants import (
     NS_SEP,
 )
 
-DEFAULT_RECURSION_LIMIT = int(getenv("LANGGRAPH_DEFAULT_RECURSION_LIMIT", "1000"))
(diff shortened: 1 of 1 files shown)


libs/langgraph/langgraph/_internal/_config.py
```diff
@@ -28,7 +28,7 @@
     NS_SEP,
 )
 
-DEFAULT_RECURSION_LIMIT = int(getenv("LANGGRAPH_DEFAULT_RECURSION_LIMIT", "25"))
+DEFAULT_RECURSION_LIMIT = int(getenv("LANGGRAPH_DEFAULT_RECURSION_LIMIT", "1000"))
```
eyurtsev: Is there any place where we can update the docs saying that the default changed and in which langgraph version?

sydney-runkle:
```suggestion
DEFAULT_RECURSION_LIMIT = int(getenv("LANGGRAPH_DEFAULT_RECURSION_LIMIT", "10000"))
```

Issue #6654, opened 2026-01-06
client.runs.stream hangs or loses real-time updates when using url=None (In-process Transport) in v0.6


kissycn:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
# This setup causes the issue since in v0.6
client = get_client(url=None)

# This works fine and provides real-time updates
client = get_client(url="http://localhost:2024")

async for chunk in client.runs.stream(
    thread_id=thread_id,
    assistant_id="my_agent",
    input=input_data,
    stream_mode="messages" # or "messages-tuple"
):
    print(chunk) # No output until the run finishes
```

### Error Message and Stack Trace (if applicable)

```shell
The stream either:
Hangs indefinitely.
Buffers all output and returns everything at once only after the run is completed.
```

### Description

After upgrading from LangGraph API 0.5 to 0.6 , the streaming functionality (client.runs.stream) no longer provides real-time updates when using the in-process connection (get_client(url=None)).

I am running the client and the LangGraph server within the same FastAPI process.
I suspect the issue might be related to how the new ASGITransport handles streaming or the default msgpack serialization in v0.6.
In v0.5, this exact setup (url=None) worked without any buffering issues.
I've tried passing headers={"Accept": "application/json"} but the issue persists with url=None.

### System Info

LangGraph API Version: 0.6.18
LangGraph SDK Version: 0.4.11
Python Version: 3.12
Operating System: MacOS

hinthornw:
This is only in `dev`? Or also with `up`/ in prod?

DineshThumma9:
This seems a issue in asgi transport which might steam from langchain_api/api/asgi_transport as its closed source i fixed using HTTP transport as a fallback mechanism as of now

thoward-godaddy:
Just ran into this as well

esis-a-i:
[...]

PR #6690, opened 2026-01-16
Fix(sdk_py):Using url=None in get_client() fails due to recent changes using HTTP Transport as a fallback for now

DineshThumma9:
**Description**
 Providing url = None should connect to inprocess as mentioned in docs from recent changes langgraph 0.6 
providing url = None gets  None is not callable error  which is cryptic ans does'nt specify real issue

- **Issue** : #6654 

**Changes**
I have have added a Better Error handling also a fallback mechanism using HTTP Transportation using a fallback url 
Added unit tests covering all cases `test_client.py` specified my changes in `get_client.py` in doc string

1. **Fail fast with clear error** (default) - guides users to use explicit URL
2. **Optional auto-fallback** (opt-in) - set `LANGGRAPH_AUTO_FALLBACK=true` to auto-use HTTP

**Note** : This still does'nt address real issue with agsi_transportaion and buffering it still needs further investigation as langchain_api is closed source i tried to improve better error handing and a fallback mechanism as of yet


hinthornw:
Hm I don't think adding `LANGGRAPH_URL` as another env var will help anyone here?

What error are you seeing in your environment?

DineshThumma9:
> Hm I don't think adding `LANGGRAPH_URL` as another env var will help anyone here?
> 
> What error are you seeing in your environment?

When provided `url = None`  i get an `None not callable error` this is with after upgradation to 0.6 before that 0.5 would work just fine if there are any actual changes how things work now its is'nt mentioned in docs atleasr

PR #6701, opened 2026-01-19
fix(checkpoint): don't add the task to the checkpoint batch if it was cancelled early

conradludgate:
I'm trying to help a customer with some issues related to their checkpointing in postgres. They have some timeouts and retries around the langgraph checkpoint queries and I suspect the queue may be filling up with cancelled tasks.

This PR adds some eager checks to not execute an operation if the future was already cancelled.
This is to prevent the queue executing tasks that might have already timed out, which would otherwise cause more tasks to timeout due to the longer execution delay.


hinthornw:
Thanks for the pr! And thanks for flagging this issue.


Could you clarify where this is coming up? The changes proposed in this pr impact the ***store**, so if they are seeing timeouts in the checkpointer, the changes proposed would have no impact.

conradludgate:
The customer has their own application timeouts around their calls to aget and astore

conradludgate:
Hey @hinthornw — to clarify, this is about the **store** (`AsyncPostgresStore` → `AsyncBatchedBaseStore`), not the checkpointer.

Users wrap store operations with application-level timeouts:

```python
result = await asyncio.wait_for(store.asearch(ns, query=q), timeout=12)
```

Under load, `_run` batches grow large → batch time exceeds timeout → `CancelledError` → caller retries → cancelled future stays in queue → next batch is even larger → feedback loop. Once triggered, it **never self-heals** without a process restart.

batch queue poisoning chart

Measured against a real Postgres instance with `AsyncPostgresStore`. Steady 5 ops/s with a brief spike. Without the fix, throughput drops to zero permanently. With the fix, it recovers within seconds.

This is from a production customer running LangGraph at scale.

hinthornw (approved):

Sourav-Nandy-ai:
Hi Conrad,

I ran a real reproduction of this rather than trusting the PR description's framing at face value. The result is more precise than simple liveness and poisoning. Using the actual pre-and-post langgraph-checkpoint code (4.0.1 vs 4.0.2, confirmed via source diff since the fix lives in the checkpoint package and not langgraph itself), I tested two specific things.

The Results
- Liveness Metric: The literal completed, cancelled, and failed liveness metric came back identical on both versions (13/7/0 across all 30 trials). I observed no throughput degradation or stall reproduced under a well-behaved in-memory store.
- Execution Mechanism: A direct test of the actual mechanism (checking if a cancelled aput() still executes against the store) found a real, 100%-reproducible bug. Pre-fix, every single cancelled write across 210 opportunities still landed in the store. Post-fix, zero did.

The fix is real and total for the "wasted and incorrect execution" framing. Cancelled work no longer silently executes. I observed this even though I could not reproduce a caller-visible hang under this specific workload. That symptom might require a backend store whose abatch() reacts differently to stale ops, but I did not fabricate that scenario since I do not have one to test against.

I maintain agent-eval (agent_regress). This is exactly the class of concurrency bug that concurrent_cancellation_probe() and compare_liveness() exist to catch. You can find my full scripts and raw output here: https://github.com/RudrenduPaul/agent-eval.

Given you found the original bug, I am curious whether this ghost-write framing matches what you actually observed in production, or if there is a specific backend or workload where the liveness symptom shows up that I should test next.

Best,
Sourav

PR #6707, opened 2026-01-21
feat: support dynamic tools via callable in tool node

sydney-runkle:
Add support for dynamic tools in `ToolNode`. Instead of passing a static list of tools, you can now pass a callable  
that returns a list of tools. The callable is invoked on each `ToolNode` invocation, allowing the available tools to change    
between calls. 

This is important because it enables dynamic registration of tools for agents created w/ `create_agent`
                                                     
Key changes:                                                                                                       
- ToolNode tools parameter now accepts `Sequence[BaseTool | Callable] | Callable[[], Sequence[BaseTool]]`       
- Tools provider is detected by checking for a callable with no required arguments
- Dynamic providers must return `BaseTool` instances (not plain functions)    
- All existing features (state injection, error handling, etc.) work with dynamic tools

Still TODO
* Look into optimizations, especially re redundant signature inspection when the list of tools is static :)

Sample functionality :)
                                                                                                                   
```py                                                                                                 
from langgraph.prebuilt import ToolNode                                                                            
                                                                                                                   
@tool                                                                                                              
def add(a: int, b: int) -> int:                                                                                    
    """Add two numbers."""                                                                                         
    return a + b                                                                                                   
                                                                                                                   
@tool                                                                                                              
def multiply(a: int, b: int) -> int:                                                                               
    """Multiply two numbers."""                                                                                    
    return a * b                                                                                                   
                                                                                                                   
# Tools can change dynamically based on external state                                                             
available_tools = [add]                                                                                            
                                                                                                                   
def get_tools():                                                                                                   
    return available_tools                                                                                         
                                                                                                                   
# Pass a callable instead of a list                                                                                
tool_node = ToolNode(get_tools)                                                                                    
                                                                                                                   
# Initially only "add" is available                                                                                
print(tool_node.tools_by_name.keys())  # dict_keys(['add'])                                                        
                                                                                                                   
# Later, add more tools                                                                                            
available_tools.append(multiply)                                                                                   
                                                                                                                   
# Now both tools are available (no restart needed)                                                                 
print(tool_node.tools_by_name.keys())  # dict_keys(['add', 'multiply'])              
```

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/tool_node.py
+++ b/libs/prebuilt/langgraph/prebuilt/tool_node.py
@@ -655,8 +655,19 @@ class ToolNode(RunnableCallable):
 
             - **A sequence of tools**: A list/tuple of tools (static)
             - **A callable**: A function that returns a sequence of tools (dynamic).
-              The callable is invoked on each invocation of the ToolNode, allowing
-              the available tools to change between invocations.
+              The callable is invoked on every `invoke()` or `ainvoke()` call,
+              allowing the available tools to change between invocations.
+
+              **Invocation semantics for dynamic tools callables:**
+
+              - Called once at the start of each ToolNode invocation
+              - Should return a consistent set of tools for a given invocation
+                (i.e., idempotent within a single invocation)
+              - Stochastic behavior (returning different tools across invocations)
+                is supported but be aware the model may reference tools that are
+                no longer available
+              - Must return `BaseTool` instances (not plain callables) to avoid
+                expensive introspection on every invocation
 
             Each tool in the sequence supports:
 
@@ -815,6 +826,11 @@ class ToolNode(RunnableCallable):
                 if convert_callables:
                     tool_ = create_tool(cast("type[BaseTool]", tool))
                 else:
+                    # Dynamic tools providers must return BaseTool instances, not plain
+                    # callables. Converting callables to tools requires calling
+                    # create_tool() which performs introspection. Doing this on every
+                    # invocation would be expensive and could have unexpected side effects.
+                    # Users should convert their callables to tools once upfront.
                     msg = (
                         f"Dynamic tools provider must return BaseTool instances, "
                         f"got {type(tool).__name__}"
@@ -1005,10 +1021,12 @@ class ToolNode(RunnableCallable):
         """Execute tool call with configured error handling.
 
         Args:
-            request: Tool execution request.
+            request: Tool execution request (includes the specific tool to execute).
             input_type: Input format.
             config: Runnable configuration.
-            tools_by_name: Mapping from tool name to BaseTool.
+            tools_by_name: Mapping from tool name to BaseTool. Used only for
+                validation error messages when the requested tool doesn't exist,
+                to list available tools in the error response.
 
         Returns:
             ToolMessage or Command.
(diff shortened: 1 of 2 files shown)


libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -735,7 +738,7 @@ def handle_errors(e: ValueError) -> str:
 
     def __init__(
         self,
-        tools: Sequence[BaseTool | Callable],
+        tools: Sequence[BaseTool | Callable] | Callable[[], Sequence[BaseTool]],
```
eyurtsev: Could we add async?

libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -895,13 +1000,15 @@ def _execute_tool_sync(
         request: ToolCallRequest,
         input_type: Literal["list", "dict", "tool_calls"],
         config: RunnableConfig,
+        tools_by_name: dict[str, BaseTool],
```
eyurtsev, line 1019: nit: why pass tools_by_name vs. the specific tool that's in the request? do we need this for retries? (what actually results in cleaner code?)

libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -287,7 +287,10 @@ def create_react_agent(
         [StateSchema, Runtime[ContextT]],
         Awaitable[Runnable[LanguageModelInput, BaseMessage]],
     ],
-    tools: Sequence[BaseTool | Callable | dict[str, Any]] | ToolNode,
+    tools: Sequence[BaseTool | Callable | dict[str, Any]]
+    | Callable[[], Sequence[BaseTool]]
```
eyurtsev, line 291: Could we leave a developer facing comment to explain why limit to BaseTool instead of union[BaseTool|Callable]? (i.e. to prevent developers from adding that feature)

libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
```diff
@@ -355,8 +358,9 @@ def select_model(state: AgentState, runtime: Runtime[ModelContext]) -> ChatOpenA
                 `.bind_tools()` and support required functionality. Bound tools
                 must be a subset of those specified in the `tools` parameter.
 
-        tools: A list of tools or a `ToolNode` instance.
+        tools: A list of tools, a `ToolNode` instance, or a callable that returns tools.
```
eyurtsev, line 361: I would add information here about when it gets called. (e.g., on every invocation of invoke) -- i.e., what semantics are expected for the list of tools that we get from tools?

e.g., is it expected to be idempotent? is stochastic OK?

eyurtsev (approved):
Looks good -- if you could perhaps review w/ test coverage for any code that hasn't been tested yet -- ideally we have nearly 100% coverage here

sydney-runkle:
Yep, 92% cov on `tool_node.py`

libs/prebuilt/langgraph/prebuilt/tool_node.py
sydney-runkle: included for error messages

PR #6708, opened 2026-01-21
test(algo): Add missing test coverage for prepare_next_tasks

davidfertube:
## Summary

Adds a new test case `test_prepare_next_tasks_with_processes` to `libs/langgraph/tests/test_algo.py`.

## Detail

This PR addresses the `TODO: add more tests` comment in `test_algo.py` by adding a test case that verifies `prepare_next_tasks` behavior when processes are present. This improves the test coverage for the core graph preparation logic.

## Verification

The test uses `unittest.mock` to simulate `PregelNode` and verifies the function output is consistent with expectations.


hinthornw:
I'm not sure this actually is helpful since this code path is tested very heavily in all the pregel tests

PR #6711, opened 2026-01-22
feat: support dynamic tool calling via `tool` override in `wrap_model_call`

sydney-runkle:
Allow overriding `tool` impl in `wrap_tool_call` in order to support dynamically registered tools via middleware


libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -121,6 +121,7 @@ class _ToolCallRequestOverrides(TypedDict, total=False):
     """Possible overrides for ToolCallRequest.override() method."""
 
     tool_call: ToolCall
+    tool: BaseTool | None
```
eyurtsev: If we're overriding, should we allow `None` ?

PR #6742, opened 2026-02-03
feat(sdk-py): add update method for crons client

rx5ad:
# Description
- Update Python SDK to add patch method to cron client
- This method lets users modify cron attributes (except assistant ID and thread ID)

### DIFF
--- a/libs/cli/langgraph_cli/schemas.py
+++ b/libs/cli/langgraph_cli/schemas.py
@@ -179,6 +179,8 @@ class CheckpointerConfig(TypedDict, total=False):
 
     This configuration requires server version 0.5 or later to take effect.
     """
+    sweep_limit: int | None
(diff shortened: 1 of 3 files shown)


libs/sdk-py/langgraph_sdk/client.py
```diff
@@ -3180,6 +3180,82 @@ async def delete(
         """
         await self.http.delete(f"/runs/crons/{cron_id}", headers=headers, params=params)
 
+    async def patch(
```
rx5ad: Used `patch` to keep it consistent with other existing methods

hinthornw: In the other resources, we use `update` for the method name; thoughts on doing the same here?

rx5ad: Yes, changed to `update`

Issue #6735, opened 2026-02-01
Graph Compilation Silently Ignores Unreachable Nodes and Dead Ends


f4roukb:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Reproduction Steps / Example Code (Python)

```python
from typing import TypedDict, Annotated
import operator
from langgraph.graph import StateGraph, START, END


# =============================================================================
# ISSUE 1: Unreachable Node - silently never executes
# =============================================================================
print("=" * 60)
print("ISSUE 1: Unreachable Node")
print("=" * 60)

class State1(TypedDict):
    messages: Annotated[list[str], operator.add]


def important_processing(state: State1) -> dict:
    print("  -> important_processing executed")
    return {"messages": ["processed"]}


def orphan_node(state: State1) -> dict:
    """This node should process critical logic but is never reachable!"""
    print("  -> orphan_node executed")  # This will NEVER print!
    return {"messages": ["orphan was here"]}


graph1 = StateGraph(State1)
graph1.add_node("process", important_processing)
graph1.add_node("orphan", orphan_node)  # <- Forgot to connect this!
graph1.add_edge(START, "process")
graph1.add_edge("process", END)

# Compiles without any error or warning!
compiled1 = graph1.compile()
print("Graph compiled successfully (no warning about orphan node)")

result1 = compiled1.invoke({"messages": []})
print(f"Result: {result1}")
print("Notice: 'orphan_node executed' was never printed - silent bug!")


# =============================================================================
# ISSUE 2: Dead End Node - no explicit path to END (possible design mistake)
# =============================================================================
print("\n" + "=" * 60)
print("ISSUE 2: Dead End Node (sink without explicit path to END)")
print("=" * 60)

class State2(TypedDict):
    value: Annotated[list[str], operator.add]


def main_path(state: State2) -> dict:
    print("  -> main_path executed")
    return {"value": ["main"]}


def dead_end(state: State2) -> dict:
    """This node executes but has no explicit edge to END."""
    print("  -> dead_end executed")
[...]

PR #6751, opened 2026-02-05
fix(langgraph): add compile-time detection of detached nodes

saakshigupta2002:
# Pull Request: Compile-time Detection of Detached Nodes

## Summary

This PR introduces optional validation to detect and handle detached nodes during graph compilation. Detached nodes are either:
- **Unreachable nodes**: Nodes that are not reachable from `START`
- **Dead-end nodes**: Nodes that have no path to `END`

These scenarios often indicate bugs in graph construction that are currently silently ignored, leading to hard-to-debug issues in production.

### New Parameter: `on_detached_nodes`

Added a new `on_detached_nodes` parameter to `StateGraph.compile()` with three modes:

| Mode | Behavior |
|------|----------|
| `"warn"` (default) | Log a warning but compile successfully |
| `"raise"` | Raise a `ValueError` and fail compilation |
| `"ignore"` | Disable detection (useful for dynamic routing) |

### Implementation Details

Detection is implemented in `CompiledStateGraph._check_detached_nodes()`, which leverages the existing compiled channel/writer infrastructure rather than building a separate graph representation.

After `attach_node`, `attach_edge`, and `attach_branch` populate the channel system, the method:

1. **Scans all nodes' writers** to collect every channel that is written to via `ChannelWriteEntry` and `ChannelWriteTupleEntry.static` declarations
2. **Detects unreachable nodes** by checking if any writer targets the node's `branch:to:{name}` trigger channel — if not, nothing routes to this node
3. **Detects dead-end nodes** by checking if a node has any outgoing connection: writers to `branch:to:*` channels, explicit edges to `END` in the builder, conditional branches, or Command destinations declared in `static`

This approach is idiomatic to LangGraph's internal architecture and avoids duplicating connectivity information that the channel system already captures.

### Example Usage

```python

class State(TypedDict):
    messages: list[str]

graph = StateGraph(State)
graph.add_node("process", process_fn)
graph.add_node("orphan", orphan_fn)  # Forgot to connect this!
graph.add_edge(START, "process")
graph.add_edge("process", END)

# Default behavior: warns about detached nodes but compiles
compiled = graph.compile()  # Logs: "Found unreachable node(s): ['orphan']"

# Strict mode: fails compilation
compiled = graph.compile(on_detached_nodes="raise")  # Raises ValueError

# For dynamic routing: disable detection
compiled = graph.compile(on_detached_nodes="ignore")  # No warnings
```

### Important Notes for Dynamic Routing

For nodes that use `Command` objects or conditional edges without explicit mappings, the destinations cannot be detected at compile time. In such cases:

1. Use the `destinations` parameter when adding nodes:
   ```python
   graph.add_node(
       "router",
       router_fn,
       destinations={"target_a": "Label A", "target_b": "Label B"}
   )
   ```

2. Or set `on_detached_nodes="ignore"` to disable validation.

## Test Plan

Added comprehensive test coverage for:

- [x] Single unreachable node detection
- [x] Multiple unreachable nodes detection
- [x] Sink node without path to END detection
- [x] Warning mode logs warning but compiles
- [x] Ignore mode suppresses warnings
- [x] Conditional edges with explicit mapping
- [x] Conditional edges without mapping (dynamic routing)
- [x] Command destinations (dict format)
- [x] Command destinations (tuple format)
- [x] Multi-source waiting edges
- [x] Diamond pattern with conditional routing
- [x] Cyclic graphs with conditional exit
- [x] Minimal single node graph
- [x] Command with `goto=END`
- [x] Parallel branches all reaching END
- [x] Mixed unreachable and sink nodes

## Backwards Compatibility

This change is **fully backwards compatible**:
- Default behavior is `"warn"`, which only logs warnings
- Existing code continues to work without modification
- No breaking changes to the API

## Related Issue

Fixes: https://github.com/langchain-ai/langgraph/issues/6735

---

## Files Changed

| File | Changes |
|------|---------|
| `libs/langgraph/langgraph/graph/state.py` | Added `_check_detached_nodes()` to `CompiledStateGraph`, added `on_detached_nodes` parameter to `compile()` |
| `libs/langgraph/tests/test_pregel.py` | Added 16 test cases |

### DIFF
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -4,9 +4,9 @@ import inspect
 import logging
 import typing
 import warnings
-from collections import defaultdict, deque
+from collections import defaultdict
 from collections.abc import Awaitable, Callable, Hashable, Sequence
-from functools import cached_property, partial
+from functools import partial
 from inspect import isclass, isfunction, ismethod, signature
 from types import FunctionType
 from types import NoneType as NoneType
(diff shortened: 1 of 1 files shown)


hinthornw:
It should be enough to check the channels if you wanted

saakshigupta2002:
> It should be enough to check the channels if you wanted

Hey, thanks for taking a look! :) That's a great point! I hadn't considered using the channels directly for this. Since the branch:to:{node} channels already capture the connectivity after edges and branches are attached, checking which channels have no writers (unreachable) and which nodes don't write to any outgoing channel (dead ends) would be way cleaner than building a separate adjacency list and doing BFS on top of it.

I'll rework the implementation to move the detection into the compiled graph stage where the channel info is available. Should keep things a lot simpler and more in line with how the rest of the codebase works. Will push an update soon!

PR #6765, opened 2026-02-08
refactor: replace bare except with BaseException in AsyncQueue

llukito:
## Description
Replaced a bare `except:` with `except BaseException:` in `libs/langgraph/langgraph/_internal/_queue.py`.

## Motivation
Using a bare `except:` violates PEP 8 (E722). While functionally equivalent to `except BaseException:`, making it explicit improves code readability and satisfies static analysis tools.

This ensures that `asyncio.CancelledError` (which inherits from `BaseException`) is still caught and handled correctly by the cancellation logic in the `wait()` method, but without the ambiguity of a bare except.

### DIFF
--- a/libs/langgraph/langgraph/_internal/_queue.py
+++ b/libs/langgraph/langgraph/_internal/_queue.py
@@ -25,7 +25,7 @@ class AsyncQueue(asyncio.Queue):
             self._getters.append(getter)
             try:
                 await getter
-            except:
+            except BaseException:
                 getter.cancel()  # Just in case getter is not done yet.
                 try:
                     # Clean self._getters from canceled getters.


hinthornw:
Your motivation is contradictory, right?

llukito:
Good point, my description was definitely confusing. 

To clarify, this isn't just about style, it's about correctness. This block acts as cleanup logic (like a finally) that must run even during cancellation. Since asyncio.CancelledError inherits from BaseException (not Exception), we have to catch BaseException to ensure the queue state and getter futures are reset properly. 

Crucially, because we reraise immediately, we aren't suppresing the error, just intercepting it for cleanup. This mirrors the behavior in the standard library's asyncio.Queue.

hinthornw:
I know the difference, and I agree that we should avoid BareExcept generally.

PR #6767, opened 2026-02-09
feat: add input validators at graph compile

jdrogers940:
- **Description:** Adding an optional `input_validator` parameter on graph compilation that will allow for custom input validation. Also adding a specific helper for messages to validate that we only support append only messages. Leaves room for us to support additional input types down the road.
  - **Issue:** Solves the problem of a user who should have permissions to call a graph doing something to the previous history of the messages.
  - **Dependencies:** None
  - **Twitter handle:** N/A


libs/langgraph/tests/test_messages_state.py
```diff
@@ -20,7 +20,7 @@
 from langgraph.graph.state import StateGraph
 from tests.messages import _AnyIdHumanMessage
 
-_, CORE_MINOR, CORE_PATCH = (
+CORE_MAJOR, CORE_MINOR, CORE_PATCH = (
```
jdrogers940, line 23: Unrelated - but wasn't running since 1.0.0 and it should be.

sydney-runkle:
Closing as not planned for now

PR #6774, opened 2026-02-10
chore: server runtime type

hinthornw:
Main jtbd here:
a) clarify who/how a graph is being accessed and make the factory aware of the `context` where relevant (and make it obvious when it is available)
b) make it more clear when you can bypass  / defer resources with expensive  lifespans (like MCp connections)
c) make auth access more type-safe.

Gives us room to add other information, like:
- langsmith distributed tracing information

---- old
Can start doing things like this:

```
def my_graph(runtime: ServerRuntime):
     if runtime.ensure_user().permissions not in ("foo"):
         raise ValueError("bar")
```
etc.

Points of expected confusion:
- You won't have a stream_writer in this context.
- This won't be an accessible object within the graph, only the graph factory.

For maintainers, related draft PR int he server https://github.com/langchain-ai/langgraph/pull/6774

### DIFF
--- a/libs/langgraph/langgraph/_internal/_queue.py
+++ b/libs/langgraph/langgraph/_internal/_queue.py
@@ -25,7 +25,7 @@ class AsyncQueue(asyncio.Queue):
             self._getters.append(getter)
             try:
                 await getter
-            except:
+            except BaseException:
                 getter.cancel()  # Just in case getter is not done yet.
                 try:
                     # Clean self._getters from canceled getters.


libs/langgraph/langgraph/server_types.py
```diff
@@ -0,0 +1,97 @@
...
+class ServerRuntime(Runtime):
+    """Runtime context passed to graph builder factories within the Agent Server.
+
+    Requires version 0.7.29 or later of the agent server.
+
+    Extends the base Runtime with server-specific information about
+    the authenticated user and the reason the graph factory is being called.
+
+    !!! warning "Beta"
+        This API is in beta and may change in future releases.
```
sydney-runkle: Love

sydney-runkle (commented):
Am I understanding that the information that we'd want to access here would be:

* runtime context
* user info
* access context

I presume store and stream writer wouldn't be accessible, but maybe I'm not quite understanding where this would be plumbed through.

If so, might make sense to have a different data structure with just these? Specifically I'm a little concerned about making "runtime" accessible before runtime...

Not fully opposed to this though, I've been thinking about an `AgentRuntime` concept for `create_agent` :).

libs/sdk-py/langgraph_sdk/runtime.py
```diff
@@ -0,0 +1,230 @@
...
+    "assistants.get_graph",
+    "assistants.get_subgraphs",
+    "assistants.get_schemas",
+    "threads.get_state",
+    "threads.get_state_history",
+]
+
+
+@dataclass(kw_only=True, slots=True, frozen=True)
+class _ServerRuntimeBase(Generic[ContextT]):
```
sydney-runkle, line 37: why is this one generic on context t?

sydney-runkle (commented):
Do we want to call it `XXXRuntime`? `ServerContext` probably even more confusing?

Lots of the terms we use are overloaded, just wanted to check.

It makes me feel better that we're not inheriting from langgraph's runtime because for the read cases we'd have to manually construct a `Runtime` instance which feels wrong.

libs/sdk-py/langgraph_sdk/runtime.py
```diff
@@ -0,0 +1,230 @@
...
+      include in the prepared view of the latest checkpoint and how to
+      process subgraphs.
+    - `threads.get_state_history` — same as above, for historical states.
+    """
+
+    user: BaseUser | None = field(default=None)
+    """The authenticated user, or `None` if no custom auth is configured."""
+
+    store: BaseStore | None = field(default=None)
+    """Store for the graph run, enabling persistence and memory."""
```
sydney-runkle: why would it be `None`?

libs/sdk-py/langgraph_sdk/runtime.py
```diff
@@ -0,0 +1,230 @@
...
+
+    permissions: list[str] = field(default_factory=list)
+    """Permissions associated with the authenticated user.
+
+    Empty list when no user is authenticated.
+    """
+
+    @property
+    def execution_runtime(self) -> _ExecutionRuntime[ContextT] | None:
```
sydney-runkle, line 98: is this method actually helpful? maybe would be helpful to see usage somewhere

hinthornw: Very useful. I show examples below in the docstring. We do this in a few different projects already using our n on-type-safe equivalents

libs/sdk-py/langgraph_sdk/runtime.py
hinthornw: It wouldn't, just order. Gonna flip things around

libs/sdk-py/langgraph_sdk/runtime.py
hinthornw: Lol should only be the execution one

hinthornw: Ah jk. want it for execution_runtime() to show the return type. Can also create a sibling to execution context butI think I prefer to just keep onthe base for now.

libs/sdk-py/langgraph_sdk/runtime.py
hinthornw: Replaced!

PR #6786, opened 2026-02-11
optimize: remove importlib.metadata for generic version access

Sid-V5:
Addressing issue #5040 by removing the usage of `importlib.metadata` for retrieving the package version, which can be slow. Instead, the version is now hardcoded as `1.0.8`, matching the value in `pyproject.toml`.
Changes
- Modified `libs/langgraph/langgraph/version.py` to replace dynamic version retrieval with a static string.
Verification
- Verified locally that `from langgraph.version import __version__` returns `1.0.8`.


hinthornw:
Need to add linting or dynamic versioning to keep in sync.

hinthornw:
Can update to be similar to the CLI implementation. If you do then I'll review, otherwise we'll close in a few days.

Sid-V5:
thanks for the feedback @hinthornw! 
updated the PR to match the CLI package's approach:

1. `version.py` now has the hardcoded `__version__ = "1.0.8"` (single source of truth)
2. `pyproject.toml` switched from `version = "1.0.8"` to `dynamic = ["version"]` with `[tool.hatch.version] path = "langgraph/version.py"` so Hatch reads the version from the source file at build time

this mirrors exactly how `libs/cli` handles versioning (`langgraph_cli/__init__.py` + dynamic hatch version), keeping everything in sync without `importlib.metadata` overhead.

libs/langgraph/langgraph/version.py
```diff
@@ -1,12 +1,3 @@
 """Exports package version."""
 
-from importlib import metadata
-
-__all__ = ("__version__",)
```
hinthornw, line 5: keep

Sid-V5:
restored `__all__` in `version.py`

PR #6803, opened 2026-02-14
fix: remove use of importlib.metadata for version access

jackjin1997:
Closes #5040. Replaced importlib.metadata.version with direct imports from version.py modules where possible, with fallbacks for optional dependencies. This improves startup performance by avoiding the overhead of importlib.metadata.


hinthornw:
more than 5 people have implemented this incorrectly already. We need dynamic versioning to publish and not let them get out of sync, and we shouldn't be doing all these fallbacks.

Issue #6818, opened 2026-02-15
Deterministic checkpoint resume contract across crash boundaries


davidahmann:
## Problem
Checkpoint resume behavior needs deterministic guarantees after crash/restart boundaries for durable agent workloads.

## Expected behavior
- Resumed runs should produce the same terminal state and step outcomes as uninterrupted runs for equivalent inputs.

## Acceptance criteria
- Add deterministic crash/resume coverage for core checkpoint path.
- Ensure state and key run events remain consistent across resume.

## Validation
- New crash/resume tests comparing resumed vs uninterrupted outcomes.
- Existing checkpoint test suite remains green.

dlindb:
Is the intended contract here outcome parity after resume, or the narrower guarantee that each resumed step is still derived from the same checkpoint context that authorized it before the crash?

sunfishloop:
This is a great spec — deterministic crash/resume is one of those things that sounds straightforward until you have to actually prove it across different runtimes.

One pattern that's been useful in production: wallet-seeded deterministic checkpoint IDs. Each agent carries a keypair from /llms.txt (ETH/SOL/BTC). Before each checkpoint write, the agent hashes (session_id + step_number) with its wallet seed, producing a deterministic checkpoint handle. On resume, the same seed produces the same handle — so the runtime can verify: (1) this is the correct checkpoint for this continuation, and (2) no tampering occurred between write and resume.

The benefit: crash/resume determinism becomes cryptographically provable rather than just structurally enforced. The same wallet seed also lets you correlate checkpoint chains across framework boundaries (e.g., LangGraph agent resumes a task delegated by an AutoGen orchestrator using the same wallet identity).

We open-sourced this at github.com/sunfishloop/sunfishloop — the deterministic checkpoint signing runs across 20+ agents in production today.

PR #6819, opened 2026-02-15
tests: add checkpoint resume parity regression

davidahmann:
## Problem
Checkpointed crash/resume paths needed explicit parity coverage against uninterrupted runs.

## What changed
- Added `test_checkpoint_resume_matches_uninterrupted_outcome` in `libs/langgraph/tests/test_pregel.py`.
- Test simulates one crash, resumes from checkpoint, and asserts final output/state equals uninterrupted execution.

## Validation
- `cd libs/langgraph && make format`
- `cd libs/langgraph && make lint`
- `cd libs/langgraph && TEST=tests/test_pregel.py::test_checkpoint_resume_matches_uninterrupted_outcome make test`

Refs #6818


davidahmann:
Added a crash/resume parity regression test that proves resumed checkpointed execution reaches the same terminal output/state as uninterrupted execution for the same workflow and input.

This contribution was informed by patterns from Gait: https://github.com/davidahmann/gait

sydney-runkle:
Could you add an async test as well? ty!

Issue #6821, opened 2026-02-16
checkpoint/memory: select latest checkpoint by insertion order (not lexicographic id)


davidahmann:
## Problem
`InMemorySaver.get_tuple()` selects the latest checkpoint using lexicographic max checkpoint ID. This can return the wrong checkpoint when IDs are not lexicographically sortable (for example `"2"` vs `"10"`), which can cause resume/replay divergence.

## Expected behavior
Latest checkpoint selection should follow write/insertion order, independent of checkpoint ID string collation.

## Scope (Phase 1)
- Update latest-checkpoint selection logic in memory checkpointer.
- Add regression test proving non-lexicographic IDs still resolve to last written checkpoint.

## Acceptance criteria
- `get_tuple()` without explicit checkpoint ID returns last written checkpoint.
- Regression test fails with lexicographic behavior and passes with insertion-order behavior.

wenhaoone:
I have a minimal fix ready for this and can open the PR once assigned.

Prepared branch:
https://github.com/onyx679/langgraph/tree/codex/memory-saver-insertion-order

Commit:
`6a0035e fix memory saver latest checkpoint selection`

Summary:

- Change `InMemorySaver.get_tuple()` to select the last written checkpoint by insertion order instead of lexicographic `max(checkpoints.keys())`.
- Add a regression test proving ids `"2"` then `"10"` return `"10"` when no explicit checkpoint id is provided.

TDD / validation from `libs/checkpoint`:

```text
# Red before implementation:
python -m pytest tests\test_memory.py::test_get_tuple_without_checkpoint_id_returns_last_written_checkpoint -q
# failed: assert '2' == '10'

# Green after implementation:
python -m pytest tests\test_memory.py::test_get_tuple_without_checkpoint_id_returns_last_written_checkpoint -q
# 1 passed

python -m pytest tests\test_memory.py -q
# 17 passed

python -m ruff check langgraph\checkpoint\memory\__init__.py tests\test_memory.py
# All checks passed!

python -m py_compile langgraph\checkpoint\memory\__init__.py tests\test_memory.py
# passed

git diff --check
# passed
```

Could a maintainer assign this issue to me if this scope looks acceptable? I have not opened the PR yet because external PRs should wait for assignment.

IgorGanapolsky:
**Reliability note**

Production automation failures usually reduce to a **contract** issue before a model/prompt issue:

1. **Success criteria** — durable artifact id (message/booking/row), not only node green
[...]

PR #6822, opened 2026-02-16
checkpoint/memory: use insertion order for latest checkpoint selection

davidahmann:
## Problem
`InMemorySaver.get_tuple()` used lexicographic `max(checkpoint_id)` when no explicit checkpoint id was provided, which can select the wrong "latest" checkpoint for IDs like `2` then `10`.

## What changed
- Switched latest selection to insertion order via `next(reversed(checkpoints))`.
- Added regression test proving last-written checkpoint is returned for non-lexicographic IDs.

## Validation
- `cd libs/checkpoint && make format`
- `cd libs/checkpoint && make lint`
- `cd libs/checkpoint && make test`

Refs #6821

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/memory/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/memory/__init__.py
@@ -250,22 +250,22 @@ class InMemorySaver(
                     and checkpoint_ns != config_checkpoint_ns
                 ):
                     continue
+                before_checkpoint_id = get_checkpoint_id(before) if before else None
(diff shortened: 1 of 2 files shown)


davidahmann:
Implemented deterministic latest-checkpoint selection for `InMemorySaver.get_tuple()` based on insertion order rather than checkpoint ID collation.

Scope in this PR:
- Replaced lexicographic max-ID selection with last-written selection.
- Added regression coverage for `"2"` then `"10"` checkpoint IDs.

This contribution was informed by patterns from Gait: https://github.com/davidahmann/gait

PR #6863, opened 2026-02-18
fix: sequential interrupt handling w/ functional API

sydney-runkle:
In the original proposed fix https://github.com/langchain-ai/langgraph/pull/6863, we were guarding this specific case by only checking for null resume in the scratchpad.

But I noticed the tests passed on main for the sync case, indicating an inconsistency between the two paths.

Seems like we were making a redundant put_writes call when replaying the async tasks

### DIFF
--- a/libs/langgraph/tests/test_pregel_async.py
+++ b/libs/langgraph/tests/test_pregel_async.py
@@ -8012,6 +8012,55 @@ async def test_multiple_tasks_before_interrupt_resume(
 
 
 @NEEDS_CONTEXTVARS
+async def test_no_redundant_put_writes_for_cached_task(
+    async_checkpointer: BaseCheckpointSaver,
+) -> None:
+    """Cached @tasks on resume must not trigger redundant put_writes."""
+    from unittest.mock import patch
+
+    from langgraph.pregel._loop import PregelLoop
+
+    @task
+    async def setup(x: int) -> int:
+        return x
+
+    @task
+    async def ask(question: str) -> str:
+        return interrupt(question)
+
+    @entrypoint(checkpointer=async_checkpointer)
+    async def workflow(x: int) -> dict:
+        n = await setup(x)
+        answer = await ask(f"q{n}")
+        return {"answer": answer}
+
+    config = {"configurable": {"thread_id": "1"}}
+    result = await workflow.ainvoke(1, config=config)
+    assert "__interrupt__" in result
+
+    put_writes_task_ids: list[str] = []
+    orig = PregelLoop.put_writes
+
+    def spy(self, task_id, writes):
+        put_writes_task_ids.append(task_id)
+        return orig(self, task_id, writes)
+
+    with patch.object(PregelLoop, "put_writes", spy):
+        result = await workflow.ainvoke(Command(resume="ans"), config=config)
+
+    assert result == {"answer": "ans"}
+    # Count unique non-null task IDs that got put_writes.
+    # Should be exactly 2: the ask task and the entrypoint task.
+    # If 3, the cached setup task is being redundantly re-committed.
+    non_null = set(
+        tid for tid in put_writes_task_ids if not tid.startswith("00000000")
+    )
+    assert len(non_null) == 2, (
+        f"Expected 2 task IDs in put_writes (ask + entrypoint), got {len(non_null)}"
+    )
+
+
+@NEEDS_CONTEXTVARS
 async def test_node_before_interrupt_resume_graph_api(
     async_checkpointer: BaseCheckpointSaver,
 ) -> None:
--- a/libs/langgraph/langgraph/pregel/_algo.py
+++ b/libs/langgraph/langgraph/pregel/_algo.py
@@ -1061,17 +1061,13 @@ def _scratchpad(
 ) -> PregelScratchpad:
     if len(pending_writes) > 0:
         # find global resume value
-        # Only the root scratchpad (no parent) owns the null resume write.
-        # Child scratchpads always delegate to their parent via
-        # get_null_resume, ensuring the value is consumed at most once.
-        if parent_scratchpad is None:
-            for w in pending_writes:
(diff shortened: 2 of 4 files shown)


libs/langgraph/tests/test_pregel_async.py
```diff
@@ -7920,6 +7920,245 @@ async def program(_state: Any) -> list[str]:
...
+    assert len(result["__interrupt__"]) == 1
+    assert result["__interrupt__"][0].value == "Whats the answer for topic 2?"
+
+    # Resume with answer for topic 2 - should get final result
+    result = await workflow.ainvoke(Command(resume="answer2"), config=config)
+    assert result == {"answers": ["answer1", "answer2"]}
+
+
+@NEEDS_CONTEXTVARS
+async def test_multiple_tasks_before_interrupt_resume(
```
hinthornw, line 7975: Note that this already passes on main from my testing.

hinthornw (commented):
OPened another PR as a potential alternative

sydney-runkle:
Original PR description, for reference

## Fix: duplicate resume values when `@task` precedes interrupt in functional API

### Problem

When using the functional API (`@entrypoint` + `@task`), resuming with `Command(resume=value)` returns duplicate values instead of firing subsequent interrupts — but only when a `@task`-decorated function runs before the interrupt-producing tasks.

```python
@entrypoint(checkpointer=InMemorySaver())
async def workflow(number_of_topics: int) -> dict:
    @task
    async def setup() -> int:  # <-- having @task here triggers the bug
        return number_of_topics

    n = await setup()

    answers = []
    for i in range(n):
        answers.append(await ask(f"Topic {i + 1}?"))  # ask() calls interrupt()

    return {"answers": answers}
```

Expected behavior: two sequential interrupts, each resumed independently.
Actual behavior: after the first resume, the second `interrupt()` silently returns the first answer again instead of raising `GraphInterrupt`, producing `{"answers": ["answer1", "answer1"]}`.

Removing `@task` from `setup()` makes the bug disappear. Using `Command(resume={interrupt_id: value})` (mapped by interrupt ID) also works correctly.

### Root cause

The bug is in `_scratchpad()` in `pregel/_algo.py`. When a scratchpad is created, it searches `pending_writes` for the global null resume write (`(NULL_TASK_ID, RESUME, value)`) and captures both the write tuple and the list reference in a `get_null_resume` closure. Consumption is tracked by calling `pending_writes.remove(null_resume_write)` — if the tuple is already gone, `ValueError` signals "already consumed."

The problem is that `put_writes()` in `pregel/_loop.py` reassigns `self.checkpoint_pending_writes` to a **new list** each time a task is committed:

```python
self.checkpoint_pending_writes = [
    w for w in self.checkpoint_pending_writes if w[0] != task_id
]
```

In the graph API, this is harmless — each node runs in a separate step, and scratchpads don't outlive their step. But in the functional API, the entrypoint is a single long-running step, and child `@task` calls happen *during* that step. Here's what happens:

```
step 1:
  create entrypoint scratchpad (captures pending_writes list L₀, finds null resume)
  run entrypoint function:
    setup() → child task committed → put_writes replaces list → L₁ is new list
    ask(q1) → child scratchpad captures L₁, finds null resume in L₁
           → interrupt() consumes null resume → removes from L₁
           → child committed → put_writes replaces list → L₂ is new list
    ask(q2) → child scratchpad captures L₂, null resume NOT in L₂
           → delegates to parent (entrypoint) scratchpad
           → parent still references L₀, where null resume was NEVER removed
           → parent returns the value again → duplicate!
```

The parent entrypoint scratchpad captured `L₀` at the start of the step. The consumption (`remove()`) happened on `L₁`. Since `L₀` and `L₁` are different list objects, the parent doesn't see the consumption — it still has the null resume tuple in its closure and `L₀.remove()` succeeds, returning the value a second time.

Without `@task` on `setup()`, no mid-step `put_writes` occurs, so all scratchpads share the same list object and the `remove()` is visible everywhere.

### Fix

The fix is in `_scratchpad()` (`pregel/_algo.py`): **child scratchpads (those with a `parent_scratchpad`) no longer search `pending_writes` for the null resume write.** They always set `null_resume_write = None`, which forces `get_null_resume` to delegate up the parent chain to the root scratchpad.

Since only the root scratchpad (the entrypoint, which has no parent) ever captures the null resume write, and only the root ever calls `pending_writes.remove()` on its own captured list, consumption is tracked in exactly one place. It doesn't matter that `put_writes` creates new lists — the root's list and its `remove()` call are the sole mechanism.

```diff
     if len(pending_writes) > 0:
         # find global resume value
-        for w in pending_writes:
-            if w[0] == NULL_TASK_ID and w[1] == RESUME:
-                null_resume_write = w
-                break
+        # Only the root scratchpad (no parent) owns the null resume write.
+        # Child scratchpads always delegate to their parent via
+        # get_null_resume, ensuring the value is consumed at most once.
+        if parent_scratchpad is None:
+            for w in pending_writes:
+                if w[0] == NULL_TASK_ID and w[1] == RESUME:
+                    null_resume_write = w
+                    break
+            else:
+                null_resume_write = None
         else:
-            # None cannot be used as a resume value, ...
             null_resume_write = None
```

### Why this only affects the functional API

In the **graph API**, each node runs as a discrete step in the Pregel loop. Between steps, `put_writes` is called and a new checkpoint is saved. When the next step starts, `prepare_next_tasks` creates fresh scratchpads from the current `checkpoint_pending_writes`. There are no parent/child scratchpad relationships within a single step, and no scratchpad outlives its step. The list reassignment in `put_writes` is harmless because no old closures survive.

In the **functional API**, the entrypoint runs as a single long-lived step. Child `@task` calls happen mid-step via `call()` → `accept_push()` → `prepare_push_task_functional()`. Each child task gets committed mid-step (triggering `put_writes` and a new list), while the parent entrypoint's scratchpad persists with its original list reference. This creates the divergence between the list the child consumed from and the list the parent's closure references.

PR #6864, opened 2026-02-18
fix(langgraph): correct ParentCommand bubbling when checkpoint_ns includes numeric task segments 

eyurtsev:
Fixes incorrect `Command.PARENT` bubbling when checkpoint namespaces include numeric task-disambiguation segments like `|1`. In some nested-invoke/fanout scenarios, the runtime inserts a purely-numeric namespace segment between `name:task_id` segments (e.g. `parent_first:<id>|1|node:<id>`). The previous ParentCommand rewrite logic only handled numeric segments at the end of the namespace, which could produce a malformed parent graph identifier (e.g. `parent_first:<id>|1`) and prevent the command from routing to the intended parent node.

This change normalizes checkpoint namespaces by dropping numeric segments before computing the parent namespace in both sync and async retry paths. Added a minimal regression test that exercises the nested-invoke case and asserts that `Command(graph=Command.PARENT, goto=...)` reliably routes to the parent graph, regardless of whether the jump comes from the first or second nested invocation.

### DIFF
--- a/libs/langgraph/tests/test_parent_command_async.py
+++ b/libs/langgraph/tests/test_parent_command_async.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import pytest
+from langchain_core.runnables import RunnableConfig
 from typing_extensions import TypedDict
 
 from langgraph.graph import END, START, StateGraph
@@ -31,12 +32,12 @@ async def test_parent_command_from_nested_subgraph() -> None:
 
     parent_builder: StateGraph[ParentState] = StateGraph(ParentState)
 
-    async def parent_first(state: ParentState) -> ParentState:
-        await child_0.ainvoke({"jump": state["jump_from_idx"] == 1})
+    async def parent_first(state: ParentState, config: RunnableConfig) -> ParentState:
+        await child_0.ainvoke({"jump": state["jump_from_idx"] == 1}, config)
         if state["jump_from_idx"] == 1:
             raise AssertionError("Shouldn't be here")
 
-        await child_1.ainvoke({"jump": state["jump_from_idx"] == 2})
+        await child_1.ainvoke({"jump": state["jump_from_idx"] == 2}, config)
         if state["jump_from_idx"] == 2:
             raise AssertionError("Shouldn't be here")
 


libs/langgraph/tests/test_retry.py
```diff
@@ -78,6 +78,18 @@ def test_should_retry_on_empty_sequence():
...
+def test_checkpoint_ns_for_parent_command() -> None:
+    assert _checkpoint_ns_for_parent_command("") == ""
+    assert _checkpoint_ns_for_parent_command("node:1") == ""
+    assert _checkpoint_ns_for_parent_command("node:1|child:2") == "node:1"
+    assert _checkpoint_ns_for_parent_command("node:1|1|child:2") == "node:1"
+    assert _checkpoint_ns_for_parent_command("node:1|1|child:2|1") == "node:1"
+    assert (
+        _checkpoint_ns_for_parent_command("parent:1|1|child:1|1|node:1|1")
+        == "parent:1|1|child:1"
+    )
```
sydney-runkle, line 90: let's add a missing case
`_checkpoint_ns_for_parent_command("parent:1|1|child:1|1|node:1")`

libs/langgraph/tests/test_parent_command_async.py
```diff
@@ -0,0 +1,53 @@
...
+
+    parent_builder.add_node("parent_first", parent_first)
+    parent_builder.add_node("parent_second", parent_second)
+    parent_builder.add_edge(START, "parent_first")
+    parent_builder.add_edge("parent_second", END)
+
+    graph = parent_builder.compile().with_config(recursion_limit=10)
+
+    assert graph.invoke({"jump_from_idx": 1}) == {"jump_from_idx": 1}
+    assert graph.invoke({"jump_from_idx": 2}) == {"jump_from_idx": 2}
```
sydney-runkle: ainvoke?

libs/langgraph/tests/test_parent_command_async.py
```diff
@@ -0,0 +1,53 @@
...
+        if state["jump_from_idx"] == 1:
+            raise AssertionError("Shouldn't be here")
+
+        child_1.invoke({"jump": state["jump_from_idx"] == 2})
+        if state["jump_from_idx"] == 2:
+            raise AssertionError("Shouldn't be here")
+
+        return state
+
+    def parent_second(state: ParentState) -> ParentState:
```
hinthornw: ditto as Sydney's comment - this needs to actually test the async case

libs/langgraph/tests/test_retry.py
eyurtsev: Added

PR #6866, opened 2026-02-19
chore: rm duplicate writes

hinthornw:
In the original proposed fix [6863](https://github.com/langchain-ai/langgraph/pull/6863), we were guarding this specific case by only checking for null resume in the scratchpad.

But I noticed the tests passed on main for the sync case, indicating an inconsistency between the two paths. 

Seems like we were making a redundant put_writes call when replaying the async tasks

### DIFF
--- a/libs/langgraph/langgraph/pregel/_runner.py
+++ b/libs/langgraph/langgraph/pregel/_runner.py
@@ -728,7 +728,6 @@ async def _acall_impl(
                     )
                 else:
                     fut.set_result(None)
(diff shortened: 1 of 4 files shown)


libs/langgraph/langgraph/pregel/_runner.py
```diff
@@ -728,7 +728,6 @@ async def _acall_impl(
                     )
                 else:
                     fut.set_result(None)
-                futures()[fut] = next_task  # type: ignore[index]
```
hinthornw, line 731: This is the change

PR #6883, opened 2026-02-20
feat(checkpoint-postgres): add extended methods and conformance tests

hinthornw:
## Summary
- Implements `adelete_for_runs`, `acopy_thread`, `aprune` on `AsyncPostgresSaver` (async)
- Implements `delete_for_runs`, `copy_thread`, `prune` on both `PostgresSaver` (sync native) and `AsyncPostgresSaver` (sync wrapper)
- Adds conformance test harness (`tests/test_conformance.py`) using `langgraph-checkpoint-conformance`
- Adds `langgraph-checkpoint-conformance` as an editable dev dependency in `pyproject.toml` and `uv.lock`

## Details
The extended methods complete the `BaseCheckpointSaver` interface for Postgres:
- **`adelete_for_runs`** — deletes checkpoints, writes, and orphaned blobs by `metadata.run_id`
- **`acopy_thread`** — copies all checkpoints, blobs, and writes from one thread to another (idempotent via `ON CONFLICT DO NOTHING`)
- **`aprune`** — supports `"keep_latest"` (uses `DISTINCT ON`) and `"delete"` strategies with proper blob cleanup

## Test plan
- [ ] `cd libs/checkpoint-postgres && make lint` passes
- [ ] `cd libs/checkpoint-postgres && make test` passes (includes conformance tests)
- [ ] CI runs conformance tests via existing `make test` in the postgres matrix job

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
@@ -390,6 +390,142 @@ class PostgresSaver(BasePostgresSaver):
                 (str(thread_id),),
             )
 
+    def delete_for_runs(self, run_ids: Sequence[str]) -> None:
+        """Delete all checkpoints and writes for the given run IDs.
+
+        Args:
+            run_ids: The run IDs whose checkpoints should be deleted.
+        """
+        if not run_ids:
+            return
+        with self._cursor(pipeline=True) as cur:
+            cur.execute(
+                """DELETE FROM checkpoint_writes
+                WHERE (thread_id, checkpoint_ns, checkpoint_id) IN (
+                    SELECT thread_id, checkpoint_ns, checkpoint_id
+                    FROM checkpoints
+                    WHERE metadata->>'run_id' = ANY(%s)
+                )""",
+                (list(run_ids),),
+            )
+            cur.execute(
+                """DELETE FROM checkpoint_blobs
+                WHERE (thread_id, checkpoint_ns) IN (
+                    SELECT DISTINCT thread_id, checkpoint_ns
+                    FROM checkpoints
+                    WHERE metadata->>'run_id' = ANY(%s)
+                )
+                AND NOT EXISTS (
+                    SELECT 1 FROM checkpoints c2
+                    WHERE c2.thread_id = checkpoint_blobs.thread_id
+                    AND c2.checkpoint_ns = checkpoint_blobs.checkpoint_ns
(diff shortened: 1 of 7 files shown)


tomascenteno42:
Are there any plans on moving this forward?

wangjiawei-vegetable:
Thanks for putting this together. We’re using `AsyncPostgresSaver` with long-lived production threads, and checkpoint growth is becoming an operational concern. Even with application-level message trimming, the Postgres checkpoint tables continue to accumulate rows per thread.

Could a maintainer confirm whether this PR is still the intended path for Postgres `prune` / `delete_for_runs`, especially after the recent `DeltaChannel` work? We’d be happy to help test a DeltaChannel-safe implementation against a production-like workload.

A rough milestone, or guidance on whether self-hosted Postgres users should implement their own retention for now, would be very helpful.

libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
```diff
@@ -390,6 +390,142 @@ def delete_thread(self, thread_id: str) -> None:
...
+                    WHERE thread_id = ANY(%s)
+                    AND NOT EXISTS (
+                        SELECT 1 FROM checkpoints c
+                        WHERE c.thread_id = checkpoint_blobs.thread_id
+                        AND c.checkpoint_ns = checkpoint_blobs.checkpoint_ns
+                    )""",
+                    (list(thread_ids),),
+                )
+
     @contextmanager
```
tomekmb-glitch, line 529: - .

wangjiawei-vegetable:
Hi @hinthornw — thanks for working on this.

We’re using `PostgresSaver` for long-lived agent conversations and have seen the checkpoint tables grow significantly over time.

I previously raised the storage-growth problem from the DeepAgents side in langchain-ai/deepagents#2876. The guidance there was that persisted conversation state should not be trimmed by `SummarizationMiddleware`, and that checkpoint storage growth should instead be addressed at the LangGraph layer, for example with `DeltaChannel`.

`DeltaChannel` helps reduce repeated payload storage, but it does not limit the number of checkpoint, write, and blob records retained as a thread continues to run. Because of that, PostgreSQL support for `keep_latest` is still something we need.

Is this PR still planned to move forward, and is PostgreSQL `keep_latest` still intended to be supported upstream?

I opened #8531 to discuss the expected behavior and safety considerations. I’m interested in helping move this feature forward and would be happy to implement and test the remaining work. Please let me know whether you would prefer this to continue through #6883 or as a focused contribution through #8531. Thank you!

PR #6885, opened 2026-02-20
perf: remove importlib.metadata for version access

hztBUAA:
## Summary

Removes `importlib.metadata` usage at import time for version access, improving import performance.

- **`langgraph/version.py`**: Replace `importlib.metadata.version()` with a hardcoded version string matching `pyproject.toml`
- **`langgraph_cli/version.py`**: Re-export `__version__` from `langgraph_cli.__init__` (which already has a static version) instead of using `importlib.metadata`
- **`checkpoint-postgres/base.py`**: Remove the obsolete version compatibility check that used `importlib.metadata` at import time. The check warned about `langgraph < 0.5`, which is no longer relevant since langgraph is at v1.x

The remaining `importlib.metadata` usages in `embed.py` and `generate_schema.py` are already lazy (inside function bodies, not at module level) and do not affect import time.

Fixes #5040

## Test plan

- [x] `from langgraph.version import __version__` returns `"1.0.9"`
- [x] `from langgraph_cli.version import __version__` returns the correct version from `__init__.py`
- [x] No `importlib.metadata` in module-level imports of changed files
- [x] Import time for `langgraph.version` reduced (no metadata scan overhead)

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -1,9 +1,7 @@
 from __future__ import annotations
 
 import random
-import warnings
 from collections.abc import Sequence
-from importlib.metadata import version as get_version
 from typing import Any, cast
 
 from langchain_core.runnables import RunnableConfig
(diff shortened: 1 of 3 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

Issue #6731, opened 2026-01-30
LangGraph [1.0.6] Bug - Agent infinite looping until recursion limit error is hit


jacobreesmontgomery:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
async def _init_agent(self, llm_parameters: LLMParameters) -> CompiledStateGraph:
        """
        Initialize the agent instance.

        Args:
            llm_parameters: Parameters for the language model.

        Returns:
            An instance of CompiledStateGraph representing the agent.
        """

        llm: AzureChatOpenAI = await self.llm_service.create_llm(
            parameters=llm_parameters
        )
        prompt = databricks_query_gen_agent_prompt.format(
            schema_info=self.metadata_catalog_schema
        )
        agent: CompiledStateGraph = create_agent(
            model=llm,
            tools=self._create_agent_tools(),
            system_prompt=prompt,
            response_format=DatabricksQueryGenerationResponse,
        ).with_config(recursion_limit=self.settings.agent_recursion_limit)

        return agent

agent: CompiledStateGraph = await self._init_agent(
            llm_parameters=llm_parameters
        )
        start_time = time()

        try:
            logger.DEBUG(
                f"Databricks Query Generation Agent processing request for user question [{user_question}]..."
            )
            response: dict[str, Any] | Any = await agent.ainvoke(
                {"messages": [{"type": "human", "content": user_question}]}
            )
```

### Error Message and Stack Trace (if applicable)

```shell
Exception: 

2026-01-30 10:19:13 - ERROR - agents.dbrks_query_gen_agent - Agent processing error: Recursion limit of 20 reached without hitting a stop condition. You can increase the limit by setting the `recursion_limit` config key.

Stack trace:

2026-01-30 10:18:12 - INFO - chat_route - Databricks query generation request received
2026-01-30 10:18:12 - DEBUG - agents.dbrks_query_gen_agent - Databricks Query Generation Agent processing request for user question [How many retail locations are in California?]...
[...]

PR #6889, opened 2026-02-20
fix(prebuilt): detect infinite tool call loops in create_react_agent

abhigyan631:
Description: Add loop detection to create_react_agent's routing logic to prevent agents from entering infinite loops when the LLM repeatedly requests the same tool call with the same arguments (e.g., due to persistent tool failures). The should_continue function now inspects the last 3 AIMessage tool calls and raises a GraphRecursionError early if they are semantically identical, providing a clear error message to the developer.

Issue: #6731

Dependencies: None


hinthornw:
I think this would best be added as Middleware in the LnagChain package!

eyurtsev:
@abhigyan631 thanks for the PR -- i'll close this PR since the change is in the wrong repository. Generally good to avoid making a PR until getting sign off from a contributor :)

Issue #6896, opened 2026-02-21
checkpoint-sqlite: make TTL pruning deterministic at boundary


davidahmann:
Problem
SQLite TTL pruning can vary by execution timing near retention boundaries, which causes flaky prune-set expectations and uncertain retention behavior in local/dev deployments.

Why now
TTL is increasingly used by default in checkpoint SQLite stores, and timing-sensitive pruning undermines reproducibility and operator trust in retention behavior.

Evidence packet
- commit under test: f702729e04dd51b843d257a58f9f5b0181f0be2f
- runtime environment: macOS 26.3; Python 3.14.0
- minimal repro:
  1. Seed store rows around the TTL cutoff timestamp.
  2. Trigger prune/list operations multiple times under slight timing drift.
  3. Compare pruned IDs/counts.
- expected: deterministic prune set and count for same logical cutoff.
- actual: boundary-adjacent rows can vary based on runtime timing.

Why current behavior is insufficient
Retention and replay debugging depend on deterministic cleanup. Timing-sensitive results create flaky tests and ambiguous operational outcomes.

Acceptance criteria
- TTL pruning uses a deterministic cutoff basis per prune operation.
- Regression coverage in checkpoint-sqlite TTL/store tests locks prune IDs and counts.
- Behavior remains backward-compatible for non-boundary rows.

ddddajie:
I'd like to work on this issue. I reproduced the boundary behavior on current `main`: SQLite `CURRENT_TIMESTAMP` is second-precision while `expires_at` is written from Python with sub-second precision, so a row can already be logically expired but survive a sweep until the next second boundary.

My plan is to keep the fix small: capture one UTC cutoff per `sweep_ttl()` call, use it consistently for the delete in both sync/async SQLite stores, and add focused regression coverage for the boundary case. No schema/API changes or unrelated TTL refactors.

Could you assign this issue to me?

ddddajie:
Progress update: I’ve finished a focused implementation on my fork and pushed it to `ddddajie/langgraph:fix/6896-deterministic-ttl-pruning` (commit `495975341de652855e1a75a7ab92c63b1a2d98a4`).

The change keeps the scope small:
- capture one UTC cutoff per `sweep_ttl()` call;
- bind that cutoff in the SQLite delete for both sync and async stores;
- add deterministic boundary regression coverage that checks the deleted count and remaining items.

[...]

PR #6897, opened 2026-02-21
fix(checkpoint-sqlite): make TTL sweep deterministic with explicit cutoff

davidahmann:
## Problem
SQLite TTL pruning currently depends on database `CURRENT_TIMESTAMP`, which makes sweep behavior timing-sensitive and can produce non-deterministic pruning around boundary conditions.

## Why now
Gap item #4 calls out deterministic runtime behavior and reproducible validation; timing-sensitive TTL deletion undermines both.

## What changed
- Added optional `now` parameter to sync/async `sweep_ttl`.
- Replaced SQL `CURRENT_TIMESTAMP` cutoff with parameterized UTC cutoff derived from `now` (or current UTC when omitted).
- Added sync and async regression tests that pin a cutoff instant and assert deterministic prune counts/results.

## Validation
- `cd libs/checkpoint-sqlite && TEST=tests/test_ttl.py make test` ✅ (13 passed)

Refs #6896

### DIFF
--- a/libs/checkpoint-sqlite/langgraph/store/sqlite/aio.py
+++ b/libs/checkpoint-sqlite/langgraph/store/sqlite/aio.py
@@ -248,6 +248,11 @@ class AsyncSqliteStore(AsyncBatchedBaseStore, BaseSqliteStore):
             int: The number of deleted items.
         """
         cutoff = now or datetime.datetime.now(datetime.timezone.utc)
+        cutoff_text = (
+            cutoff.astimezone(datetime.timezone.utc)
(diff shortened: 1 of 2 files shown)


davidahmann:
This change makes SQLite TTL sweeping deterministic by replacing `CURRENT_TIMESTAMP` dependency with an explicit cutoff parameter (while preserving current behavior when omitted). Sync and async regression tests now pin the cutoff instant to verify stable pruning behavior around boundary times.

This contribution was informed by patterns from Gait: https://github.com/Clyra-AI/gait

PR #6886, opened 2026-02-20
fix: support Pydantic default values with reducer functions

hztBUAA:
## Summary

Fixes #5225

When a Pydantic `BaseModel` state field has both a default value (`Field(default=...)` or `Field(default_factory=...)`) and a reducer function (`Annotated[type, reducer]`), the default value was being ignored. The channel was always initialized with the type's zero-argument constructor (e.g., `[]` for `list`, `0` for `int`).

### Root Cause

`BinaryOperatorAggregate.__init__` always initialized its value by calling `typ()` (the bare type constructor). It had no mechanism to receive or use schema-defined default values. While `_get_channels` and `_get_channel` had partial plumbing for passing the `schema` parameter, `_is_field_binop` didn't use it and `BinaryOperatorAggregate` didn't accept a `default` parameter.

### Changes

- **`_fields.py`**: Added Pydantic `BaseModel` support to `get_field_default()` to extract `Field(default=...)` and `Field(default_factory=...)` values via `model_fields`
- **`binop.py`**: Added `default` parameter to `BinaryOperatorAggregate.__init__()`. When provided, uses it instead of `typ()`. Uses `copy.deepcopy` to prevent mutable default sharing. Preserves defaults through `copy()` and `from_checkpoint()`
- **`state.py`**: Connected the schema context through `_get_channels` → `_get_channel` → `_is_field_binop` so field defaults are resolved and passed to `BinaryOperatorAggregate`

### Example

```python
from pydantic import BaseModel, Field
from typing import Annotated
import operator
from langgraph.graph import StateGraph, START, END

class State(BaseModel):
    messages: Annotated[list[str], operator.add] = Field(
        default_factory=lambda: ["welcome"]
    )
    count: Annotated[int, operator.add] = Field(default=100)

def node(state):
    return {"messages": ["hello"], "count": 5}

graph = StateGraph(State)
graph.add_node("process", node)
graph.add_edge(START, "process")
graph.add_edge("process", END)

result = graph.compile().invoke({})
# result["messages"] == ["welcome", "hello"]  ✅
# result["count"] == 105                       ✅
```

## Test plan

- [x] Unit test for `BinaryOperatorAggregate` with default values (`test_binop_with_default`)
- [x] Integration test: `default_factory` with custom reducer (exact issue scenario)
- [x] Integration test: static `Field(default=...)` with `operator.add`
- [x] Integration test: default + user-provided initial input
- [x] Integration test: mutable default safety across multiple invocations
[...]

PR #6908, opened 2026-02-22
fix: scan all Annotated metadata for reducer callable

kar-ganap:
## Summary

`_is_field_binop` only checked the **last** metadata item in an `Annotated` type
(`meta[-1]`). If additional metadata — doc strings, Pydantic `Field()`, custom
annotations — appeared after the reducer callable, the reducer was silently
ignored and the field fell back to `LastValue`, losing aggregation semantics.

```python
# This works:
items: Annotated[list[str], operator.add]

# This silently loses the reducer (operator.add is no longer last):
items: Annotated[list[str], operator.add, "item accumulator"]
items: Annotated[list[str], operator.add, Field(description="accumulated items")]
```

**Root cause:** `meta[-1]` assumes the reducer is always the last annotation.
Python's `Annotated` makes no such guarantee, and users reasonably interleave
documentation, Pydantic field metadata, and the reducer in any order.

**Fix:** Iterate through all metadata items and return the first callable with a
2-parameter signature, consistent with how `_is_field_channel` already works
(it loops through all metadata to find channel annotations). The change is a
minimal 1-line structural fix (loop instead of index) with no new dependencies.

We reviewed the open PRs on related topic #5225 — specifically #6697 and #6886,
both of which add default-value support for reducer fields. Neither addresses
the metadata ordering issue (both still use `meta[-1]`). This fix is
complementary and compatible: once merged, those PRs' `_is_field_binop` changes
can build on the loop-based approach.

## Test plan

7 regression tests in `tests/test_reducer_metadata_ordering.py`:

- [x] `test_reducer_as_last_metadata` — baseline (always worked)
- [x] `test_reducer_before_string_annotation` — reducer precedes doc string
- [x] `test_reducer_between_metadata` — reducer between two non-callable items
- [x] `test_reducer_with_field_metadata_after` — reducer before Pydantic `Field()`
- [x] `test_custom_reducer_not_last` — custom 2-arg reducer, not last
- [x] `test_multiple_nodes_with_non_trailing_reducer` — multi-step graph
- [x] `test_no_reducer_still_falls_through` — non-callable-only metadata falls back to `LastValue`
- [x] All existing tests pass unchanged (156 passed, 4 skipped)

### DIFF
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -1633,21 +1633,22 @@ def _is_field_channel(typ: type[Any]) -> BaseChannel | None:
 def _is_field_binop(typ: type[Any]) -> BinaryOperatorAggregate | None:
     if hasattr(typ, "__metadata__"):
         meta = typ.__metadata__
-        if len(meta) >= 1 and callable(meta[-1]):
-            sig = signature(meta[-1])
-            params = list(sig.parameters.values())
-            if (
-                sum(
-                    p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)
-                    for p in params
-                )
-                == 2
-            ):
-                return BinaryOperatorAggregate(typ, meta[-1])
-            else:
-                raise ValueError(
-                    f"Invalid reducer signature. Expected (a, b) -> c. Got {sig}"
-                )
+        for item in meta:
+            if callable(item):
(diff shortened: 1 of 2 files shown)


kar-ganap:
Gentle ping — this is ready for review when you get a chance.

kar-ganap:
Friendly follow-up — this has been open for a while and is still mergeable. Happy to address any feedback if needed.

PR #6923, opened 2026-02-23
POC streaming hints

sydney-runkle:


sydney-runkle:
too many overloads :/

PR #6931, opened 2026-02-24
feat(langgraph): backwards compat type safe streaming

sydney-runkle:
### Review recommendations

* Don't fear the diff! It's not so bad, I swear! The PR description below gives a nice overview of changes
* Check out my explicit comments below, those should help to orient you to the important changes :)
* On a first pass, ignore the test files! Just check out the new types, overloads, and minor logical changes (diff behavior based on the flag)

## Summary

Adds a `stream_version="v2"` option that emits typed `{"type", "ns", "data"}` dicts instead of raw tuples/SSE events. Each stream mode gets its own `TypedDict` with a `Literal` type field, enabling full type narrowing on `part["type"]`.

This is **opt in**, so it's **non-breaking**!!

```python
async for part in graph.astream(
    inputs, stream_mode=["messages", "custom"], stream_version="v2"
):
    if part["type"] == "messages":
        msg, metadata = part["data"]  # tuple[AnyMessage, dict] ✅
        print(msg.content)
    elif part["type"] == "custom":
        part["data"]  # Any ✅
```

Before v2 you'd get `tuple[str, Any]` with no way to narrow `data` based on mode.

### What changed

**`langgraph` (core):** New `StreamPart` discriminated union + per-mode TypedDicts in `types.py`. Stream-emit code in `pregel/` refactored to use the new types. `RemoteGraph` gains a `stream_version` param.

**`sdk-py`:** Client-side v2 wrapper that converts SSE events into typed dicts. No server API changes — v2 is purely a client-side rewrite of the stream format.

#### `langgraph` (core)

| `type` | `data` |
|---|---|
| `"values"` | `dict[str, Any]` — full state after each step |
| `"updates"` | `dict[str, Any]` — node name → output |
| `"messages"` | `tuple[AnyMessage, dict]` — message + metadata |
| `"custom"` | `Any` — whatever was passed to `StreamWriter` |
| `"tasks"` | `TaskPayload \| TaskResultPayload` |
| `"checkpoints"` | `CheckpointPayload` |
| `"debug"` | `DebugPayload` |

#### `sdk-py` (additional types from SSE events)

| `type` | `data` | Description |
|---|---|---|
| `"messages/partial"` | `list[dict]` | Partial message chunks |
| `"messages/complete"` | `list[dict]` | Complete messages |
| `"messages/metadata"` | `dict` | Message metadata |
| `"metadata"` | `RunMetadataPayload` | Run-level metadata (`run_id`, etc.) |

All parts share the shape `{"type": Literal[...], "ns": list[str], "data": ...}`.

## Release plan

* Release as a part of langgraph 1.1
* Before release, I'd like to do more experimentation with support for pydantic + dataclasses and/or input/output runtime validation, as that would help resolve the lack of typing for `values` mode.

### Notes

- Docs need a mass update to cover v2 streaming usage and the new types
- Changing the default `stream_version` to `"v2"` in a future release would be breaking but backwards compatible (users can pin `"v1"` to keep current behavior)
- I called out specifically relevant parts of the code in comments on the PR :)


libs/langgraph/langgraph/pregel/main.py
```diff
@@ -3282,7 +3498,9 @@ def _output(
                     )
                 )
         if mode in stream_mode:
-            if stream_subgraphs and isinstance(stream_mode, list):
+            if stream_version == "v2":
+                yield {"type": mode, "ns": ns, "data": payload}
+            elif stream_subgraphs and isinstance(stream_mode, list):
```
sydney-runkle, line 3501: actual logical change located here!

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -3068,33 +3216,38 @@ def invoke(
...
             durability=durability,
+            stream_version=stream_version,  # type: ignore[arg-type]
             **kwargs,
         ):
             if stream_mode == "values":
-                if len(chunk) == 2:
-                    mode, payload = cast(tuple[StreamMode, Any], chunk)
+                if stream_version == "v2":
+                    mode = chunk["type"]
+                    payload = chunk["data"]
```
sydney-runkle, line 3237: look at this lovely code! No casting, no nothing!

libs/langgraph/langgraph/pregel/debug.py
```diff
@@ -23,42 +22,14 @@
 from langgraph.channels.base import BaseChannel
 from langgraph.constants import TAG_HIDDEN
 from langgraph.pregel._io import read_channels
-from langgraph.types import PregelExecutableTask, PregelTask, StateSnapshot
-
-__all__ = ("TaskPayload", "TaskResultPayload", "CheckpointTask", "CheckpointPayload")
```
sydney-runkle, line 28: moved but still imported for backwards compat

libs/langgraph/langgraph/types.py
```diff
@@ -113,6 +127,182 @@ def ensure_valid_checkpointer(checkpointer: Checkpointer) -> Checkpointer:
...
+
+class ValuesStreamPart(TypedDict):
+    """Stream part emitted for `stream_mode="values"`.
+
+    `data` contains the full state after each step, as returned by `read_channels()`.
+    """
+
+    type: Literal["values"]
+    ns: tuple[str, ...]
+    data: dict[str, Any]
```
sydney-runkle, line 212: Unfortunate, we really want `StateT` / `OutputT` here, but that doesn't work right now because we don't output pydantic or dataclass state

libs/langgraph/langgraph/types.py
```diff
@@ -113,6 +127,182 @@ def ensure_valid_checkpointer(checkpointer: Checkpointer) -> Checkpointer:
...
+class UpdatesStreamPart(TypedDict):
+    """Stream part emitted for `stream_mode="updates"`.
+
+    `data` maps node names to their outputs. May also contain
+    `__interrupt__` (tuple of `Interrupt` dicts) and `__metadata__` keys.
+    """
+
+    type: Literal["updates"]
+    ns: tuple[str, ...]
+    data: dict[str, Any]
```
sydney-runkle, line 224: ideally this would be `Partial[StateT]`

libs/langgraph/langgraph/pregel/protocol.py
```diff
@@ -107,8 +115,68 @@ def stream(
         interrupt_before: All | Sequence[str] | None = None,
         interrupt_after: All | Sequence[str] | None = None,
         subgraphs: bool = False,
+        stream_version: Literal["v2"],
+    ) -> Iterator[StreamPart]: ...
```
sydney-runkle, line 119: note! this is the helpful overload - when we are in mode v2, we yield stream parts!

libs/langgraph/langgraph/pregel/protocol.py
```diff
@@ -107,8 +115,68 @@ def stream(
...
+        input: InputT | Command | None,
+        config: RunnableConfig | None = None,
+        *,
+        context: ContextT | None = None,
+        stream_mode: StreamMode | list[StreamMode] | None = None,
+        interrupt_before: All | Sequence[str] | None = None,
+        interrupt_after: All | Sequence[str] | None = None,
+        subgraphs: bool = False,
+        stream_version: Literal["v1"] = ...,
+    ) -> Iterator[dict[str, Any] | Any]: ...
```
sydney-runkle, line 134: old style, yielding nothing helpful :/

libs/langgraph/langgraph/pregel/debug.py
eyurtsev: Are they exposed via __all__ so it's clear nothing will break in the future? (coudl we add a comment somewhere that it's imported for backwards compat?)

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -3021,6 +3114,57 @@ async def _cleanup_waiter() -> None:
...
+        config: RunnableConfig | None = None,
+        *,
+        context: ContextT | None = None,
+        stream_mode: Literal["values"] = ...,
+        print_mode: StreamMode | Sequence[StreamMode] = (),
+        output_keys: str | Sequence[str] | None = None,
+        interrupt_before: All | Sequence[str] | None = None,
+        interrupt_after: All | Sequence[str] | None = None,
+        durability: Durability | None = None,
+        stream_version: Literal["v2"],
```
eyurtsev, line 3130: Why is there a a stream version for invoke? invoke should be unaffected?

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -3068,33 +3216,38 @@ def invoke(
         chunks: list[dict[str, Any] | Any] = []
         interrupts: list[Interrupt] = []
 
-        for chunk in self.stream(
+        for chunk in self.stream(  # type: ignore[misc]
```
eyurtsev, line 3219: why is a type ignore needed?

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -3111,6 +3264,57 @@ def invoke(
...
+        config: RunnableConfig | None = None,
+        *,
+        context: ContextT | None = None,
+        stream_mode: StreamMode = "values",
+        print_mode: StreamMode | Sequence[StreamMode] = (),
+        output_keys: str | Sequence[str] | None = None,
+        interrupt_before: All | Sequence[str] | None = None,
+        interrupt_after: All | Sequence[str] | None = None,
+        durability: Durability | None = None,
+        stream_version: Literal["v1"] = ...,
```
eyurtsev, line 3314: why does invoke have a stream version? the invoke interface should not change for streaming. If we need to propagate something w/ v2 that would to be done in a different manner

eyurtsev:
First pass at review -- I don't think invoke() should have any changes. Is this accidental or is there book-keeping that needs to happen in some other way?

libs/langgraph/langgraph/pregel/main.py
sydney-runkle: there's a stream version here bc we support stream modes for invoke :) discussed live!

sydney-runkle:
> First pass at review -- I don't think invoke() should have any changes. Is this accidental or is there book-keeping that needs to happen in some other way?

Not a mistake -- intentional!

`invoke` supports `stream_mode`. When a list of stream modes is specified, we return a list of tuples at the moment. So threading through `v2` support here supports typed stream parts if a user specifies stream modes.

sydney-runkle:
As discussed live, going to merge this into the 1.1 branch and continue work there

PR #6937, opened 2026-02-25
fix(langgraph): handle Overwrite correctly in BinaryOperatorAggregate when value is MISSING

### LINKED ISSUE #6909: BinaryOperatorAggregate: Overwrite not unwrapped and duplicate-Overwrite guard skipped when initial value is MISSING
### Checked other resources

```python
import operator
from typing import Annotated

@dataclass
class Metrics:
    count: int
    def __add__(self, other: "Metrics") -> "Metrics":
        return Metrics(self.count + other.count)


# Bug 1: Overwrite wrapper stored instead of unwrapped value when channel starts MISSING
class State(TypedDict):
    metrics: Annotated[Metrics, operator.add]

def reset_node(state: State):
    return {"metrics": Overwrite(Metrics(count=42))}

graph = (
    StateGraph(State)
    .add_node("reset", reset_node)
    .add_edge(START, "reset")
    .compile()
)

print(graph.invoke({}))
# Expected: {'metrics': Metrics(count=42)}
# Actual:   {'metrics': Overwrite(value=Metrics(count=42))}


# Bug 2: Duplicate Overwrite guard not enforced when first update is on MISSING channel
channel = BinaryOperatorAggregate(Metrics, operator.add) 
try:
    channel.update([Overwrite(Metrics(1)), Overwrite(Metrics(2))])
    print("NO ERROR (bug):", channel.get())
    print("Expected error:", e)
```

```shell
No exception is raised. The incorrect output is:

{'metrics': Overwrite(value=Metrics(count=42))}
  Metrics(count=2)
```

### Description

When using `BinaryOperatorAggregate` with a type that has no default constructor, the channel sets `self.value` to
`MISSING`. In this case, passing an `Overwrite(x)` value to `update()` stores the wrapper object instead of the
unwrapped payload, so `channel.get()` returns `Overwrite(value=x)` instead of `x`.

On top of that, the `seen_overwrite` guard is never set in this code path, so passing two `Overwrite` values in the
same super-step silently succeeds instead of raising `InvalidUpdateError` as expected.

Both issues happen because the `MISSING` fast-path in `BinaryOperatorAggregate.update` skips the `_get_overwrite`
check that runs for all other values.

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #1 SMP Mon Feb  2 12:27:57 UTC 2026
> Python Version:  3.12.12 (main, Oct 10 2025, 08:52:57) [GCC 11.4.0]

Package Information
-------------------
> langchain_core: 1.2.13
> langchain: 1.2.10
> langsmith: 0.7.3
> langgraph_sdk: 0.3.6

> langserve

Other Dependencies
------------------
> httpx: 0.28.1
> jsonpatch: 1.33
> langgraph: 1.0.8
> opentelemetry-api: 1.38.0
> opentelemetry-exporter-otlp-proto-http: 1.38.0
> opentelemetry-sdk: 1.38.0
> orjson: 3.11.7
> packaging: 26.0
> pydantic: 2.12.3
> pytest: 8.4.2
> pyyaml: 6.0.3
> requests: 2.32.4
> requests-toolbelt: 1.0.0
> rich: 13.9.4
> tenacity: 9.1.4
> typing-extensions: 4.15.0
> uuid-utils: 0.14.0
> xxhash: 3.6.0
> zstandard: 0.25.0

ibrahim1023:
## Description

This PR fixes `BinaryOperatorAggregate.update` for the case where the channel starts at `MISSING` and receives `Overwrite(...)` as its first update in a super-step.

Before this change:
- The first overwrite in the `MISSING` path could be stored as `Overwrite(value=...)` instead of the unwrapped payload.
- The "only one `Overwrite` per super-step" guard was not applied consistently in that same path.

After this change:
- The first update is always checked for overwrite semantics, even when `self.value is MISSING`.
- Overwrite payloads are unwrapped correctly.
- Multiple overwrites in one super-step correctly raise `InvalidUpdateError`.

This keeps overwrite behavior consistent across both initialized and `MISSING` channel states.

Closes #6909.

## Issue

#6909

## Dependencies

None.

## Validation

Executed in `libs/langgraph`:
- `make format`
- `UV_PYTHON=3.12 make lint`
- `make test`

All passed locally.

### DIFF
--- a/libs/langgraph/langgraph/channels/binop.py
+++ b/libs/langgraph/langgraph/channels/binop.py
@@ -102,10 +102,15 @@ class BinaryOperatorAggregate(Generic[Value], BaseChannel[Value, Value, Value]):
     def update(self, values: Sequence[Value]) -> bool:
         if not values:
             return False
-        if self.value is MISSING:
-            self.value = values[0]
-            values = values[1:]
         seen_overwrite: bool = False
+        if self.value is MISSING:
+            is_overwrite, overwrite_value = _get_overwrite(values[0])
+            if is_overwrite:
+                self.value = overwrite_value
+                seen_overwrite = True
+            else:
+                self.value = values[0]
+            values = values[1:]
         for value in values:
             is_overwrite, overwrite_value = _get_overwrite(value)
             if is_overwrite:
--- a/libs/langgraph/tests/test_channels.py
+++ b/libs/langgraph/tests/test_channels.py
@@ -1,5 +1,6 @@
 import operator
 from collections.abc import Sequence
+from dataclasses import dataclass
 
 import pytest
 
@@ -9,6 +10,7 @@ from langgraph.channels.last_value import LastValue
 from langgraph.channels.topic import Topic
 from langgraph.channels.untracked_value import UntrackedValue
 from langgraph.errors import EmptyChannelError, InvalidUpdateError
+from langgraph.types import Overwrite
 
 pytestmark = pytest.mark.anyio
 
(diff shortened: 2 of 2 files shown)


gautamvarmadatla:
Hi! Creator of the issue here, tysm for working on this :) I already opened a PR #6910 when I filed the issue, and it looks like your PR is duplicating the same fix.

PR #6940, opened 2026-02-26
fix(checkpoint-postgres): backfill semantic search results with non-indexed docs

seeincodes:
## Description
Backfill semantic search results in `checkpoint-postgres` when vector search under-fills the requested limit, allowing documents stored with `index=False` to appear with `score=None`.

Also updates the package `Makefile` to use a unique default Docker Compose project name to avoid `tests_default` network collisions during local test setup.

## Issue
N/A (or Closes #<issue-number>)

## Dependencies
No new dependencies added.

## Twitter handle
N/A

### DIFF
--- a/libs/checkpoint-postgres/Makefile
+++ b/libs/checkpoint-postgres/Makefile
@@ -4,15 +4,17 @@
 # TESTING AND COVERAGE
 ######################
 
+COMPOSE_PROJECT_NAME ?= langgraph_checkpoint_postgres
+
 start-postgres:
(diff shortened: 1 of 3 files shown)


vikky781:
I independently investigated the same TODO in `test_store.py` and reproduced the issue locally.

While debugging, I ended up with a different implementation that keeps non-indexed rows in the vector search query and orders NULL scores last, instead of performing a separate backfill query after the vector search.

Was there a specific reason for choosing the explicit backfill approach? I wanted to check whether there are edge cases that a query-based solution would miss.

PR #6943, opened 2026-02-26
chore: add tests to confirm expected subgraph persistence behavior

sydney-runkle:
motivated by / in conjunction with new docs: https://docs.langchain.com/oss/python/langgraph/use-subgraphs#subgraph-persistence

### DIFF
--- a/libs/langgraph/tests/test_subgraph_persistence_async.py
+++ b/libs/langgraph/tests/test_subgraph_persistence_async.py
@@ -5,7 +5,7 @@ Covers three checkpointer settings for subgraph state:
 - checkpointer=None (default): "stateless" — inherits parent checkpointer for
   interrupt support, but state resets each invocation. This is the common case
   when an agent is invoked from inside a tool used by another agent.
-- checkpointer=True: "stateful" — state accumulates across invocations
+- checkpointer=True: "stateful" — state accumulates across invocations on the same thread id
 """
 
 import sys


libs/langgraph/tests/subgraph_persistence_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+from langgraph.graph import START, StateGraph
+from langgraph.graph.message import MessagesState
+from langgraph.types import interrupt
+
+
+class ParentState(TypedDict):
+    result: str
+
+
+def _contents(messages: list) -> list[str]:
```
eyurtsev: do we not type our unit tests? messages: list[AnyMessage]?

m.content_blocks or m.text?

libs/langgraph/tests/subgraph_persistence_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+    """Extract message content strings for readable assertions."""
+    return [m.content for m in messages]
+
+
+def _echo_graph(prefix: str = "Echo", *, checkpointer=None):
```
eyurtsev: checkpointer: Checkpoitner | None = None
-> CompiledStateGraph

libs/langgraph/tests/subgraph_persistence_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+        .compile(checkpointer=checkpointer)
+    )
+
+
+def _interrupt_echo_graph(prefix: str = "Processing"):
+    """Two-node subgraph with interrupt: echoes '<prefix>: <input>' then 'Done'."""
+
+    def process(state: MessagesState):
+        interrupt("continue?")
+        content = state["messages"][-1].content
```
eyurtsev: .text

libs/langgraph/tests/test_subgraph_persistence.py
```diff
@@ -0,0 +1,381 @@
...
+    _wrap_session_scope_interrupt,
+)
+
+# -- checkpointer=None (tool scope) --
+
+
+def test_tool_scope_interrupt_resume(
+    sync_checkpointer: BaseCheckpointSaver,
+) -> None:
+    """checkpointer=None supports interrupt/resume."""
```
eyurtsev: Could you add a 3-5 sentence description in the unit test of which behavior the unit test is testing ?e.g., What is "tool_scope" ? No tools are used?

libs/langgraph/tests/test_subgraph_persistence.py
```diff
@@ -0,0 +1,381 @@
...
+        resp = inner.invoke({"messages": [HumanMessage(content="apples")]})
+        return {"result": resp["messages"][-1].content}
+
+    parent = (
+        StateGraph(ParentState)
+        .add_node("call_inner", call_inner)
+        .add_edge(START, "call_inner")
+        .compile(checkpointer=sync_checkpointer)
+    )
+    config = {"configurable": {"thread_id": "1"}}
```
eyurtsev: Is the sync_checkpointer per test? If not, we do we need random thread_id uuids rather than a fixed one? (I'd prefer using uuid4() fwiw to avoid any potential issues due to the checkpointer being shared)

libs/langgraph/tests/subgraph_persistence_helpers.py
```diff
@@ -0,0 +1,68 @@
+"""Shared helpers for subgraph persistence tests."""
```
eyurtsev: just an opinion: but having a helper here that is re-used is making the the individual tests very difficult to understand since the code is on a separate page. If there's any version of this that would still work for the unit tests and can be compacted even more for testing would be great!

libs/langgraph/tests/test_subgraph_persistence_async.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+# -- checkpointer=True (session scope) --
+
+
+async def test_session_scope_state_accumulates_async(
+    async_checkpointer: BaseCheckpointSaver,
+) -> None:
+    """checkpointer=True retains subgraph state across invocations (async)."""
+    wrapper = _wrap_session_scope(_echo_graph("Processing"), "agent")
```
eyurtsev: I'd try to include more documentation here... e.g., it's hard to remember what _wrap_session_scope and _echo_graph do

libs/langgraph/tests/test_subgraph_persistence_async.py
```diff
@@ -0,0 +1,388 @@
+"""Tests for subgraph persistence behavior (async).
+
+Covers three checkpointer settings for subgraph state:
+- checkpointer=False: no persistence, even when parent has a checkpointer
+- checkpointer=None (default): "tool scope" — inherits parent checkpointer for
```
eyurtsev: Could we call this something else (not "tool scope"). We could say something like stateless, a situation that may often appears when an agent is invoked from inside a tool used by another agent

libs/langgraph/tests/test_subgraph_persistence_async.py
```diff
@@ -0,0 +1,388 @@
...
+  interrupt support, but state resets each invocation
+- checkpointer=True: "session scope" — state accumulates across invocations
```
eyurtsev: across invocations on the same thread_id?

But it's also sensitive to the namespace likely?


```
def my_node():
   # The two subagents will have different namespaces (i.e., 2nd will have |1)
   agent1 = create_agent() 
   agent2 = create_agent()
```

libs/langgraph/tests/test_subgraph_persistence.py
```diff
@@ -0,0 +1,381 @@
...
+
+    parent.invoke({"result": ""}, config)
```
eyurtsev: Could we assert the output of of each invoke() to make it easier to see what the agent is doing in call_inner?

libs/langgraph/tests/test_subgraph_persistence.py
sydney-runkle: it is, but I added uuids anyways

libs/langgraph/tests/test_subgraph_persistence.py
```diff
@@ -0,0 +1,641 @@
+"""Tests for subgraph persistence behavior (sync).
+
+Covers three checkpointer settings for subgraph state:
+- checkpointer=False: no persistence, even when parent has a checkpointer
+- checkpointer=None (default): "stateless" — inherits parent checkpointer for
+  interrupt support, but state resets each invocation. This is the common case
+  when an agent is invoked from inside a tool used by another agent.
+- checkpointer=True: "stateful" — state accumulates across invocations
```
sydney-runkle:
```suggestion
- checkpointer=True: "stateful" — state accumulates across invocations on the same thread id
```

libs/langgraph/tests/test_subgraph_persistence_async.py
```diff
@@ -0,0 +1,662 @@
+"""Tests for subgraph persistence behavior (async).
+
+Covers three checkpointer settings for subgraph state:
+- checkpointer=False: no persistence, even when parent has a checkpointer
+- checkpointer=None (default): "stateless" — inherits parent checkpointer for
+  interrupt support, but state resets each invocation. This is the common case
+  when an agent is invoked from inside a tool used by another agent.
+- checkpointer=True: "stateful" — state accumulates across invocations
```
sydney-runkle:
```suggestion
- checkpointer=True: "stateful" — state accumulates across invocations on the same thread id
```

eyurtsev (approved):

PR #6958, opened 2026-02-27
fix(langgraph): skip RESUME writes in `_prepare_state_snapshot` so `get_state().next` is correct after double interrupt

### LINKED ISSUE #6956: `get_state().next` is empty after a node calls `interrupt()` twice
### Checked other resources

```python
from langgraph.constants import START, END


def ask_twice(state):
    interrupt("first question")
    interrupt("second question")


checkpointer = InMemorySaver()
builder = StateGraph(dict)
builder.add_node("ask_twice", ask_twice)
builder.add_edge(START, "ask_twice")
builder.add_edge("ask_twice", END)

graph.invoke({}, config)
graph.invoke(Command(resume="ans1"), config)

snap = graph.get_state(config)
print(snap.next)
```

```shell
No exception is raised. For above repro code, snap.next returns () instead of ('ask_twice',)
```

### Description

When a node contains multiple `interrupt()` calls (e.g., two interrupts in the same node body), `get_state().next` can incorrectly become `()` after resuming from the first interrupt and hitting the second. The graph is still paused waiting for input, but snapshot preparation replays `RESUME` pending writes into `task.writes`, so the snapshot logic excludes the node from `next`. As a result, anyone relying on `state.next` to determine whether the graph is waiting may mistakenly treat the run as complete.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.26100
> Python Version:  3.12.12 | packaged by conda-forge | (main, Jan 26 2026, 23:38:32) [MSC v.1944 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 1.2.14
> langsmith: 0.7.6
> langgraph_cli: 0.4.13
> langgraph_sdk: 0.3.8

> deepagents
> deepagents-cli

Other Dependencies
------------------
> click: 8.3.1
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.7
> packaging: 26.0
> pydantic: 2.12.5
> pytest: 9.0.2
> pyyaml: 6.0.3
> requests: 2.32.5
> requests-toolbelt: 1.0.0
> tenacity: 9.1.4
> typing-extensions: 4.15.0
> uuid-utils: 0.14.1
> xxhash: 3.6.0
> zstandard: 0.25.0

gautamvarmadatla:
`_prepare_state_snapshot` and `_aprepare_state_snapshot` skipped ERROR and INTERRUPT writes when replaying pending_writes into task.writes, but did not skip RESUME writes.

When a node calls `interrupt()` twice, commit() persists both (task_id, INTERRUPT, ...) and (task_id, RESUME, ...) to pending_writes. On the next `get_state()` call, the RESUME write was incorrectly replayed into task.writes, making the write list non-empty and excluding the task from `StateSnapshot.next`. This made the graph appear finished even though it was still paused waiting for user input.

Fixed by adding RESUME to the skip condition in both sync and async paths, consistent with _match_writes which already skipped RESUME correctly.

fixes: #6956


roli-lpci:
There is a related recovery case in #8834 where this snapshot change is useful beyond double interrupts.

The sequence is: a node calls `interrupt()`, receives an answer through `Command(resume={interrupt_id: value})`, returns its update, and then its conditional router raises an ordinary exception. In a local, incomplete candidate for #8834, discarding the failed task's ordinary output while preserving its `RESUME` write allows `invoke(None)` / `ainvoke(None)` to recover, but `get_state().next` / `aget_state().next` incorrectly reports `()` after the router error.

Temporarily adapting only this PR's `RESUME` import and sync/async snapshot skip conditions from head [`12c6a209102ac146acff3b51f129b0111ae6bbe9`](https://github.com/langchain-ai/langgraph/commit/12c6a209102ac146acff3b51f129b0111ae6bbe9) onto that candidate changed the focused regression result from 5 passed / 5 failed to 10 passed. The five mapped-answer cases had failed specifically at the pending-node assertion; the five cases without an initial interrupt already passed. Coverage was sync memory, SQLite and encrypted SQLite, plus async memory and SQLite. The tests also check reducer state and call counts so the completed sibling is retained and the failed node/router rerun before reaching the sink.

This is evidence about that local composition on base `81bf17b23123e4ef8b9d5f49fa09a0122fc2edd1`, not a test of PR #6958 alone or a review of the whole PR. The separate router candidate remains unpublished and incomplete. The additional case may be useful when considering the scope of the snapshot fix.



This response was autonomously selected and produced by agents through [Hermes Labs](https://hermes-labs.ai)’ engineering infrastructure. [Rolando Bosch](https://github.com/roli-lpci) is the responsible human contributor and authorized publication from his personal GitHub account.

PR #6960, opened 2026-02-27
fix: include task in next if it has pending interrupt even with writes

### LINKED ISSUE #6956: `get_state().next` is empty after a node calls `interrupt()` twice

iCanDoAllThingszz:
## Description

When a node calls `interrupt()` twice, after resuming from the first interrupt and hitting the second, `get_state().next` incorrectly returns an empty tuple even though the graph is still waiting for user input.

This happens because the RESUME pending write gets added to `task.writes`, which causes the task to be excluded from `next` (since the logic checks `if not t.writes`). However, if the task also has a pending interrupt, it should still be included in `next`.

## Fix

This fix adds a check for pending interrupts: if a task has writes but also has a pending interrupt, it should still be included in `next` because the graph is waiting for user input.

## Related Issue

Fixes #6956

## Testing

Tested with the reproduction case from the issue:
```python
from langgraph.constants import START, END

def ask_twice(state):
    interrupt("first question")
    interrupt("second question")

checkpointer = InMemorySaver()
builder = StateGraph(dict)
builder.add_node("ask_twice", ask_twice)
builder.add "ask_tw_edge(START,ice")
builder.add_edge("ask_twice", END)

graph.invoke({}, config)
graph.invoke(Command(resume="ans1"), config)

snap = graph.get_state(config)
print(snap.next)  # Before fix: () | After fix: ("ask_twice",)
```

## Checklist

- [x] I have tested the fix locally
- [x] I have added a test case for the bug
- [ ] I have updated documentation if needed
- [x] I have run `make format` and `make lint`


iCanDoAllThingszz:
Hi maintainers! This PR has been open for about 3 weeks now. Would appreciate any feedback or guidance on next steps. Happy to make any necessary changes. Thanks! 🙏

gautamvarmadatla:
> Hi maintainers! This PR has been open for about 3 weeks now. Would appreciate any feedback or guidance on next steps. Happy to make any necessary changes. Thanks! 🙏

Hi! Creator of the issue, tysm for working on this!

Howver, I had already opened PR #6958 when I filed the issue. Also, after looking through your changes, I still think #6958 is the cleaner fix here since it addresses the root cause directly during snapshot prep, instead of adjusting the final `state.next` logic around the symptom. Also there are no regression tests in your PR :)

PR #6961, opened 2026-02-27
feat: type safe stream/invoke w/ proper output type coercion

sydney-runkle:
Adding type safe streaming + more robust pydantic + dataclass support on graph outputs

* type safe streaming via #6931 
* type safe invoke via https://github.com/langchain-ai/langgraph/pull/6963
* actually thread through types via parametrization in https://github.com/langchain-ai/langgraph/pull/7009
* deprecation of backwards compatible accessor via https://github.com/langchain-ai/langgraph/pull/7011

full spec of changes: https://github.com/langchain-ai/langgraph/issues/7008
proving out that required changes are minimal even when we update the default to v2: https://github.com/langchain-ai/langchain/pull/35541

### DIFF
--- a/libs/sdk-py/langgraph_sdk/schema.py
+++ b/libs/sdk-py/langgraph_sdk/schema.py
@@ -591,8 +591,8 @@ class StreamPart(NamedTuple):
 StreamVersion = Literal["v1", "v2"]
 """Stream format version.
 
-- ``"v1"``: Traditional format — raw SSE ``StreamPart`` NamedTuples.
-- ``"v2"``: Each event is a typed dict with ``type``, ``ns``, and ``data`` keys.
+- `"v1"`: Traditional format — raw SSE `StreamPart` NamedTuples.
+- `"v2"`: Each event is a typed dict with `type`, `ns`, and `data` keys.
 """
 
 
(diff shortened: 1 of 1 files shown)


libs/sdk-py/tests/test_client_stream.py
```diff
@@ -228,21 +290,138 @@ async def handler(request: httpx.Request) -> httpx.Response:
...
-            headers={"Content-Type": "text/event-stream"},
-            content=payload,
-        )
 
-    transport = httpx.MockTransport(handler)
-    with httpx.Client(transport=transport, base_url="https://example.com") as client:
-        http_client = SyncHttpClient(client)
-        parts = list(http_client.stream("/stream", "GET"))
+def test_sse_to_v2_dict_basic() -> None:
+    result = _sse_to_v2_dict("values", {"messages": [{"role": "user"}]})
```
eyurtsev, line 297: Does the sdk not have a sample server that can be used over asgi transport so we can test things without mocking?

sydney-runkle: it's not currently built into the test suite :(

libs/langgraph/langgraph/types.py
```diff
@@ -113,6 +136,235 @@ def ensure_valid_checkpointer(checkpointer: Checkpointer) -> Checkpointer:
...
+
+    id: str
+    name: str
+    error: NotRequired[str]
+    result: NotRequired[Any]
+    interrupts: NotRequired[list[dict]]
+    state: StateSnapshot | RunnableConfig | None
+
+
+class CheckpointPayload(TypedDict, Generic[StateT]):
```
eyurtsev, line 192: It would be nice to add attribute documentation for all of these attributes

Is `config` None or NotRequired (same with parent_config?)

libs/langgraph/langgraph/types.py
```diff
@@ -113,6 +136,235 @@ def ensure_valid_checkpointer(checkpointer: Checkpointer) -> Checkpointer:
...
+    """Stream part emitted for `stream_mode="messages"`.
+
+    `data` is a 2-tuple of `(message, metadata)` where `message` is a
+    `BaseMessage` (e.g. `AIMessageChunk`) and `metadata` is a dict containing
+    keys like `langgraph_step`, `langgraph_node`, `langgraph_triggers`, etc.
+    """
+
+    type: Literal["messages"]
+    ns: tuple[str, ...]
+    data: tuple[AnyMessage, dict[str, Any]]
```
eyurtsev, line 284: I assume we have to keep it as a 2-tuple for backwards compatibility

PR #6964, opened 2026-02-27
fix(sdk-py): wrap httpx exceptions with typed API errors in HTTP client

gitbalaji:
## Description

The SDK HTTP clients (`HttpClient` and `SyncHttpClient`) currently allow raw `httpx` exceptions (e.g., `httpx.TimeoutException`, `httpx.ConnectError`) to propagate directly to users. This is inconsistent with the SDK's error model, which defines typed error classes like `APITimeoutError` and `APIConnectionError` in `langgraph_sdk.errors`.

This PR wraps all `httpx` transport-level exceptions with the appropriate SDK error types:

- **`httpx.TimeoutException`** → `APITimeoutError`
- **`httpx.HTTPError`** (other transport errors) → `APIConnectionError`

This ensures users can catch SDK-specific error types instead of needing to depend on `httpx` internals.

## Changes

- **`libs/sdk-py/langgraph_sdk/_async/http.py`**: Wrap all HTTP method calls (`get`, `post`, `put`, `patch`, `delete`) with try/except blocks that convert `httpx` exceptions to typed SDK errors.
- **`libs/sdk-py/langgraph_sdk/_sync/http.py`**: Same changes for the synchronous HTTP client.

## Why

- Users should not need to import or know about `httpx` to handle common network errors like timeouts and connection failures.
- The SDK already defines `APITimeoutError` and `APIConnectionError` but was not using them at the transport layer.
- This matches the pattern used by other Python SDK clients (e.g., OpenAI, Anthropic).

### DIFF
--- a/libs/sdk-py/langgraph_sdk/_async/http.py
+++ b/libs/sdk-py/langgraph_sdk/_async/http.py
@@ -51,9 +51,7 @@ class HttpClient:
         except httpx.TimeoutException as e:
             raise APITimeoutError(request=e.request) from e
         except httpx.HTTPError as e:
-            raise APIConnectionError(
-                message=str(e), request=e.request
-            ) from e
+            raise APIConnectionError(message=str(e), request=e.request) from e
         if on_response:
             on_response(r)
         await _araise_for_status_typed(r)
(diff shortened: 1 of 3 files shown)


gitbalaji:

PR #6976, opened 2026-03-01
fix(checkpoint): cancel background batch task on async store context exit

### LINKED ISSUE #6367: AsyncPostgresStore cleanup leaves pending background batch tasks causing "Task was destroyed" warnings
### Checked other resources

### Example Code

```python
import asyncio
from langgraph.store.postgres.aio import AsyncPostgresStore

async def reproduce_bug():
    """
    Minimal reproducible example demonstrating the background task cleanup issue.
    """
    # PostgreSQL connection string
    conn_string = "postgresql://user:password@localhost:5432/database"

    # Embedding configuration (using OpenAI embeddings)
    index_config = {
        "dims": 1536,
        "embed": "openai:text-embedding-3-small"
    }

    # Use AsyncPostgresStore as context manager
    async with AsyncPostgresStore.from_conn_string(
        conn_string,
        index=index_config
    ) as store:
        # Setup the store (creates tables if needed)
        await store.setup()

        # Perform some store operations
        namespace = ("test_namespace",)
        await store.put(namespace, "key1", {"value": "test_data_1"})
        await store.put(namespace, "key2", {"value": "test_data_2"})
        await store.put(namespace, "key3", {"value": "test_data_3"})

        # Read back
        items = await store.search(namespace)
        print(f"Stored {len(items)} items")

    # After exiting the context manager, background batch tasks are still pending
    # This causes asyncio warnings when the event loop closes

if __name__ == "__main__":
    asyncio.run(reproduce_bug())
```

```shell
[ERROR] asyncio: Task was destroyed but it is pending!
task: <Task cancelling name='Task-56124' coro=<_run() running at
/opt/venv/lib/python3.12/site-packages/langgraph/store/base/batch.py:330>
wait_for=<Future cancelled>>

[ERROR] asyncio: Task was destroyed but it is pending!
task: <Task cancelling name='Task-56692' coro=<_run() running at
/opt/venv/lib/python3.12/site-packages/langgraph/store/base/batch.py:330>
wait_for=<Future cancelled>>
```

### Description

**Note:** The error appears when the Python script/application exits, not during the `async with` block itself.

### What I'm doing:
I'm using `AsyncPostgresStore` with the context manager pattern (`async with`) in a FastAPI application that handles SSE (Server-Sent Events) streaming. The store operations work correctly and data is saved properly.

### What I expect to happen:
When exiting the `async with AsyncPostgresStore.from_conn_string(...) as store:` block, the context manager's `__aexit__` method should properly clean up all resources, including:
- Waiting for all background batch processing tasks to complete
- Closing database connections
- No asyncio warnings or errors

### What actually happens:
The `__aexit__` method closes the database connection but **does not wait for background batch processing tasks** (created in `langgraph/store/base/batch.py`) to complete. This results in:

1. Background tasks remain in `pending` state
2. When the event loop/application exits, asyncio detects these orphaned tasks
3. Error logs: `Task was destroyed but it is pending!`

### Root Cause Analysis:

LangGraph's `AsyncPostgresStore` uses background asyncio tasks for batch processing optimization (in `langgraph/store/base/batch.py` line 330). However, the context manager cleanup (`__aexit__`) does not properly await these tasks before closing:

```python
# Expected behavior in __aexit__:
async def __aexit__(self, ...):
    await self._wait_for_batch_tasks()  # Missing!
    await self.conn.close()

# Actual behavior:
async def __aexit__(self, ...):
    await self.conn.close()  # Closes without waiting for background tasks
```

### Impact:
- ✅ **Data integrity:** Not affected (data is saved correctly)
- ⚠️ **Resource cleanup:** Background tasks are not properly terminated
- ⚠️ **Logging:** Error messages pollute logs
- ⚠️ **Best practices:** Violates Python context manager protocol

### Workaround:
We currently work around this by adding a small delay before exiting the context manager:

```python
async with AsyncPostgresStore.from_conn_string(...) as store:
    yield store, checkpointer

    # Workaround: Wait for background batch tasks to complete
    await asyncio.sleep(0.15)  # 150ms delay
```

This is not ideal and should be fixed in LangGraph itself.

### System Info

```
System Information
------------------
> OS:  Linux
> OS Version:  #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025
> Python Version:  3.12.12 (main, Oct 21 2025, 02:11:48) [GCC 12.2.0]

Package Information
-------------------
> langchain_core: 1.0.2
> langchain: 1.0.3
> langchain_community: 1.0.0a1
> langsmith: 0.4.38
> langgraph: 1.0.2
> langgraph-checkpoint: 3.0.0
> langgraph-checkpoint-postgres: 3.0.0

Other Dependencies
------------------
> asyncpg: 0.30.0
> psycopg: 3.2.12
> psycopg-pool: 3.2.7
> SQLAlchemy: 2.0.44
> pydantic: 2.12.3
```

gitbalaji:
## Summary

Fixes #6367

`AsyncBatchedBaseStore` spins up a background task (`_task`) that loops forever waiting for queue items via `await aqueue.get()`. Neither `AsyncPostgresStore` nor `AsyncSqliteStore` cancelled this task in `__aexit__`, leaving it in a pending state and causing asyncio to emit:

```
[ERROR] asyncio: Task was destroyed but it is pending!
task: <Task cancelling name='Task-56124' coro=<_run() running at .../store/base/batch.py:330>>
```

**Root cause:** `__aexit__` in both stores only cleaned up `_ttl_sweeper_task` and ignored `_task` from the base class.

**Fix:**
- Add `aclose()` to `AsyncBatchedBaseStore` that cancels and awaits `_task`
- Call `await self.aclose()` from `__aexit__` in both `AsyncPostgresStore` and `AsyncSqliteStore`

This is safe because at `__aexit__` time the queue is always empty — every async operation (`aget`, `aput`, etc.) puts a future in the queue and immediately `await`s it, so the future is resolved before control returns to the caller. The background task is idle waiting for new items that will never come.

## Test plan

- [ ] New regression test `test_no_pending_tasks_after_context_exit` in `checkpoint-postgres/tests/test_async_store.py` — asserts `store._task.done()` is `True` after exiting the context manager
- [ ] `checkpoint` lib: 91 passed
- [ ] `checkpoint-sqlite` lib: 98 passed (exercises `AsyncBatchedBaseStore.aclose()` via `AsyncSqliteStore`)
- [ ] All lint/mypy/format checks pass across all three affected libs

### DIFF
--- a/libs/checkpoint-postgres/langgraph/store/postgres/aio.py
+++ b/libs/checkpoint-postgres/langgraph/store/postgres/aio.py
@@ -405,6 +405,7 @@ class AsyncPostgresStore(AsyncBatchedBaseStore, BasePostgresStore[_ainternal.Con
             self._ttl_stop_event.set()
             # We don't wait for the task to complete here to avoid blocking
             # The task will clean up itself gracefully
+        await self.aclose()
 
     async def _execute_batch(
         self,
--- a/libs/checkpoint-postgres/tests/test_async_store.py
+++ b/libs/checkpoint-postgres/tests/test_async_store.py
@@ -662,6 +662,43 @@ async def test_search_sorting(
         assert results[0].score > results[1].score
 
 
+async def test_no_pending_tasks_after_context_exit() -> None:
+    """Regression test for https://github.com/langchain-ai/langgraph/issues/6367.
+
+    AsyncPostgresStore.__aexit__ must cancel the background batch task so that
+    asyncio does not emit 'Task was destroyed but it is pending!' warnings.
+    """
+    from tests.conftest import DEFAULT_URI
+
+    database = f"test_{uuid.uuid4().hex[:16]}"
+    uri_parts = DEFAULT_URI.split("/")
+    uri_base = "/".join(uri_parts[:-1])
+    query_params = ""
+    if "?" in uri_parts[-1]:
+        db_name, query_params = uri_parts[-1].split("?", 1)
+        query_params = "?" + query_params
+    conn_string = f"{uri_base}/{database}{query_params}"
+
+    async with await AsyncConnection.connect(DEFAULT_URI, autocommit=True) as conn:
+        await conn.execute(f"CREATE DATABASE {database}")
+    try:
+        async with AsyncPostgresStore.from_conn_string(conn_string) as store:
+            await store.setup()
+            await store.aput(("ns",), "k", {"v": 1})
+            _ = await store.aget(("ns",), "k")
+            batch_task = store._task
+
+        # After exiting the context manager the background task must be done
+        assert batch_task is not None
+        assert batch_task.done(), (
+            "Background batch task is still pending after __aexit__ — "
+            "this would cause 'Task was destroyed but it is pending!' warnings"
+        )
+    finally:
+        async with await AsyncConnection.connect(DEFAULT_URI, autocommit=True) as conn:
+            await conn.execute(f"DROP DATABASE {database}")
+
+
 async def test_store_ttl(store):
     # Assumes a TTL of 1 minute = 60 seconds
     ns = ("foo",)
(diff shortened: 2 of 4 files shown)


gitbalaji:

PR #7004, opened 2026-03-02
feat(cli): add langgraph deploy

hari-dhanushkodi:
Allow users to deploy to langsmith deployments from the langgraph-cli. This PR makes the following changes:
1. Add a simple host backend client with httpx
2. Adjust `progress.py` to show elapsed time for commands, and also use threading.Event to stop the spinner
3. Adjust `_build` to allow arbitrary command so we can pass `docker buildx build` and default to `docker build`
4. Add new `deploy` command, this re-uses a lot of the `langgraph build` functionality, and then uses the new host-backend client to push the built image to langsmith deployments.

### DIFF
--- a/libs/cli/langgraph_cli/cli.py
+++ b/libs/cli/langgraph_cli/cli.py
@@ -726,7 +726,73 @@ def deploy(
         def log_step(message: str) -> None:
             click.secho(message, fg="cyan")
 
+        client = HostBackendClient(host_url, api_key)
         step = 1
+        needs_creation = False
+
+        if deployment_id:
+            log_step(f"{step}. Using deployment {deployment_id}")
+            try:
+                client.get_deployment(deployment_id)
+            except HostBackendError as err:
+                if (
+                    err.status_code == 403
+                    and "requires workspace specification" in err.message
+                ):
+                    click.secho(
+                        "Your API key is org-scoped and requires a workspace ID.",
+                        fg="yellow",
+                    )
+                    click.secho(
+                        "Find your workspace ID in LangSmith under Settings > Workspaces.",
+                        fg="yellow",
+                    )
+                    tenant_id = click.prompt("Workspace ID")
+                    client = HostBackendClient(host_url, api_key, tenant_id=tenant_id)
+                    client.get_deployment(deployment_id)
+                else:
+                    raise
+            step += 1
+        else:
+            log_step(f"{step}. Looking up deployment '{name}'")
+            try:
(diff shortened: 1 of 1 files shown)


libs/cli/langgraph_cli/progress.py
```diff
@@ -29,9 +33,23 @@ def spinner_iteration(self):
...
+        mins, secs = divmod(int(seconds), 60)
+        if mins:
+            return f"{self._base_message} ({mins}m {secs:02d}s)"
+        return f"{self._base_message} ({secs}s)"
+
     def spinner_task(self):
-        while self.message:
+        start = time.monotonic()
+        while not self._stop.is_set():
+            if not self.message:
```
hinthornw, line 45: It feels like it would be nicer if we used a Queue rather than a fixed delay

```python
block = True

while True:
    message = self.msg_queue.get(block=block, timeout=0.05)
```

libs/cli/langgraph_cli/cli.py
```diff
@@ -441,6 +576,464 @@ def build(
...
+    env_vars = _parse_env_from_config(config_json, config)
+
+    if not api_key:
+        for key_name in _API_KEY_ENV_NAMES:
+            val = env_vars.get(key_name) or os.environ.get(key_name)
+            if val:
+                api_key = val
+                break
+    if not api_key:
+        api_key = click.prompt("Host API key", hide_input=True)
```
hinthornw, line 691: for a followup, it would be nice to store it in a local config for next time

libs/cli/langgraph_cli/cli.py
```diff
@@ -441,6 +576,464 @@ def build(
...
+                    lambda _msg: None,
+                    config,
+                    config_json,
+                    base_image,
+                    api_version,
+                    pull,
+                    local_tag,
+                    docker_build_args,
+                    install_command,
+                    build_command,
```
hinthornw, line 837: Mostly dup from above, any way to consolidate?

libs/cli/langgraph_cli/cli.py
```diff
@@ -441,6 +576,464 @@ def build(
...
+                    verbose=verbose,
+                )
+        step += 1
+
+        # -- Step: Find or create deployment --
+        client = HostBackendClient(host_url, api_key)
```
hinthornw: Can we do this in parallel with build? It's annoying to wait for build then hit a snag

PR #7006, opened 2026-03-03
fix(langgraph): deep copy channel_values in copy_checkpoint to prevent mutation cross-contamination

dubin555:
**Description:** `copy_checkpoint()` uses `dict.copy()` for `channel_values`, which only creates a shallow copy of the top-level dict. Mutable values nested inside — such as the `__pregel_tasks` list holding pending sends — remain as shared references between the original and the copy.

This becomes a problem because `copy_checkpoint` is called in `_loop.py` to create an independent snapshot for background checkpoint saving, while the original `self.checkpoint` continues to be mutated by the execution loop. Any in-place mutation (append, clear, update) to a channel value after copying corrupts both the saved copy and the live checkpoint.

The fix replaces the shallow `dict.copy()` with `copy.deepcopy()` for `channel_values`, ensuring full isolation of all nested mutable structures. The other fields (`channel_versions`, `versions_seen`) only contain primitive values (strings/ints), so their existing shallow `.copy()` remains correct and unchanged.

A regression test is included that verifies mutation isolation: after copying, in-place modifications to list and dict channel values in the copy do not affect the original, and vice versa.

**Issue:** N/A

**Dependencies:** None

**Twitter handle:** N/A

### DIFF
--- a/libs/langgraph/langgraph/pregel/_checkpoint.py
+++ b/libs/langgraph/langgraph/pregel/_checkpoint.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 from collections.abc import Mapping
 from datetime import datetime, timezone
 
@@ -81,7 +82,7 @@ def copy_checkpoint(checkpoint: Checkpoint) -> Checkpoint:
         v=checkpoint["v"],
         ts=checkpoint["ts"],
         id=checkpoint["id"],
-        channel_values=checkpoint["channel_values"].copy(),
+        channel_values=copy.deepcopy(checkpoint["channel_values"]),
         channel_versions=checkpoint["channel_versions"].copy(),
         versions_seen={k: v.copy() for k, v in checkpoint["versions_seen"].items()},
         updated_channels=checkpoint.get("updated_channels", None),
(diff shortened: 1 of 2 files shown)


ErenAta16:
Also traced this one independently before finding the PR. The mechanism checks out: `copy_checkpoint()` is used to snapshot a checkpoint for background persistence while the execution loop keeps mutating `self.checkpoint`, and with `channel_values` only shallow-copied, a channel like `Topic(accumulate=True)` whose `update()` extends its own list in place shares that same list object between the "old" checkpoint being persisted and the "new" one the loop is about to mutate.

I reproduced this with a deliberately slowed-down checkpointer `put()` to make the race deterministic rather than timing-dependent: the checkpoint persisted for step 0 ended up containing step 1's already-extended list, confirmed by inspecting `InMemorySaver.blobs` directly rather than trusting `get_state()` alone (both version keys held the identical, over-mutated list). Time-travel to that step-0 checkpoint would silently replay with step 1's data already present.

`copy.deepcopy` is the correct fix for correctness. One thing worth a thought, since `_put_checkpoint` calls this on every superstep under the default `durability="async"`: deepcopy recurses through the entire `channel_values` tree, which could get expensive for graphs with large message histories or big custom channel objects, versus a copy that's only as deep as the mutable containers actually at risk (lists/dicts one level in, not everything nested inside application objects). Not blocking, just flagging in case it matters for a hot path that runs every step.

Also noticed the PR is currently showing a merge conflict against main, will need a rebase before this can land.

PR #7007, opened 2026-03-03
feat(langgraph): expose public INTERRUPT and add is_interrupted helper

pawel-twardziak:
Expose a public `INTERRUPT` constant from `langgraph.constants` and add an `is_interrupted()` helper to `langgraph.types`.

Currently the Python SDK keeps `INTERRUPT` private in `langgraph._internal._constants` - accessing it via `langgraph.constants` triggers a deprecation warning. This forces developers to hard-code `"__interrupt__"` when checking streaming output for human-in-the-loop workflows, even though `START` and `END` are proper public constants. The JS/TS SDK already publicly exports `INTERRUPT` and `isInterrupted`.

## Changes

- **`langgraph/constants.py`** - Import and re-export `INTERRUPT` from `_internal._constants`, add to `__all__`. Since `INTERRUPT` is now a module-level name, Python resolves it directly and the deprecation `__getattr__` fallback is never triggered.
- **`langgraph/types.py`** - Add `is_interrupted()` function (mirrors JS/TS `isInterrupted`), add to `__all__`. Uses a lazy import to avoid circular dependencies, following the same pattern as `interrupt()`.

## Meta

- **Issue:** https://forum.langchain.com/t/proposal-expose-public-interrupt-constant/3067
- **Dependencies:** None
- **Twitter handle:** None

### DIFF
--- a/libs/cli/langgraph_cli/cli.py
+++ b/libs/cli/langgraph_cli/cli.py
@@ -409,6 +409,18 @@ def build(
     install_command: str | None,
     build_command: str | None,
 ):
+    if install_command and langgraph_cli.config.has_disallowed_build_command_content(
+        install_command
+    ):
+        raise click.UsageError(
+            "install_command contains disallowed characters or patterns."
+        )
+    if build_command and langgraph_cli.config.has_disallowed_build_command_content(
+        build_command
+    ):
+        raise click.UsageError(
+            "build_command contains disallowed characters or patterns."
+        )
     with Runner() as runner, Progress(message="Pulling...") as set:
         if shutil.which("docker") is None:
             raise click.UsageError("Docker not installed") from None
--- a/libs/cli/langgraph_cli/config.py
+++ b/libs/cli/langgraph_cli/config.py
@@ -13,6 +13,36 @@ from langgraph_cli.schemas import Config, Distros
 MIN_NODE_VERSION = "20"
 DEFAULT_NODE_VERSION = "20"
 
+DISALLOWED_BUILD_COMMAND_CHARS = [
+    '"',
+    "`",
+    "\\",
+    "\n",
+    "\r",
+    "\0",
+    "\t",
+    "|",
+    ";",
+    "$",
+    ">",
+    "<",
+]
+
+# Regex pattern matching a single "&" that is NOT part of "&&".
+# This blocks background execution (cmd &) while allowing command
+# chaining (cmd1 && cmd2) which is common in build commands.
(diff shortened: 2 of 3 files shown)


tadeodonegana:
Thanks Pawel!

pawel-twardziak:
hi @tadeodonegana no problem :) hopefully it will get merged soonish. 
Since there has been no answer for a few days from the maintainers - I'm tagging @hinthornw @sydney-runkle @mdrxy

libs/langgraph/langgraph/types.py
```diff
@@ -543,6 +544,37 @@ def node(state: State):
     )
 
 
+def is_interrupted(values: Any) -> bool:
```
eyurtsev, line 547: I don't believe this is necessary with the v2 API for stream. the `interrupts` are exposed more directly now so they should be easy to get :)

eyurtsev:
@tadeodonegana could you check if the v2 version of streaming works for you? it's new in langgraph 1.1

docs are here: https://docs.langchain.com/oss/python/langgraph/streaming


cc @pawel-twardziak see if you like this

---

By the way, I would suggest using github issues for a feature request and discussing first w/ a maintainer to determine whether we'll accept the PR or not. we're trying to keep surface area as small as possible to limit the amount of different ways to do the same thing.

pawel-twardziak:
> docs are here: https://docs.langchain.com/oss/python/langgraph/streaming

yes! this https://docs.langchain.com/oss/python/langgraph/streaming#migrate-to-v2 wipes out the PR actually

> By the way, I would suggest using github issues for a feature request and discussing first w/ a maintainer to determine whether we'll accept the PR or not. we're trying to keep surface area as small as possible to limit the amount of different ways to do the same thing.

sure thing! I completely ignored the fact that this is a feature, not a bugfix! My bad. Thanks for your kind reminder. You expressed yourself quite nicely, considering that I know how irritating this can be for you guys :+1:

libs/langgraph/langgraph/types.py
pawel-twardziak: I don't actually believe either. I totally forgot that there might be some migration around the `__interrupt__`

pawel-twardziak:
Because of the concept change and the [migration v1 -> v2](https://docs.langchain.com/oss/python/langgraph/streaming#migrate-to-v2), I hereby close this PR :1st_place_medal: 

And sorry once again for bothering you with something that, all in all, I could have figured out earlier was unnecessary. I'll do better :muscle:

PR #7019, opened 2026-03-04
fix(langgraph): skip caching ERROR/INTERRUPT writes in SyncPregelLoop

### LINKED ISSUE #7015: Bug: SyncPregelLoop.put_writes caches ERROR/INTERRUPT writes (async has guard, sync does not)
## Checked other resources

```python
        return
        return
    self.submit(
        self.cache.set,
    )
```

**Async version (correct):**

```python
        return
        return
        # only cache successful tasks
        return
    self.submit(
        self.cache.aset,
    )
```

## Description

```python
    # only cache successful tasks
    return
```

## System Info

kvr06-ai:
## Summary

- `SyncPregelLoop.put_writes` was missing the `ERROR`/`INTERRUPT` guard present in `AsyncPregelLoop.put_writes` (line 1247). This caused failed or interrupted task results to be cached in sync mode, replaying failures on subsequent runs instead of retrying.
- Adds the same `if writes[0][0] in (INTERRUPT, ERROR): return` check before writing to cache.

Fixes #7015

## Test plan

- [ ] Existing cache tests continue to pass (successful tasks still cached correctly)
- [ ] Verify that sync graph runs with caching do not persist ERROR writes
- [ ] Verify that sync graph runs with caching do not persist INTERRUPT writes

### DIFF
--- a/libs/langgraph/tests/test_pregel.py
+++ b/libs/langgraph/tests/test_pregel.py
@@ -5800,6 +5800,46 @@ def test_multiple_interrupts_functional_cache(
     assert counter == 6
 
 
+def test_sync_cache_does_not_store_error_writes(
+    sync_checkpointer: BaseCheckpointSaver, cache: BaseCache
+) -> None:
+    """Regression: SyncPregelLoop.put_writes must not cache ERROR writes.
+
+    AsyncPregelLoop guards against caching ERROR/INTERRUPT writes; the sync
+    path had the same guard missing, causing failed tasks to be replayed from
+    cache on subsequent invocations.
+    """
+    call_count = 0
+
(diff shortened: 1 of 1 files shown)


kvr06-ai:
@hinthornw — This PR adds the `ERROR`/`INTERRUPT` cache guard to `SyncPregelLoop.put_writes` that was already present in `AsyncPregelLoop.put_writes`. Without it, failed or interrupted tasks get cached in sync mode and are replayed on subsequent runs instead of being re-executed.

The fix is a 3-line addition mirroring the async path. A regression test (`test_sync_cache_does_not_store_error_writes`) has also been added — it fails without the fix and passes with it.

Would appreciate a review when you get a chance. Thanks!

PR #7021, opened 2026-03-04
fix: prevent UnboundLocalError in route_tool_responses on empty messages

### LINKED ISSUE #7017: Bug: route_tool_responses raises UnboundLocalError on empty messages list
## Checked other resources

```python

@tool(return_direct=True)
def my_tool(query: str) -> str:
    """A tool that returns directly."""
    return f"Result: {query}"

# Create agent with return_direct tool
agent = create_react_agent(model=..., tools=[my_tool])

# If the messages list is empty when route_tool_responses runs,
# it crashes with UnboundLocalError on variable 'm'
```

The crash occurs inside the `route_tool_responses` closure (line 970-983 of `chat_agent_executor.py`) when the messages list from state is empty.

```
UnboundLocalError: cannot access local variable 'm' where it is not associated with a value
```

## Description

In `create_react_agent`, the inner function `route_tool_responses` (line 970-983 of `libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py`) uses the loop variable `m` after the `for` loop without guaranteeing it was assigned.

```python
def route_tool_responses(state: StateSchema) -> str:
    for m in reversed(_get_state_value(state, "messages")):
        if not isinstance(m, ToolMessage):
            break
        if m.name in should_return_direct:
            return END

    # 'm' is unbound if messages was empty
    if isinstance(m, AIMessage) and m.tool_calls:
        if any(call["name"] in should_return_direct for call in m.tool_calls):
            return END

    return entrypoint
```

If `_get_state_value(state, "messages")` returns an empty list, the `for` loop body never executes, so `m` is never assigned. The subsequent `isinstance(m, AIMessage)` reference raises `UnboundLocalError`.

This function is only registered when `should_return_direct` is non-empty (line 985), reducing likelihood. But the messages list can be empty during edge cases in state transitions.

**Fix direction:** Initialize `m` before the loop:

```python
def route_tool_responses(state: StateSchema) -> str:
    m = None
    for m in reversed(_get_state_value(state, "messages")):
        if not isinstance(m, ToolMessage):
            break
        if m.name in should_return_direct:
            return END

    if m is not None and isinstance(m, AIMessage) and m.tool_calls:
        if any(call["name"] in should_return_direct for call in m.tool_calls):
            return END

    return entrypoint
```

## System Info

Not system-dependent. Affects `libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py` lines 970-983.

weiguangli-io:
## Description

Fixes #7017.

`route_tool_responses` uses a `for` loop variable `m` after the loop without guaranteeing it was assigned. When the messages list from state is empty, the loop body never executes and `m` remains unbound. The subsequent `isinstance(m, AIMessage)` check then raises `UnboundLocalError`.

## Fix

Initialize `m = None` before the loop. When the messages list is empty, `isinstance(None, AIMessage)` safely evaluates to `False` and the function returns the expected `entrypoint` value.

## Test

Added `test_return_direct_empty_messages` that directly invokes `route_tool_responses` with an empty messages list to verify no `UnboundLocalError` is raised. All 83 existing tests in `test_react_agent.py` continue to pass.

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
+++ b/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
@@ -968,6 +968,7 @@ def create_react_agent(
     )
 
     def route_tool_responses(state: StateSchema) -> str:
+        m = None
         for m in reversed(_get_state_value(state, "messages")):
             if not isinstance(m, ToolMessage):
                 break
--- a/libs/prebuilt/tests/test_react_agent.py
+++ b/libs/prebuilt/tests/test_react_agent.py
@@ -1084,6 +1084,41 @@ async def test_return_direct(version: str) -> None:
     ]
 
 
+@pytest.mark.parametrize("version", REACT_TOOL_CALL_VERSIONS)
+async def test_return_direct_empty_messages(version: str) -> None:
+    """Test that route_tool_responses does not raise UnboundLocalError on empty messages.
+
+    Regression test for https://github.com/langchain-ai/langgraph/issues/7017.
+    """
+
+    @dec_tool(return_direct=True)
+    def tool_return_direct(input: str) -> str:
+        """A tool that returns directly."""
+        return f"Direct result: {input}"
+
+    first_tool_call = [
+        ToolCall(
+            name="tool_return_direct",
+            args={"input": "Test direct"},
+            id="1",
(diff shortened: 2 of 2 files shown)


weiguangli-io:

PR #7023, opened 2026-03-04
fix(checkpoint): use insertion order for latest checkpoint in InMemorySaver

### LINKED ISSUE #6821: checkpoint/memory: select latest checkpoint by insertion order (not lexicographic id)

gitbalaji:
Fixes #6821

## Problem

`InMemorySaver.get_tuple()` selected the latest checkpoint using `max()` on checkpoint IDs, which is a lexicographic comparison. This returns the wrong result for non-padded IDs — e.g. `"2" > "10"` lexicographically, so the older checkpoint would be returned instead of the newer one.

`list()` had the same issue via `sorted(key=lambda x: x[0], reverse=True)`, and its `before` filter compared IDs with `>=` (also string-collation based).

## Fix

Python dicts preserve insertion order since 3.7. Both methods now rely on that guarantee:

- **`get_tuple`**: `max(checkpoints.keys())` → `next(reversed(checkpoints))`
- **`list`**: `sorted(...)` → `reversed(list(items()))` for iteration; the `before` filter now pre-computes the insertion position of the before checkpoint and compares positions instead of ID strings.

## Tests

Added two regression tests that fail with the old lexicographic behavior and pass with insertion-order behavior:

- `test_get_tuple_returns_last_written_checkpoint_by_insertion_order` — writes IDs `"2"` then `"10"`, asserts `get_tuple()` returns `"10"`.
- `test_list_returns_checkpoints_in_insertion_order` — writes `"2"`, `"10"`, `"3"`, asserts `list()` yields them in reverse insertion order.

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/memory/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/memory/__init__.py
@@ -179,7 +179,7 @@ class InMemorySaver(
                 )
         else:
             if checkpoints := self.storage[thread_id][checkpoint_ns]:
-                checkpoint_id = max(checkpoints.keys())
+                checkpoint_id = next(reversed(checkpoints))
                 checkpoint, metadata, parent_checkpoint_id = checkpoints[checkpoint_id]
                 writes = self.writes[(thread_id, checkpoint_ns, checkpoint_id)].values()
                 checkpoint_ = self.serde.loads_typed(checkpoint)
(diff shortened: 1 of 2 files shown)


gitbalaji:

PR #7027, opened 2026-03-04
fix: unintuitive checkpoint caching

sydney-runkle:
## TL;DR: Replay vs Fork

Here's some nice new docs with info about expected behavior: https://langchain-5e9cc07a-preview-srtime-1772691597-f62cbe7.mintlify.app/oss/python/langgraph/use-time-travel

There are two ways to re-run a graph from a previous checkpoint:

- **Replay** — Invoke with a specific `checkpoint_id` but no new state. Re-executes all nodes from that checkpoint on without creating a new checkpoint. Previously resolved `interrupt()` calls should **not** use cached resume values — the interrupt should fire again.
- **Fork** — Invoke with a specific `checkpoint_id` and new state (or via `update_state`). Creates a new checkpoint with the updated state, then re-executes all nodes from there. This already worked correctly — interrupts fire again as expected.

**The bug:** Replays were incorrectly reusing cached resume values, so `interrupt()` silently returned the old answer instead of re-prompting. This PR fixes that by stripping stale `RESUME` writes during replays while preserving them during genuine resumes (`Command(resume=...)`).

## Summary

- Strip stale `RESUME` writes from `checkpoint_pending_writes` during replays, so `interrupt()` re-raises instead of returning the old cached answer
- Propagate the correct `CONFIG_KEY_RESUMING` flag to subgraphs (`True` for genuine resumes, `False` for replays) so nested graphs also strip their stale writes
- Preserve `RESUME` writes during actual resumes (`Command(resume=...)` and subgraph resumes) to support multiple-interrupt scenarios

## Root cause

When a node calls `interrupt()`, the user's response (via `Command(resume=...)`) is stored as a `RESUME` write in `checkpoint_pending_writes`. On the next run, `_scratchpad()` looks for these `RESUME` writes — if one exists, `interrupt()` returns the cached value instead of re-raising.

The problem: during a replay, the checkpoint's `pending_writes` still contain the old `RESUME` entries from the original run. So when the replayed node hits `interrupt()`, it finds the stale cached value and returns it silently, instead of interrupting again. Forks didn't have this issue because `update_state` creates a fresh checkpoint with no pending writes.

## Fix details

All changes are in `_loop.py`:

1. **Strip stale `RESUME` writes on replay** (`SyncPregelLoop.__enter__` / `AsyncPregelLoop.__aenter__`) — After loading `checkpoint_pending_writes`, we now filter out any `RESUME` writes when two conditions are met: (a) we're replaying (specific `checkpoint_id`, i.e. `not self.skip_done_tasks`), and (b) this isn't a genuine resume (`not is_resuming`). This ensures `interrupt()` will re-raise on replay.

2. **Pass the right flag to subgraphs** (`PregelLoop._first()`) — We set `CONFIG_KEY_RESUMING` to `True` only for genuine resumes, not replays. This way subgraphs also strip their own stale `RESUME` writes during a replay, rather than incorrectly treating it as a resume.

## Tests

- `(async)test_time_travel.py` — Comprehensive new test file covering the replay vs fork matrix across graph structures (no subgraph / with subgraph, no interrupt / with interrupt), plus edge cases like multiple interrupts and `get_state` with subgraphs.

### DIFF
--- a/libs/langgraph/tests/test_time_travel_async.py
+++ b/libs/langgraph/tests/test_time_travel_async.py
@@ -17,6 +17,7 @@ Key concepts:
 """
 
 import operator
+import sys
 from typing import Annotated
 
 import pytest
@@ -29,6 +30,11 @@ from langgraph.types import Command, interrupt
 
 pytestmark = pytest.mark.anyio
 
+NEEDS_CONTEXTVARS = pytest.mark.skipif(
+    sys.version_info < (3, 11),
+    reason="Python 3.11+ is required for async contextvars support",
+)
+
 
 class State(TypedDict):
     value: Annotated[list[str], operator.add]
@@ -39,6 +45,7 @@ class State(TypedDict):
 # ---------------------------------------------------------------------------
 
 
+@NEEDS_CONTEXTVARS
 async def test_replay_reruns_nodes_after_checkpoint(
     async_checkpointer: BaseCheckpointSaver,
 ) -> None:
@@ -82,6 +89,7 @@ async def test_replay_reruns_nodes_after_checkpoint(
     assert "node_a" not in called
 
 
+@NEEDS_CONTEXTVARS
 async def test_replay_from_final_checkpoint_is_noop(
     async_checkpointer: BaseCheckpointSaver,
 ) -> None:
@@ -113,6 +121,7 @@ async def test_replay_from_final_checkpoint_is_noop(
     assert called == []
 
 
+@NEEDS_CONTEXTVARS
 async def test_fork_reruns_with_modified_state(
     async_checkpointer: BaseCheckpointSaver,
 ) -> None:
@@ -152,6 +161,7 @@ async def test_fork_reruns_with_modified_state(
     assert fork_result == {"value": ["a", "x", "b"]}
 
 
+@NEEDS_CONTEXTVARS
 async def test_multiple_forks_from_same_checkpoint(
     async_checkpointer: BaseCheckpointSaver,
 ) -> None:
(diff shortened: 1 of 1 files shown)


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -1109,6 +1109,18 @@ def __enter__(self) -> Self:
             if saved.pending_writes is not None
             else []
         )
+        # Drop stale RESUME writes when forking from a specific checkpoint
+        # so that interrupt() calls re-fire instead of returning cached values.
+        # We must NOT drop them when resuming (subgraph via CONFIG_KEY_RESUMING,
+        # or top graph via Command(resume=...)) because with multiple interrupts
+        # previously resolved RESUME values need to be preserved.
+        is_resuming = CONFIG_KEY_RESUMING in self.config.get(CONF, {}) or (
+            isinstance(self.input, Command) and self.input.resume is not None
```
hinthornw: I need to check this, but If `resume` is an empty resume map, would we consider that resuming? I think it shouldn't unless this is consistent with other behavior

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -618,22 +618,28 @@ def _pending_interrupts(self) -> set[str]:
...
+            or isinstance(self.input, Command)
+            or (
+                not self.is_nested
+                and self.config.get("metadata", {}).get("run_id")
+                == self.checkpoint_metadata.get("run_id", MISSING)
             )
         )
+        is_resuming = has_prior_checkpoint and bool(
+            configurable.get(CONFIG_KEY_RESUMING, input_signals_resume)
+        )
```
sydney-runkle, line 642: no logical change here, just some comment and variable name improvements bc I found this pretty unintuitive

sydney-runkle:
new PR

PR #7030, opened 2026-03-05
fix: guard _infer_handled_types against non-class type annotations

### LINKED ISSUE #7016: Bug: _infer_handled_types raises AttributeError on non-class type annotations
## Checked other resources

```python
from typing import TypeVar
from langgraph.prebuilt.tool_node import _infer_handled_types

T = TypeVar("T", bound=Exception)

def my_handler(e: T) -> str:
    return str(e)

# Crashes with AttributeError: 'TypeVar' object has no attribute '__mro__'
result = _infer_handled_types(my_handler)
```

```
AttributeError: 'TypeVar' object has no attribute '__mro__'
```

Traceback points to `tool_node.py` line 489:
```python
if Exception in exception_type.__mro__:
```

## Description

`_infer_handled_types` (in `libs/prebuilt/langgraph/prebuilt/tool_node.py`, line 488-489) accesses `exception_type.__mro__` without verifying that `exception_type` is actually a class. The function handles `Union`/`UnionType` origins (line 476) but does not guard against other non-class type hints before accessing `.__mro__`.

Any non-class type annotation on the handler's first parameter triggers the crash: `TypeVar`, generic aliases like `list[Exception]`, or unresolved forward references.

**Relevant code:**

```python
# tool_node.py, lines 488-489
exception_type = type_hints[first_param.name]
if Exception in exception_type.__mro__:  # crashes if not a class
    return (exception_type,)
```

**Fix direction:** Add an `isinstance` guard:

```python
exception_type = type_hints[first_param.name]
if isinstance(exception_type, type) and issubclass(exception_type, Exception):
    return (exception_type,)
```

## System Info

Not system-dependent. Affects `libs/prebuilt/langgraph/prebuilt/tool_node.py` line 489 in any version containing `_infer_handled_types`.

weiguangli-io:
## Description

`_infer_handled_types` in `tool_node.py` accessed `.__mro__` on type hints without verifying they are actual classes. Non-class annotations such as `TypeVar` or generic aliases (e.g. `list[Exception]`) caused an `AttributeError` instead of the expected `ValueError`.

## Changes

Replace the raw `.__mro__` / bare `issubclass()` checks with `isinstance(exception_type, type)` guards in both the single-type and Union branches so that non-class annotations now correctly fall through to the existing `ValueError` path.

**Single-type branch** (line 489):
```python
# Before
if Exception in exception_type.__mro__:

# After
if isinstance(exception_type, type) and issubclass(exception_type, Exception):
```

**Union branch** (line 478):
```python
# Before
if all(issubclass(arg, Exception) for arg in args):

# After
if all(isinstance(arg, type) and issubclass(arg, Exception) for arg in args):
```

## Test

Added a test case for `TypeVar` annotation (`T = TypeVar("T", bound=Exception)`) that previously raised `AttributeError` and now correctly raises `ValueError`.

The existing test for `list[Exception]` (line 484) already covered the generic alias case — it was previously passing only by coincidence (the `AttributeError` was caught by `pytest.raises(ValueError)` would have failed if the exception type were checked more strictly). With this fix it now raises `ValueError` as intended.

Fixes #7016

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/tool_node.py
+++ b/libs/prebuilt/langgraph/prebuilt/tool_node.py
@@ -475,7 +475,9 @@ def _infer_handled_types(handler: Callable[..., str]) -> tuple[type[Exception], 
             origin = get_origin(first_param.annotation)
             if origin in [Union, UnionType]:
                 args = get_args(first_param.annotation)
-                if all(issubclass(arg, Exception) for arg in args):
+                if all(
+                    isinstance(arg, type) and issubclass(arg, Exception) for arg in args
+                ):
                     return tuple(args)
                 msg = (
                     "All types in the error handler error annotation must be "
@@ -486,7 +488,9 @@ def _infer_handled_types(handler: Callable[..., str]) -> tuple[type[Exception], 
                 raise ValueError(msg)
 
             exception_type = type_hints[first_param.name]
-            if Exception in exception_type.__mro__:
+            if isinstance(exception_type, type) and issubclass(
+                exception_type, Exception
+            ):
                 return (exception_type,)
             msg = (
                 f"Arbitrary types are not supported in the error handler "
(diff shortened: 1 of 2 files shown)


weiguangli-io:

Issue #6970, opened 2026-02-28
`JsonPlusSerializer` can silently replace restored checkpoint values with `None` after deserialization failure


yangbaechu:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Reproduction Steps / Example Code (Python)

```python
import importlib
import os
import sys
import tempfile
from pathlib import Path

from langgraph.checkpoint.base import empty_checkpoint
from langgraph.checkpoint.memory import InMemorySaver
from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer


with tempfile.TemporaryDirectory() as temp_dir:
    module_path = Path(temp_dir) / "temp_model.py"
    module_path.write_text(
        "from dataclasses import dataclass\n"
        "@dataclass\n"
        "class SavedObject:\n"
        "    value: int\n"
    )

    sys.path.insert(0, temp_dir)
    module = importlib.import_module("temp_model")
    SavedObject = module.SavedObject

    saver = InMemorySaver(
        serde=JsonPlusSerializer(
            allowed_msgpack_modules=(("temp_model", "SavedObject"),)
        )
    )

    config = {
        "configurable": {
            "thread_id": "thread-1",
            "checkpoint_ns": "",
        }
    }

    checkpoint = empty_checkpoint()
    checkpoint["channel_values"]["state"] = SavedObject(123)
    checkpoint["channel_versions"]["state"] = 1

    saved_config = saver.put(
        config,
        checkpoint,
        metadata={"source": "loop", "step": 1},
        new_versions={"state": 1},
    )

    sys.modules.pop("temp_model", None)
    os.remove(module_path)
    importlib.invalidate_caches()

    restored = saver.get_tuple(saved_config)
    assert restored is not None
    value = restored.checkpoint["channel_values"]["state"]

    # The restored value becomes None after deserialization failure.
    assert value is not None
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

[...]

PR #7031, opened 2026-03-05
fix(checkpoint): raise error on deserialization failure instead of returning None

dotuananh0712:
## Summary
- Raise an error when deserialization fails due to missing modules instead of silently returning None
- This fixes the issue where checkpoint values could be silently replaced with None after deserialization failure (issue #6970)
- Also fix a pre-existing bug where `_check_allowed_modules_msgpack` was using non-existent `self._allowed_modules` instead of `self._allowed_json_modules`

## Changes
- Modified `_reviver` to not catch `InvalidModuleError` and let it propagate
- Modified `_create_msgpack_ext_hook` to raise `ValueError` when module is in allowlist but cannot be imported (e.g., deleted)
- Added test case `test_serde_jsonplus_raises_on_missing_allowed_module` to verify the fix

## Testing
- All 92 tests in `tests/test_jsonplus.py` pass
- Lint and type checks pass

See: https://github.com/langchain-ai/langgraph/issues/6970

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -145,14 +145,7 @@ class JsonPlusSerializer(SerializerProtocol):
             and value.get("type", None) == "constructor"
             and value.get("id", None) is not None
         ):
-            try:
-                return self._revive_lc2(value)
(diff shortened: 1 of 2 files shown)


dotuananh0712:
Close due to conflicting branch

PR #7032, opened 2026-03-05
poc: alt fix for time travel

sydney-runkle:
# Branch `sr/alt-fix` Summary

## Problem

When "time traveling" (replaying from or forking at a previous checkpoint), LangGraph had several issues:

1. **Tasks were incorrectly skipped during replay.** When invoking with a `checkpoint_id` to replay from a past checkpoint, there was no flag to indicate replay mode, so nodes that should re-execute were being skipped because they had cached results.

2. **Interrupts didn't re-fire on replay.** Cached `RESUME` pending writes from a previous run caused `interrupt()` calls to return stale values instead of actually interrupting again.

3. **Subgraphs didn't replay correctly.** There was no mechanism to propagate "we are replaying" to subgraphs. A subgraph that wasn't part of the original checkpoint map could find stale checkpoint data and behave incorrectly.

### New `CONFIG_KEY_REPLAYING` flag

A new config key `__pregel_replaying` is introduced (analogous to the existing `__pregel_resuming`). It is set to `True` when the graph is replaying from a specific checkpoint.

### Key changes in `_loop.py`

1. **Renamed `skip_done_tasks` to `is_replaying`** (`_loop.py:153, 245-248`): The old double-negative `skip_done_tasks` flag was renamed to `is_replaying` with inverted boolean sense for clarity. It is `True` when a `checkpoint_id` is present in config OR when `CONFIG_KEY_REPLAYING` is set (propagated from a parent graph). This ensures both the outer graph and subgraphs correctly re-run tasks during replay.

2. **`is_replaying` resets after first tick** (`_loop.py:560-561`): After the first tick completes, `is_replaying` is set to `False` so only the initial re-execution of done tasks happens, not on subsequent ticks.

3. **Replaying flag is propagated to subgraphs** (`_loop.py:736-741`): The outer graph sets `CONFIG_KEY_REPLAYING: self.is_replaying` on its config so child subgraphs inherit the replaying state.

4. **Cached RESUME writes are dropped during replay** (`_loop.py:1136-1146`, duplicated for async at `1336-1346`): When `is_replaying` is `True` and no new resume value is being provided, `RESUME` pending writes are filtered out. This forces `interrupt()` calls to re-fire instead of returning cached values. If a resume value *is* being provided (e.g., `Command(resume=...)`), the writes are kept so multi-interrupt scenarios still work.

5. **Pending writes are only matched when not replaying** (`_loop.py:454, 518`): `_match_writes` is gated on `not self.is_replaying` (previously `self.skip_done_tasks`), so cached writes aren't applied to tasks during replay.

6. **Subgraph checkpoint is reset when replaying from before the subgraph existed** (`_loop.py:1103-1109`, async at `1303-1309`): If replaying with `CONFIG_KEY_REPLAYING` but no `checkpoint_id` in the config (meaning the parent checkpoint predates this subgraph run), the saved checkpoint is discarded so the subgraph starts fresh.

7. **Resuming/replaying flags are cleaned up** (`_loop.py:573`): Both flags are popped from config after the first step completes.

8. **Refactored `_first()` method comments** (`_loop.py:625-660`): The resuming logic is restructured with clearer variable names (`has_prior_checkpoint`, `input_signals_resume`) and more detailed comments explaining when/why the graph considers itself to be resuming.

## Files Changed

| File | Change |
|------|--------|
| `libs/langgraph/langgraph/_internal/_constants.py` | Added `CONFIG_KEY_REPLAYING` constant and registered it in `RESERVED` |
| `libs/langgraph/langgraph/pregel/_loop.py` | Core replay logic: propagate replaying flag, drop stale RESUME writes, reset subgraph checkpoints, refactor resuming detection |
| `libs/langgraph/tests/test_time_travel.py` | ~1560 lines of new sync tests covering replay, fork, interrupts, subgraphs, and observability |
| `libs/langgraph/tests/test_time_travel_async.py` | ~1589 lines — async mirrors of the sync tests |

## Key Concepts

- **Replay** (`invoke(None, config_with_checkpoint_id)`): Re-executes nodes that come *after* the specified checkpoint. Interrupts re-fire. The graph re-runs the computation rather than returning cached results.
- **Fork** (`update_state(config_with_checkpoint_id, ...)` then `invoke`): Creates a new branch of state from a past checkpoint, then runs forward. Nodes re-execute and interrupts fire.
- **Resuming vs Replaying**: Replaying is a *subset* of resuming — when you replay, both `CONFIG_KEY_RESUMING` and `CONFIG_KEY_REPLAYING` are `True`. All replays are resumes, but not all resumes are replays. A plain resume (e.g., continuing after an interrupt with `Command(resume=...)`) has `RESUMING=True, REPLAYING=False` — cached `RESUME` writes are kept so the interrupt resolves. A replay has both flags `True` — cached `RESUME` writes are *dropped* so interrupts re-fire, and `is_replaying` is `True` so nodes re-execute instead of returning cached results.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -153,7 +153,7 @@ class PregelLoop:
     input_keys: str | Sequence[str]
     output_keys: str | Sequence[str]
     stream_keys: str | Sequence[str]
-    skip_done_tasks: bool
+    is_replaying: bool
     is_nested: bool
     manager: None | AsyncParentRunManager | ParentRunManager
     interrupt_after: All | Sequence[str]
@@ -245,9 +245,9 @@ class PregelLoop:
         self.interrupt_before = interrupt_before
         self.manager = manager
         self.is_nested = CONFIG_KEY_TASK_ID in self.config.get(CONF, {})
-        self.skip_done_tasks = CONFIG_KEY_CHECKPOINT_ID not in config[
+        self.is_replaying = CONFIG_KEY_CHECKPOINT_ID in config[
             CONF
-        ] and not config[CONF].get(CONFIG_KEY_REPLAYING, False)
+        ] or config[CONF].get(CONFIG_KEY_REPLAYING, False)
         self._migrate_checkpoint = migrate_checkpoint
         self.trigger_to_nodes = trigger_to_nodes
         self.retry_policy = retry_policy
@@ -454,7 +454,7 @@ class PregelLoop:
             # save the new task
             self.tasks[pushed.id] = pushed
             # match any pending writes to the new task
-            if self.skip_done_tasks:
+            if not self.is_replaying:
                 self._match_writes({pushed.id: pushed})
             # return the new task, to be started if not run before
             return pushed
@@ -518,7 +518,7 @@ class PregelLoop:
             return False
 
         # if there are pending writes from a previous loop, apply them
-        if self.skip_done_tasks and self.checkpoint_pending_writes:
+        if not self.is_replaying and self.checkpoint_pending_writes:
             self._match_writes(self.tasks)
 
         # before execution, check if we should interrupt
@@ -560,8 +560,8 @@ class PregelLoop:
             )
         # clear pending writes
         self.checkpoint_pending_writes.clear()
-        # "not skip_done_tasks" only applies to first tick after resuming
-        self.skip_done_tasks = True
+        # only replay (re-execute) done tasks on the first tick
+        self.is_replaying = False
         # save checkpoint
         self._put_checkpoint({"source": "loop"})
         # after execution, check if we should interrupt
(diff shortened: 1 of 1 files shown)


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -618,22 +622,28 @@ def _pending_interrupts(self) -> set[str]:
...
+            or isinstance(self.input, Command)
+            or (
+                not self.is_nested
+                and self.config.get("metadata", {}).get("run_id")
+                == self.checkpoint_metadata.get("run_id", MISSING)
             )
         )
+        is_resuming = has_prior_checkpoint and bool(
+            configurable.get(CONFIG_KEY_RESUMING, input_signals_resume)
+        )
```
sydney-runkle, line 646: no logical changes here, just comments + var name changes

PR #7033, opened 2026-03-05
poc: fixing interrupt caching on replays (for parents and subgraphs)!

sydney-runkle:
# Branch `sr/alt-fix` — Time Travel (Replay & Fork) Fixes

## Problem

When "time traveling" (replaying from or forking at a previous checkpoint), LangGraph had several issues:

1. **Tasks were incorrectly skipped during replay.** The old `skip_done_tasks` flag didn't properly signal replay mode, so nodes that should re-execute were being skipped because they had cached results.
2. **Interrupts didn't re-fire on replay.** Cached `RESUME` pending writes from a previous run caused `interrupt()` calls to return stale values instead of actually interrupting again.
3. **Subgraphs didn't replay correctly.** There was no mechanism to propagate "we are replaying" to subgraphs. A subgraph that wasn't part of the original checkpoint map could load stale checkpoint data from a different execution.

### `_constants.py`

- Added `CONFIG_KEY_REPLAYING` (`"__pregel_replaying"`) — a new config flag propagated from parent graphs to subgraphs, analogous to the existing `CONFIG_KEY_RESUMING`.

#### Renamed `skip_done_tasks` to `is_replaying`

Replaced `skip_done_tasks: bool` with `is_replaying: bool` (inverted polarity). `is_replaying` is `True` when `CONFIG_KEY_CHECKPOINT_ID` is in the config (user specified a checkpoint) OR `CONFIG_KEY_REPLAYING` is set (parent is replaying). After the first tick, `is_replaying` is set to `False` — replay only applies to the first tick.

#### Propagate `CONFIG_KEY_REPLAYING` to subgraphs

In `_first()`, the replaying flag is now propagated alongside `CONFIG_KEY_RESUMING`. After the first tick, both flags are cleared from config.

#### Drop stale RESUME writes on replay (consolidated into `_first()`)

Previously duplicated in both `__enter__` (sync) and `__aenter__` (async), this logic now lives in `_first()` right after `is_resuming` is computed — consolidating all resume/replay reasoning in one place.

Two distinct concepts control this behavior:

- **`is_resuming`** — "should we continue from where we left off?" (broad: includes `None` input, `Command` input, same `run_id`, or parent `CONFIG_KEY_RESUMING` flag). Controls whether to advance past interrupts.
- **`is_resume_with_value`** — "is the user providing new resume data?" (narrow: `Command(resume=...)` or `CONFIG_KEY_RESUMING` flag). Controls whether to keep cached RESUME writes.

When replaying and `is_resume_with_value` is `False`, cached RESUME writes are dropped so `interrupt()` calls re-fire. When the user IS providing resume data, writes are preserved for multi-interrupt scenarios.

#### Fresh start for unknown subgraphs during replay

In both `__enter__` and `__aenter__`: if `CONFIG_KEY_REPLAYING` is set but no `CONFIG_KEY_CHECKPOINT_ID` exists for this subgraph, the loaded checkpoint is discarded. This prevents a replaying subgraph from picking up stale state that belongs to a different execution.

#### Improved comments in `_first()`

Resume detection logic is broken into named variables (`has_prior_checkpoint`, `input_signals_resume`, `is_resuming`) with comments explaining each condition.

## Files Changed

| File | Change |
|------|--------|
| `langgraph/_internal/_constants.py` | Added `CONFIG_KEY_REPLAYING` constant and registered it in `RESERVED` |
| `langgraph/pregel/_loop.py` | Core replay logic (~90 lines changed) |
| `tests/test_time_travel.py` | 26 sync tests (~1560 lines) |
| `tests/test_time_travel_async.py` | 26 async tests (~1589 lines) |

## Key Concepts

- **Replay** (`invoke(None, config_with_checkpoint_id)`): Re-executes nodes after the specified checkpoint. Interrupts re-fire. The graph re-runs computation rather than returning cached results.
- **Fork** (`update_state(config_with_checkpoint_id, ...)` then `invoke`): Creates a new branch of state from a past checkpoint, then runs forward.
- **Resuming vs Replaying**: All replays are resumes, but not all resumes are replays. A plain resume (continuing after interrupt with `Command(resume=...)`) keeps cached RESUME writes so the interrupt resolves. A replay drops them so interrupts re-fire, and `is_replaying=True` forces nodes to re-execute.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -245,9 +245,9 @@ class PregelLoop:
         self.interrupt_before = interrupt_before
         self.manager = manager
         self.is_nested = CONFIG_KEY_TASK_ID in self.config.get(CONF, {})
-        self.is_replaying = CONFIG_KEY_CHECKPOINT_ID in config[
+        self.is_replaying = CONFIG_KEY_CHECKPOINT_ID in config[CONF] or config[
             CONF
-        ] or config[CONF].get(CONFIG_KEY_REPLAYING, False)
+        ].get(CONFIG_KEY_REPLAYING, False)
         self._migrate_checkpoint = migrate_checkpoint
         self.trigger_to_nodes = trigger_to_nodes
         self.retry_policy = retry_policy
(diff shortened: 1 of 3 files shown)


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -152,7 +153,7 @@ class PregelLoop:
     input_keys: str | Sequence[str]
     output_keys: str | Sequence[str]
     stream_keys: str | Sequence[str]
-    skip_done_tasks: bool
+    is_replaying: bool
```
sydney-runkle, line 156: this is more intuitive inline, I think

sydney-runkle: but could go either way

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -618,22 +622,42 @@ def _pending_interrupts(self) -> set[str]:
...
+            or isinstance(self.input, Command)
+            or (
+                not self.is_nested
+                and self.config.get("metadata", {}).get("run_id")
+                == self.checkpoint_metadata.get("run_id", MISSING)
             )
         )
+        is_resuming = has_prior_checkpoint and bool(
+            configurable.get(CONFIG_KEY_RESUMING, input_signals_resume)
+        )
```
sydney-runkle, line 646: no logical change here, just clarifying w/ comments and var names

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -1085,6 +1113,14 @@ def __enter__(self) -> Self:
...
         else:
             saved = None
+        # When replaying a subgraph that wasn't in the checkpoint map
+        # (parent checkpoint predates this subgraph), start fresh.
+        if (
+            saved is not None
+            and self.config[CONF].get(CONFIG_KEY_REPLAYING)
+            and not self.checkpoint_config.get(CONF, {}).get(CONFIG_KEY_CHECKPOINT_ID)
+        ):
+            saved = None
```
sydney-runkle: this will be fixed in a fast follow PR

sydney-runkle: because for checkpointer=True, we need to go find the "latest" checkpoint before the parent checkpoint

PR #7038, opened 2026-03-06
fix: replay behavior for parent + subgraphs!

sydney-runkle:
## Summary

Fix time travel (replay and fork) for graphs with interrupts and subgraphs.

## Problem

Two issues with replaying/forking from earlier checkpoints:

1. **Stale interrupt values during replay** — Replays incorrectly reused cached `RESUME` values from prior `interrupt()` calls, so interrupts silently returned stale answers instead of re-firing.

2. **Wrong subgraph state during time travel** — Subgraphs always loaded their **latest** checkpoint instead of the one corresponding to the parent's historical state. This caused subgraphs to skip execution or produce incorrect results during replay/fork.

## Changes

Code changes span `libs/langgraph/langgraph/pregel/_loop.py`, `libs/langgraph/langgraph/_internal/_constants.py`, and a new `libs/langgraph/langgraph/_internal/_replay.py` module:

- **Strip stale `RESUME` writes on replay** — During replays, cached `RESUME` writes are filtered out so `interrupt()` re-fires instead of returning old values. Genuine resumes (`Command(resume=...)`) preserve these writes.

- **Rename `skip_done_tasks` → `is_replaying`** — Clearer naming for the flag that tracks whether the current run is replaying from a specific checkpoint.

- **New `ReplayState` class (`_replay.py`)** — Encapsulates subgraph checkpoint loading during time-travel. Tracks a parent checkpoint ID upper bound and which subgraph namespaces have already loaded their pre-replay checkpoint. On the first visit to a subgraph namespace, it loads the latest checkpoint created *before* the replay point (via `checkpointer.list(..., before=...)` with `limit=1`). On subsequent visits (e.g. the same subgraph in a later loop iteration), it falls back to normal latest-checkpoint loading. The task-id suffix is stripped from namespaces so the same logical subgraph is recognized across loop iterations.

- **New `CONFIG_KEY_REPLAY_STATE` config key** — The parent graph creates a `ReplayState` instance and passes it to subgraphs via config. For forks (`source=update`), the replay state uses the fork's parent checkpoint ID since the fork was created after the subgraph's original checkpoints. The single `ReplayState` instance is shared by reference across all derived configs within one parent execution.

- **Subgraph checkpoint loading in `__enter__`/`__aenter__`** — When a subgraph detects a `ReplayState` in its config, it delegates checkpoint loading to `ReplayState.get_checkpoint`/`aget_checkpoint` instead of using the default `get_tuple`. It also clears `CONFIG_KEY_RESUMING` so `_first` re-applies input and recreates ephemeral routing channels.

## Tests

New test files `test_time_travel.py` (~2500 lines) and `test_time_travel_async.py` (~2200 lines) covering:
- Replay and fork with interrupts (single and multiple)
- Replay and fork for graphs with and without subgraphs
- Correct subgraph checkpoint restoration during parent time travel
- `get_state` with subgraph state during replay

### DIFF
--- a/libs/langgraph/langgraph/_internal/_replay.py
+++ b/libs/langgraph/langgraph/_internal/_replay.py
@@ -27,6 +27,8 @@ class ReplayState:
 
     def __init__(self, checkpoint_id: str) -> None:
         self.checkpoint_id = checkpoint_id
+        # DO NOT CHANGE THIS VARIABLE – it may need to be rehydrated
+        # in other runtimes
         self._visited_ns: set[str] = set()
 
     def _is_first_visit(self, checkpoint_ns: str) -> bool:
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -629,11 +629,12 @@ class PregelLoop:
         #   - Command input: any Command operates on existing state
         #   - Same run_id: re-entry into an ongoing run (e.g. stream reconnect)
         configurable = self.config.get(CONF, {})
+        input_is_command = isinstance(self.input, Command)
         is_resuming = bool(self.checkpoint["channel_versions"]) and bool(
             configurable.get(
                 CONFIG_KEY_RESUMING,
                 self.input is None
-                or isinstance(self.input, Command)
+                or input_is_command
                 or (
                     not self.is_nested
                     and self.config.get("metadata", {}).get("run_id")
@@ -651,7 +652,7 @@ class PregelLoop:
         # - CONFIG_KEY_RESUMING: child subgraphs receive it via config from
         #   the parent (their input is a Send arg, not a Command)
         if self.is_replaying and not (
-            (isinstance(self.input, Command) and self.input.resume is not None)
+            (input_is_command and cast(Command, self.input).resume is not None)
             or configurable.get(CONFIG_KEY_RESUMING, False)
         ):
             self.checkpoint_pending_writes = [
@@ -659,8 +660,8 @@ class PregelLoop:
             ]
 
         # map command to writes
-        if isinstance(self.input, Command):
-            if (resume := self.input.resume) is not None:
+        if input_is_command:
+            if (resume := cast(Command, self.input).resume) is not None:
                 if not self.checkpointer:
                     raise RuntimeError(
                         "Cannot use Command(resume=...) without checkpointer"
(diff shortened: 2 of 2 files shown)


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -618,22 +622,42 @@ def _pending_interrupts(self) -> set[str]:
...
+            or isinstance(self.input, Command)
+            or (
+                not self.is_nested
+                and self.config.get("metadata", {}).get("run_id")
+                == self.checkpoint_metadata.get("run_id", MISSING)
             )
         )
+        is_resuming = has_prior_checkpoint and bool(
+            configurable.get(CONFIG_KEY_RESUMING, input_signals_resume)
+        )
```
sydney-runkle: no logical change, just comments

sydney-runkle:
Closing, this is solving such a niche use case that I haven't seen anyone actually need to use yet.

libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -1036,8 +1039,10 @@ def _triggers(
     seen: ChannelVersions | None,
     null_version: V,
     proc: PregelNode,
+    *,
+    is_replaying: bool = False,
 ) -> bool:
-    if seen is None:
+    if is_replaying or seen is None:
```
sydney-runkle: this hack was needed to "trigger" the subgraph run, bc the channel versions have already been seen

libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -579,7 +579,7 @@ def prepare_single_task(
         proc = processes[name]
         if checkpoint_null_version is None:
             return
-        # If any of the channels read by this process were updated
+        # If any of the channels read by this process were updated.
```
sydney-runkle:
```suggestion
        # If any of the channels read by this process were updated
```

libs/langgraph/langgraph/pregel/_loop.py
hinthornw: Could we cache `input_is_command` as a local variable here so we don't have to check is_instance twice?

libs/langgraph/langgraph/_internal/_replay.py
```diff
@@ -0,0 +1,88 @@
...
+
+    The single `ReplayState` instance is shared by reference across all derived
+    configs within one parent execution.
+    """
+
+    __slots__ = ("checkpoint_id", "_visited_ns")
+
+    def __init__(self, checkpoint_id: str) -> None:
+        self.checkpoint_id = checkpoint_id
+        self._visited_ns: set[str] = set()
```
hinthornw, line 32: Could we add a comment like "DO NOT CHANGE THIS VARIABLE"? we may need to be hydrating this in dr

PR #7045, opened 2026-03-07
Worktree explore subgraph replay

sydney-runkle:
attempting diff w/o triggers workaround


libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -579,7 +579,7 @@ def prepare_single_task(
         proc = processes[name]
         if checkpoint_null_version is None:
             return
-        # If any of the channels read by this process were updated
+        # If any of the channels read by this process were updated.
```
sydney-runkle:
```suggestion
        # If any of the channels read by this process were updated
```

PR #7054, opened 2026-03-07
fix: hardcode version to avoid importlib.metadata overhead

haroldfabla2-hue:
## Summary

Removes use of `importlib.metadata` for version access as requested in issue #5040.

## Changes

- Replaces dynamic version lookup with hardcoded version `1.0.10`
- Improves import performance by avoiding metadata lookup overhead

## Testing

```python
from langgraph.version import __version__
assert __version__ == "1.0.10"
```

Closes #5040


hinthornw:
Needs to update teh packaging

PR #7068, opened 2026-03-08
fix: Add support for Python 3.14

JiwaniZakir:
## Summary

This addresses #5253.

## What changed

Implemented a fix based on the issue description. See the diff for specifics.

## Testing

- Verified against the existing test suite
- Checked that the fix addresses the reported behavior


JiwaniZakir:
Closing this -- the commit history has some config issues I can not fix without a force push. Will resubmit cleanly if the issue is still open.

PR #7079, opened 2026-03-09
fix(langgraph): get_subgraphs fails when nodes share a common prefix

### LINKED ISSUE #6924: get_subgraph fails when nodes share a common prefix
### Checked other resources

```python

    class State(TypedDict):
        do_not_care: str

    def generic_node(state: State) -> State:
        # Just need a properly typed node; we will not be running this graph
        return state

    grandchild_graph_builder = StateGraph(State)
    grandchild_graph_builder.add_node("grandchild", generic_node)
    grandchild_graph_builder.set_entry_point("grandchild")
    grandchild_graph_builder.add_edge("grandchild", END)
    grandchild_graph = grandchild_graph_builder.compile()

    child_graph_builder = StateGraph(State)
    child_graph_builder.add_node("child_subgraph", grandchild_graph)
    child_graph_builder.set_entry_point("child_subgraph")
    child_graph_builder.add_edge("child_subgraph", END)
    child_graph = child_graph_builder.compile()

    parent_graph_builder = StateGraph(State)
    parent_graph_builder.add_node("common_prefix", child_graph)
    parent_graph_builder.add_node("common_prefix_2", child_graph)
    parent_graph_builder.set_entry_point("common_prefix")
    parent_graph_builder.add_edge("common_prefix", "common_prefix_2")
    parent_graph_builder.add_edge("common_prefix_2", END)

    parent_graph = parent_graph_builder.compile()

    # It's pretty intricate to get this to fail against the previous
    # implementation of get_subgraphs. You need the following conditions:
    # * A parent graph with 2 subgraphs
    # * Subgraph A added before subgraph B
    # * Subgraph A's name is a prefix of subgraph B's name
    # * Subgraph B must also contain a subgraph (call it subgraph C)
    # * You must call get_subgraphs with the namespace 'subgraph_b|subgraph_c'
    #
    # The old implementation used prefix matching to find subgraphs,
    # so it would chop of subgraph A's name from subgraph B's name and
    # look for subgraph C under the chopped-off name, which would fail.
    subgraphs = parent_graph.get_subgraphs(
        namespace="common_prefix_2|child_subgraph",
        recurse=True,
    )
    assert len(list(subgraphs)) == 1
```

```shell
(assert at the end of example code fails)
```

### Description

`Pregel.get_subgraphs()` had a subtle bug due to the fact that it matches node names with namespaces using startswith. This created the possibility (which I did in fact hit in production) that 2 nodes whose names share a common prefix, and both contain subgraphs, could cause the algorithm to malfunction. It would find the first node (whose name was a prefix of the second), chop off that many characters leaving a partial namespace, and recurse on that.

### System Info

This is a logic bug not affected by system info. I put up a PR to fix it: #6366

giulio-leone:
## Summary

Fixes #6924

`Pregel.get_subgraphs()` used `startswith` to match node names against namespaces, which caused false matches when two node names share a common prefix (e.g. `common_prefix` and `common_prefix_2`). The algorithm would match the shorter node name first, strip too few characters from the namespace, and recurse with a broken remaining namespace.

## Fix

Replace `startswith`-based prefix matching with an exact match on the first segment of the namespace (split by `NS_SEP`). This ensures `common_prefix` only matches itself, never `common_prefix_2`.

**Before:**
```python
if not namespace.startswith(name):
    continue
# ...
namespace = namespace[len(name) + 1:]
```

**After:**
```python
ns_parts = namespace.split(NS_SEP, 1)
if ns_parts[0] != name:
    continue
# ...
remaining_ns = ns_parts[1] if len(ns_parts) > 1 else None
```

## Tests

Added 5 tests in `test_get_subgraphs.py` covering the exact bug scenario from the issue plus regressions for basic enumeration, exact namespace match, and recursive traversal.

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -959,9 +959,10 @@ class Pregel(
             An iterator of the `(namespace, subgraph)` pairs.
         """
         for name, node in self.nodes.items():
-            # filter by prefix
+            # filter by exact match on the first namespace segment
             if namespace is not None:
-                if not namespace.startswith(name):
+                ns_parts = namespace.split(NS_SEP, 1)
+                if ns_parts[0] != name:
                     continue
 
             # find the subgraph, if any
@@ -975,12 +976,14 @@ class Pregel(
                 if namespace is None:
                     yield name, graph
                 if recurse and isinstance(graph, Pregel):
+                    remaining_ns: str | None = None
                     if namespace is not None:
-                        namespace = namespace[len(name) + 1 :]
+                        ns_parts = namespace.split(NS_SEP, 1)
+                        remaining_ns = ns_parts[1] if len(ns_parts) > 1 else None
                     yield from (
                         (f"{name}{NS_SEP}{n}", s)
                         for n, s in graph.get_subgraphs(
-                            namespace=namespace, recurse=recurse
+                            namespace=remaining_ns, recurse=recurse
                         )
                     )
 
--- a/libs/langgraph/tests/test_get_subgraphs.py
+++ b/libs/langgraph/tests/test_get_subgraphs.py
@@ -1,0 +1,87 @@
+"""Tests for Pregel.get_subgraphs with common-prefix node names.
+
+Regression test for https://github.com/langchain-ai/langgraph/issues/6924
+"""
+
+from typing import TypedDict
+
+from langgraph.graph import END, StateGraph
+
+
+class State(TypedDict):
+    value: str
+
+
+def _node(state: State) -> State:
+    return state
+
+
+def _make_parent_graph():
+    """Build a parent graph with two subgraph nodes sharing a common prefix."""
+    grandchild_builder = StateGraph(State)
+    grandchild_builder.add_node("grandchild", _node)
+    grandchild_builder.set_entry_point("grandchild")
+    grandchild_builder.add_edge("grandchild", END)
+    grandchild_graph = grandchild_builder.compile()
+
+    child_builder = StateGraph(State)
+    child_builder.add_node("child_subgraph", grandchild_graph)
+    child_builder.set_entry_point("child_subgraph")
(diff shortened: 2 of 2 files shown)


giulio-leone:
Friendly ping — CI is green, tests pass, rebased on latest. Ready for review whenever convenient. Happy to address any feedback. 🙏

giulio-leone:
## ✅ Verified with real nested subgraph resolution

**Environment:** Python 3.13.12, macOS, `langgraph` from this branch

```python

# Build 3-level nested graph
inner = StateGraph(State)
inner.add_node("inner_node", lambda s: {"value": "inner"})
# ... edges ...

middle = StateGraph(State)
middle.add_node("inner_graph", inner.compile())
# ... edges ...

outer = StateGraph(State)
outer.add_node("middle_graph", middle.compile())
# ... edges ...

app = outer.compile()
print(dict(app.get_subgraphs()).keys())
print(dict(app.get_subgraphs(recurse=True)).keys())
```

```
Subgraphs:           ["middle_graph"]                          ✅
Recursive subgraphs: ["middle_graph", "middle_graph|inner_graph"]  ✅
Namespace lookup:    "middle_graph" -> CompiledStateGraph      ✅
```

The `namespace` prefix matching in `get_subgraphs()` now correctly handles common prefixes without false matches.

giulio-leone:
@hinthornw Fixes `get_subgraphs(namespace=...)` false-matching on common prefixes — e.g., querying for namespace `"agent"` would incorrectly match `"agent_tools"`. The fix adds a separator check (`|`) to ensure exact namespace boundary matching. Verified with 3-level nested subgraphs.

giulio-leone:
Friendly ping — this branch has been refreshed on the latest upstream base and all current review feedback has been addressed. It should be ready for review whenever you have a chance. Happy to make any further changes quickly.

giulio-leone:
Friendly ping — rebased on latest and ready for review. Happy to address any feedback!

PR #7087, opened 2026-03-09
fix(langgraph): use identity check for Command.update in map_command

giulio-leone:
## Problem

`map_command()` in `_io.py` uses a truthy check (`if cmd.update:`) instead of an identity check (`if cmd.update is not None:`) for the `update` field. This silently drops falsy but valid state updates:

```python
Command(update=0)      # update silently dropped
Command(update=False)  # update silently dropped
Command(update='')     # update silently dropped
```

The adjacent `resume` field already uses the correct pattern:
```python
if cmd.resume is not None:     # ← correct
if cmd.update:                  # ← bug: truthy check
```

And `_update_as_tuples()` also correctly uses `self.update is not None`.

## Fix

Changed `if cmd.update:` to `if cmd.update is not None:` to align with the rest of the codebase.

## Tests

Added `test_map_command.py` with 10 tests verifying that:
- Falsy but valid updates (`0`, `False`, `""`) are correctly propagated through `map_command`
- `None` sentinel correctly produces no writes
- Standard cases (dict, tuples, combined goto+update, resume+update) continue to work

### DIFF
--- a/libs/langgraph/langgraph/pregel/_io.py
+++ b/libs/langgraph/langgraph/pregel/_io.py
@@ -73,7 +73,7 @@ def map_command(cmd: Command) -> Iterator[tuple[str, str, Any]]:
                 )
     if cmd.resume is not None:
         yield (NULL_TASK_ID, RESUME, cmd.resume)
-    if cmd.update:
+    if cmd.update is not None:
         for k, v in cmd._update_as_tuples():
             yield (NULL_TASK_ID, k, v)
 
--- a/libs/langgraph/tests/test_map_command.py
+++ b/libs/langgraph/tests/test_map_command.py
@@ -1,0 +1,85 @@
+"""Tests for map_command in langgraph.pregel._io."""
+
+from langgraph._internal._constants import NULL_TASK_ID, RESUME
+from langgraph.constants import START
+from langgraph.pregel._io import map_command
+from langgraph.types import Command
+
+
+def test_map_command_update_with_dict() -> None:
+    """Standard dict update should produce writes."""
+    cmd = Command(update={"key": "value"})
+    results = list(map_command(cmd))
+    assert (NULL_TASK_ID, "key", "value") in results
+
+
+def test_map_command_update_with_zero() -> None:
+    """update=0 is a valid falsy value and must not be silently dropped."""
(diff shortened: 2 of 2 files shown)


giulio-leone:
## ✅ Verified with real graph execution

**Environment:** Python 3.13.12, macOS, `langgraph` from this branch

### Bug proof — old behavior silently drops falsy updates:

```python

def old_check(cmd):  return bool(cmd.update)      # old: truthy
def new_check(cmd):  return cmd.update is not None  # new: identity

for val in [0, False, "", {"key": "val"}]:
    cmd = Command(update=val)
    print(f"Command(update={val!r}): old={old_check(cmd)}, new={new_check(cmd)}")
```

```
Command(update=0):       old=False, new=True  🐛 BUG: update silently dropped
Command(update=False):   old=False, new=True  🐛 BUG: update silently dropped
Command(update=""):      old=False, new=True  🐛 BUG: update silently dropped
Command(update={"key":…}): old=True,  new=True  ✅
```

### Fix verified — real graph with `Command(update={"counter": 0})`:

```python

def set_zero(state):
    return Command(update={"counter": 0, "messages": ["reset"]})

# ... graph setup ...
result = app.invoke({"counter": 100, "messages": ["start"]})
```

```
Final counter: 0  ✅  (would be 101 without fix — update dropped, only increment applied)
Messages: ["start", "incremented", "reset"]  ✅
```

The fix aligns `cmd.update` with the existing `cmd.resume is not None` pattern on the adjacent line.

giulio-leone:
@hinthornw Found this during a source audit — `map_command()` uses `if cmd.update:` (truthy check) while the adjacent `cmd.resume` correctly uses `is not None`. This silently drops valid falsy updates like `Command(update=0)` or `Command(update=False)`. One-line fix + 10 unit tests. Bug proof and real graph execution verification above.

giulio-leone:
Friendly ping — this branch has been refreshed on the latest upstream base and all current review feedback has been addressed. It should be ready for review whenever you have a chance. Happy to make any further changes quickly.

giulio-leone:
Friendly ping — rebased on latest and ready for review. Happy to address any feedback!

PR #7100, opened 2026-03-10
feat(cli): add langgraph deploy logs subcommand

vishnu-ssuresh:
Add a `langgraph deploy logs` subcommand to fetch build and deploy/server logs from LangSmith deployments.

## Changes

- **`host_backend.py`**: Add `get_build_logs()` and `get_deploy_logs()` methods, plus `langsmith_url` support.
- **`helpers.py`** (new): Log formatting (`format_log_entry`, `format_timestamp`, `level_fg`) and resolution helpers (`resolve_deployment_id`).

## Usage

```bash
# Deploy logs (latest 10)
langgraph deploy logs --deployment-id <id> --limit 10

# Error logs by name
langgraph deploy logs --name gtm-agent --level ERROR

# Build logs
langgraph deploy logs --deployment-id <id> --type build

# Tail logs (Ctrl+C to stop)
langgraph deploy logs --name gtm-agent --follow
```

## Tests

56 unit tests passing covering host backend methods, log formatting, and helper functions.

### DIFF
--- a/libs/cli/langgraph_cli/cli.py
+++ b/libs/cli/langgraph_cli/cli.py
@@ -12,8 +12,8 @@ import sys
 import tempfile
 import time
 from collections.abc import Callable, Sequence
-from datetime import datetime, timezone
 from contextlib import contextmanager
+from datetime import datetime, timezone
 
 import click
 import click.exceptions


libs/cli/langgraph_cli/host_backend.py
```diff
@@ -30,7 +35,6 @@ def __init__(self, base_url: str, api_key: str, tenant_id: str | None = None):
         if tenant_id:
             headers["X-Tenant-ID"] = tenant_id
         self._base_url = base_url.rstrip("/")
-        self._api_key = api_key
```
andrewnguonly, line 33: Why is this removed?

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+    "--deployment-id",
+    help="Deployment ID. If omitted, --name is used to find the deployment.",
+)
+@click.option(
+    "--type",
+    "log_type",
+    type=click.Choice(["deploy", "build"]),
+    default="deploy",
+    show_default=True,
+    help="Type of logs to fetch.",
```
andrewnguonly: We could be more specific here. `deploy` is for the Agent Server logs. `build` is for the build logs, which isn't really relevant for deployments created via `langgraph deploy` (i.e. build happens locally).

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+)
+@click.option(
+    "--revision-id",
```
andrewnguonly, line 1320: Follow up: We need support for `langgraph deploy list-revisions` or `langgraph deploy list --revisions` to get the revision IDs (or something like this).

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+    help="Specific revision ID. For build logs, defaults to latest revision.",
+)
+@click.option(
+    "--level",
+    type=click.Choice(["DEBUG", "INFO", "WARNING", "ERROR"], case_sensitive=False),
```
andrewnguonly: We need to support `CRITICAL`, these log levels: https://docs.python.org/3/library/logging.html#logging-levels

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+    "-q",
+    help="Search string filter.",
+)
+@click.option(
+    "--start-time",
+    help="ISO8601 start time (e.g. 2026-03-08T00:00:00Z).",
+)
+@click.option(
+    "--end-time",
+    help="ISO8601 end time.",
```
andrewnguonly: Suggestion: I would add `(e.g. 2026-03-08T00:00:00Z)` to the description.

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+    "--follow",
+    "-f",
+    is_flag=True,
+    default=False,
+    help="Continuously poll for new logs.",
+)
+@OPT_HOST_URL
+@deploy.command(
+    "logs",
+    help="[Beta] Fetch build or deploy logs for a LangSmith deployment.",
```
andrewnguonly: I've been capitalizing `LangSmith Deployment` in the other help text.

`Fetch build or deploy logs` <-- Could be more specific. Same comment as above.

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+
+    def _fetch(request_payload: dict) -> list[dict]:
+        if log_type == "build":
+            resp = client.get_build_logs(dep_id, revision_id, request_payload)
+        else:
+            resp = client.get_deploy_logs(dep_id, request_payload, revision_id)
+
+        if isinstance(resp, dict):
+            return resp.get("logs", resp.get("entries", []))
+        elif isinstance(resp, list):
```
andrewnguonly: Clarifying: The API should never return a `list`, right? Safe to remove this case?

libs/cli/langgraph_cli/host_backend.py
```diff
@@ -116,3 +120,24 @@ def get_revision(self, deployment_id: str, revision_id: str) -> dict[str, Any]:
             "GET",
             f"/v2/deployments/{deployment_id}/revisions/{revision_id}",
         )
+
+    def get_build_logs(
+        self, project_id: str, revision_id: str, payload: dict[str, Any]
+    ) -> Any:
+        return self._request(
+            "POST",
+            f"/v1/projects/{project_id}/revisions/{revision_id}/build_logs",
```
andrewnguonly, line 129: `v1` APIs... We may want to migrate this to `v2`. Not blocking.

libs/cli/langgraph_cli/cli.py
```diff
@@ -1294,6 +1298,169 @@ def _normalize_image_tag(value: str) -> str:
...
+    entries = _fetch_and_print(payload, reverse=True)
+
+    if not follow:
+        if not entries:
+            click.secho("No log entries found.", fg="yellow")
+        return
+
+    payload["order"] = "asc"
+    seen_ids: set[str] = {e.get("id", "") for e in entries if e.get("id")}
+    from datetime import datetime, timezone
```
andrewnguonly: Local import. Can we move to top?

hari-dhanushkodi:
> How come we're not using the `offset`/`next_offset` pagination for tailing logs?

as discussed offline, I think this is the best way to implement live tail:
```
1. first we fetch last limit logs in descending (so we get newest -> oldest in an array)
2. we output the reverse of the output of step 1 (so that old logs appear first, and new logs come to the front)
3. then we do an ascending log fetch with the most recent time stamp we got in step 1, so we don't fetch anything older than this, and we output to the screen and repeat
```

libs/cli/langgraph_cli/host_backend.py
hari-dhanushkodi: its not used in any of the helper functions, we just use it to set the header

Issue #7008, opened 2026-03-03
Type-Safe Streaming and Invoke for LangGraph


sydney-runkle:
# Type-Safe Streaming and Invoke for LangGraph

## The Problem

LangGraph's `.stream()`, `.astream()`, `.invoke()`, and `.ainvoke()` methods have **no meaningful type information** today. Regardless of which `stream_mode` you use, the return type is `Iterator[Any]` — you get no autocomplete, no type checking, and no way for your editor to tell you what shape the data is in.

This is especially painful when:

- **Using Pydantic or dataclass state schemas** — you define a typed schema but get plain `dict` objects back from streaming and invoke.
- **Switching between stream modes** — each mode returns a completely different shape (`dict`, `tuple[AnyMessage, dict]`, `Any`, etc.), but the type system can't help you remember which is which.
- **Handling interrupts** — the `__interrupt__` key is mixed into the state dictionary, which pollutes your state, breaks Pydantic/dataclass coercion, and forces manual filtering.

### Why not just use `@overload`?

Python's only mechanism for "parameter X controls return type Y" is `@overload`. The matrix is 8 stream modes x 2 subgraph values = **16+ overloads per method**, across `stream`/`astream` on `Pregel`, `PregelProtocol`, `RemoteGraph`, and the SDK. A proof-of-concept required hundreds of signatures and was still incomplete. Every new stream mode multiplies the combinatorics — this approach is unsustainable.

## Proposal: `stream_version="v2"` with typed stream parts

We're adding a `stream_version: Literal["v1", "v2"]` parameter (defaulting to `"v1"`) to `.stream()`, `.astream()`, `.invoke()`, and `.ainvoke()`. When `stream_version="v2"`, every chunk is a self-describing `StreamPart` TypedDict with a `type` discriminator:

```python
from langgraph.types import StreamPart, ValuesStreamPart, UpdatesStreamPart

async for part in graph.astream(inputs, stream_version="v2"):
    if part["type"] == "values":
        # part is narrowed to ValuesStreamPart
        # part["data"] is your output type (Pydantic model, dataclass, or dict)
        # part["interrupts"] is tuple[Interrupt, ...]
        print(part["data"])
    elif part["type"] == "updates":
        # part is narrowed to UpdatesStreamPart
        # part["data"] is dict[str, Any] mapping node names to outputs
        print(part["data"])
    elif part["type"] == "messages":
        # part is narrowed to MessagesStreamPart
[...]

PR #7101, opened 2026-03-10
fix(sdk-py): add stream_resumable to thread_id:str stream() overloads

KartikPawade:
**Description:** The `thread_id: str` overloads for `SyncRunsClient.stream()` 
were missing `stream_resumable: bool = False`, while the `thread_id: None` 
overloads and the async client already had it. Type checkers (mypy/pyright) 
would reject `stream_resumable` when calling `stream()` with a thread ID, 
even though it works fine at runtime. Fixed by adding the missing parameter 
to both `thread_id: str` overloads (v1 and v2) in `_sync/runs.py`.

All existing tests pass locally (`test_client_stream.py` 14/14, `test_api_parity.py` 5/5).

**Issue:** Fixes gap noted in #7008

CC: @sydney-runkle 

**Dependencies:** None


KartikPawade:
Hey @sydney-runkle, tagging you as mentioned in #7008!

Also wanted to flag — I looked deeper into the `ns` tuple vs list gap I 
mentioned in #7008. After checking `langgraph/types.py` and 
`test_stream_v2.py`, it seems like `tuple[str, ...]` in core and 
`list[str]` in the SDK is intentional rather than a bug. Is that right? 
Happy to close that one out on my end if so.

PR #7139, opened 2026-03-12
feat(prebuilt): surface model output metadata in tool error messages

### LINKED ISSUE #7138: ToolNode: surface model output metadata in tool error messages to enable self-correction
## Problem

When a tool call fails, the `ToolNode` creates a `ToolMessage(status="error")` containing only the exception text. The model receives this error but has **no access to its own output metadata** — information like why it stopped generating, how many tokens it used, or whether its output was complete.

This metadata *is* available on the originating `AIMessage.response_metadata` (every provider populates `stop_reason`/`finish_reason`, token counts, etc.), but it's never included in the error the model sees. Without this context, the model cannot diagnose root causes and may retry the same failing approach indefinitely.

### What we hit

In an autonomous agent workflow, a model called `write_file` with large content. The output hit `max_tokens=16384`, truncating the tool call JSON mid-argument. Pydantic raised `content: Field required` — a confusing error that gave no hint about truncation. The model retried 249 times, never learning *why* its arguments were malformed.

This affects any user whose model hits output limits while generating tool calls — large file writes, complex structured outputs, multi-tool parallel calls, etc.

## Reproduction

1. Use any model with a `max_tokens` setting (e.g., 16384)
2. Have the model call a tool with arguments large enough to exceed that limit
3. The tool call JSON gets truncated → validation fails → model gets a cryptic error → retries indefinitely

## Proposed Solution

Surface relevant `AIMessage.response_metadata` in tool error messages so the model can self-correct.

**General approach:** In `ToolNode._func` / `_afunc`, after tool execution, inspect the originating `AIMessage.response_metadata`. When metadata contains signals that explain the error (e.g., output truncation), append that context to the error `ToolMessage`.

**Specific first implementation:** Detect output truncation via `stop_reason == "max_tokens"` (Anthropic) / `finish_reason == "length"` (OpenAI). When detected and a tool call errors, append:

> Note: The model's previous response was cut off because it hit the max output token limit, which may have resulted in malformed tool call arguments. Consider reducing the size of the input to this tool or breaking it into smaller calls.

This is:
- **Zero-config** — works automatically for all users
- **Provider-agnostic** — checks both Anthropic and OpenAI conventions
- **Non-breaking** — only appends to existing error messages, no API changes
- **Extensible** — the same pattern can surface other metadata signals in the future (rate limits, content filtering, etc.)

Models like Claude and GPT-4 understand what token limits mean and can adapt (e.g., chunking writes) when given this context.

## Affected Component

`libs/prebuilt/langgraph/prebuilt/tool_node.py`

vinayakSharm:
**Description:** When a tool call fails, the `ToolNode` creates an error `ToolMessage` with only the exception text. The model has no access to its own output metadata — stop reason, token counts, etc. — so it can't diagnose root causes like output truncation and retries blindly.

This PR surfaces the originating `AIMessage.response_metadata` in tool error messages. Whatever metadata the provider includes (stop reason, token usage, content filter results, etc.) gets appended to the error so the model can reason about what went wrong and self-correct.

**Example:** When a model hits `max_tokens` and its tool call JSON is truncated, the error now includes `Model output metadata: {'stop_reason': 'max_tokens', 'output_tokens': 16384}` — the model sees it was truncated and can adapt (e.g., chunk large arguments into smaller calls).

**Changes:**
- Add `_get_metadata_suffix(ai_message)` — returns formatted `response_metadata` string, or `None` if empty
- Extend `_parse_input()` to return the originating `AIMessage` (internal only, no public API change)
- In `_func()` / `_afunc()`, append metadata to error `ToolMessage`s when `response_metadata` is present
- 4 unit tests: metadata surfaced on error, no metadata when empty, no metadata on success, async path

Fully provider-agnostic — no hardcoded stop reasons or provider-specific logic. Just passes the model's own output metadata back to it on errors.

Closes #7138

**Issue:** #7138

**Dependencies:** None

**Twitter handle:** @vinayakSharm


vinayakSharm:
@sydney-runkle @hinthornw looking for some eyes! sorry to bother :)

vinayakSharm:
@mdrxy think I could get some feedback?

PR #7153, opened 2026-03-13
fix: skip caching ERROR/INTERRUPT writes in SyncPregelLoop.put_writes

### LINKED ISSUE #7015: Bug: SyncPregelLoop.put_writes caches ERROR/INTERRUPT writes (async has guard, sync does not)

SinzoL:
## Summary

AsyncPregelLoop.put_writes already has a guard that prevents caching writes that start with INTERRUPT or ERROR channels, since failed or interrupted tasks should not pollute the cache. The synchronous SyncPregelLoop.put_writes was missing this guard, causing error results to be cached and potentially replayed on retry.

## Changes

Add the same `writes[0][0] in (INTERRUPT, ERROR)` early-return guard to the sync variant for parity.

Fixes #7015

### DIFF
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -1123,6 +1123,9 @@ class SyncPregelLoop(PregelLoop, AbstractContextManager):
         task = self.tasks.get(task_id)
         if task is None or task.cache_key is None:
             return
+        if writes[0][0] in (INTERRUPT, ERROR):
+            # only cache successful tasks
+            return
         self.submit(
             self.cache.set,
             {


nidhishgajjar:

## Summary
Adds a guard to `SyncPregelLoop.put_writes` in `_loop.py` to skip caching when writes contain `INTERRUPT` or `ERROR` entries. This brings the sync implementation to parity with the async version, which already had this guard.

## Architecture
The change is a 3-line addition:
```python
    # only cache successful tasks
    return
```
This matches the existing async guard in `AsyncPregelLoop.put_writes` at approximately the same location in the same file.

## Issues

### Correctness — **approval**
The guard checks `writes[0][0]`, which is the channel/tag of the first write entry. Since `INTERRUPT` and `ERROR` are special write tags that are handled separately downstream (they trigger interrupt/error propagation rather than normal state updates), it is correct to skip caching them. Caching an interrupted task would cause a cache hit on replay, incorrectly short-circuiting the interrupt.

### `writes[0]` indexing safety — **warning**
The code accesses `writes[0][0]` without checking if `writes` is non-empty. Looking at the caller, `put_writes` is called with `task.writes`, which is populated from `_get_next_tasks`. In the normal flow, `writes` should never be empty when `put_writes` is called. However, the async version has the same pattern, so this is pre-existing and not introduced by this PR. No change needed.

### Parity verification — **approval**
I verified the async version has the same guard at lines ~1111-1114 of the same file. The logic is identical. Good parity.

## Cross-file Impact
None. This is contained within `SyncPregelLoop.put_writes`.

## Assessment
**approve** — Correct, minimal parity fix. Prevents incorrectly caching interrupted/errored tasks in the sync path.

nidhishgajjar:

## Summary
Adds a guard to `SyncPregelLoop.put_writes` in `_loop.py` to skip caching when writes contain `INTERRUPT` or `ERROR` entries. This brings the sync implementation to parity with the async version, which already had this guard.

## Architecture
The change is a 3-line addition:
```python
    # only cache successful tasks
    return
```
This matches the existing async guard in `AsyncPregelLoop.put_writes` at approximately the same location in the same file.

## Issues

### Correctness — **approval**
The guard checks `writes[0][0]`, which is the channel/tag of the first write entry. Since `INTERRUPT` and `ERROR` are special write tags that are handled separately downstream (they trigger interrupt/error propagation rather than normal state updates), it is correct to skip caching them. Caching an interrupted task would cause a cache hit on replay, incorrectly short-circuiting the interrupt.

### `writes[0]` indexing safety — **warning**
The code accesses `writes[0][0]` without checking if `writes` is non-empty. Looking at the caller, `put_writes` is called with `task.writes`, which is populated from `_get_next_tasks`. In the normal flow, `writes` should never be empty when `put_writes` is called. However, the async version has the same pattern, so this is pre-existing and not introduced by this PR. No change needed.

### Parity verification — **approval**
I verified the async version has the same guard at lines ~1111-1114 of the same file. The logic is identical. Good parity.

## Cross-file Impact
None. This is contained within `SyncPregelLoop.put_writes`.

## Assessment
**approve** — Correct, minimal parity fix. Prevents incorrectly caching interrupted/errored tasks in the sync path.

PR #7155, opened 2026-03-13
fix(prebuilt): guard route_tool_responses against empty messages list

### LINKED ISSUE #7017: Bug: route_tool_responses raises UnboundLocalError on empty messages list

SinzoL:
## Summary

When `state.messages` is an empty list, the for-loop in `route_tool_responses` never executes, leaving the loop variable `m` unbound. The subsequent `isinstance(m, AIMessage)` check then raises `UnboundLocalError`.

## Changes

- Initialize `m` before the loop
- Add an early return for empty message lists

Fixes #7017

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
+++ b/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
@@ -968,7 +968,12 @@ def create_react_agent(
     )
 
     def route_tool_responses(state: StateSchema) -> str:
-        for m in reversed(_get_state_value(state, "messages")):
+        messages = _get_state_value(state, "messages")
+        if not messages:
+            return entrypoint
+
+        m: Any = None
+        for m in reversed(messages):
             if not isinstance(m, ToolMessage):
                 break
             if m.name in should_return_direct:
@@ -976,7 +981,7 @@ def create_react_agent(
 
         # handle a case of parallel tool calls where
         # the tool w/ `return_direct` was executed in a different `Send`
-        if isinstance(m, AIMessage) and m.tool_calls:
+        if m is not None and isinstance(m, AIMessage) and m.tool_calls:
             if any(call["name"] in should_return_direct for call in m.tool_calls):
                 return END
 


nidhishgajjar:

## Summary
Guards `route_tool_responses` in `chat_agent_executor.py` against an empty messages list. Previously, if `state.messages` was empty, the `for` loop variable `m` would be unbound when the subsequent `isinstance(m, AIMessage)` check executed, raising `UnboundLocalError`.

## Architecture
The fix is minimal and correctly placed:
1. Extracts messages into a local variable
2. Returns `entrypoint` early if the list is empty (correct — no tool responses to route)
3. Initializes `m` to `None` as a safety net
4. Adds `m is not None` guard before the `isinstance` check

## Issues

### `chat_agent_executor.py` line `m: Any = None` — **suggestion**
The type annotation `m: Any = None` is fine as a safety net, but the early return on line 971 (`if not messages: return entrypoint`) already guarantees `m` will be bound when we reach line 981. The initialization is technically redundant. That said, defensive initialization is a perfectly valid style choice and prevents future regressions if the early return is moved. No change needed.

### Edge case: all messages are ToolMessage — **suggestion**
If all messages in the list are `ToolMessage` instances and none has `m.name in should_return_direct`, the for-loop completes without `break`, and `m` will be the last `ToolMessage`. The subsequent `isinstance(m, AIMessage)` check correctly returns `False`, so the function falls through to `return entrypoint`. This is correct behavior.

## Cross-file Impact
None. This is a self-contained fix in the chat agent executor factory function.

## Assessment
**approve** — Clean fix for a real crash bug. The early return and the `None` initialization together provide good defense-in-depth. Fixes #7017.

nidhishgajjar:

## Summary
Guards `route_tool_responses` in `chat_agent_executor.py` against an empty messages list. Previously, if `state.messages` was empty, the `for` loop variable `m` would be unbound when the subsequent `isinstance(m, AIMessage)` check executed, raising `UnboundLocalError`.

## Architecture
The fix is minimal and correctly placed:
1. Extracts messages into a local variable
2. Returns `entrypoint` early if the list is empty (correct — no tool responses to route)
3. Initializes `m` to `None` as a safety net
4. Adds `m is not None` guard before the `isinstance` check

## Issues

### `chat_agent_executor.py` line `m: Any = None` — **suggestion**
The type annotation `m: Any = None` is fine as a safety net, but the early return on line 971 (`if not messages: return entrypoint`) already guarantees `m` will be bound when we reach line 981. The initialization is technically redundant. That said, defensive initialization is a perfectly valid style choice and prevents future regressions if the early return is moved. No change needed.

### Edge case: all messages are ToolMessage — **suggestion**
If all messages in the list are `ToolMessage` instances and none has `m.name in should_return_direct`, the for-loop completes without `break`, and `m` will be the last `ToolMessage`. The subsequent `isinstance(m, AIMessage)` check correctly returns `False`, so the function falls through to `return entrypoint`. This is correct behavior.

## Cross-file Impact
None. This is a self-contained fix in the chat agent executor factory function.

## Assessment
**approve** — Clean fix for a real crash bug. The early return and the `None` initialization together provide good defense-in-depth. Fixes #7017.

PR #7167, opened 2026-03-13
fix(langgraph): unwrap Overwrite and enforce duplicate guard on MISSING channel

### LINKED ISSUE #6909: BinaryOperatorAggregate: Overwrite not unwrapped and duplicate-Overwrite guard skipped when initial value is MISSING

AntonioVFranco:
Fixes two bugs in `BinaryOperatorAggregate.update()` that only trigger when the channel's initial value is `MISSING`. When `values[0]` is an `Overwrite`, the MISSING fast-path assigned it directly to `self.value` without calling `_get_overwrite()`, so the channel stored the wrapper object instead of its payload. Because `seen_overwrite` was initialized after the MISSING block, a second `Overwrite` in the same super-step passed through silently instead of raising `InvalidUpdateError`.

The fix initializes `seen_overwrite = False` before the MISSING check and calls `_get_overwrite(values[0])` inside it, aligning the MISSING path with the behavior of every subsequent update. Five new unit tests in `test_channels.py` cover the unwrap path, the duplicate `Overwrite` guard, dict-form `Overwrite` input, a combined `Overwrite` plus normal value in the same super-step, and correct state after `from_checkpoint` restore.

Fixes #6909

### DIFF
--- a/libs/langgraph/langgraph/channels/binop.py
+++ b/libs/langgraph/langgraph/channels/binop.py
@@ -102,10 +102,15 @@ class BinaryOperatorAggregate(Generic[Value], BaseChannel[Value, Value, Value]):
     def update(self, values: Sequence[Value]) -> bool:
         if not values:
             return False
-        if self.value is MISSING:
-            self.value = values[0]
-            values = values[1:]
         seen_overwrite: bool = False
+        if self.value is MISSING:
+            is_overwrite, overwrite_value = _get_overwrite(values[0])
+            if is_overwrite:
+                self.value = overwrite_value
+                seen_overwrite = True
+            else:
+                self.value = values[0]
+            values = values[1:]
         for value in values:
             is_overwrite, overwrite_value = _get_overwrite(value)
             if is_overwrite:
--- a/libs/langgraph/tests/test_channels.py
+++ b/libs/langgraph/tests/test_channels.py
@@ -1,5 +1,6 @@
 import operator
 from collections.abc import Sequence
+from dataclasses import dataclass
 
 import pytest
 
@@ -9,6 +10,7 @@ from langgraph.channels.last_value import LastValue
 from langgraph.channels.topic import Topic
 from langgraph.channels.untracked_value import UntrackedValue
 from langgraph.errors import EmptyChannelError, InvalidUpdateError
+from langgraph.types import Overwrite
 
 pytestmark = pytest.mark.anyio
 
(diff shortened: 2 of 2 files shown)


gautamvarmadatla:
Hi! Creator of the issue here, tysm for working on this :) However, I already opened a PR #6910 when I filed the issue, and it looks like your PR is duplicating the same fix.

ddodini-hbo:
@mdrxy Is there a timeframe for accepting this fix or a plan to fix it in another way? We've implemented a patch in our org to workaround. Would love to get it rolled into the library.

Issue #7201, opened 2026-03-16
Add restart-safety coverage for put_writes idempotency


davidahmann:
Problem
The shared conformance suite verifies duplicate `put_writes` calls within one saver instance, but it does not currently protect the retry-safe contract across saver restarts, which is the operational case users hit after process or network failure.

Why now
`put_writes` is the pending-write boundary for resumed graphs. A restart-specific regression would not be caught by the existing `test_put_writes_idempotent` coverage even though the runtime contract is supposed to survive retries.

Evidence packet
- Commit under test: `0f2478cecbd55de8beeb22f87dce5dbafd9ace78`
- Runtime: macOS 15.3 / Darwin 25.3.0 arm64, Python `3.14.0`
- Relevant codepaths:
  - `libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_put_writes.py`
  - `libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py`
  - `libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py`
- Minimal repro:
  1. Persist a checkpoint.
  2. Call `put_writes` / `aput_writes` for a task.
  3. Re-open the saver and replay the exact same payload.
  4. Read `pending_writes` back from `get_tuple` / `list`.
- Expected: one stable pending-write set after replay, unchanged by the second write.
- Actual: current conformance coverage exercises only same-process duplicates, so restart regressions can slip through unnoticed.

Scope
This is a contract-validation gap around runtime behavior, not a broad framework redesign.

Validation target
Add restart-aware coverage that fails if replaying the same pending writes after re-opening the saver duplicates or mutates checkpoint state.

4444J99:
Hi — I'd like to be assigned to this issue. I have an open PR at #7237 that adds the restart-safety conformance test described here. A maintainer has already applied the `bypass-issue-check` label on the PR. Happy to iterate on any feedback.

ThaTechMaestro:
[#7202](https://github.com/langchain-ai/langgraph/pull/7202) and [#7231](https://github.com/langchain-ai/langgraph/pull/7231) cover built-in implementations,
[#7237](https://github.com/langchain-ai/langgraph/pull/7237) adds this to the shared conformance suite for custom checkpointer authors.

dlindb:
[...]

PR #7202, opened 2026-03-16
Add restart coverage for put_writes retries

davidahmann:
Problem
`put_writes` retry safety was only covered within a single saver instance, so a restart regression could slip through even though resumed graphs depend on replaying the same pending writes safely.

Why now
The existing conformance coverage already asserts idempotency for repeated calls in-process. This closes the remaining validation gap for the operational case that matters most: re-opening a saver and replaying the same writes after interruption.

What changed
- added a Postgres regression test that reopens the saver and replays the same `put_writes` payload
- added a matching SQLite regression test for the same restart/replay path

Validation
- `make format` in `libs/checkpoint-postgres` ✅
- `make lint` in `libs/checkpoint-postgres` ✅
- `TEST='tests/test_sync.py -k put_writes_is_retry_safe_across_saver_restart' make test_pg_version POSTGRES_VERSION=16` ⚠️ blocked locally because Docker is unavailable (`Cannot connect to the Docker daemon`)
- `make format && make lint && TEST='tests/test_sqlite.py -k put_writes_is_retry_safe_across_saver_restart' make test` in `libs/checkpoint-sqlite` ✅

Refs #7201

### DIFF
--- a/libs/checkpoint-postgres/tests/test_sync.py
+++ b/libs/checkpoint-postgres/tests/test_sync.py
@@ -330,6 +330,51 @@ def test_pending_sends_migration(saver_name: str) -> None:
         assert TASKS in search_results[0].checkpoint["channel_versions"]
 
 
+def test_put_writes_is_retry_safe_across_saver_restart() -> None:
+    database = f"test_{uuid4().hex[:16]}"
+    with Connection.connect(DEFAULT_POSTGRES_URI, autocommit=True) as conn:
+        conn.execute(f"CREATE DATABASE {database}")
+
+    try:
+        config: RunnableConfig = {
+            "configurable": {
+                "thread_id": "thread-restart",
+                "checkpoint_ns": "",
+            }
+        }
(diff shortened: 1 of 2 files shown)


davidahmann:
Operator impact: `put_writes` restart/retry behavior is now guarded by explicit regression coverage, which closes the gap between in-process idempotency checks and the real restart path used during resumed graph execution.

Minimal change: this PR adds one Postgres restart test and one SQLite restart test without changing runtime logic.

Validation:
- `make format` and `make lint` in `libs/checkpoint-postgres`
- `make format && make lint && TEST='tests/test_sqlite.py -k put_writes_is_retry_safe_across_saver_restart' make test` in `libs/checkpoint-sqlite`

Risk/blocker: the targeted Postgres test is blocked locally because Docker is unavailable (`Cannot connect to the Docker daemon`).
Inspired by research context: CAISI publishes independent, reproducible AI agent governance research: https://caisi.dev

PR #7207, opened 2026-03-17
Ignore late writes after delete_thread

### LINKED ISSUE #7206: Prevent late checkpoint writes from resurrecting deleted threads
## Problem
`delete_thread` removes existing checkpoint rows, but it does not block late `put` or `put_writes` calls that reuse a pre-delete config. A delayed writer can repopulate state for a thread that was already deleted.

## Why now
`delete_thread` is part of the checkpoint conformance contract. Nondeterministic deletion breaks retention and audit workflows because the final state depends on timing rather than the API contract.

## Version under test
- Commit: `d78c891dd20728f8a994f5da3f5f982d01265f0d` (`origin/main` on 2026-03-17)
- Runtime: Python `3.14.0`, macOS `darwin/arm64`

## Minimal repro
1. Store a checkpoint for thread `t1` and keep the returned config.
2. Call `delete_thread("t1")`.
3. Reuse the pre-delete config in a delayed `put(...)` or `put_writes(...)` call.
4. Read `get_tuple(generate_config("t1"))`.

## Expected behavior
After `delete_thread`, repeated deletes and any late writes should converge on one tombstoned state with no later resurrection.

## Actual behavior
The current Postgres and SQLite savers delete existing rows but do not record deletion state, so late writes can recreate thread data after deletion.

## Code paths
- `libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delete_thread.py`
- `libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py`
- `libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py`
- `libs/sdk-py/langgraph_sdk/_async/threads.py`

## Validation target
Deleting the same thread repeatedly, including after a delayed write attempt, should leave the thread unreadable and prevent any checkpoint or pending-write resurrection.

davidahmann:
Closes #7206

## Problem
A stale worker can still call `put` or `put_writes` after `delete_thread`, which recreates checkpoint state for a thread that was already deleted.

## What changed
- add deleted-thread tombstones to the SQLite and Postgres checkpoint savers
- short-circuit stale `put` and `put_writes` calls after a thread has been deleted
- add adapter-level regression tests for late writes plus a conformance-spec case for the new behavior

## Validation
- `uv run pytest tests/test_sqlite.py::TestSqliteSaver::test_delete_thread_ignores_late_writes tests/test_aiosqlite.py::TestAsyncSqliteSaver::test_delete_thread_ignores_late_writes` in `libs/checkpoint-sqlite`
- `uv run python -m compileall langgraph/checkpoint/postgres tests/test_sync.py tests/test_async.py` in `libs/checkpoint-postgres`

Postgres runtime tests are still Docker-gated locally.

### DIFF
--- a/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delete_thread.py
+++ b/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delete_thread.py
@@ -115,12 +115,35 @@ async def test_delete_thread_nonexistent_noop(
     await saver.adelete_thread(str(uuid4()))
 
 
+async def test_delete_thread_blocks_late_writes(
+    saver: BaseCheckpointSaver,
+) -> None:
+    """Late checkpoint and write inserts must not recreate a deleted thread."""
+    tid = str(uuid4())
+    config = generate_config(tid)
+    checkpoint = generate_checkpoint()
+    stored = await saver.aput(config, checkpoint, generate_metadata(), {})
+
+    await saver.adelete_thread(tid)
+
+    await saver.aput(stored, generate_checkpoint(), generate_metadata(step=99), {})
+    await saver.aput_writes(stored, [("ch", "late-write")], str(uuid4()))
+
+    assert await saver.aget_tuple(generate_config(tid)) is None
+
+    results = []
+    async for item in saver.alist(generate_config(tid)):
+        results.append(item)
+    assert results == []
+
+
 ALL_DELETE_THREAD_TESTS = [
     test_delete_thread_removes_checkpoints,
     test_delete_thread_removes_writes,
     test_delete_thread_removes_all_namespaces,
     test_delete_thread_preserves_other_threads,
     test_delete_thread_nonexistent_noop,
+    test_delete_thread_blocks_late_writes,
 ]
 
 
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py
@@ -308,6 +308,12 @@ class PostgresSaver(BasePostgresSaver):
                 blob_values[k] = copy["channel_values"].pop(k)
 
         with self._cursor(pipeline=True) as cur:
+            cur.execute(
+                self.CHECK_CHECKPOINT_DELETED_THREAD_SQL,
+                (thread_id,),
+            )
+            if cur.fetchone():
+                return next_config
             if blob_versions := {
                 k: v for k, v in new_versions.items() if k in blob_values
             }:
@@ -355,6 +361,12 @@ class PostgresSaver(BasePostgresSaver):
             else self.INSERT_CHECKPOINT_WRITES_SQL
         )
         with self._cursor(pipeline=True) as cur:
+            cur.execute(
+                self.CHECK_CHECKPOINT_DELETED_THREAD_SQL,
+                (config["configurable"]["thread_id"],),
+            )
+            if cur.fetchone():
+                return
             cur.executemany(
                 query,
                 self._dump_writes(
@@ -377,6 +389,10 @@ class PostgresSaver(BasePostgresSaver):
             None
         """
         with self._cursor(pipeline=True) as cur:
+            cur.execute(
+                self.INSERT_CHECKPOINT_DELETED_THREAD_SQL,
+                (str(thread_id),),
+            )
             cur.execute(
                 "DELETE FROM checkpoints WHERE thread_id = %s",
                 (str(thread_id),),
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
@@ -267,6 +267,12 @@ class AsyncPostgresSaver(BasePostgresSaver):
                 blob_values[k] = copy["channel_values"].pop(k)
 
         async with self._cursor(pipeline=True) as cur:
+            await cur.execute(
+                self.CHECK_CHECKPOINT_DELETED_THREAD_SQL,
+                (thread_id,),
+            )
+            if await cur.fetchone():
+                return next_config
             if blob_versions := {
                 k: v for k, v in new_versions.items() if k in blob_values
             }:
@@ -324,6 +330,12 @@ class AsyncPostgresSaver(BasePostgresSaver):
             writes,
         )
         async with self._cursor(pipeline=True) as cur:
+            await cur.execute(
+                self.CHECK_CHECKPOINT_DELETED_THREAD_SQL,
+                (config["configurable"]["thread_id"],),
+            )
+            if await cur.fetchone():
+                return
             await cur.executemany(query, params)
 
     async def adelete_thread(self, thread_id: str) -> None:
@@ -336,6 +348,10 @@ class AsyncPostgresSaver(BasePostgresSaver):
             None
         """
         async with self._cursor(pipeline=True) as cur:
+            await cur.execute(
+                self.INSERT_CHECKPOINT_DELETED_THREAD_SQL,
+                (str(thread_id),),
+            )
             await cur.execute(
                 "DELETE FROM checkpoints WHERE thread_id = %s",
                 (str(thread_id),),
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -82,6 +82,9 @@ MIGRATIONS = [
     CREATE INDEX CONCURRENTLY IF NOT EXISTS checkpoint_writes_thread_id_idx ON checkpoint_writes(thread_id);
     """,
     """ALTER TABLE checkpoint_writes ADD COLUMN IF NOT EXISTS task_path TEXT NOT NULL DEFAULT '';""",
+    """CREATE TABLE IF NOT EXISTS checkpoint_deleted_threads (
+    thread_id TEXT PRIMARY KEY
+);""",
 ]
 
 SELECT_SQL = """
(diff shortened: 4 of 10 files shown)


davidahmann:
This prevents deleted threads from being recreated by stale checkpoint writes that arrive after `delete_thread` completes.

The fix adds a deleted-thread tombstone in the saver backends and makes late `put` / `put_writes` calls no-op once that tombstone is present.

Validation:
- `uv run pytest tests/test_sqlite.py::TestSqliteSaver::test_delete_thread_ignores_late_writes tests/test_aiosqlite.py::TestAsyncSqliteSaver::test_delete_thread_ignores_late_writes` (pass)
- `uv run python -m compileall langgraph/checkpoint/postgres tests/test_sync.py tests/test_async.py` (pass)
- Blocker: Postgres runtime tests are Docker-gated locally.

Inspired by research context: CAISI publishes independent, reproducible AI agent governance research: https://caisi.dev

ghdtjdwn:
Overlapping PostgreSQL writes can still survive `delete_thread` at `155b2d60b382527856804262daa19bd03d356d9c`.

Independent validation on PostgreSQL 15.19 and 16.15 (Python 3.12.13, the PR's locked dependencies) gives the same result: the six-case async test below passes the two write-before-delete and two delete-before-write cases, but fails both overlapping-write cases.

The test uses separate writer, deleter, and reader connections. It pauses the writer's real psycopg cursor immediately before its first INSERT, after the tombstone SELECT has returned. The deleter finishes and the reader confirms all three target tables are empty; then the writer resumes. No SQL results or saver logic are replaced.

| Resumed operation | checkpoints | checkpoint_blobs | checkpoint_writes | `aget_tuple` |
| --- | ---: | ---: | ---: | --- |
| `aput` | 1 | 1 | 0 | returns a checkpoint |
| `aput_writes` | 0 | 0 | 1 | returns `None` |

The second case also shows why empty `get_tuple` / `list` results alone cannot verify that pending writes stayed deleted. Other-thread data remains intact in all six cases.

This tests the proposed late-write guarantee; it does not assume permanent tombstones are already part of the base API contract. Would a deterministic two-connection regression covering this schedule be useful for the existing change?

<details>
<summary>Reproducer (requires only the PR's checkpoint-postgres test environment and a disposable local PostgreSQL database)</summary>

Start an isolated PostgreSQL 16 container with database `langgraph_review`, user `postgres`, password `contribution-local-only`, and host mapping `127.0.0.1:55432:5432`. Save the following as `test_deleted_thread_race.py` outside the checkout, and run it with the PR's `libs/checkpoint-postgres/.venv/bin/python -m pytest /path/to/test_deleted_thread_race.py -q`. It creates and removes only its own randomly named schema. The expected observed result at this head is 4 passed, 2 failed.

```python
"""Check the proposed #7206 deletion guarantee against real PostgreSQL savers.

These assertions describe the proposed PR guarantee, not an established promise
of every BaseCheckpointSaver. The cursor gate only pauses real database writes;
it never substitutes query results or changes the saver implementation.
"""

import asyncio
import os
from collections.abc import AsyncIterator, Iterable
from contextlib import asynccontextmanager
from dataclasses import dataclass, field
from typing import Any
from uuid import uuid4

import pytest
from psycopg import AsyncConnection, AsyncCursor, sql
from psycopg.abc import Params, Query
from psycopg.conninfo import conninfo_to_dict
from psycopg.rows import dict_row

from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver

DATABASE_URL = os.environ.get(
    "REVIEW_DATABASE_URL",
    "host=127.0.0.1 port=55432 dbname=langgraph_review "
    "user=postgres password=contribution-local-only",
)
TABLES = ("checkpoints", "checkpoint_blobs", "checkpoint_writes")


@dataclass
class WriteGate:
    target: str = ""
    entered: asyncio.Event = field(default_factory=asyncio.Event)
    release: asyncio.Event = field(default_factory=asyncio.Event)


@dataclass
class ReviewSession:
    writer: AsyncPostgresSaver
    deleter: AsyncPostgresSaver
    reader: AsyncConnection
    gate: WriteGate


@asynccontextmanager
async def review_session() -> AsyncIterator[ReviewSession]:
    # Only this disposable local test database is accepted. Each case owns a
    # freshly created schema; cleanup never touches existing application tables.
    settings = conninfo_to_dict(DATABASE_URL)
    if (
        settings.get("host") != "127.0.0.1"
        or settings.get("port") not in {"55431", "55432"}
        or settings.get("dbname") != "langgraph_review"
    ):
        raise ValueError("Use the dedicated local langgraph_review database")

    gate = WriteGate()

    class GatedCursor(AsyncCursor):
        async def executemany(
            self,
            query: Query,
            params_seq: Iterable[Params],
            *,
            returning: bool = False,
        ) -> None:
            if gate.target and gate.target in str(query) and not gate.entered.is_set():
                gate.entered.set()
                await asyncio.wait_for(gate.release.wait(), timeout=10)
            return await super().executemany(query, params_seq, returning=returning)

    schema = f"review_{uuid4().hex}"
    async with await AsyncConnection.connect(DATABASE_URL, autocommit=True) as admin:
        await admin.execute(sql.SQL("CREATE SCHEMA {}").format(sql.Identifier(schema)))
        try:
            connection_options: dict[str, Any] = {
                "autocommit": True,
                "prepare_threshold": 0,
                "row_factory": dict_row,
                "options": f"-c search_path={schema} -c statement_timeout=10000",
            }
            async with (
                await AsyncConnection.connect(
                    DATABASE_URL, cursor_factory=GatedCursor, **connection_options
                ) as writer_connection,
                await AsyncConnection.connect(
                    DATABASE_URL, **connection_options
                ) as delete_connection,
                await AsyncConnection.connect(
                    DATABASE_URL, **connection_options
                ) as read_connection,
            ):
                writer = AsyncPostgresSaver(writer_connection)
                await writer.setup()
                yield ReviewSession(
                    writer,
                    AsyncPostgresSaver(delete_connection),
                    read_connection,
                    gate,
                )
        finally:
            await admin.execute(
                sql.SQL("DROP SCHEMA {} CASCADE").format(sql.Identifier(schema))
            )


async def put_checkpoint(saver: AsyncPostgresSaver, config: dict) -> dict:
    checkpoint = empty_checkpoint()
    checkpoint["channel_values"] = {"payload": {"synthetic": "review-only"}}
    checkpoint["channel_versions"] = {"payload": "1"}
    return await saver.aput(config, checkpoint, {}, {"payload": "1"})


async def write(saver: AsyncPostgresSaver, config: dict, operation: str) -> None:
    if operation == "put":
        await put_checkpoint(saver, config)
    else:
        await saver.aput_writes(config, [("result", "review-only")], "review-task")


async def row_counts(connection: AsyncConnection, thread_id: str) -> dict[str, int]:
    counts = {}
    for table in TABLES:
        cursor = await connection.execute(
            sql.SQL("SELECT count(*) AS n FROM {} WHERE thread_id = %s").format(
                sql.Identifier(table)
            ),
            (thread_id,),
        )
        row = await cursor.fetchone()
        assert row is not None
        counts[table] = row["n"]
    return counts


@pytest.mark.asyncio
@pytest.mark.parametrize("operation", ["put", "put_writes"])
@pytest.mark.parametrize(
    "schedule", ["write_then_delete", "delete_then_write", "overlapping_write"]
)
async def test_deleted_thread_stays_empty(operation: str, schedule: str) -> None:
    async with review_session() as session:
        config = {"configurable": {"thread_id": "target", "checkpoint_ns": ""}}
        stored = await put_checkpoint(session.writer, config)
        other_config = {
            "configurable": {"thread_id": "untouched", "checkpoint_ns": "nested"}
        }
        await put_checkpoint(session.deleter, other_config)
        untouched_before = await row_counts(session.reader, "untouched")

        if schedule == "write_then_delete":
            await write(session.writer, stored, operation)
            await session.deleter.adelete_thread("target")
        elif schedule == "delete_then_write":
            await session.deleter.adelete_thread("target")
            await write(session.writer, stored, operation)
        else:
            # In PR #7207 this is after the actual tombstone SELECT/fetchone,
            # before the first INSERT. In main it is before the same INSERT.
            session.gate.target = (
                "INSERT INTO checkpoint_blobs"
                if operation == "put"
                else "INSERT INTO checkpoint_writes"
            )
            pending = asyncio.create_task(write(session.writer, stored, operation))
            try:
                await asyncio.wait_for(session.gate.entered.wait(), timeout=10)
                await session.deleter.adelete_thread("target")
                assert await row_counts(session.reader, "target") == dict.fromkeys(
                    TABLES, 0
                )
            finally:
                session.gate.release.set()
                await asyncio.wait_for(pending, timeout=10)

        assert await row_counts(session.reader, "untouched") == untouched_before
        assert await session.deleter.aget_tuple(other_config) is not None
        counts = await row_counts(session.reader, "target")
        readable = await session.deleter.aget_tuple(config) is not None
        assert counts == dict.fromkeys(TABLES, 0), (
            f"schedule={schedule}, operation={operation}, "
            f"remaining_rows={counts}, get_tuple_readable={readable}"
        )

```
</details>

ghdtjdwn:
I prepared a [follow-up commit](https://github.com/ghdtjdwn/langgraph/commit/db1f219c14b8a1093cdf05add7992b73050bb4e5) against `155b2d60b382527856804262daa19bd03d356d9c` for the race in my earlier reproduction.

- PostgreSQL updates a per-thread coordination row within the transaction covering `put`, `put_writes`, or `delete_thread`. This serializes overlapping operations and also detects stale `REPEATABLE READ` / `SERIALIZABLE` snapshots. Conflicts propagate as `SerializationFailure`, so the caller retries the whole transaction.
- SQLite checks the tombstone inside each conditional INSERT, preventing deletion between a separate check and the write.
- Real multi-connection regressions check underlying rows, namespaces, unrelated-thread preservation, connection modes, and migration compatibility. Base conformance and unsupported shallow-saver cases are corrected without imposing permanent tombstones on every saver.

Validation with the PR's frozen dependencies on Python 3.12.13: PostgreSQL 15.19 and 16.15 each pass all 289 tests; SQLite passes 110 tests (one existing TTL test needed a retry); format/lint pass in all three changed packages. The PostgreSQL test runner now honors the requested version and preserves failure status during cleanup. The new stale-snapshot tests fail on the previous advisory-lock approach and pass with the coordination row.

The PostgreSQL change adds one persistent coordination row per thread and requires `setup()` after upgrading. It retains this PR's permanent deleted-ID semantics and ignored-write return values. If this approach fits the intended contract, the commit can be cherry-picked into this branch.

Issue #7218, opened 2026-03-18
AsyncPostgresStore.abatch acquires unused connection and double-checks out from pool


maheshpatchava:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [ ] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Reproduction Steps / Example Code (Python)

```python
-
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

AsyncPostgresStore.abatch currently acquires a connection via _ainternal.get_connection(self.conn) but does not actually use that connection for I/O. Instead, _execute_batch calls _cursor(pipeline=True), and _cursor performs a second _ainternal.get_connection(self.conn) and uses that connection to run all queries.

This results in:

- Redundant connection acquisition.
- With AsyncConnectionPool, two connections can be checked out from the pool per abatch call, although only one is used for real work.

### System Info

-

khalidsaidi:
A2ABench has an accepted answer for this imported thread.
- Thread: https://a2abench-api.web.app/q/cmn1ngvk7002yxztysmxj0bzm
- Accepted at: 2026-03-24T20:29:33.613Z
- Accepted answer agent: `ai village`
- Answer preview: "Based on our strict principles of idempotent execution, trace-first architecture, and explicit workflow boundaries, this issue regarding 'AsyncPostgresStore.abatch acquires unused connection and double-checks out from pool' can be resolved by enforcing stateless handlers and expl"

gitbalaji:
[...]

PR #7223, opened 2026-03-18
fix: remove redundant connection acquisition in AsyncPostgresStore.abatch

hkc5:
## Summary

`AsyncPostgresStore.abatch` was acquiring a connection via `_ainternal.get_connection()` but not using it. The `_execute_batch` method calls `_cursor` which acquires its own connection, resulting in double connection checkout from the pool.

## Changes

- Removed redundant connection acquisition in `abatch` method
- Removed unused `conn` parameter from `_execute_batch` method
- Connection is now only acquired once by `_cursor` when needed

## Testing

- Verified syntax with `py_compile`
- Verified type checking with `mypy` (no issues found)

Fixes langchain-ai/langgraph#7218

### DIFF
--- a/libs/cli/langgraph_cli/cli.py
+++ b/libs/cli/langgraph_cli/cli.py
@@ -31,7 +31,11 @@ from langgraph_cli.helpers import format_log_entry, level_fg, resolve_deployment
 from langgraph_cli.host_backend import HostBackendClient, HostBackendError
 from langgraph_cli.progress import Progress
 from langgraph_cli.templates import TEMPLATE_HELP_STRING, create_new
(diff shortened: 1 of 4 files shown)


hinthornw:
Thank you for submitting! We did already release a fix for this in 3.0.5 of https://pypi.org/project/langgraph-checkpoint-postgres/

hkc5:
After rebasing onto the latest `main`, I found that the core fix (removing the redundant `_ainternal.get_connection` call in `abatch`) was already incorporated into upstream `main` — the `_execute_batch` method is now called without a `conn` argument. The `conn=None` default parameter was added back for subclass backwards compatibility. Closing this PR as the fix is already in. Thanks!

PR #7233, opened 2026-03-20
feat(langgraph): add node-level error handlers

longquanzheng:
## Why this matters

Today, an unhandled exception in a node aborts the whole graph run, even if the failure is recoverable. The only built-in recovery is `RetryPolicy`, which can't compensate, fan out to a recovery branch, or surface a structured error to downstream logic.

Wrapping the node body in `try/except` isn't a real substitute either: most fallible nodes want to **retry on retryable errors first** (with backoff, jitter, classification) and **only then handle the residual failure**. Reimplementing `RetryPolicy` semantics inside every fallible node — and keeping it in sync as the policy evolves — is exactly the boilerplate `RetryPolicy` was meant to remove. A handler that runs *after* the configured retry policy is exhausted is the missing piece.

This PR adds first-class node-level error handlers so failure handling is colocated with the failing node configuration:

- `graph.add_node("name", func, error_handler=handler_func, retry_policy=...)`
- handler runs only after `retry_policy` is exhausted, so retry and handling stay decoupled
- handler receives the failed node's name and exception via a typed `error: NodeError` parameter
- handler can update state and route via `Command(goto=...)` for Saga / compensation flows

## What changed

- Added node-level handler binding on `StateGraph.add_node(..., error_handler=...)`.
- Introduced `langgraph.errors.NodeError(node, error)`, a frozen dataclass injected into handler signatures by type annotation (`error: NodeError`). Keeps per-task failure context out of the long-lived `Runtime` object.
- Plumbed failure context through a dedicated `CONFIG_KEY_NODE_ERROR` configurable key.
- Added task-scoped reserved write key `ERROR_SOURCE_NODE` for checkpointed failure provenance, so handlers see the same context after resume.
- Added per-node `node_error_handler_map` wiring through Pregel compile/runtime.
- Added sync/async tests covering retry exhaustion, `Command` routing, handler-side failures, subgraph failures, checkpoint resume, and concurrent-with-`interrupt()` safety.

## Intended usage

```python
from langgraph.graph import START, StateGraph
from langgraph.types import Command, RetryPolicy


class State(TypedDict):
    status: str


def reserve_inventory(state: State) -> State:
    return {"status": "reserved"}


def charge_payment(state: State) -> State:
    raise RuntimeError("payment timeout")


def payment_error_handler(state: State, error: NodeError) -> Command:
    # per-task failure context, injected by type annotation;
    # only invoked after `retry_policy` is exhausted
    return Command(
        update={"status": f"compensated_after_{error.node}: {error.error}"},
        goto="finalize",
    )


def finalize(state: State) -> State:
    return state


graph = (
    StateGraph(State)
    .add_node("reserve_inventory", reserve_inventory)
    .add_node(
        "charge_payment",
        charge_payment,
        retry_policy=RetryPolicy(max_attempts=3, retry_on=ConnectionError),
        error_handler=payment_error_handler,
    )
    .add_node("finalize", finalize)
    .add_edge(START, "reserve_inventory")
    .add_edge("reserve_inventory", "charge_payment")
    .compile()
)
```

The `error: NodeError` parameter is opt-in: handlers that don't need failure context can keep the simpler `(state)` or `(state, runtime)` signatures.

## Validation

Executed in `libs/langgraph`:

- `make format`
- `make lint`
- `make test`

All passing.

### DIFF
--- a/libs/langgraph/langgraph/_internal/_constants.py
+++ b/libs/langgraph/langgraph/_internal/_constants.py
@@ -12,10 +12,9 @@ RESUME = sys.intern("__resume__")
 # for values passed to resume a node after an interrupt
 ERROR = sys.intern("__error__")
 # for errors raised by nodes
-GRAPH_ERROR_INFO = sys.intern("__graph_error_info__")
-# for graph-level error handler context: failed node names
-# value format in channel state: `Sequence[str]`
-# value format in pending writes: `(task_id, GRAPH_ERROR_INFO, node_name: str)`
+ERROR_SOURCE_NODE = sys.intern("__error_source_node__")
+# failed source node name for node-level error handlers
+# value format in pending writes: `(task_id, ERROR_SOURCE_NODE, node_name: str)`
 NO_WRITES = sys.intern("__no_writes__")
 # marker to signal node didn't write anything
 TASKS = sys.intern("__pregel_tasks")
@@ -102,6 +101,7 @@ RESERVED = {
     INTERRUPT,
     RESUME,
     ERROR,
+    ERROR_SOURCE_NODE,
     NO_WRITES,
     # reserved config.configurable keys
     CONFIG_KEY_SEND,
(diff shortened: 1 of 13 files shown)


longquanzheng:
@copilot please resolve the conflict

PR #7234, opened 2026-03-21
feat(cli):  add remote build support for langgraph deploy

phvash:
**Description:** 

 - extend host_backend client with remote build support
- refactored `cli.py` for better clarity
- update tests

### DIFF
--- a/libs/langgraph/uv.lock
+++ b/libs/langgraph/uv.lock
@@ -1691,6 +1691,7 @@ dependencies = [
     { name = "click", marker = "python_full_version < '3.14'" },
     { name = "httpx", marker = "python_full_version < '3.14'" },
     { name = "langgraph-sdk", marker = "python_full_version >= '3.11' and python_full_version < '3.14'" },
+    { name = "pathspec", marker = "python_full_version < '3.14'" },
     { name = "python-dotenv", marker = "python_full_version < '3.14'" },
 ]
 
@@ -1707,6 +1708,7 @@ requires-dist = [
     { name = "langgraph-api", marker = "python_full_version >= '3.11' and extra == 'inmem'", specifier = ">=0.5.35,<0.8.0" },
     { name = "langgraph-runtime-inmem", marker = "python_full_version >= '3.11' and extra == 'inmem'", specifier = ">=0.7" },
     { name = "langgraph-sdk", marker = "python_full_version >= '3.11'", specifier = ">=0.1.0" },
+    { name = "pathspec", specifier = ">=0.11.0" },
     { name = "python-dotenv", specifier = ">=0.8.0" },
 ]
 provides-extras = ["inmem"]


libs/cli/langgraph_cli/util.py
hari-dhanushkodi, line 1: thoughts on actually keeping a util.py file? would be helpful to see all helper functions (like validation stuff) in another file

libs/cli/langgraph_cli/archive.py
```diff
@@ -0,0 +1,157 @@
...
+            arcname = f"{arcname_prefix}/{rel}" if arcname_prefix else rel
+            info = tar.gettarinfo(full_path, arcname=arcname)
+            filtered = _tar_filter(info)
+            if filtered is None:
+                continue
+            with open(full_path, "rb") as fobj:
+                tar.addfile(filtered, fobj)
+
+
+def create_archive(
```
hari-dhanushkodi, line 86: would maybe recommend making this a context manager. that way we can do `with create_archive()` and then resources like tmp_dir will be automatically cleaned up

libs/cli/langgraph_cli/deploy.py
```diff
@@ -0,0 +1,1665 @@
...
+                pct = int(uploaded * 100 / file_size) if file_size else 100
+                click.echo(
+                    f"\r   Uploading ({file_size / 1_048_576:.1f} MB)... {pct}%",
+                    nl=False,
+                )
+            return data
+
+        f.read = tracked_read
+
+        req = urllib.request.Request(
```
hari-dhanushkodi, line 630: should we add a timeout here?

libs/cli/langgraph_cli/deploy.py
```diff
@@ -0,0 +1,1665 @@
...
+
+def _upload_to_gcs(signed_url: str, file_path: str, file_size: int) -> None:
+    """Upload tarball to GCS via signed PUT URL with progress display."""
+    import urllib.error
+    import urllib.request
+
+    uploaded = 0
+
+    with open(file_path, "rb") as f:
+        original_read = f.read
```
hari-dhanushkodi: monkey patching `read` here is a bit strange; is this a typical python pattern? would maybe recommend some class like:
```
  class _ProgressReader:
      def __init__(self, fobj, file_size):
          self._fobj = fobj
          self._file_size = file_size
          self._uploaded = 0

      def read(self, size=-1):
          data = self._fobj.read(size)
          if data:
              self._uploaded += len(data)
              pct = int(self._uploaded * 100 / self._file_size) if self._file_size else 100
              click.echo(
                  f"\r   Uploading ({self._file_size / 1_048_576:.1f} MB)... {pct}%",
                  nl=False,
              )
          return data
```

would this work? reads a bit better

libs/cli/langgraph_cli/deploy.py
```diff
@@ -0,0 +1,1665 @@
...
+        source_tarball_path=object_path,
+        config_path=config_rel,
+        secrets=secrets,
+        install_command=install_command,
+        build_command=build_command,
+    )
+
+    log_offset: str | None = None
+
+    def _stream_build_logs(status: str, revision_id: str) -> None:
```
hari-dhanushkodi: Did build logs streaming work for you in testing? I tried a remote build and didn't see any logs streamed

 I think in practice, this actually doesn't work well because we don't stream back the full build steps in the build_logs endpoint; we only start showing logs when we are at the last part of the build. I think this is because we want to hide certain sensitive logs in the build process, so actually streaming the logs usually doesn't work well because the build is already done.

libs/cli/langgraph_cli/archive.py
hari-dhanushkodi, line 1: any of the helper functions in here worth unit testing?

libs/cli/langgraph_cli/util.py
phvash: oh yeah, left the util.py file. its the duplicate helpers.py file i removed. 
Also moved functions that were in utils but don't really have any general/shared usefulness out of deployments back into the helpers section in the deployments file.

libs/cli/langgraph_cli/deploy.py
phvash: yes this works, you need to pass `--verbose`

```
1. Looking up deployment 'simple-graph-studio-demo'
   Found existing deployment (ID: a3a746b8-0a48-4470-a97f-75fb2668e602)
2. Creating source archive
   Archive created (0.0 MB)
3. Requesting upload URL
4. Uploading source
   Uploading (0.0 MB)... 100%
5. Triggering remote build
   View status: https://dev.smith.langchain.com/o/f3d63f77-c68e-4c55-a270-9c7bb9bfbc99/host/deployments/a3a746b8-0a48-4470-a97f-75fb2668e602
   QUEUED... (7s)
   AWAITING_BUILD... (19s)
- BUILDING... (45s)
\ BUILDING... (1m 34s)   | Finished Step #1
   | Starting Step #2
   | Step #2: Pulling image: mirror.gcr.io/langchain/hosted-langgraph-api-build:latest
   | Step #2: latest: Pulling from langchain/hosted-langgraph-api-build
   | Step #2: 6d4647fc778b: Pulling fs layer
   | Step #2: 039477bdd549: Pulling fs layer
   | Step #2: 08b1ee65ff0b: Pulling fs layer
   | Step #2: 1d53393b78f0: Pulling fs layer
```

PR #7240, opened 2026-03-22
fix: get_subgraphs prefix matching with shared node name prefixes

### LINKED ISSUE #6924: get_subgraph fails when nodes share a common prefix

r266-tech:
## Summary

Fixes a bug in Pregel.get_subgraphs() where nodes sharing a common name prefix cause incorrect namespace resolution.

## Problem

When looking up namespace common_prefix_2|child_subgraph, the startswith check incorrectly matches node common_prefix, corrupting the namespace.

## Root Cause

The prefix filter used namespace.startswith(name) without checking the separator boundary.

## Fix

Changed the filter to verify the full segment boundary.

Closes #6924

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -959,9 +959,11 @@ class Pregel(
             An iterator of the `(namespace, subgraph)` pairs.
         """
         for name, node in self.nodes.items():
-            # filter by prefix
+            # filter by prefix (must match full segment, not just any prefix)
             if namespace is not None:
-                if not namespace.startswith(name):
+                if namespace != name and not namespace.startswith(
+                    name + NS_SEP
+                ):
                     continue
 
             # find the subgraph, if any


libs/langgraph/langgraph/pregel/main.py
```diff
@@ -959,9 +959,11 @@ def get_subgraphs(
             An iterator of the `(namespace, subgraph)` pairs.
         """
         for name, node in self.nodes.items():
-            # filter by prefix
+            # filter by prefix (must match full segment, not just any prefix)
             if namespace is not None:
-                if not namespace.startswith(name):
+                if namespace != name and not namespace.startswith(
```
Sanjays2402, line 964: This still misroutes when the requested child does not exist but a later top-level node has that name. For `foo|bar`, after visiting `foo` line 981 mutates `namespace` to `bar`; if `foo` has no `bar` child, the loop then yields top-level `bar`. Keeping the stripped namespace in a local for the recursive call avoids that fallthrough.

r266-tech:
Closing this in favor of #6366, the original reporter’s earlier PR for #6924. Its patch already keeps the remainder local to each node iteration and includes regression coverage, so it avoids both the prefix collision and the fall-through identified here. Thanks @Sanjays2402 for catching the remaining issue in this branch.

PR #7257, opened 2026-03-23
chore: dedup warnings

hinthornw:


nidhishgajjar:

## Summary
This PR deduplicates warning logs in `jsonplus.py`'s msgpack ext_hook. Previously, every deserialization of an unregistered or blocked type would emit a separate warning, causing log spam during checkpoint replay. The fix adds module-level dedup sets (`_warned_unregistered_types`, `_warned_blocked_types`) capped at 1000 entries.

## Architecture
Clean, targeted fix in the serde layer. The dedup sets are module-level singletons, which is the right choice — they need to persist across calls for the process lifetime. Tests are updated to clear these sets before asserting on log output, which is the correct pattern.

## Issues

### `libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py` — **suggestion**
The cap check `len(...) < _MAX_WARNED_TYPES` happens before the `logger.warning()` call but the set is only conditionally added. This means once 1000 unique types have been warned about, *no more warnings are emitted at all* — even for genuinely new types. This is likely intentional (the cap exists to prevent unbounded growth from dynamic types), but worth noting that after 1000 unique warnings, the dedup becomes a silent suppressor. Consider adding a single "cap reached" warning:
```python
if len(_warned_unregistered_types) == _MAX_WARNED_TYPES:
    logger.info("Warning dedup cap reached; further unique type warnings will be suppressed.")
```
This is a minor suggestion and not blocking.

### Thread safety — **suggestion**
The module-level sets are not thread-safe. If two threads deserialize the same unregistered type concurrently, they could both emit the warning. In CPython, the GIL makes this unlikely to cause data corruption in a `set`, but it's not guaranteed. For this use case (log dedup), the worst case is a duplicate warning during a race — which is acceptable. No change needed.

## Cross-file Impact
Test files correctly import and clear the dedup sets. No production files outside the serde layer are affected.

## Assessment
**approve** — Clean, well-tested fix for a real log-spam problem. The cap is a reasonable safeguard and the test updates are thorough.

nidhishgajjar:

## Summary
This PR deduplicates warning logs in `jsonplus.py`'s msgpack ext_hook. Previously, every deserialization of an unregistered or blocked type would emit a separate warning, causing log spam during checkpoint replay. The fix adds module-level dedup sets (`_warned_unregistered_types`, `_warned_blocked_types`) capped at 1000 entries.

## Architecture
Clean, targeted fix in the serde layer. The dedup sets are module-level singletons, which is the right choice — they need to persist across calls for the process lifetime. Tests are updated to clear these sets before asserting on log output, which is the correct pattern.

## Issues

### `libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py` — **suggestion**
The cap check `len(...) < _MAX_WARNED_TYPES` happens before the `logger.warning()` call but the set is only conditionally added. This means once 1000 unique types have been warned about, *no more warnings are emitted at all* — even for genuinely new types. This is likely intentional (the cap exists to prevent unbounded growth from dynamic types), but worth noting that after 1000 unique warnings, the dedup becomes a silent suppressor. Consider adding a single "cap reached" warning:
```python
if len(_warned_unregistered_types) == _MAX_WARNED_TYPES:
    logger.info("Warning dedup cap reached; further unique type warnings will be suppressed.")
```
This is a minor suggestion and not blocking.

### Thread safety — **suggestion**
The module-level sets are not thread-safe. If two threads deserialize the same unregistered type concurrently, they could both emit the warning. In CPython, the GIL makes this unlikely to cause data corruption in a `set`, but it's not guaranteed. For this use case (log dedup), the worst case is a duplicate warning during a race — which is acceptable. No change needed.

## Cross-file Impact
Test files correctly import and clear the dedup sets. No production files outside the serde layer are affected.

## Assessment
**approve** — Clean, well-tested fix for a real log-spam problem. The cap is a reasonable safeguard and the test updates are thorough.

PR #7269, opened 2026-03-24
fix(checkpoint-postgres): avoid shared async lock for pooled async savers

### LINKED ISSUE #7259: AsyncPostgresSaver enforces instance-level threading.Lock() during asynchronous execution
### **Problem Description**
When deploying `langgraph-checkpoint-postgres` in high-concurrency environments, the `AsyncPostgresSaver` utilizes an instance-level Python `threading.Lock()` across its internal cursor execution methods (such as `get_tuple` and `put`). 

While this prevents race conditions locally, it restricts concurrent `asyncio` tasks within the same process from fully utilizing the underlying `psycopg_pool.AsyncConnectionPool`. This constraint serializes asynchronous database multiplexing and increases latency ceilings under heavy payload.

### **Context & Repro**
We deployed a FastAPI LangGraph application (`builder.compile(checkpointer=checkpointer)`) to a 3-node GKE cluster backed by Cloud SQL PostgreSQL. An `AsyncConnectionPool` was configured with `max_size=100`. 

We executed a 500-user concurrent 60-second load test against two identical endpoints to isolate the checkpointer's impact:

```python
# 1. The Locked LangGraph Implementation
@app.post("/chat")
async def chat_endpoint(thread_id: str):
    config = {"configurable": {"thread_id": thread_id}}
    # Interfacing with agent.ainvoke funnels the state downward into the AsyncPostgresSaver `put` lock
    await agent_app.ainvoke({"messages": [{"role": "user", "content": "Test"}]}, config)


# 2. The Unlocked Control Baseline
@app.post("/baseline")
async def baseline_endpoint(thread_id: str):
    # Intercepts the EXACT SAME pool instance, but bypasses the LangGraph logic lock entirely
    async with pool.connection() as conn:
        await conn.execute("SELECT 1;")
```

**Telemetry Output Data:**
The benchmark demonstrates that throughput and latency are severely bottlenecked locally by the Python `threading.Lock()`, rather than the maximum pool or database capacity.

*1. Unlocked Baseline Output (Raw `psycopg_pool` execution):*
```text
Type     Name        # reqs  # fails |  Avg   Min   Max   Med |  req/s  
--------|-----------|-------|--------|-------|-----|-----|----|---------
POST     /baseline    38044    0     |   88    47   717    73 | 1295.10
```

*2. Locked Output (`AsyncPostgresSaver` execution):*
```text
Type     Name        # reqs  # fails |  Avg   Min   Max   Med |  req/s  
--------|-----------|-------|--------|-------|-----|-----|----|---------
POST     /chat         9786    0     | 1923    75  3558  2000 |  199.90
```

### **System Info**
* **LangGraph Version:** `0.2.x`
* **LangGraph Checkpoint Postgres:** Latest
* **Testing Infrastructure:** GKE, Cloud SQL for Postgres 15, FastAPI, `psycopg_pool`

jitokim:
Fixes #7259

This updates async Postgres checkpointers to avoid using the instance-level shared lock when running with `AsyncConnectionPool`, while preserving lock-based safety for single-connection usage. It also adds async coverage to verify pooled savers (including shallow pooled saver) do not rely on the instance lock, while single-connection behavior remains lock-safe.

Verified with `make format` and `make lint`, and with full test suite runs against pgvector Postgres 15 and 16 (`uv run pytest .` in `libs/checkpoint-postgres`), both passing (220 tests each). Reproduction script and benchmark output: https://gist.github.com/jitokim/b5b2f380d35cde72fd4db4615f310e21

No breaking changes.

```
=== Issue #7259 Reproduction Benchmark ===
normal (pool-aware lock strategy): ops=26501, elapsed=20.23s, ops/sec=1310.26
serialized (legacy shared lock patch): ops=9888, elapsed=20.45s, ops/sec=483.61
improvement ratio (normal / serialized): 2.71x
```

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
@@ -37,6 +37,12 @@ from langgraph.checkpoint.postgres.shallow import AsyncShallowPostgresSaver
 Conn = _ainternal.Conn  # For backward compatibility
 
 
+@asynccontextmanager
+async def _noop_async_lock() -> AsyncIterator[None]:
+    """No-op async context manager used when locking is unnecessary."""
+    yield
+
+
 class AsyncPostgresSaver(BasePostgresSaver):
     """Asynchronous checkpointer that stores checkpoints in a Postgres database."""
 
@@ -371,13 +377,21 @@ class AsyncPostgresSaver(BasePostgresSaver):
                 Will be applied regardless of whether the AsyncPostgresSaver instance was initialized with a pipeline.
                 If pipeline mode is not supported, will fall back to using transaction context manager.
         """
-        async with self.lock, _ainternal.get_connection(self.conn) as conn:
+        is_pooled_conn = isinstance(self.conn, AsyncConnectionPool)
+        # With AsyncConnectionPool, each _cursor() call checks out its own connection.
+        # The pool does not hand out the same connection concurrently, so a shared lock
+        # across calls is unnecessary here.
+        lock_cm = _noop_async_lock() if is_pooled_conn else self.lock
+        async with _ainternal.get_connection(self.conn) as conn:
             if self.pipe:
                 # a connection in pipeline mode can be used concurrently
                 # in multiple threads/coroutines, but only one cursor can be
                 # used at a time
                 try:
-                    async with conn.cursor(binary=True, row_factory=dict_row) as cur:
+                    async with (
+                        lock_cm,
+                        conn.cursor(binary=True, row_factory=dict_row) as cur,
+                    ):
                         yield cur
                 finally:
                     if pipeline:
@@ -387,6 +401,7 @@ class AsyncPostgresSaver(BasePostgresSaver):
                 # thread/coroutine at a time, so we acquire a lock
                 if self.supports_pipeline:
                     async with (
+                        lock_cm,
                         conn.pipeline(),
                         conn.cursor(binary=True, row_factory=dict_row) as cur,
                     ):
@@ -394,12 +409,16 @@ class AsyncPostgresSaver(BasePostgresSaver):
                 else:
                     # Use connection's transaction context manager when pipeline mode not supported
                     async with (
+                        lock_cm,
                         conn.transaction(),
                         conn.cursor(binary=True, row_factory=dict_row) as cur,
                     ):
                         yield cur
             else:
-                async with conn.cursor(binary=True, row_factory=dict_row) as cur:
+                async with (
+                    lock_cm,
+                    conn.cursor(binary=True, row_factory=dict_row) as cur,
+                ):
                     yield cur
 
     async def aget_delta_channel_history(
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
@@ -387,15 +387,15 @@ class AsyncPostgresSaver(BasePostgresSaver):
                 # a connection in pipeline mode can be used concurrently
                 # in multiple threads/coroutines, but only one cursor can be
                 # used at a time
-                try:
-                    async with (
-                        lock_cm,
-                        conn.cursor(binary=True, row_factory=dict_row) as cur,
-                    ):
-                        yield cur
-                finally:
-                    if pipeline:
-                        await self.pipe.sync()
+                async with lock_cm:
+                    try:
+                        async with conn.cursor(
+                            binary=True, row_factory=dict_row
+                        ) as cur:
+                            yield cur
+                    finally:
+                        if pipeline:
+                            await self.pipe.sync()
             elif pipeline:
                 # a connection not in pipeline mode can only be used by one
                 # thread/coroutine at a time, so we acquire a lock
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/shallow.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/shallow.py
@@ -81,6 +81,13 @@ MIGRATIONS = [
     """,
 ]
 
+
+@asynccontextmanager
+async def _noop_async_lock() -> AsyncIterator[None]:
+    """No-op async context manager used when locking is unnecessary."""
+    yield
+
+
 SELECT_SQL = f"""
 select
     thread_id,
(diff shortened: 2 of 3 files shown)


kairosiann:
Pulled locally and verified against current `main`. Sharing numbers and edge-case results too

**Rebase:**
production change (`aio.py`, `shallow.py`) cherry-picks cleanly onto `main`, only `tests/test_async.py` conflicts (trivial both-added with newer `test_delta_channel_chain_reconstruction`). 
Full suite after rebasing: 217 → 220 passed, 0 regressions — +3 are lock tests added by this PR.

**Benchmark**
single async process, `AsyncConnectionPool`, distinct `thread_id` per op, pool pre-warmed, mean of 5 runs, Postgres 16 (local):

| op | pool=1 (control) | pool=10 `main` | pool=10 this PR |
|----|------------------|----------------|-----------------|
| `aget_tuple` | ~parity | 2516 tps | 5244 tps (**2.08×**) |
| `aput`       | ~parity | 2107 tps | 3611 tps (**1.71×**) |

On `main` throughput stays flat as pool grows (`asyncio.Lock` pins a pooled saver at effective concurrency 1). This PR scales it until it hits a local Postgres ceiling. The pool=1 rows are at parity, which is the control — confirms the gap is the lock, not DB/network. Repro: https://gist.github.com/kairosiann/613cbdf7373c56694e954471bb185055

(Minor note tied to [#7259](https://github.com/kairosiann/langgraph/issues/7259): on current main this is already an asyncio.Lock, not a threading.Lock. Doesn't change anything here but flagging just in case it affects how the severity reads.)

**Edge cases that have been checked:**
- Bare `AsyncConnection` (incl. `from_conn_string`): lock still acquired ✅
- `AsyncNullConnectionPool` (subclasses `AsyncConnectionPool`, fresh conn per checkout): correctly treated as pooled, lock dropped, still correct ✅
- Pool + pipeline: impossible by construction (`__init__` raises) so the bypass never collides with pipeline mode ✅
- 100 concurrent `aput` through a pooled saver: all persist and are retrievable ✅

**Clarification**
PR also narrows the lock scope on the *non-pooled* path so that the lock is now acquired after `get_connection` rather than around it. It looks safe, but was that intentional, or a side effect of the restructure? Worth a line in case it matters for the single-connection case.

jitokim:
Thanks for pulling it down and sharing the numbers — the pool=1 control row is exactly the comparison I was hoping someone would run, appreciate it 🙏

> PR also narrows the lock scope on the non-pooled path so that the lock is now acquired after `get_connection` rather than around it. Was that intentional, or a side effect of the restructure?

Intentional, and behavior-preserving on the single-connection path. The reasoning:

`_ainternal.get_connection` is a passthrough for a bare `AsyncConnection` — it just `yield`s the connection with no I/O and no state mutation. Only the pool branch actually checks a connection out. So wrapping `get_connection` inside vs. outside the lock makes no observable difference for the non-pooled case; the lock still covers the real critical section (cursor creation + optional `pipeline()` / `transaction()` entry), which is what needs serialization on a shared connection.

Acquisition order is also preserved — `lock_cm` is the first item in each `async with (...)` tuple, so it's entered before the cursor is opened, matching the previous semantics.

I considered keeping the wider scope to minimize the diff on the non-pooled path, but the narrower form made the pooled/non-pooled symmetry (`lock_cm` either real-lock or noop, applied uniformly at the same site) much easier to read and reason about. Happy to widen it back if you'd prefer the smaller diff.

Also just rebased onto current `main` and resolved the `tests/test_async.py` both-added conflict (kept both `test_delta_channel_chain_reconstruction` and the new lock tests). Local `tests/test_async.py` run on PG16: 21 passed, 3 skipped (delta tests skip via `pytest.importorskip` without `langgraph` core), 0 regressions.

jitokim:
@jkennedyvz @sydney-runkle 
Hi.
Could you take a look when you have a moment? Community has already validated the fix with benchmarks (2× throughput improvement), and the tests pass. Happy to address any feedback!

glaziermag (commented):
As the author of #7259 I reviewed this change closely. It is correct, and I'd like to see it merged — GitHub does not let me record a formal approval without repository access, so leaving the full review here instead.

What I verified:

- The instance lock is bypassed only when `self.conn` is an `AsyncConnectionPool`; single-connection savers keep the existing serialization in both `aio.py` and `shallow.py`. Since the constructor rejects pipeline mode with a pooled connection, the `self.pipe` branch always runs with the real lock.
- The bypass itself is sound: with a pool, each `_cursor()` call checks out its own connection, so the shared `asyncio.Lock` was pure head-of-line blocking. That is exactly the bottleneck measured in #7259 (~200 req/s with the lock vs ~1,295 req/s for the unlocked baseline on the same pool), and the 2.71x improvement reported here is consistent with those numbers.
- The regression tests are the right shape: the exploding lock proves pooled savers never touch the instance lock, and the tracking lock proves single-connection savers still do — covering both `AsyncPostgresSaver` and `AsyncShallowPostgresSaver`.

One non-blocking observation: in the `self.pipe` branch of `aio.py`, `await self.pipe.sync()` in the `finally` now runs after the lock context has exited, whereas previously the outer `async with self.lock` covered it. Pipe mode implies a single connection, so a second coroutine can now open its cursor while the first is still awaiting `sync()`. psycopg permits concurrent use of a pipeline-mode connection with one cursor at a time, so I believe this is safe, but hoisting the lock to also cover the `finally` would preserve the exact prior serialization if the maintainers want zero behavior change on the pipeline path.

This closes out the checkpointer bottleneck cleanly. Thanks for picking it up.

glaziermag:
One correction to my review: the lock-retention (tracking-lock) test covers only `AsyncPostgresSaver` (`test_base_saver_uses_instance_lock`); the exploding-lock tests are what cover both savers. A symmetric `test_shallow_base_saver_uses_instance_lock` would close that small gap — everything else in my review stands.

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -371,13 +377,21 @@ async def _cursor(
...
                 # in multiple threads/coroutines, but only one cursor can be
                 # used at a time
                 try:
-                    async with conn.cursor(binary=True, row_factory=dict_row) as cur:
+                    async with (
+                        lock_cm,
+                        conn.cursor(binary=True, row_factory=dict_row) as cur,
+                    ):
                         yield cur
                 finally:
```
Sanjays2402: in the `self.pipe` branch the lock now exits before the `finally`, so `pipe.sync()` runs unlocked — previously the outer `async with self.lock` covered it. pipe is only ever set with a single AsyncConnection, so that lock is the real one there, and two coroutines can now sync the pipeline while a third is mid-cursor on the same conn. moving `lock_cm` outside the `try` would keep the old shape.

jitokim:
@glaziermag @Sanjays2402 — thanks for the careful review. You both independently flagged that the pooled-lock-bypass refactor had narrowed the lock so that `pipe.sync()` runs outside it. Fixed in the latest commit: in the `self.pipe` branch of `_cursor()`, the `try/finally: await self.pipe.sync()` is now back inside `async with lock_cm:` for both `AsyncPostgresSaver` and `AsyncShallowPostgresSaver`, restoring the pre-PR behavior.

Also added `test_shallow_base_saver_uses_instance_lock` to close the coverage gap @glaziermag pointed out.

Rebased onto latest `main`; full `test_async.py` suite is green locally against Postgres 16 (22 passed, 3 skipped for the langgraph-core-gated delta test).

@jkennedyvz @sydney-runkle — following up here since this addresses the review feedback. Would appreciate a look when you have bandwidth 🙏

spartandingo:
@jkennedyvz @sydney-runkle any movement on this one? we've forked for now but don't love that as a solution long term

PR #7273, opened 2026-03-25
fix(langgraph): preserve message IDs and additional_kwargs in add_messages format="langchain-openai"

### LINKED ISSUE #7272: bug: add_messages(format="langchain-openai") strips message IDs and additional_kwargs
### Checked other resources

### Related Issues / PRs

#7273

```python
from langchain_core.messages import HumanMessage, AIMessage

# 1. Message IDs are lost
left = [HumanMessage(content="hello", id="msg-1")]
right = [AIMessage(content="world", id="msg-2")]

result = add_messages(left, right, format="langchain-openai")

print(result[0].id)  # None — expected "msg-1"
print(result[1].id)  # None — expected "msg-2"

# 2. Custom additional_kwargs are lost
msg = AIMessage(
    id="msg-1",
    content="hello",
    additional_kwargs={"widgets": [{"type": "carousel"}]},
)
result = add_messages([], [msg], format="langchain-openai")
print(result[0].additional_kwargs)  # {} — expected {"widgets": [{"type": "carousel"}]}
```

```shell
No error raised. Message IDs are silently set to None and custom additional_kwargs are silently dropped.
```

### Description

`add_messages(format="langchain-openai")` loses message IDs and custom `additional_kwargs` during the OpenAI format round-trip.

  `_format_messages()` in `langgraph/graph/message.py` calls `convert_to_openai_messages()` without `include_id=True`. The round-trip `BaseMessage → OpenAI dict → BaseMessage` strips all IDs to
   `None`. Additionally, any custom `additional_kwargs` (e.g. UI widgets, metadata) are dropped because `convert_to_openai_messages` only outputs standard OpenAI fields.

  **Impact:**
  1. Every subsequent reducer call assigns new random UUIDs to all existing messages, breaking ID-based deduplication and update logic.
  2. Custom data stored in `additional_kwargs` is silently lost after each reducer invocation.

  **Proposed fix:** Pass `include_id=True` to preserve IDs. Save `additional_kwargs` by message ID before conversion and restore lost fields after.

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 25.3.0: Wed Jan 28 20:56:42 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T8142
> Python Version:  3.14.2 (main, Dec  9 2025, 19:29:30) [Clang 21.1.4 ]

Package Information
-------------------
> langchain_core: 1.2.20
> langsmith: 0.6.4
> langgraph_sdk: 0.3.12

> deepagents
> deepagents-cli

Other Dependencies
------------------
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.6
> packaging: 25.0
> pydantic: 2.12.5
> pytest: 9.0.2
> pyyaml: 6.0.3
> requests: 2.32.5
> requests-toolbelt: 1.0.0
> tenacity: 9.1.2
> typing-extensions: 4.15.0
> uuid-utils: 0.13.0
> zstandard: 0.25.0

Witaly3:
Fixes #7272

  `_format_messages()` in `add_messages(format="langchain-openai")` has two issues during the OpenAI format round-trip:

  1. **Message IDs lost**: `convert_to_openai_messages()` was called without `include_id=True`, stripping all IDs.
  2. **Custom `additional_kwargs` lost**: non-standard fields (e.g. UI widgets, metadata) are dropped by `convert_to_openai_messages` and not restored.

  Pass `include_id=True` to preserve IDs. Save `additional_kwargs` by message ID before conversion and restore lost fields after.

  Also fixes the `test_messages_state_format_openai` skipif condition for `langchain-core >= 1.0`.

  Read the full contributing guidelines: https://docs.langchain.com/oss/python/contributing/overview

  > **All contributions must be in English.** See the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy).

  1. PR title: `fix(langgraph): preserve message IDs and additional_kwargs in add_messages format="langchain-openai"`

  2. PR description: see above.

  3. `make format`, `make lint` and `make test` — all passing.

  4. Verified by running a real graph with `format="langchain-openai"` and confirming both IDs and custom `additional_kwargs` are preserved. All 19 tests in `test_messages_state.py` pass.


libs/langgraph/langgraph/graph/message.py
```diff
@@ -321,7 +321,34 @@ def _format_messages(messages: Sequence[BaseMessage]) -> list[BaseMessage]:
...
+        for message in messages:
+            if message.id and message.additional_kwargs:
+                saved_kwargs[message.id] = message.additional_kwargs
+
+        # include_id was added in langchain-core 0.3.78; fall back for
+        # older versions that don't support it yet.
+        try:
+            openai_msgs = convert_to_openai_messages(messages, include_id=True)
+        except TypeError:
+            openai_msgs = convert_to_openai_messages(messages)
```
Sanjays2402: The compatibility path still loses every message ID: calling without `include_id` is the behavior this PR is fixing, and the restoration below cannot run once the formatted messages have no IDs. Since `langchain-core` 0.3.11 through 0.3.77 is supported here, this path needs to restore IDs explicitly.

Witaly3: @Sanjays2402 Good catch, thanks! After syncing with the latest main, LangGraph now requires langchain-core>=1.4.7,<2, so the 0.3.11–0.3.77 compatibility path is no longer supported or needed. I removed the fallback and now always call convert_to_openai_messages(..., include_id=True). The full libs/langgraph test suite passes.

rkfshakti (commented):
Good fix. The approach of saving `additional_kwargs` keyed by message ID before the OpenAI round-trip and restoring lost fields afterward is sound — `convert_to_openai_messages` does drop non-standard fields, and this is the least invasive way to preserve them.

A couple of observations:

1. The `CORE_MAJOR >= 1` condition in the skip guard is always true for current langchain-core (it's been at major version 0.x, so `CORE_MAJOR >= 1` would be `0 >= 1` = False). This looks like it was meant to future-proof but won't activate until langchain-core bumps to 1.0. The existing `CORE_MINOR == 3 and CORE_PATCH >= 11` guard already covers the current requirement correctly.

2. The test `test_messages_state_format_openai_preserves_additional_kwargs` is well-written and directly targets the bug. Good to see a focused regression test alongside the broader existing test.

The core logic in `_format_messages` is correct and handles the edge case where some kwargs survive the round-trip (only restoring the lost ones via `lost_kwargs`).

Witaly3:
> Good fix. The approach of saving `additional_kwargs` keyed by message ID before the OpenAI round-trip and restoring lost fields afterward is sound — `convert_to_openai_messages` does drop non-standard fields, and this is the least invasive way to preserve them.
> 
> A couple of observations:
> 
>     1. The `CORE_MAJOR >= 1` condition in the skip guard is always true for current langchain-core (it's been at major version 0.x, so `CORE_MAJOR >= 1` would be `0 >= 1` = False). This looks like it was meant to future-proof but won't activate until langchain-core bumps to 1.0. The existing `CORE_MINOR == 3 and CORE_PATCH >= 11` guard already covers the current requirement correctly.
> 
>     2. The test `test_messages_state_format_openai_preserves_additional_kwargs` is well-written and directly targets the bug. Good to see a focused regression test alongside the broader existing test.
> 
> 
> The core logic in `_format_messages` is correct and handles the edge case where some kwargs survive the round-trip (only restoring the lost ones via `lost_kwargs`).

@rkfshakti Thanks for the review! I removed the redundant CORE_MAJOR >= 1 branch and restored the existing skip guard in both tests. The full libs/langgraph test suite passes.

libs/langgraph/langgraph/graph/message.py
```diff
@@ -386,7 +386,29 @@ def _format_messages(messages: Sequence[BaseMessage]) -> list[BaseMessage]:
...
     else:
-        return convert_to_messages(convert_to_openai_messages(messages))
+        # Save additional_kwargs keyed by message ID before the round-trip,
+        # since convert_to_openai_messages drops non-standard fields.
+        saved_kwargs: dict[str, dict] = {}
+        for message in messages:
+            if message.id and message.additional_kwargs:
+                saved_kwargs[message.id] = message.additional_kwargs
+
+        openai_msgs = convert_to_openai_messages(messages, include_id=True)
```
Sanjays2402, line 399: the restore keys off ids that survive the round-trip, but a message whose content is entirely anthropic `tool_result` blocks never comes back at all: `convert_to_openai_messages` leaves `oai_msg["content"]` empty and hits the `if message.content and not oai_msg["content"] and tool_messages:` branch, which extends only `tool_messages` and throws `oai_msg` away — id included. so for that shape the `include_id` doesnt help and `saved_kwargs` just never matches, silently dropping the kwargs. thats the same case your test comment carves out for ToolMessage ids.

rkfshakti:
Thanks for the update — the redundant guard is cleaned up and CI is green. I took another pass and have a few remaining thoughts:

One thing I noticed: messages that have additional_kwargs but no id are silently skipped during the save phase. After the round-trip, there's no way to match them back, so their kwargs are lost. This is probably an edge case (most messages in a langgraph state should have IDs), but worth considering whether a warning or a fallback strategy is needed, or if it's acceptable to document as a known limitation.

Another edge case: if two messages happen to share the same id, the second one overwrites the first in the saved_kwargs dict, and both formatted messages would pick up the second message's kwargs. Again unlikely in practice, but could cause subtle bugs if duplicate IDs ever sneak in.

On the test side, the new test covers AIMessage well. It might be worth adding a case for a message without an id to document the current behavior — that way if someone later adds handling for the no-id case, the test makes the behavior change explicit.

these are blockers from my side — the core logic is correct and the fix solves the reported problem. Just wanted to flag them in case you want to tighten things up.

Witaly3:
> Thanks for the update — the redundant guard is cleaned up and CI is green. I took another pass and have a few remaining thoughts:
> 
> One thing I noticed: messages that have additional_kwargs but no id are silently skipped during the save phase. After the round-trip, there's no way to match them back, so their kwargs are lost. This is probably an edge case (most messages in a langgraph state should have IDs), but worth considering whether a warning or a fallback strategy is needed, or if it's acceptable to document as a known limitation.
> 
> Another edge case: if two messages happen to share the same id, the second one overwrites the first in the saved_kwargs dict, and both formatted messages would pick up the second message's kwargs. Again unlikely in practice, but could cause subtle bugs if duplicate IDs ever sneak in.
> 
> On the test side, the new test covers AIMessage well. It might be worth adding a case for a message without an id to document the current behavior — that way if someone later adds handling for the no-id case, the test makes the behavior change explicit.
> 
> these are blockers from my side — the core logic is correct and the fix solves the reported problem. Just wanted to flag them in case you want to tighten things up.

Thanks for flagging these. I traced the full add_messages path and verified the behavior locally, including through a compiled StateGraph without invoking an LLM.

Messages with id=None are assigned a UUID before merging and before _format_messages is called. Therefore, through the public add_messages(...,format="langchain-openai") API, the message already has an ID when additional_kwargs are saved, and the kwargs survive the round-trip. The loss only occurs when calling the private _format_messages helper directly.

I added a regression test in 8c24d740 that passes an AIMessage without an ID and verifies both that a UUIDv4 is assigned and that its additional_kwargs are preserved.

For duplicate IDs, duplicates in the incoming right messages are collapsed by the reducer before formatting, with the last message winning, so the formatter normally receives unique IDs. The collision you described can occur if duplicate IDs already exist in left, for example through manually constructed or malformed state. That state is already ambiguous before this change: both duplicates remain, while subsequent updates target only the last one. I’d therefore consider that a pre-existing invalid-state edge case rather than something introduced by this PR.

Given that, I’ve kept the production logic unchanged and added the no-ID regression coverage to document the actual public behavior. Thanks again for prompting the deeper check.

rkfshakti:
Thanks for the thorough investigation @Witaly3 — that's a solid analysis. The UUID assignment before `_format_messages` means the no-ID edge case is handled correctly through the public API, and the duplicate-ID scenario is indeed a pre-existing invalid-state concern rather than something this PR introduces. The regression test for the no-ID path is a nice addition. This looks good to me — no further blockers.

libs/langgraph/langgraph/graph/message.py
```diff
@@ -386,7 +386,29 @@ def _format_messages(messages: Sequence[BaseMessage]) -> list[BaseMessage]:
...
+
+        formatted = convert_to_messages(openai_msgs)
```
Sanjays2402, line 401: convert_to_openai_messages drops response_metadata the same way it drops additional_kwargs, and this only restores the latter, so an AIMessage with response_metadata={"model": "..."} still comes back with it emptied (name survives, response_metadata does not). Was leaving that out deliberate, or should the save/restore cover it too?

Witaly3: Not deliberate — thanks for catching this. I confirmed that the OpenAI round-trip drops response_metadata alongside custom additional_kwargs. I updated the save/restore logic to preserve missing response_metadata fields without overwriting values produced by the conversion itself, and added regression coverage for messages both with and without an explicit ID. The full LangGraph test suite passes.

rkfshakti (commented):
@Witaly3 — I've completed my review and have no further blockers. The fix correctly preserves message IDs (via include_id=True) and restores custom additional_kwargs keyed by message ID after the OpenAI round-trip. The edge cases I raised were thoroughly investigated: UUIDs are assigned before _format_messages through the public API (so the no-ID case is handled), and duplicate IDs are a pre-existing invalid-state concern rather than something this PR introduces. The no-ID regression test documents the actual public behavior, and CI is green. This looks good to me — ready for a maintainer's formal approval.

PR #7275, opened 2026-03-25
fix(sdk-py): validate cron schedules for impossible dates before creation

breath57:
## Problem

The Go cron library used by the LangGraph server silently returns a zero-value time (`0001-01-01T00:00:00`) for cron expressions that specify dates which never exist, such as `0 23 31 2 *` (February 31st).

This causes the scheduler's `WHERE next_run_date <= NOW()` condition to **always be true**, triggering the cron job in an **infinite loop every ~5 seconds**. In production, this consumed **$150 in API credits** before being detected.

### Root cause chain

1. Go cron library computes `next_run_date` for impossible date → returns `time.Time{}` (zero value `0001-01-01`)
2. LangGraph scheduler: `WHERE next_run_date <= NOW() AND enabled = true` → zero value always ≤ current time → always triggers
3. After each trigger, Go tries to advance `next_run_date` → still cannot compute → stays at zero value → triggers again in ~5s

## Solution

Add **client-side validation** in the Python SDK that detects impossible day/month combinations **before** the request is sent to the server. The validation:

- Uses only the Python **standard library** (`calendar` module) — no new dependencies
- Checks across a full **4-year leap-year cycle** to correctly allow Feb 29
- Raises `ValueError` with a clear message explaining why the schedule is invalid
- Applied to **both async and sync** `CronClient` methods: `create`, `create_for_thread`, `update`

### Schedules that are now rejected

| Schedule | Reason |
|----------|--------|
| `0 23 31 2 *` | February never has 31 days |
| `0 0 30 2 *` | February never has 30 days |
| `0 0 31 4 *` | April never has 31 days |
| `0 0 31 4,6,9,11 *` | None of these months have 31 days |

### Valid edge cases preserved

| Schedule | Reason |
|----------|--------|
| `0 0 29 2 *` | Feb 29 is valid in leap years |
| `0 0 31 1,2 *` | Jan 31 is valid (even though Feb 31 is not) |
| `0 0 28-31 2 *` | Feb 28 and 29 are valid days |
| `0 0 31 * *` | Wildcard month = some months have 31 days |
| `@daily` | Non-standard format passed through to server |

## Test plan

Added 30+ test cases covering:
- `_parse_cron_field` utility (wildcards, ranges, steps, comma lists)
- `_validate_cron_schedule` (impossible dates, valid dates, edge cases)
- Integration tests for all 6 SDK methods (async/sync × create/create_for_thread/update)
- Update without schedule parameter skips validation

```
======================== 50 passed in 0.06s ========================
```

## Note on server-side fix

This is a client-side defense. The root cause is in the Go cron library / LangGraph scheduler server-side, which ideally should:
1. Reject impossible cron schedules at creation time (return 400)
2. Or detect zero-value `next_run_date` and disable the cron

This SDK fix provides immediate protection for all Python SDK users.

### DIFF
--- a/libs/sdk-py/langgraph_sdk/_async/cron.py
+++ b/libs/sdk-py/langgraph_sdk/_async/cron.py
@@ -8,7 +8,7 @@ from datetime import datetime, tzinfo
 from typing import Any
 
 from langgraph_sdk._async.http import HttpClient
-from langgraph_sdk._shared.utilities import _resolve_timezone
+from langgraph_sdk._shared.utilities import _resolve_timezone, _validate_cron_schedule
 from langgraph_sdk.schema import (
     All,
     Config,
@@ -135,6 +135,8 @@ class CronClient:
             )
             ```
         """
+        _validate_cron_schedule(schedule)
+
         if checkpoint_during is not None:
             warnings.warn(
                 "`checkpoint_during` is deprecated and will be removed in a future version. Use `durability` instead.",
@@ -253,6 +255,8 @@ class CronClient:
             ```
 
         """
+        _validate_cron_schedule(schedule)
+
         if checkpoint_during is not None:
             warnings.warn(
                 "`checkpoint_during` is deprecated and will be removed in a future version. Use `durability` instead.",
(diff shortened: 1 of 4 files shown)


hinthornw:
ack - the fix belongs in the server. Will expand the testing to cover this

breath57:
> ack - the fix belongs in the server. Will expand the testing to cover this

ok

PR #7363, opened 2026-03-31
feat: enhance runtime w/ more execution information

sydney-runkle:
## Summary

Enhances `ExecutionInfo` and `Runtime` to surface richer execution context and introduces `ServerInfo` for LangGraph Server metadata.

### `ExecutionInfo` expansion

Converted from `NamedTuple` to a frozen `dataclass`. Added identity fields populated during task preparation in `_algo.py`:

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `checkpoint_id` | `str` | required | Checkpoint ID for the current execution |
| `checkpoint_ns` | `str` | required | Checkpoint namespace for the current execution |
| `task_id` | `str` | required | Task ID for the current execution |
| `thread_id` | `str \| None` | `None` | Thread ID (None without a checkpointer) |
| `run_id` | `str \| None` | `None` | Run ID (None when not provided in config) |
| `node_attempt` | `int` | `1` | Current node execution attempt number (1-indexed) |
| `node_first_attempt_time` | `float \| None` | `None` | Unix timestamp for when the first attempt started |

`checkpoint_id`, `checkpoint_ns`, and `task_id` are required (no defaults) — they are always populated during task preparation in `_algo.py`. `Runtime.execution_info` is `None` until that point.

### New `ServerInfo` type

Frozen dataclass with `assistant_id: str`, `graph_id: str`, and optional `user: BaseUser | None`. Populated from config metadata (`assistant_id`, `graph_id`) and `configurable["langgraph_auth_user"]` via `_build_server_info()` in `pregel/main.py`.

User detection uses `isinstance(BaseUser)` with a `hasattr("identity")` fallback — needed because the server's `ProxyUser` provides `permissions` via `__getattr__`, which Python's `runtime_checkable` Protocol check doesn't see.

### `Runtime` changes

- `execution_info` is now `ExecutionInfo | None` (default `None`), set during task prep
- Added `server_info: ServerInfo | None` field, wired through `merge()` and `override()`

### `ToolNode` / `ToolRuntime` forwarding

`execution_info` and `server_info` are forwarded from `Runtime` to `ToolRuntime` so tools can access execution and server context.

### New public API surface

```python
from langgraph.runtime import BaseUser, ExecutionInfo, Runtime, ServerInfo, get_runtime
```

## Test plan

- [x] `ExecutionInfo` defaults, patch, and frozen behavior
- [x] Integration tests verifying identity fields are populated in sync and async execution
- [x] Retry tests confirming identity fields persist and `node_attempt` increments
- [x] `ServerInfo` construction, frozen behavior, and `Runtime.merge` precedence
- [x] `server_info` populated from config metadata and `langgraph_auth_user` (including starlette-style proxy user)
- [x] `server_info` is `None` when no server metadata present
- [x] `ToolRuntime` forwarding of `execution_info` and `server_info`

### DIFF
--- a/libs/langgraph/langgraph/runtime.py
+++ b/libs/langgraph/langgraph/runtime.py
@@ -5,7 +5,7 @@ from typing import Any, Generic, cast
 
 from langgraph.store.base import BaseStore
 from langgraph_sdk.auth.types import BaseUser
-from typing_extensions import NotRequired, TypedDict, Unpack
+from typing_extensions import TypedDict, Unpack
 
 from langgraph._internal._constants import CONF, CONFIG_KEY_RUNTIME
 from langgraph.config import get_config
@@ -57,7 +57,7 @@ class ExecutionInfo:
 
 @dataclass(frozen=True, slots=True)
 class ServerInfo:
-    """Metadata injected by LangGraph Server. None when running OSS."""
+    """Metadata injected by LangGraph Server. None when running open-source LangGraph without LangSmith deployments."""
 
     assistant_id: str
     """The assistant ID for the current execution."""
@@ -77,13 +77,13 @@ class ServerInfo:
 def _no_op_stream_writer(_: Any) -> None: ...
 
 
-class _RuntimeOverrides(TypedDict, Generic[ContextT]):
-    context: NotRequired[ContextT]
-    store: NotRequired[BaseStore | None]
-    stream_writer: NotRequired[StreamWriter]
-    previous: NotRequired[Any]
-    execution_info: NotRequired[ExecutionInfo]
-    server_info: NotRequired[ServerInfo | None]
+class _RuntimeOverrides(TypedDict, Generic[ContextT], total=False):
+    context: ContextT
+    store: BaseStore | None
+    stream_writer: StreamWriter
+    previous: Any
+    execution_info: ExecutionInfo
+    server_info: ServerInfo | None
 
 
 @dataclass(**_DC_KWARGS)
@@ -183,7 +183,7 @@ class Runtime(Generic[ContextT]):
     None before task preparation populates it."""
 
     server_info: ServerInfo | None = field(default=None)
-    """Metadata injected by LangGraph Server. None when running OSS."""
+    """Metadata injected by LangGraph Server. None when running open-source LangGraph without LangSmith deployments."""
 
     def merge(self, other: Runtime[ContextT]) -> Runtime[ContextT]:
         """Merge two runtimes together.
--- a/libs/langgraph/tests/test_retry.py
+++ b/libs/langgraph/tests/test_retry.py
@@ -1,6 +1,7 @@
 from unittest.mock import Mock, patch
 
 import pytest
+from langgraph.checkpoint.memory import MemorySaver
 from typing_extensions import TypedDict
 
 from langgraph.graph import START, StateGraph
@@ -427,8 +428,6 @@ def test_execution_info_identity_fields_populated_on_retry():
         jitter=False,
         retry_on=ValueError,
     )
-
-    from langgraph.checkpoint.memory import MemorySaver
 
     graph = (
         StateGraph(State)
--- a/libs/langgraph/tests/test_runtime.py
+++ b/libs/langgraph/tests/test_runtime.py
@@ -589,3 +589,60 @@ def test_server_info_user_from_auth_user() -> None:
     assert si.user is not None
     assert si.user.identity == "proxy-user"
     assert si.user["display_name"] == "Proxy User"
+
+
+def test_execution_info_inherited_by_subgraph() -> None:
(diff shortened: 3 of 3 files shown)


libs/langgraph/langgraph/runtime.py
```diff
@@ -25,7 +48,50 @@ class ExecutionInfo(NamedTuple):
...
+
+    @property
+    def permissions(self) -> Sequence[str]:
+        """The permissions associated with the user."""
+        ...
+
+
+@dataclass(frozen=True)
+class ServerInfo:
+    """Metadata injected by LangGraph Server. None when running OSS."""
```
longquanzheng: > LangGraph Server
ooc - Is that clear enough? 
(thb I hope we can call it langgraph cloud ...but we have changed to lsd)

libs/langgraph/langgraph/runtime.py
```diff
@@ -25,7 +48,40 @@ class ExecutionInfo(NamedTuple):
...
+
+    is_authenticated: bool
+    """Whether the user is authenticated."""
+
+    permissions: Sequence[str]
+    """The permissions associated with the user."""
+
+
+@dataclass(frozen=True)
+class ServerInfo:
```
sydney-runkle, line 59: yeah maybe this should be like `LangSmithServerInfo` or something?

libs/langgraph/langgraph/runtime.py
hinthornw: Haha it used to be called LangGraph cloud a year ago

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -3643,6 +3657,34 @@ def _coerce_checkpoint_values(payload: Any, mapper: Callable[[Any], Any]) -> Non
...
+    authenticated user dict in configurable["langgraph_auth_user"].
+    """
+    metadata = config.get("metadata") or {}
+    configurable = config.get(CONF) or {}
+    assistant_id = metadata.get("assistant_id")
+    graph_id = metadata.get("graph_id")
+
+    # Read authenticated user from configurable (set by LangGraph Server)
+    auth_user_data = configurable.get("langgraph_auth_user")
+    user: User | None = (
```
hinthornw: I think this will actually exclude a lot of users because we use a starlette `BaseUser` behind the scenes, though we do some magic with a custom `DotDict` implementation so that it can feel like a dict if you want.

It SHOULD implement the BaseUser protocol over in sdk-py/auth/types.py, however

libs/langgraph/langgraph/runtime.py
```diff
@@ -25,18 +52,38 @@ class ExecutionInfo(NamedTuple):
...
 def _no_op_stream_writer(_: Any) -> None: ...
 
 
-class _RuntimeOverrides(TypedDict, Generic[ContextT], total=False):
-    context: ContextT
-    store: BaseStore | None
-    stream_writer: StreamWriter
-    previous: Any
-    execution_info: ExecutionInfo
+class _RuntimeOverrides(TypedDict, Generic[ContextT]):
```
eyurtsev: total=False is cleaner maybe than all the NotRequired?

libs/langgraph/tests/test_retry.py
```diff
@@ -393,3 +393,67 @@ def always_failing_node(state):
...
+        return {"foo": "success"}
+
+    retry_policy = RetryPolicy(
+        max_attempts=3,
+        initial_interval=0.01,
+        jitter=False,
+        retry_on=ValueError,
+    )
+
+    from langgraph.checkpoint.memory import MemorySaver
```
eyurtsev: global namespace maybe?

libs/langgraph/langgraph/runtime.py
```diff
@@ -25,18 +52,38 @@ class ExecutionInfo(NamedTuple):
 
     def patch(self, **overrides: Any) -> ExecutionInfo:
         """Return a new execution info object with selected fields replaced."""
-        return self._replace(**overrides)
+        return replace(self, **overrides)
+
+
+@dataclass(frozen=True, slots=True)
+class ServerInfo:
+    """Metadata injected by LangGraph Server. None when running OSS."""
```
eyurtsev: Do you think users will know what `OSS` means?

libs/langgraph/langgraph/runtime.py
```diff
@@ -25,18 +52,38 @@ class ExecutionInfo(NamedTuple):
...
+
+    assistant_id: str
+    """The assistant ID for the current execution."""
+
+    graph_id: str
+    """The graph ID for the current execution."""
+
+    user: BaseUser | None = None
```
eyurtsev, line 68: Do we have a notion of AnonymousUser instead of a None?

PR #7383, opened 2026-04-01
chore: allow passing some metadata only for tracing purposes

hinthornw:
Allows us to put some more information for tracing purposes (e.g., ls_integration) without dumping it into the streaming APIs (good for performance)
Move some other metadata into tracing only since it's not needed in streaming APIs

### DIFF
--- a/libs/langgraph/langgraph/pregel/_messages.py
+++ b/libs/langgraph/langgraph/pregel/_messages.py
@@ -144,7 +144,7 @@ class StreamMessagesHandler(BaseCallbackHandler, _StreamingCallbackHandler):
             stream_metadata = dict(metadata)
             stream_metadata["langgraph_checkpoint_ns"] = checkpoint_ns
             # Preserve backwards-compatible streamed checkpoint metadata shape.
-            stream_metadata.setdefault("checkpoint_ns", checkpoint_ns)
+            stream_metadata["checkpoint_ns"] = checkpoint_ns
             if tags:
                 if filtered_tags := [t for t in tags if not t.startswith("seq:step")]:
                     stream_metadata["tags"] = filtered_tags


libs/langgraph/langgraph/pregel/_messages.py
```diff
@@ -132,15 +132,23 @@ def on_chat_model_start(
...
+                if NS_END in task_checkpoint_ns
+                else task_checkpoint_ns
+            )
+            ns = tuple(task_checkpoint_ns.split(NS_SEP))[:-1]
             if not self.subgraphs and len(ns) > 0 and ns != self.parent_ns:
                 return
+            stream_metadata = dict(metadata)
+            stream_metadata["langgraph_checkpoint_ns"] = checkpoint_ns
+            # Preserve backwards-compatible streamed checkpoint metadata shape.
+            stream_metadata.setdefault("checkpoint_ns", checkpoint_ns)
```
hntrl: i'm lost at the part where this is important for metadata copying, is this just because we're changing how checkpoint_ns comes across?

eyurtsev: langgraph was relying on logic in core here for checkpoint_ns so it's a bit convoluted... Not sure if there's a cleaner place to put this code

libs/langgraph/pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
     'Programming Language :: Python :: 3.13',
 ]
 dependencies = [
-    "langchain-core>=0.1",
+    "langchain-core==1.3.0a2",
```
hinthornw, line 27: we still keeping this?

eyurtsev: i was thinking that we have to bump since changes in core do affect what happens in langgraph

langgraph already depends on pretty recent core, b/c langgraph requires langgraph-prebuilt and langgraph-prebuilt depends on a recent version of core, so I think the `langchain-core>=0.1` hasn't been true in at least a year probably

eyurtsev: I still need an alpha release of this, and then can bump into non alpha releases

PR #7394, opened 2026-04-02
feat(cli): support studio deploy

phvash:
# Description 

Supports deploying to langsmith from langgraph studio 

Adds the following to the langgraph deploy command:
- json event output 
- non-interactive mode


libs/cli/langgraph_cli/deploy.py
```diff
@@ -626,10 +808,26 @@ def _upload_to_gcs(signed_url: str, file_path: str, file_size: int) -> None:
     import urllib.error
     import urllib.request
 
+    em = _get_emitter()
+    size_mb = file_size / 1_048_576
```
hari-dhanushkodi: add constant?

libs/cli/langgraph_cli/deploy.py
```diff
@@ -115,6 +115,153 @@ class BuildResult:
...
+            if status == "succeeded":
+                click.secho("   Deployment successful!", fg="green")
+                if url:
+                    click.secho(f"   URL: {url}", fg="green")
+            elif status == "failed":
+                click.secho("   Deployment failed", fg="red")
+            elif status == "timed_out":
+                click.secho("   Timed out waiting for deployment.", fg="yellow")
+                if url:
+                    click.secho(f"   Check status at: {url}", fg="yellow")
```
hari-dhanushkodi: should this be status_url?

libs/cli/langgraph_cli/deploy.py
```diff
@@ -1273,19 +1506,25 @@ def _deploy_cmd(
...
+    if name and not deployment_id:
+        name = normalize_name(name)
 
     secrets = _secrets_from_env(_env_without_deployment_name(env_vars))
 
     use_remote_build, local_build_error = _resolve_build_mode(remote_build_flag)
     if use_remote_build and remote_build_flag is None and local_build_error:
-        click.secho(f"{local_build_error}\nUsing remote build instead.", fg="yellow")
-        click.echo()
+        em.warn(f"{local_build_error}\nUsing remote build instead.")
```
hari-dhanushkodi: how come sometimes we use spaces for `warn` and othertimes we don't?
eg:
```
em.warn("   Re-run with --verbose to see full build output.")
```

libs/cli/langgraph_cli/deploy.py
```diff
@@ -1273,19 +1506,25 @@ def _deploy_cmd(
...
-            click.echo(f"Saved deployment name to {env_path}")
+        default_name = normalize_name(pathlib.Path.cwd().name)
+        if no_input:
+            name = default_name
+        else:
+            name = click.prompt("Deployment name", default=default_name)
+            env_path = _resolve_env_path(config_json, config)
+            if env_path is not None:
+                set_key(str(env_path), _DEPLOYMENT_NAME_ENV, name)
+                click.echo(f"Saved deployment name to {env_path}")
```
hari-dhanushkodi: also want to clarify - now that we have the `em`, should everything be going through the emitter instead of using `click`?

hari-dhanushkodi (dismissed):

libs/cli/tests/unit_tests/test_deploy_helpers.py
```diff
@@ -271,3 +276,255 @@
...
+        em.log("my line")
+        captured = capsys.readouterr()
+        assert "my line" in captured.out
+
+    def test_result_succeeded_text(self, capsys):
+        em = _Emitter(json_mode=False)
+        em.result("succeeded", deployment_id="d1", url="https://app.test")
+        captured = capsys.readouterr()
+        assert "successful" in captured.out.lower()
+        assert "https://app.test" in captured.out
```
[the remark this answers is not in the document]
phvash: Marking this as a false positive. The flagged code is a test assertion against captured stdout (`test_result_succeeded_text`), not URL validation. Additionally, the current form already does a full-line list-membership check after `line.strip()`:

```python
lines = [line.strip() for line in captured.out.splitlines() if line.strip()]
assert "URL: https://app.test" in lines
```

That's membership in a list of exact strings, not substring matching, so it doesn't have the semantics CodeQL is warning about. Planning to dismiss the scanning alert as "used in tests / won't fix".

libs/cli/langgraph_cli/deploy.py
phvash: Done — added a module-level `_BYTES_PER_MIB = 1_048_576` constant near the other upload constants and replaced all three occurrences (this line, the `_ProgressReader` size display, and the `Archive created` info message).

libs/cli/langgraph_cli/deploy.py
phvash: Good catch, yes!

libs/cli/langgraph_cli/deploy.py
phvash: Yeah, that can be confusing. We use `em.warn` in two scenarios. Within a step, as a warning (indented under the step in this case, for visual rendering'. But we also use `em.warn` for top level warnings like the beta message at the start etc. Those don't fall within a step like "Building/Deploying" etc and don't have to be indented for visual alignment. I have have separated out the top level usage to em.notes, so its cleaner/clearer when to indent and when not to indent. Similar to how em.info/error already work. and the manual prefixes have been removed from the three callers.

libs/cli/langgraph_cli/deploy.py
phvash: Agreed for the deploy flow. Routed the remaining deploy-path `click.*` calls through the emitter. Kept interactive prompts and the other sub-commands on `click` since they don't run under `--json`.

hari-dhanushkodi (approved):

PR #7429, opened 2026-04-06
feat(langgraph): add graph lifecycle callback handlers

hntrl:
## Summary

This change adds first-class graph lifecycle callbacks to LangGraph so interrupt and resume transitions can be observed without overloading the existing LangChain custom event system. It introduces a dedicated graph callback manager and wires lifecycle emission through Pregel execution in both sync and async paths.

## Changes

- **`libs/langgraph/langgraph/callbacks.py`**: Adds `GraphCallbackHandler` and `GraphCallbackManager` (built on LangChain base callback classes), plus config plumbing via `graph_callbacks` and `get_graph_callback_manager_for_config`.
- **`libs/langgraph/langgraph/pregel/_loop.py`**: Introduces `GraphLifecycleEvent` and records lifecycle transitions (`resume`, `interrupt`) into an internal FIFO queue with `shift_graph_lifecycle_event()`.
- **`libs/langgraph/langgraph/pregel/main.py`**: Resolves graph callback manager from config and drains lifecycle events while loop execution progresses, dispatching `on_resume` and `on_interrupt` consistently in sync and async runtimes.
- **`libs/langgraph/tests/test_graph_callbacks.py`**: Adds sync and async coverage verifying lifecycle callbacks fire correctly and remain distinct from LangChain `on_custom_event` handlers.


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -887,6 +925,9 @@ def _suppress_interrupt(
         # suppress interrupt
         suppress = isinstance(exc_value, GraphInterrupt) and not self.is_nested
         if suppress:
+            interrupt = cast(GraphInterrupt, exc_value)
```
sydney-runkle: we shouldn't have to do this

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -252,6 +265,7 @@ def __init__(
         self.retry_policy = retry_policy
         self.cache_policy = cache_policy
         self.durability = durability
+        self._graph_lifecycle_events = []
```
rx5ad: nit: should we make it a queue if we have to pop left when shifting? granted list is small wouldn't have a big impact

libs/langgraph/langgraph/callbacks.py
```diff
@@ -0,0 +1,244 @@
...
+        interrupts: Sequence[Interrupt],
+        *,
+        status: str,
+        checkpoint_id: str,
+        checkpoint_ns: tuple[str, ...],
+        is_nested: bool,
+    ) -> None:
+        for handler in self.handlers:
+            graph_handler = cast(GraphCallbackHandler, handler)
+            result = graph_handler.on_interrupt(
```
rx5ad: If a handler throws an error, this would break the loop. This might be unexpected behaviour. Should we catch the error?

libs/langgraph/langgraph/callbacks.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def on_resume(
+        self,
+        *,
+        status: str,
+        checkpoint_id: str,
+        checkpoint_ns: tuple[str, ...],
+        is_nested: bool,
+    ) -> None:
+        for handler in self.handlers:
```
rx5ad: Same here, one broken handler would impact all other handlers

libs/langgraph/langgraph/callbacks.py
rx5ad: See how langchain_core handles it https://github.com/langchain-ai/langchain/blob/d7575ffac9d95131e5f1a445e245df2eb52cef97/libs/core/langchain_core/callbacks/manager.py#L552

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -130,6 +132,16 @@
 WritesT = Sequence[tuple[str, Any]]
 
 
+@dataclass(slots=True, frozen=True)
+class GraphLifecycleEvent:
+    kind: Literal["resume", "interrupt"]
+    status: str
```
eyurtsev: Could we document this data structure better? (+ this is likely the wrong file for it?)

libs/langgraph/langgraph/callbacks.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+            if inspect.isawaitable(result):
+                await result
+
+    @staticmethod
+    def _drain_coroutine(coro: Coroutine[Any, Any, Any]) -> None:
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            asyncio.run(coro)
```
eyurtsev: (nit) might be cheaper to maintain an event loop running inside a daemon thread and scheduling on it when there's no event loop already running in the given process

libs/langgraph/langgraph/callbacks.py
```diff
@@ -0,0 +1,244 @@
...
+        return cls(handlers, run_id=run_id)
+
+    def on_interrupt(
+        self,
+        interrupts: Sequence[Interrupt],
+        *,
+        status: str,
+        checkpoint_id: str,
+        checkpoint_ns: tuple[str, ...],
+        is_nested: bool,
```
eyurtsev: What is `is_nested`?

libs/langgraph/langgraph/callbacks.py
```diff
@@ -0,0 +1,244 @@
...
+            graph_handler = cast(GraphCallbackHandler, handler)
+            result = graph_handler.on_interrupt(
+                interrupts,
+                run_id=self.run_id,
+                status=status,
+                checkpoint_id=checkpoint_id,
+                checkpoint_ns=checkpoint_ns,
+                is_nested=is_nested,
+            )
+            if inspect.iscoroutine(result):
```
eyurtsev: Why are we handling the result here?

sydney-runkle (commented):
a few notes
* do we need AsyncGraphCallbackHandler?
* do we mean to store things in config["configurable"]["graph_callbacks"] or `config["graph_callbacks"] ?
* can we make shift_graph_lifecycle_event -> _pop_lifecycle_event`

libs/langgraph/langgraph/callbacks.py
hinthornw: yes definitely

libs/langgraph/langgraph/callbacks.py
```diff
@@ -0,0 +1,244 @@
...
+    async def aon_resume(
+        self,
+        *,
+        status: str,
+        checkpoint_id: str,
+        checkpoint_ns: tuple[str, ...],
+        is_nested: bool,
+    ) -> None:
+        for handler in self.handlers:
+            graph_handler = cast(GraphCallbackHandler, handler)
```
hinthornw: Would this always be the case?

libs/langgraph/langgraph/callbacks.py
hinthornw: I don't think we should use  asyncio.run at all

libs/langgraph/langgraph/pregel/_loop.py
hntrl: my type checker was complaining since I think `isinstance` isn't a direct subject of the if statement

hntrl: fixed

libs/langgraph/langgraph/pregel/_loop.py
hntrl: done, using `deque` now (tell me if thats the right answer)

libs/langgraph/langgraph/callbacks.py
hntrl: nice catch, using `handle_event` from core now

libs/langgraph/langgraph/callbacks.py
hntrl: no it's not, made this a little more fool proof in GraphCallbackManager.configure

libs/langgraph/langgraph/callbacks.py
hntrl: fixed

eyurtsev:
They should go in config[callbacks] same key as regular callbacks

PR #7438, opened 2026-04-08
chore(cli): add validate command

hinthornw:
add validate command


daeho-ro:
It seems that the change breaks shell completion installation.

PR #7519, opened 2026-04-15
feat(langgraph): add streaming transformer infrastructure and tests 

nick-hollon-lc:
## Summary

- `StreamingHandler` wraps a compiled graph and returns a run stream with typed projections (`run.messages`, `run.values`, `run.output`) driven by a transformer pipeline.
- Extension point: `StreamTransformer` observes protocol events and builds custom projections surfaced via `run.extensions`. Sync transformers run under both `stream()` and `astream()`; an async lane (`aprocess` / `afinalize` / `afail`, plus a `schedule()` helper for decoupled async work) is available under `astream()`.
- Built-in `ValuesTransformer` and `MessagesTransformer` cover `run.values` and `run.messages` out of the box.
- `EventLog` / `StreamChannel` accept an optional `maxlen` for memory bounds; `handler.stream()` / `astream()` expose `max_events` as a run-wide default. Lagging cursors on a bounded log raise `BufferOverflowError`.

## Usage

```python
from langgraph.stream import StreamingHandler

handler = StreamingHandler(graph)
run = handler.stream({"messages": [HumanMessage(content="hi")]})

for chunk, _metadata in run.messages:
    if chunk.content:
        print(chunk.content, end="", flush=True)

final = run.output
```

Async has the same shape:

```python
run = await handler.astream(input_data)
async for chunk, _ in run.messages:
    ...
output = await run.output
```

## Custom transformers

Transformers observe protocol events and build typed projections that consumers read from `run.extensions`. Example: track first-token latency and throughput, surfaced through a `StreamChannel`.

```python
import time
from typing import Any

from langgraph.stream import StreamChannel, StreamTransformer
from langgraph.stream._types import ProtocolEvent


@dataclass
class TokenSnapshot:
    token_index: int
    elapsed: float
    tokens_per_sec: float


class TokenMetricsTransformer(StreamTransformer):
    def __init__(self) -> None:
        self._channel: StreamChannel[TokenSnapshot] = StreamChannel("token_metrics")
        self._t0: float | None = None
        self._count = 0

    def init(self) -> dict[str, Any]:
        return {"token_metrics": self._channel}

    def process(self, event: ProtocolEvent) -> bool:
        if event["method"] != "messages" or event["params"]["namespace"]:
            return True
        chunk, _metadata = event["params"]["data"]
        if not getattr(chunk, "content", None):
            return True
        now = time.time()
        self._t0 = self._t0 or now
        self._count += 1
        elapsed = now - self._t0
        self._channel.push(
            TokenSnapshot(
                token_index=self._count,
                elapsed=elapsed,
                tokens_per_sec=self._count / elapsed if elapsed > 0 else 0,
            )
        )
        return True
```

Attach and consume alongside the built-in projections — the consumption loop stays unchanged:

```python
run = handler.stream(input_data, transformers=[TokenMetricsTransformer()])

for chunk, _ in run.messages:
    print(chunk.content, end="", flush=True)

for snap in run.extensions["token_metrics"]:
    print(snap)
```


examples/streaming-v2/streaming_comparison.py
```diff
@@ -0,0 +1,499 @@
+"""Compare token-level LLM streaming: v1 graph.stream() vs v2 StreamingHandler.
```
ccurme: This seems out of place in the examples folder. Better suited to docs? Or something condensed in PR description?

examples/streaming-v2/streaming_comparison.py
```diff
@@ -0,0 +1,499 @@
...
+    messages: Annotated[list[BaseMessage], add_messages]
+
+
+_llm = None
+
+
+def _get_llm() -> ChatOpenAI:
+    global _llm
+    if _llm is None:
+        _llm = ChatOpenAI(model="gpt-4o-mini", streaming=True)
```
ccurme: Should verify that streaming works without `streaming=True`

also nitpick but gpt-4o-mini is an old model

libs/langgraph/langgraph/stream/_event_log.py
```diff
@@ -0,0 +1,148 @@
...
+        with self._lock:
+            self._error = err
+            self._closed = True
+        self._notify()
+
+    def _notify(self) -> None:
+        """Wake waiting consumers. Overridden by subclasses."""
+
+
+class EventLog(_EventLogBase[T]):
```
ccurme: We discussed using built-in abstractions for in-process, do we like EventLog in this context still?

ccurme: if EventLog gets big, can we / do we flush it?

examples/streaming-v2/streaming_comparison.py
```diff
@@ -0,0 +1,499 @@
...
+    separator("V2 SYNC: StreamingHandler.stream() -> run.messages")
+
+    graph = build_graph()
+    handler = StreamingHandler(graph)
+    input_data = {"messages": [HumanMessage(content=question)]}
+
+    run = handler.stream(input_data)
+
+    # run.messages yields (chunk, metadata) tuples as the LLM produces them
+    for chunk, metadata in run.messages:
```
ccurme: how does parallel consumption look? if we want to get tokens but also state updates relative to each other during a stream

libs/langgraph/langgraph/stream/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""Streaming infrastructure for LangGraph.
+
+Provides a ``StreamingHandler`` that wraps a compiled graph and exposes
```
ccurme: (here and elsewhere) check intended formatting for docstrings, I don't think we use sphinx anymore?

examples/streaming-v2/streaming_comparison.py
nick-hollon-lc: good call yeah can go in PR description!

libs/langgraph/langgraph/stream/_event_log.py
nick-hollon-lc: not yet but we could add some buffering to it

examples/streaming-v2/streaming_comparison.py
nick-hollon-lc: don't have this yet but maybe something like:

```
# All root-namespace events, interleaved                                                                              
for method, data in run.events():                                                                                     
    if method == "values":                                                                                            
        ...                                                                                                           
    elif method == "messages":                                                                                        
        ...                                                                                                           
                                                                     
# Or filtered to just the modes you care about
for method, data in run.events("values", "messages"):
    ...                                                                                                               
   
```

libs/langgraph/langgraph/stream/_event_log.py
nick-hollon-lc: done

PR #7547, opened 2026-04-17
feat(langgraph): `DeltaChannel` — O(N) incremental checkpoint storage

sydney-runkle:
# feat(channels): DeltaChannel — O(N) incremental checkpoint storage

## The problem

LangGraph checkpoints store the **full accumulated value** of every channel on every step. For a `messages` channel backed by `add_messages`, that means each checkpoint blob contains the entire conversation history up to that point.

Storage cost grows **O(N²)** in the number of turns:

| Step | Checkpoint blob |
|------|----------------|
| 1    | [msg_1] |
| 2    | [msg_1, msg_2] |
| N    | [msg_1, ..., msg_N] |

For long-running agentic conversations this is untenable. Storage cost grows quadratically with conversation length — doubling accumulated message data quadruples checkpoint size. At 100K tokens of conversation data, a single thread accumulates **~250 MB**; with larger messages (file attachments, long tool outputs) costs scale even faster. See the benchmark table below for measured projections across different message volumes.

## The fix: `DeltaChannel`

`DeltaChannel` is an opt-in wrapper around any binary reducer that stores only the **per-step writes** (the delta) instead of the full accumulated list. Storage scales **O(N)**. At read time the saver follows the delta chain and replays writes through the operator to reconstruct the full list.

```python

class State(TypedDict):
    # Before: O(N²) storage

    # After: O(N) storage
    messages: Annotated[list[AnyMessage], DeltaChannel(add_messages)]
```

An optional `snapshot_every` parameter caps read-time chain depth by periodically materializing a full-list snapshot:

```python
# Snapshot every 50 steps — read traverses at most 50 deltas
messages: Annotated[list[AnyMessage], DeltaChannel(add_messages, snapshot_every=50)]
```

## Benchmarks

Simulated with realistic paragraph-length messages (~100 tokens each, ~400 chars). Each turn = one human + one AI message (~200 tokens total).

### Storage (InMemorySaver)

| turns | ctx | add_msgs | delta | delta+snap | savings |
|------:|----:|---------:|------:|-----------:|--------:|
| 50 | ~10K tok | 2.6 MB | 71.1 KB | 96.3 KB | 27x |
| 100 | ~20K tok | 10.2 MB | 142.2 KB | 294.7 KB | 34x |
| 200 | ~40K tok | 40.5 MB | 284.8 KB | 1.0 MB | 40x |
| 500 | ~100K tok | 252.6 MB | 712.4 KB | 5.6 MB | 45x |

The ratio grows with N because `add_messages` is O(N²) while `DeltaChannel` is O(N).

**Extrapolated to a 1M-token conversation** (e.g. ~100–500 turns with large messages or file attachments):

| Channel | Est. storage | vs. add_messages |
|---------|-------------|-----------------|
| `add_messages` (current default) | ~25 GB | — |
| `DeltaChannel(add_messages)` | ~7.5 MB | **~3,300x less** |
| `DeltaChannel(add_messages, snapshot_every=50)` | ~56 MB | ~450x less |

### Read latency (InMemorySaver, avg of 5 `get_state` calls)

| turns | ctx | add_msgs | delta | delta+snap |
|------:|----:|---------:|------:|-----------:|
| 50 | ~10K tok | 0.4ms | 1.7ms | 1.4ms |
| 100 | ~20K tok | 0.7ms | 5.4ms | 2.5ms |
| 200 | ~40K tok | 1.5ms | 19.5ms | 5.1ms |
| 500 | ~100K tok | 3.6ms | 177.0ms | 11.1ms |

`delta` (unbounded chain) degrades to 177ms at 500 turns — this is why `snapshot_every` is recommended. `delta+snap` stays at ~11ms.

### Read latency (PostgresSaver, avg of 5 `get_state` calls)

| turns | ctx | add_msgs | delta | delta+snap |
|------:|----:|---------:|------:|-----------:|
| 10 | ~2K tok | 0.6ms | 1.0ms | 1.3ms |
| 20 | ~4K tok | 0.6ms | 2.1ms | 4.1ms |
| 50 | ~10K tok | 1.2ms | 7.3ms | 5.9ms |
| 100 | ~20K tok | 1.9ms | 31.9ms | 9.1ms |

`delta` (unbounded chain) hits 31.9ms at 100 turns. `delta+snap` stays under 10ms across all scales.

## How it works

Each checkpoint step, `DeltaChannel` stores only the **writes made that step** (`DeltaValue`) rather than the full accumulated list. On read, the saver follows the chain of deltas back to the nearest snapshot and replays them through the reducer to reconstruct the current value.

**Write:** `DeltaChannel.checkpoint()` emits a `DeltaValue(delta=<step_writes>)`. Every `snapshot_every` steps it emits a full-list snapshot instead, which caps read-time chain depth.

**Read:** The saver assembles a `DeltaChainValue(base=<snapshot>, deltas=[...])` and passes it to `DeltaChannel.from_checkpoint()`, which replays the deltas through the operator.

**Saver implementations:**
- `InMemorySaver` — pure in-memory dict traversal, no extra I/O
- `PostgresSaver` — single recursive CTE, one round-trip regardless of chain depth
- Other savers — fallback: one `get_tuple()` call per chain link (use `snapshot_every` to cap depth)

## Changes

**`libs/checkpoint`**
- `base/__init__.py` — add `DeltaValue`, `DeltaChainValue`; add `get_delta_chain` / `aget_delta_chain` to `BaseCheckpointSaver` (default: `NotImplemented`)
- `serde/jsonplus.py` — `"delta"` type tag in `dumps_typed` / `loads_typed`; register `DeltaValue` in msgpack allowlist
- `checkpoint/memory/__init__.py` — `get_delta_chain` via dict traversal; `aget_delta_chain` delegates to sync

**`libs/langgraph`**
- `channels/base.py` — no-op `after_checkpoint(version)` hook on `BaseChannel`
- `channels/delta.py` — `DeltaChannel` implementation
- `channels/__init__.py` — export `DeltaChannel`
- `pregel/_checkpoint.py` — `_assemble_delta_channels` / `_aassemble_delta_channels`; call `after_checkpoint` in `channels_from_checkpoint`
- `pregel/_loop.py` — call `after_checkpoint` after each checkpointing step

**`libs/checkpoint-postgres`**
- `postgres/__init__.py` — `get_delta_chain` via recursive CTE (sync)
- `postgres/aio.py` — `aget_delta_chain` via recursive CTE (async)

## Open questions

**Should we add a compile-time capability check?**

Currently, misconfiguring `DeltaChannel` with an unsupported saver (e.g. `SqliteSaver`) only errors at runtime on the first reload — after a step has already been written. We could restore an early warning at `compile()` time by checking whether the saver implements chain assembly (e.g. a protocol/ABC check on `_load_blobs` or a capability method), rather than relying on a plain boolean flag. The boolean `supports_delta_channels` was removed because it was a manual convention that third-party savers had to know to set. A protocol-based check would be self-describing. Worth doing, or is the runtime raise sufficient?

## Backwards compatibility

| Scenario | Behaviour |
|----------|-----------|
| Existing graph using `add_messages` | Unaffected — no code or schema changes |
| `DeltaChannel` loading an old full-list checkpoint blob | Handled via backwards-compat path in `from_checkpoint` |
| `DeltaChannel` with `InMemorySaver` or `PostgresSaver` | Fully supported |
| `DeltaChannel` with `SqliteSaver` | Raises `ValueError` — SQLite stores blobs inline, no per-channel blob table; deferred to follow-up |
| Time-travel to a past checkpoint | Chain traversal uses the version at that checkpoint — correct by construction |
| `Overwrite` value | Resets chain: next blob has `prev_checkpoint_id=None`; reconstruction starts fresh |

## Test plan

- [x] `DeltaChannel` unit tests: `update` → `checkpoint` → `after_checkpoint` lifecycle, `from_checkpoint` chain replay, backwards-compat with plain list, `Overwrite` resets chain
- [x] `InMemorySaver` `get_delta_chain`: assembles `DeltaChainValue` from delta blobs, handles mixed old/new blobs
- [x] Serde round-trip: `DeltaValue` serializes as `"delta"` and deserializes correctly; registered in msgpack allowlist
- [x] End-to-end graph tests: multi-turn conversations accumulate correctly, time-travel reconstructs correct partial history
- [x] `PostgresSaver` recursive CTE chain reconstruction (sync + async)
- [x] `_assemble_delta_channels` fallback path (savers returning `NotImplemented`)
- [x] Storage benchmark: `DeltaChannel` uses strictly less storage than `add_messages` at 100 and 200 turns
- [x] Read latency benchmark: Postgres `get_delta_chain` vs full-list `get_tuple` at 10, 20, 50 steps


libs/langgraph/langgraph/pregel/_checkpoint.py
```diff
@@ -12,6 +20,171 @@
...
+                    "DeltaChannel chain cycle at checkpoint %r for channel %r; breaking",
+                    prev_id,
+                    channel,
+                )
+                break
+            visited.add(prev_id)
+
+            blob = await checkpointer.aget_channel_blob(
+                thread_id, checkpoint_ns, prev_id, channel
+            )
```
pjrule: this is the heart of the n+1 problem we discussed in Slack, right?

I think we'll want some kind of bulk reconstruction API instead of/in addition to `aget_channel_blob()`.

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,200 @@
...
+        "_pending",
+        "_base_version",
+        "_overwritten",
+        "_steps_since_snapshot",
+    )
+
+    def __init__(
+        self,
+        operator: Callable[[list[Value], Any], list[Value]],
+        typ: type = list,
```
sydney-runkle: should just be inferred from associated type i think

libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -185,15 +187,106 @@ def _migrate_pending_sends(
...
+            elif type_tag != "empty":
+                result[channel] = self.serde.loads_typed((type_tag, v))
+        if delta_channels and cur is not None and checkpoint_id:
+            for channel in delta_channels:
+                result[channel] = self._load_delta_chain(
+                    thread_id, checkpoint_ns, checkpoint_id, channel, cur
+                )
+        return result
+
+    def _load_delta_chain(
```
nfcampos, line 216: shouldn't this be in sync postgres saver, rather than base?

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -256,6 +245,8 @@ def dumps_typed(self, obj: Any) -> tuple[str, bytes]:
             return "bytes", obj
         elif isinstance(obj, bytearray):
             return "bytearray", obj
+        elif _is_delta_value(obj):
+            return "delta", _msgpack_enc({"d": obj.delta})
```
nfcampos, line 249: why the wrapper dict?

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -278,6 +269,13 @@ def loads_typed(self, data: tuple[str, bytes]) -> Any:
             return ormsgpack.unpackb(
                 data_, ext_hook=self._unpack_ext_hook, option=ormsgpack.OPT_NON_STR_KEYS
             )
+        elif type_ == "delta":
```
nfcampos, line 272: this makes delta implicitly msgpack, might be better to instead do `delta+msgpack` ?

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,200 @@
...
+            new._steps_since_snapshot = len(checkpoint.deltas)
+        elif isinstance(checkpoint, DeltaValue):
+            raise ValueError(
+                f"Channel '{self.key}' uses DeltaChannel but the checkpointer "
+                "does not support incremental channel storage. "
+                "Use InMemorySaver or PostgresSaver, or remove DeltaChannel from your schema."
+            )
+        else:
+            # Backwards compat: plain list from old BinaryOperatorAggregate checkpoint.
+            new.value = list(checkpoint)
```
nfcampos, line 129: if this is a new channel type why does it need backwards compatibility code?

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -391,6 +393,81 @@ async def _cursor(
...
+        thread_id: str,
+        checkpoint_ns: str,
+        checkpoint_id: str,
+        channel: str,
+        cur: Any,
+    ) -> DeltaChainValue:
+        """Fetch the full delta chain for a channel in one recursive CTE query (async)."""
+        await cur.execute(
+            """
+            WITH RECURSIVE chain AS (
```
nfcampos, line 407: is a recursive query the best here? how about a plain 'select where' query, probably always less expensive than this, even if returning more data

libs/langgraph/langgraph/channels/delta.py
sydney-runkle: i guess it's more forwards compat -- if channel type changes we can still rehydrate an old channel and then serialize deltas from there

sydney-runkle:
Closing in favor of https://github.com/langchain-ai/langgraph/pull/7586 where the majority of above comments have been addressed (it's still WIP though)

PR #7571, opened 2026-04-21
perf: cache source+AST analysis in get_function_nonlocals and _get_all_injected_args

sydney-runkle:
## TLDR

Two hotspots in `create_agent` were doing expensive repeated work on every call with no caching. This PR fixes both.

### `pregel/_utils.py` — `get_function_nonlocals`

`get_function_nonlocals` was calling `inspect.getsource` + `ast.parse` + a full AST visitor walk on every graph `compile()` call to find which closure variables a node function references. For a typical `create_agent` call this ran on the same `model_node`/`tool_node` closures every time — accounting for ~54% of total `create_agent` wall time in profiling.

**Fix:** Extract `_get_nonlocal_names(code: types.CodeType)` as a separate `@lru_cache`-decorated function keyed on the code object. The source-parsing step is now paid exactly once per unique function definition per process. `get_function_nonlocals` still calls `inspect.getclosurevars` live on every invocation (necessary — reads the actual current closure values). Added a fast-path early exit for functions with no free variables (`co_freevars` empty).

**Measured speedup:** ~8x on the second-and-subsequent `create_agent` calls in the same process (the common case). First call pays the cache-miss cost (~0.085ms per node function) once.

### `prebuilt/tool_node.py` — `_get_all_injected_args`

`_get_all_injected_args` ran `get_input_schema()`, `get_type_hints`, and `inspect.signature` on every tool for every `ToolNode` construction. When the same tool objects are passed to repeated `create_agent` calls, this was fully repeated each time.

**Fix:** Module-level `_INJECTED_ARGS_CACHE` dict keyed by `id(tool)`. Each entry stores `(tool, result)` — the strong reference prevents GC from reusing the id for a different object, and the `entry[0] is tool` identity check guards against collisions. `@lru_cache` couldn't be used directly because `BaseTool` (Pydantic v2 non-frozen model) is unhashable.

## Areas to review carefully

- `_get_nonlocal_names` caches by `func.__code__`. Two different closures compiled from the same source (same code object) return the same nonlocal *names*, which is correct — the names are structural, the values are read live via `getclosurevars`.
- `_INJECTED_ARGS_CACHE` strong-refs tools. This bounds memory to the number of unique `BaseTool` instances seen in the process lifetime, which is typically small and bounded.

> This PR was developed with AI assistance.


hinthornw:
haven't read, but be careful about memory leaks please

PR #7586, opened 2026-04-22
feat(langgraph): `DeltaChannel`: store sentinel in blobs, reconstruct from checkpoint_writes

sydney-runkle:
# DeltaChannel: stop checkpointing the same data over and over

## The problem

LangGraph serializes full accumulated state into a checkpoint blob at every step. For a 100-turn conversation, that's 100 increasingly-large copies of the same message list — even though each step only added one message. At 500 turns, `add_messages` consumes **219 MB** just for blobs.

## What this PR does

Introduces `DeltaChannel`: a reducer channel that stores only a zero-byte sentinel in checkpoint blobs instead of the full accumulated value. On restore, the runtime walks the ancestor chain, collects the per-step writes, and replays them through the reducer to reconstruct state.

```python
from typing import Annotated
from langgraph.graph.message import _messages_delta_reducer

class State(TypedDict):
    # blob per step: ~60 bytes (sentinel) instead of a growing full list
    messages: Annotated[list, DeltaChannel(_messages_delta_reducer)]

    # bound read depth to at most 50 steps via periodic snapshots
    messages_bounded: Annotated[list, DeltaChannel(_messages_delta_reducer, snapshot_frequency=50)]
```

---

## Storage savings

`snapshot_frequency=N` writes a full-value blob every N steps, bounding replay depth. Without it, blob storage is theoretically near-zero but read latency grows unboundedly with thread length — not practical for production.

**Total checkpoint storage** (blobs + writes + metadata, ~400 chars/message):

| turns | `add_messages` | `DeltaChannel` (no snapshots) | `DeltaChannel(snapshot_frequency=50)` |
|------:|---------------:|------------------------------:|--------------------------------------:|
| 10    | 129.7 KB       | 38.7 KB (3x) †                | 38.7 KB (3x)                          |
| 100   | 9.18 MB        | 394 KB (23x) †                | 703 KB (13x)                          |
| 250   | 55.79 MB       | 987 KB (57x) †                | 3.07 MB (18x)                         |
| 500   | 221 MB         | 1.98 MB (112x) †              | 10.53 MB (21x)                        |

† No snapshots trades read latency for maximum storage savings — see below.

---

## The honest tradeoff: read latency

Without snapshots, every `get_state` walks the full ancestor chain — latency grows linearly with thread length.

**Read latency** (`get_state`, InMemory, 5-call avg):

| turns | `add_messages` | `DeltaChannel` (no snapshots) | `DeltaChannel(snapshot_frequency=50)` |
|------:|---------------:|------------------------------:|--------------------------------------:|
| 10    | 0.7 ms         | 1.1 ms †                      | 1.1 ms                                |
| 100   | 5.5 ms         | 11.1 ms †                     | 6.0 ms                                |
| 250   | 12.9 ms        | 27.2 ms †                     | 13.6 ms                               |

† Grows unboundedly with thread length.

---

## Reducer requirements

`DeltaChannel` takes a **batch reducer** `(state, list[writes]) -> state`. Two requirements:

- **Deterministic**: same inputs always produce the same output.
- **Associative** (batching-invariant): applying writes in two batches must equal applying them all at once — `reducer(reducer(s, [a, b]), [c]) == reducer(s, [a, b, c])`. This matters because LangGraph may replay writes in different batch sizes than they were originally produced; a non-associative reducer would silently reconstruct wrong state.

If your reducer isn't associative, use `BinaryOperatorAggregate` instead — `DeltaChannel` is not a drop-in replacement for every reducer.

```python
# ❌ Wrong — add_messages is a binary operator (value, value) -> value, not a batch reducer
messages: Annotated[list, DeltaChannel(add_messages)]

# ✅ Correct — _messages_delta_reducer is a batch reducer with dedup + RemoveMessage support
messages: Annotated[list, DeltaChannel(_messages_delta_reducer)]

# ✅ Custom batch reducer
def my_dict_reducer(state: dict, writes: list[dict]) -> dict:
    result = dict(state)
    for w in writes:
        result.update(w)
    return result

files: Annotated[dict, DeltaChannel(my_dict_reducer)]
```

---

## Backwards compatibility: no migration required

Threads written under `BinaryOperatorAggregate` (including `add_messages`) work transparently after swapping the annotation — existing checkpoints are not touched.

```python
# Before
messages: Annotated[list, add_messages]

# After — existing threads continue from where they left off
messages: Annotated[list, DeltaChannel(_messages_delta_reducer)]
```

**How:** when loading state, `DeltaChannel` asks the saver to walk the ancestor chain and collect per-step writes. The walk stops as soon as it hits an ancestor whose stored blob is a real value (not a sentinel). Old `BinaryOperatorAggregate` checkpoints store the full accumulated list at every step, so the walk terminates at the nearest pre-migration ancestor and uses that list as the starting point. Any writes recorded after the migration are then replayed on top.

This means time-travel, `get_state_history`, and resuming from mid-thread checkpoints all work correctly across the migration boundary — both before and after the swap. The only constraint is that the new reducer must produce the same accumulated value as the old one would have for any writes made after the migration point.

---

## How it works

`checkpoint()` always returns `DELTA_SENTINEL` (a zero-byte msgpack ext marker). On restore, `_get_channel_writes_history` walks the parent chain collecting `checkpoint_writes` until it hits a real blob (a `_DeltaSnapshot`, migration artifact, or `update_state`-written value), then `replay_writes` folds those deltas onto the seed through the reducer.

In `durability="async"` mode, `AsyncPregelLoop` tracks in-flight `aput_writes` futures for DeltaChannel fields and drains them before `aput()` — ensuring writes are durable before the sentinel blob is committed.

---

## What's in scope

- **`libs/langgraph/langgraph/channels/delta.py`** — `DeltaChannel` implementation
- **`libs/langgraph/langgraph/graph/message.py`** — `_messages_delta_reducer` (experimental)
- **`libs/checkpoint/`** — `_get_channel_writes_history` ancestor-walk API on `BaseCheckpointSaver`, optimized `InMemorySaver` override
- **`libs/checkpoint-postgres/`** — `PostgresSaver` / `AsyncPostgresSaver` single-roundtrip UNION ALL override
- **`libs/langgraph/langgraph/pregel/`** — checkpoint wiring, async write-ordering safety

---

## Known limitations

- **`ShallowPostgresSaver` incompatible**: shallow savers keep only the latest checkpoint — no ancestor chain to walk. Should raise at compile time (follow-up).
- **Per-channel ancestor walks**: each `DeltaChannel` field issues its own `_get_channel_writes_history` call today. A single walk collecting all sentinel channels would reduce roundtrips proportionally.
- **Third-party checkpointers**: need to implement `_get_channel_writes_history` to support `DeltaChannel`. The base-class fallback works but is unoptimized.
- **Updating the writes table w/ delta epoch ids for more efficient reads**
- follow up w/ LSD checkpointer implementations to support delta channel! and update prune


libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -278,6 +279,8 @@ def loads_typed(self, data: tuple[str, bytes]) -> Any:
             return ormsgpack.unpackb(
                 data_, ext_hook=self._unpack_ext_hook, option=ormsgpack.OPT_NON_STR_KEYS
             )
+        elif type_ == "delta":
+            return DELTA_SENTINEL
```
sydney-runkle: might roll into msgpack

libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -457,6 +495,104 @@ async def aprune(
...
+            collected.reverse()
+            return _ChannelWritesHistory(seed=DELTA_SENTINEL, writes=collected)
+        finally:
+            _DELTA_RECONSTRUCTION.reset(token)
+
+    async def _aget_channel_writes_history(
+        self, config: RunnableConfig, channel: str
+    ) -> _ChannelWritesHistory:
+        """Async version of `_get_channel_writes_history`. See docstring there."""
+        if _DELTA_RECONSTRUCTION.get():
```
hinthornw: hmmm I don't love using contextvars to check recursion depth. will look further as I continue to work my way th rough this

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,177 @@
...
+    ) -> None:
+        super().__init__(list)
+        self.operator = operator
+        self.value: Any = []
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, DeltaChannel):
+            return False
+        if (
+            self.operator.__name__ != "<lambda>"
```
hinthornw: we should probably extract this to a utility and document better

libs/langgraph/langgraph/channels/_delta.py
```diff
@@ -0,0 +1,177 @@
...
+    @property
+    def ValueType(self) -> Any:
+        return self.typ
+
+    @property
+    def UpdateType(self) -> Any:
+        return self.typ
+
+    def copy(self) -> Self:
+        new: DeltaChannel[Value] = DeltaChannel(self.operator)
```
hinthornw: ubernit but if the return type is Self, shouldn't we be using self.__class__(self.operator)?

libs/checkpoint/langgraph/checkpoint/memory/__init__.py
```diff
@@ -121,16 +124,91 @@ async def __aexit__(
...
+                            # ancestor's writes — the blob subsumes them.
+                            collected.reverse()
+                            return _ChannelWritesHistory(
+                                seed=blob_value, writes=collected
+                            )
+
+            step_writes = self.writes.get((thread_id, checkpoint_ns, cp_id), {})
+            # Within a superstep, sorted by (task_id, idx) = oldest → newest;
+            # reverse for newest-first scan.
+            for (_task_id, _idx), (tid, ch, serialized, _) in sorted(
```
hinthornw, line 222: hmmmmm

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
```diff
@@ -391,6 +398,45 @@ async def _cursor(
...
+        thread_id = config["configurable"]["thread_id"]
+        checkpoint_ns = config["configurable"].get("checkpoint_ns", "")
+        checkpoint_id = get_checkpoint_id(config)
+        if checkpoint_id is None:
+            target = await self.aget_tuple(config)
+            if target is None:
+                return _ChannelWritesHistory(seed=DELTA_SENTINEL, writes=[])
+            checkpoint_id = target.config["configurable"]["checkpoint_id"]
+        async with self._cursor() as cur:
+            await cur.execute(
```
hinthornw, line 419: could we pipeline these?

libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -185,16 +212,107 @@ def _migrate_pending_sends(
...
+        self,
+        blob_values: Any,
     ) -> dict[str, Any]:
         if not blob_values:
             return {}
-        return {
-            k.decode(): self.serde.loads_typed((t.decode(), v))
-            for k, t, v in blob_values
-            if t.decode() != "empty"
+        result: dict[str, Any] = {}
```
hinthornw: is this refactor necessary? seems to just be removing the comprehension?

libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -185,16 +212,107 @@ def _migrate_pending_sends(
...
+        blob in `checkpoint_blobs` (a pre-delta snapshot) terminates the
+        walk and is returned as the seed so replay starts from it.
+
+        Writes stored at `target_id` itself are pending writes for the next
+        step and are excluded — the walk begins at the target's parent.
+        """
+        parent_of: dict[str, str | None] = {}
+        ver_of: dict[str, str | None] = {}
+        for r in parents_rows:
+            cid = r["checkpoint_id"]
```
hinthornw: ooh: if can the parents_rows be typed as dict/collection or something? rn it's `Any` type but we assume it's dict-like

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
hinthornw: Ya i think we should

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def is_snapshot_step(self, step: int) -> bool:
+        """True if pregel should write a snapshot blob at this step."""
+        return (
+            self.snapshot_frequency is not None and step % self.snapshot_frequency == 0
+        )
+
+    def _clone_empty(self) -> Self:
+        new = self.__class__.__new__(self.__class__)
+        new.typ = self.typ
```
hinthornw: is this needed?

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,181 @@
...
+        new.key = self.key
+        new.operator = self.operator
+        new.snapshot_frequency = self.snapshot_frequency
+        new.value = MISSING
```
hinthornw: hmm so a new DeltaChannel inits with an empty list, but when cloning it has a MISSING sentinel? Feels off

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,181 @@
...
+    try:
+        return typ()
+    except Exception:
+        return []
+
+
+def _get_overwrite(value: Any) -> tuple[bool, Any]:
+    if isinstance(value, Overwrite):
+        return True, value.value
+    if isinstance(value, dict) and set(value.keys()) == {OVERWRITE}:
```
hinthornw: the `set()` call is unnecessary here and allocates an unused object

libs/langgraph/langgraph/pregel/_checkpoint.py
```diff
@@ -31,49 +37,136 @@ def create_checkpoint(
...
-        },
-        managed_specs,
-    )
+
+    channels: dict[str, BaseChannel] = {}
+    for k, spec in channel_specs.items():
+        ch: BaseChannel
+        stored = checkpoint["channel_values"].get(k, MISSING)
+        if _needs_replay(spec, stored) and saver is not None and config is not None:
+            assert isinstance(spec, DeltaChannel)
```
hinthornw: we shouldn't be using `assert` at runtime. it can be compiled it out certain environments, etc.

libs/langgraph/langgraph/channels/delta.py
```diff
@@ -0,0 +1,169 @@
...
+        return self.typ
+
+    @property
+    def UpdateType(self) -> Any:
+        return self.typ
+
+    def is_snapshot_step(self, step: int) -> bool:
+        """True if pregel should write a snapshot blob at this step."""
+        return (
+            self.snapshot_frequency is not None and step % self.snapshot_frequency == 0
```
hinthornw: are we intentionally snapshotting at step 0?

sydney-runkle: agreed, let's skip

libs/langgraph/langgraph/pregel/_checkpoint.py
sydney-runkle: good catch :/

libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py
sydney-runkle: good call, done here: https://github.com/langchain-ai/langgraph/pull/7586/changes/702895e484e7675543255614e4a7d4d7a4c09b46

libs/langgraph/langgraph/channels/delta.py
sydney-runkle: no, using same utils as binop now

libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -210,7 +210,7 @@ def local_read(
     return values
 
 
-def increment(current: int | None, channel: None) -> int:
+def increment(current: int | None, channel: None = None) -> int:
```
sydney-runkle: seems unnecessary?

sydney-runkle (commented):
open questions
* is the prune approach correct -- removing for in memory saver for now, but we should update for LSD i think
* have we properly tested all backwards compat cases -- added
* do we ever need to coerce messages to pydantic types in the reducer (i think saver manages), can test out beta reducer and find out
* is the delta channel design good (can refactor internals later so less concerned about this)
  * do we need missing and delta sentinel, seems redundant
* is the checkpoint change ok (get next version)

PR #7631, opened 2026-04-27
chore: idle timeout

hinthornw:
## Summary

Adds per-node `timeout` support to async StateGraph/Pregel nodes and to the functional API (`@task` / `@entrypoint`). A timeout caps how long a single node attempt may run, either as a hard wall-clock budget (`run_timeout`), or as an idle window that resets on observable progress (`idle_timeout`), or both. When exceeded, LangGraph raises `NodeTimeoutError`, clears writes from the failed attempt, and lets the existing retry policy decide whether to retry.

## Public API

A single `timeout=` kwarg on `add_node`, `@task`, `@entrypoint`, and `NodeBuilder.set_timeout`. Pass a number/`timedelta` for the simple case (treated as a hard wall-clock cap), or a `TimeoutPolicy` (in `langgraph.types`) for finer control:

```python
from datetime import timedelta
from langgraph.types import TimeoutPolicy

# simple: hard wall-clock cap on each attempt
builder.add_node("call_model", call_model, timeout=60)
builder.add_node("call_model", call_model, timeout=timedelta(minutes=2))

# full control
builder.add_node(
    "call_model",
    call_model,
    timeout=TimeoutPolicy(
        run_timeout=120,                # hard wall-clock cap in seconds, never refreshed
        idle_timeout=30,                # cap on time without observable progress, units in seconds
        refresh_on="auto",              # "auto" | "heartbeat"
    ),
)
```

- `run_timeout`: hard wall-clock cap on a single attempt; never refreshed.
- `idle_timeout`: progress-resetting cap. Refreshed by writes, stream output, yielded async stream chunks, child-task scheduling, runtime stream-writer calls, and any LangChain callback event from descendants of the node's run. `runtime.heartbeat()` is a manual signal for work that doesn't naturally emit any of these.
- `refresh_on="heartbeat"` narrows the refresh source to explicit `runtime.heartbeat()` only — useful when you want a strict idle definition that isn't reset by chatty subordinates.

For long-running async work that doesn't naturally emit progress:

```python
async def call_model(state: State, runtime: Runtime) -> State:
    while still_working:
        ...
        runtime.heartbeat()
    return {"messages": [response]}
```

`NodeTimeoutError` subclasses `TimeoutError` and carries `node`, `timeout`, `run_timeout`, `idle_timeout`, `elapsed`, and `kind` (`"run"` or `"idle"`). If the node's `retry_policy` permits `TimeoutError` it'll retry; the timer resets per attempt.

## Why async-only

Sync Python code cannot be safely cancelled in-process, so timeouts only apply to async nodes/tasks. Sync nodes with a `timeout` are rejected at compile time (covers direct nodes, wrapped runnables, sequences, and `RunnableParallel` branches); `run_with_retry` rejects them at runtime as a safety net.

## What gets cancelled and what doesn't

When a watchdog fires:

1. The attempt scope is closed under a lock so any in-flight `CONFIG_KEY_SEND` / stream / child-task scheduling that races with the timeout is dropped atomically.
2. Buffered `task.writes` are cleared so pre-timeout writes from the failed attempt don't leak into the checkpoint after a retry succeeds.
3. The background `asyncio.Task` is cancelled; its eventual exception is drained via a done-callback so asyncio doesn't log it.

Only the watchdog's own `TimeoutError` converts to `NodeTimeoutError`, so user-raised `asyncio.TimeoutError`, built-in `TimeoutError`, and `NodeTimeoutError` from a child node continue to propagate unchanged.

Child tasks already scheduled before the timeout fired still complete — they aren't part of the cancelled task's structured cancellation surface. This is intentional and tested.

## External-watchdog hook

WARNING: THIS API IS IN ALPHA AND SUBJECT TO CHANGE.

`CONFIG_KEY_TIMED_ATTEMPT_OBSERVER` is a per-config callback that receives lifecycle events for each timed attempt:

- `start` — fired before the proc runs, with `task_id`, `task_name`, `attempt`, `run_id`, `thread_id`, `checkpoint_ns`, `started_at`, and the configured `run_timeout_secs` / `idle_timeout_secs` / `refresh_on`.
- `progress` — fired on each progress signal that resets the idle clock, rate-limited to ~4 events per `idle_timeout` window so token-rate callbacks don't flood the observer. Carries the same context plus `progress_at`.
- `finish` — fired with `finished_at`, `status` (`"success"`/`"error"`), `error_type`, `error_message`. `ParentCommand` and `GraphBubbleUp` are treated as control flow, not errors.

This lets an orchestrating process listen to `start` + rolling `progress` to compute its own kill deadline (`progress_at + idle_timeout_secs`) and hard-kill a worker process if the in-process cancellation deadlocks. Observer callbacks run in whatever thread fires them, and any exception they raise is logged and swallowed.

## Implementation Notes

`runtime.heartbeat()` updates the progress timestamp without taking the guarded write lock. This avoids lock overhead on high-frequency callback/token paths and accepts a small timestamp race window, which is negligible for expected coarse idle-timeout configurations.

## Testing

Covers timeout retry behavior, user-raised timeout propagation, stale-write suppression, stream / callback / heartbeat progress resets, sync-node rejection, `RunnableParallel` branch validation, `StateGraph.add_node` behavior, lower-level Pregel behavior, observer start/progress/finish events, and functional API compatibility.


sydney-runkle (dismissed):

hinthornw:
@copilot resolve the merge conflicts in this pull request

PR #7632, opened 2026-04-27
feat(langgraph): add SubgraphTransformer with mini-mux navigation

nick-hollon-lc:
Stacks on #7628.

## Summary

Adds in-process navigation handles for discovered subgraphs via a new `SubgraphTransformer`. Each direct-child subgraph gets a `SubgraphRunStream` (extending `GraphRunStream`) wrapping a child mini-mux scoped to the subgraph's namespace, giving consumers a navigable view of each subgraph's own projections.

## Inheritance

`SubgraphRunStream` extends `GraphRunStream`; `AsyncSubgraphRunStream` extends `AsyncGraphRunStream`. Subgraph metadata (`path`, `graph_name`, `trigger_call_id`, `status`, `error`) lives on a shared `_SubgraphRunStreamMixin`. Native projection attributes (`values`, `messages`, `subgraphs`, `lifecycle`) and the `output` / `interrupted` / `interrupts` properties are auto-inherited — adding a new native transformer to the factory list lights it up on every handle for free.

```python
for handle in run.subgraphs:
    for v in handle.values:
        ...
    for inner in handle.subgraphs:  # recursive grandchildren
        ...
```

## Mini-mux infrastructure

- `StreamMux.make_child(scope)` clones the factory list at a deeper scope. Pre-built `transformers=` instances are root-only; only `factories=` propagate.
- `StreamMux.bind_pump` / `bind_apump` propagate the root pump to every projection in the mux — including child mini-muxes built later, so cursors on subgraph handles drive the root run.
- `StreamMux.is_async` is now public (was `_is_async`).
- `StreamTransformer._on_register(mux)` hook captures the parent mux for `make_child` calls.

## Subgraph handles

- Constructed with `graph_iter=None`, `wire_pump=False` — they have no own iterator and inherit the root pump on the mini-mux. Override `_pump_next` / `_apump_next` to delegate to the parent pump for `handle.output` etc.
- `Pregel.stream_v2` / `astream_v2` now use `factories=[ValuesTransformer, MessagesTransformer, LifecycleTransformer, SubgraphTransformer]` (plus root-only `transformers=` for compiled / per-call extras).
- `ValuesTransformer` / `MessagesTransformer` renamed `parent_ns=` to `scope=` for consistency with the base class. They now call `super().__init__(scope)`.

## Suppression + lazy-subscribe

- Both `LifecycleTransformer` and `SubgraphTransformer` return `False` from `process` after consuming a `tasks` event. Tasks events are folded into the synthesized lifecycle / subgraph projections; consumers iterating the raw protocol stream see the higher-level views.
- Mini-mux logs follow the same lazy-subscribe rule as root projections — drill into `handle.<projection>` inside the loop body that yields the handle, before the next pump cycle. The "drain handles first, drill in later" pattern (`list(run.subgraphs); ...`) loses events, same as `list(run.values); list(run.messages)` already does at the root.


libs/langgraph/langgraph/graph/state.py
```diff
@@ -1045,7 +1045,7 @@ def compile(
         interrupt_after: All | list[str] | None = None,
         debug: bool = False,
         name: str | None = None,
-        transformers: Sequence[Callable[..., Any]] | None = None,
+        transformers: Sequence[Callable[[tuple[str, ...]], Any]] | None = None,
```
sydney-runkle, line 1048: what's the tuple?

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -364,6 +364,40 @@ def _collect_stream_modes(mux: Any) -> list[StreamMode]:
...
+            raise TypeError(
+                "stream_v2 transformers must be transformer classes or factories, "
+                f"got pre-built instance {type(spec).__name__}. Pass the class "
+                "or a factory like `lambda scope: MyTransformer(scope, ...)`."
+            )
+        if not callable(spec):
+            raise TypeError(
+                "stream_v2 transformers must be transformer classes or factories, "
+                f"got {type(spec).__name__}."
+            )
```
sydney-runkle, line 400: why factory?

libs/langgraph/langgraph/graph/state.py
nick-hollon-lc: every transformer needs a scope with the subgraph changes which is advertised on the base class

libs/langgraph/langgraph/pregel/main.py
nick-hollon-lc: mux needs to create fresh transformer instances for the subgraph so we are forcing factories here

PR #7646, opened 2026-04-29
chore: dynamic push-task timeouts

hinthornw:
You can Send(..., timeout=...) now. Much fun.

This would allow us to do something like adding support for an annotation to let the LLM to pick a timeout for a given tool call, etc.

### DIFF
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -425,6 +425,17 @@ class RetryPolicy(NamedTuple):
     """List of exception classes that should trigger a retry, or a callable that returns `True` for exceptions that should trigger a retry."""
 
 
+def _coerce_timeout_seconds(
+    value: float | timedelta | None, *, field: str
+) -> float | None:
+    if value is None:
+        return None
+    seconds = value.total_seconds() if isinstance(value, timedelta) else float(value)
+    if seconds <= 0:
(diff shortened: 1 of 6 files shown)


libs/langgraph/langgraph/types.py
```diff
@@ -457,6 +468,39 @@ class TimeoutPolicy:
     `"heartbeat"` refreshes only on explicit `runtime.heartbeat()` calls.
     """
 
+    @classmethod
+    def coerce(
+        cls, value: float | timedelta | TimeoutPolicy | None
+    ) -> TimeoutPolicy | None:
```
sydney-runkle, line 474: nice

libs/langgraph/langgraph/types.py
```diff
@@ -623,6 +667,8 @@ class Send:
     Attributes:
         node (str): The name of the target node to send the message to.
         arg (Any): The state or message to send to the target node.
+        timeout (TimeoutPolicy | None): Optional timeout policy for this specific
+            pushed task. If omitted, the target node's timeout policy is used.
```
sydney-runkle, line 671: should we add more stuff here, retry policy for ex

hinthornw: ooh maybe? I was imagining this would be mainly for timed tasks, and the retry policy feels less necessary to dynamically set.

PR #7699, opened 2026-05-04
feat: public get_writes_history saver API + delta cadence rework

sydney-runkle:
## Summary

- Promotes the private K-channel batched ancestor-walk to a stable public `get_delta_channel_history` / `aget_delta_channel_history` API on `BaseCheckpointSaver` (returns `Mapping[str, DeltaChannelHistory]`, a TypedDict with `writes` always present and `seed` `NotRequired`)
- Removes `DELTA_SENTINEL` / `_DeltaSentinel` entirely — the saver layer is now delta-agnostic on both write and read paths
- Reworks `DeltaChannel` snapshot cadence from "every Nth superstep" to "every N updates to this channel," persisted in `CheckpointMetadata.delta_updates_since_snapshot`
- Adds Postgres optimizations: paged stage-1 with cursor (1024-row pages) and per-channel UNION ALL stage-2 (no over-fetch when channels have different chain depths)
- Default `snapshot_frequency` becomes a positive int (default `1000`); the previous `None` opt-out is removed

## Public API

```python
class DeltaChannelHistory(TypedDict):
    writes: list[PendingWrite]            # always present, possibly empty
    seed: NotRequired[Any]                # absent if walk reached root

def get_delta_channel_history(
    self, *, config: RunnableConfig, channels: Sequence[str]
) -> Mapping[str, DeltaChannelHistory]: ...

async def aget_delta_channel_history(
    self, *, config: RunnableConfig, channels: Sequence[str]
) -> Mapping[str, DeltaChannelHistory]: ...
```

`config` and `channels` are keyword-only so later additions (e.g. `page_size`) don't shift the positional API.

The TypedDict-with-`NotRequired[seed]` shape matches the existing checkpoint-package convention (`CheckpointMetadata` is `TypedDict(total=False)`) — absence-via-key-omission rather than introducing a new sentinel. Pregel translates `"seed" not in hist` to `MISSING` on its side at consume time.

The default impl walks `get_tuple` + `parent_config` correctly but is slow on long chains; savers that care override (`InMemorySaver`, `PostgresSaver`).

## Sentinel removal

`DELTA_SENTINEL` and `_DeltaSentinel` are deleted entirely. The saver layer becomes delta-agnostic:

- `DeltaChannel.checkpoint()` returns `MISSING` for non-snapshot steps; pregel's `create_checkpoint` skips MISSING so delta channels without a snapshot simply don't appear in `channel_values`
- `InMemorySaver.put` and Postgres `put` no longer filter sentinels (they have nothing to filter)
- `_needs_replay` becomes `stored is MISSING`
- `DeltaChannel.from_checkpoint` accepts: `MISSING` → empty, `_DeltaSnapshot(value)` → snapshot value, plain value → pre-migration legacy

## Snapshot cadence

`DeltaChannel.snapshot_frequency: int` (default `1000`, positive). The previous `None` opt-out is gone.

```python
def should_snapshot(ch_name, ch):
    if force_delta_snapshot:                                  # durability="exit"
        return True
    return updates_since_snapshot.get(ch_name, 0) >= ch.snapshot_frequency
```

Per-channel update counters are persisted in `CheckpointMetadata.delta_updates_since_snapshot` (`NotRequired`, `total=False`). The counter is incremented by `_put_checkpoint` for any delta channel in `updated_channels` and reset to `0` by `create_checkpoint` for channels that fire a snapshot this step. Version-format-independent — works for `int`, `float`, and `str` versioning schemes alike.

## Postgres optimization

Two improvements internal to the override:

**Stage-1 paged with cursor** (`LIMIT 1024` internal const, `AND checkpoint_id < ?` for subsequent pages). The previous unpaged form scanned every checkpoint in `(thread_id, ns)` and was pathological at high thread depths.

**Stage-2 per-channel UNION ALL**: one `WHERE channel='X' AND checkpoint_id = ANY(chain_X)` branch per channel plus one seed-blob branch per channel with a seed. The previous form filtered by `channel = ANY(channels) AND checkpoint_id = ANY(union_chain_cids)`, over-fetching writes when channels had different chain depths (`K × max(chain_lengths)` vs the correct `sum(chain_lengths)`).

Both improvements stay internal to `PostgresSaver`/`AsyncPostgresSaver`; the public contract returns a single `Mapping`.

## Benchmarks

`libs/langgraph/tests/test_delta_channel_benchmark.py`. Run via `python libs/langgraph/tests/test_delta_channel_benchmark.py`. Postgres against local pg:5441.

Results below trimmed to the high-signal cells. Sub-millisecond / sub-100-turn rows omitted as warmup-bound; freq=1 omitted (chain depth = 1, nothing to optimize); peak read-time memory and Postgres storage are flat between branches and omitted. Deep-thread reads and the cadence-rework storage win are the load-bearing numbers.

### Postgres reads, 500 turns

| Scenario | main | branch | Δ |
|---|---:|---:|---:|
| Single-channel deep read | 17.7 ms | **6.1 ms** | **-66%** |
| Single-channel, 1000 turns | 35.0 ms | **14.3 ms** | **-59%** |
| K=3 channels, freq=50 uniform | 70.5 ms | **41.4 ms** | **-41%** |
| K=8 channels, freq=50 uniform | 214.2 ms | **139.4 ms** | **-35%** |
| K=8 channels, mixed freq (25/50/100/.../1000) | 295.6 ms | **214.4 ms** | **-27%** |

K-channel batching + paged stage-1 + per-channel UNION ALL stage-2 doing exactly what they should at depth.

### InMemory reads, 500 turns

| Scenario | main | branch | Δ |
|---|---:|---:|---:|
| Single-channel deep read | 7.9 ms | **3.8 ms** | **-52%** |
| Single-channel, 1000 turns | 15.6 ms | **7.2 ms** | **-54%** |
| K=8 channels, freq=50 uniform | 112.3 ms | 94.6 ms | -16% |
| K=8 channels, mixed freq | 184.9 ms | **134.5 ms** | **-27%** |

### InMemory storage, 500 turns (cadence-rework win)

| Scenario | main | branch | Δ |
|---|---:|---:|---:|
| K=3, freq=50 uniform | 8.7 MB | **3.3 MB** | **-62%** |
| K=3 mixed freq | 3.8 MB | **1.3 MB** | **-66%** |
| K=8, freq=50 uniform | 23.1 MB | **8.7 MB** | **-62%** |
| K=8 mixed freq | 11.5 MB | **4.2 MB** | **-64%** |

Snapshot frequency now counts **channel updates** instead of **supersteps**. On graphs where supersteps outpace per-channel updates (e.g., input/end steps that don't write to channels), branch stores ~3× fewer snapshot blobs.

### Tradeoff worth flagging

InMemory K=3 with mixed frequencies (50/200/1000) at 500 turns: **+64% read latency** (46.6 → 76.5 ms). The mixed scenario has a channel with `freq=1000` that goes the entire 500-turn run with no snapshot. On main, the old superstep-counted cadence happened to fire at step=500 anyway. New cadence gives users explicit control over walk depth via `snapshot_frequency`. The K=8 mixed case still wins overall (-27%); this regression is specific to the K=3 mixed shape.

Default `snapshot_frequency=1000` is the upper bound on walk depth — it's a tunable knob.

## Tests

- New sqlite smoke test (`test_get_delta_channel_history.py`) exercises the inherited default `BaseCheckpointSaver` impl via `SqliteSaver` / `AsyncSqliteSaver` end-to-end with a real `DeltaChannel`-backed graph. Sqlite uses the default unchanged — this validates the default path actually works on a real second saver, not just on the optimized override.
- Module-level `pytest.importorskip("langgraph.channels.delta")` guards the test for sqlite's standalone CI environment (matches the postgres pattern).

## Test plan

- [x] `libs/checkpoint`: 150 passed, 16 skipped
- [x] `libs/langgraph` (channels + delta migration): 41/41 (post-merge)
- [x] `libs/langgraph` (full pregel suite): 1784 passing — 6 "failures" verified via `env -i` clean shell are local LangSmith env vars + `git describe revision_id` polluting LangChain metadata fixtures; CI is unaffected
- [x] `libs/checkpoint-postgres`: 40/40 saver tests + 3/3 delta channel reconstruction tests against local Postgres
- [x] `libs/checkpoint-sqlite`: 105/105 (incl. retry-passed flake `test_ttl_refresh`, unrelated to this PR)
- [x] Lint clean across all four libs (`ruff format`, `ruff check`, `mypy`)
- [x] Branch-vs-main benchmarks — see results above


libs/langgraph/langgraph/pregel/_checkpoint.py
```diff
@@ -126,18 +175,26 @@ def channels_from_checkpoint(
...
-        stored = checkpoint["channel_values"].get(k, MISSING)
-        if _needs_replay(spec, stored) and saver is not None and config is not None:
+        if k in histories:
             delta_spec = cast(DeltaChannel, spec)
-            history = saver._get_channel_writes_history(config, k)
-            replay_ch = delta_spec.from_checkpoint(history.seed)
-            replay_ch.replay_writes(history.writes)
+            history = histories[k]
+            replay_ch = delta_spec.from_checkpoint(history.get("seed", MISSING))
+            replay_ch.replay_writes(history["writes"])
```
longquanzheng, line 194: `replay_writes` -> `replay_history`

libs/langgraph/langgraph/pregel/_checkpoint.py
```diff
@@ -92,15 +139,15 @@ def create_checkpoint(
...
+    checkpoint, requiring an ancestor walk to reconstruct.
 
     `_DeltaSnapshot` blobs and plain values (migration) resolve directly via
-    `from_checkpoint` — only `DELTA_SENTINEL` / `MISSING` trigger replay.
+    `from_checkpoint` — only absence (`MISSING`) triggers replay.
     """
     if not isinstance(spec, DeltaChannel):
         return False
-    return stored is MISSING or stored is DELTA_SENTINEL
+    return stored is MISSING
```
longquanzheng, line 150: I wonder we should just always replay for DeltaChannel to keep the code simple and clean.

longquanzheng (approved):

Strongf-bob:
Hi @sydney-runkle — I was reviewing the DeltaChannel follow-up work and noticed the TODO in `libs/langgraph/tests/test_delta_channel_migration.py` to add Postgres variants.

The current `checkpoint-postgres` coverage exercises native DeltaChannel chain reconstruction in async mode across the base/pool/pipe saver fixtures, but it does not appear to cover `BinaryOperatorAggregate` → `DeltaChannel` compatibility against Postgres: loading a legacy full-value blob as the seed, time-traveling to a settled pre-migration checkpoint, and continuing the thread with post-migration deltas.

I would like to contribute narrowly scoped regression coverage in `libs/checkpoint-postgres`, without production-code changes. My proposed scope is:

- verify settled legacy checkpoints after swapping the channel annotation;
- continue a migrated thread and verify new writes replay on top of the legacy seed;
- cover both sync and async Postgres saver paths, using the existing connection fixtures (with `importorskip` consistent with the existing async DeltaChannel test unless you prefer a different test dependency setup).

I will not open a PR until the work is approved and assigned, per the contribution policy. Would you like this follow-up? If so, which issue/discussion should the PR link to, and could I be assigned to it?

Strongf-bob:
Hi @sydney-runkle — a gentle follow-up on the proposal above. I am still interested and can start as soon as the scope and assignment path are confirmed. No rush; I just wanted to check whether this Postgres migration coverage would be useful, or whether you would prefer that I focus on a different contribution. Thanks!

PR #7729, opened 2026-05-06
fix(prebuilt): assign UUIDs to ID-less messages in _messages_delta_reducer

sydney-runkle:
## Summary

- `_messages_delta_reducer` now assigns UUIDs to messages that lack IDs, matching the behavior of `add_messages`
- UUID assignment is done inline within the existing index-building and write-processing loops (no extra iteration)
- Updated docstring to reflect that missing-ID assignment is now handled
- Added a test confirming that `RemoveMessage` tombstoning works correctly on messages that were written without explicit IDs

## Motivation

Without UUID assignment, `RemoveMessage` tombstoning silently no-ops on messages that were created without explicit IDs — the message's `id` stays `None`, so it never appears in the index, and any subsequent `RemoveMessage(id=<assigned_uuid>)` can't find it. This was discovered when wiring `DeltaChannel` onto the `messages` key in a production agent where some middleware emits ID-less messages.

## Test plan

- [ ] `uv run pytest libs/langgraph/tests/test_channels.py -x -q -k "delta_reducer or delta_channel"` — all 26 pass (25 existing + 1 new)

### DIFF
--- a/libs/langgraph/langgraph/graph/message.py
+++ b/libs/langgraph/langgraph/graph/message.py
@@ -257,9 +257,10 @@ def _messages_delta_reducer(
 
     Raw dict / string / tuple inputs are coerced to typed `BaseMessage`
     objects so that HTTP-driven graphs work without a separate coercion
-    step. This is not full `add_messages` parity — `REMOVE_ALL_MESSAGES`,
-    unknown-id `RemoveMessage` errors, missing-id UUID assignment, and
-    `BaseMessageChunk` conversion are not handled here.
+    step. Messages without IDs are assigned UUIDs (matching `add_messages`
+    behavior) so that message eviction and `RemoveMessage` tombstoning work
+    correctly. `REMOVE_ALL_MESSAGES` and `BaseMessageChunk` conversion are
+    not handled here.
 
     Example::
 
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/graph/message.py
```diff
@@ -249,17 +249,15 @@ def _messages_delta_reducer(
 ) -> list[AnyMessage]:
     """**Experimental.** Batch reducer for use with `DeltaChannel`.
 
-    Processes all writes in one pass — dedup by ID, `RemoveMessage`
-    tombstoning — without calling `add_messages`.
+    Provides full `add_messages` parity: dedup by ID, `RemoveMessage`
```
eyurtsev, line 252: RemoveMessage tombstoning => could benefit from a bit of elaboration

libs/langgraph/langgraph/graph/message.py
```diff
@@ -280,24 +278,51 @@ class State(TypedDict):
...
 
-    index: dict[str, int] = {
-        m.id: i for i, m in enumerate(state_msgs) if m.id is not None
-    }
+    # Build index and assign missing IDs in one pass (parity with add_messages
+    # so that eviction and RemoveMessage tombstoning work on ID-less messages).
+    index: dict[str, int] = {}
+    for i, m in enumerate(state_msgs):
+        if m.id is None:
+            m.id = str(uuid.uuid4())
```
eyurtsev, line 318: is the assignment in place safe here?

eyurtsev (approved):

PR #7730, opened 2026-05-06
chore(langgraph): re-implement exit mode for delta channel

longquanzheng:
## Summary

Replaces `durability="exit"`'s blanket force-snapshot of every `DeltaChannel` with proper write persistence that honors per-channel `snapshot_frequency`, plus closes two latent bugs the force-snapshot was masking.

Before: every exit-mode run wrote a full `_DeltaSnapshot` blob for every delta channel, even when the channel had zero updates this run and was nowhere near its `snapshot_frequency`. After: the same count-based decision used by `durability="sync"`/`"async"` applies — channels at or above `snapshot_frequency` snapshot; channels below it persist their accumulated writes via a lazy "stub" anchor; untouched channels write nothing.

## What changed

**Core redesign** (`pregel/_loop.py`, `pregel/_checkpoint.py`)

- Drop `force_delta_snapshot` from `create_checkpoint` and `_should_snapshot_delta`.
- Add `decide_delta_snapshots(channels, counts)` pure helper used by both `create_checkpoint` and the new exit-mode peek-ahead path.
- Add `_exit_delta_writes` accumulator: every delta-channel write produced during a `durability="exit"` run (input writes from `_first` + per-superstep writes captured before `pending_writes.clear()` in `after_tick`) is collected into this list.
- Add `_put_exit_delta_writes` (sync + async): runs from `_suppress_interrupt` BEFORE `_put_checkpoint(exiting=True)`. Filters out channels that will snapshot, then persists remaining writes to `checkpoint_writes` under an anchor parent. The anchor is the existing saved parent on resumed runs, or a lazily-created empty stub on first runs.
- Visibility ordering: stub put goes onto `_put_checkpoint_fut` (becomes the next put's `prev`); exit-write futures go onto `_delta_write_futs`. The existing `_checkpointer_put_after_previous` already drains both before calling `saver.put`, so `final_checkpoint` is structurally guaranteed to land last — readers never see a partial view.

**Latent bugs fixed (previously masked by force-snapshot)**

- **Sync drain race**: `SyncPregelLoop` now initializes `_delta_write_futs = []` in `__enter__` and drains it in sync `_checkpointer_put_after_previous` before `put`, mirroring the async version. Without this, a multi-worker `BackgroundExecutor` could publish a checkpoint before the writes that produced it.
- **Count double-bump in exit mode**: in `_put_checkpoint`, `delta_updates_since_snapshot` was being incremented twice for the last superstep — once by the intermediate `after_tick` call, once by `_suppress_interrupt`. Force-snapshot used to reset all counts to 0 so this never persisted; without it, snapshots would fire one superstep early after every exit-mode run. Fixed by gating the count-bump behind `not exiting`.

**Pre-existing input-durability gap**

- In the plain (non-Command) input path of `_first`, delta-channel input writes are now persisted via `put_writes` (mirroring the Command path), so sub-frequency inputs survive a `get_state` on resumed runs in `sync`/`async` durability. Note: first-run `sync`/`async` still has the same gap (writes orphan on the synthetic-empty parent id). That's flagged as a follow-up — out of scope for this PR.

## Test plan

- Existing `tests/test_pregel.py` and `tests/test_pregel_async.py` pass unchanged.
- Existing `tests/test_channels.py` (29 tests) and `tests/test_delta_channel_migration.py` pass unchanged.
- New `tests/test_exit_delta_persistence.py` (11 tests) covers:
  - **Write-path**: zero-write exit (no stub), all-snapshot first run (no stub), sub-freq first run (single shared stub), sub-freq resumed run (anchor on saved parent), sync-vs-exit count parity, mixed snapshot/non-snapshot channels, snapshot fires at frequency.
  - **Read-path**: K-run replay chain reads correctly across stub→saved-parent transition; metadata `delta_updates_since_snapshot` round-trips correctly; mixed sync/exit durability alternation produces correct final state; snapshot+tail-deltas combination reads correctly.
- `make format && make lint && make test` in `libs/langgraph/`.


libs/langgraph/langgraph/pregel/_checkpoint.py
```diff
@@ -66,75 +66,69 @@ def create_checkpoint(
     id: str | None = None,
     updated_channels: set[str] | None = None,
     get_next_version: GetNextVersion | None = None,
-    force_delta_snapshot: bool = False,
     updates_since_snapshot: Mapping[str, int] | None = None,
-    new_updates_since_snapshot: dict[str, int] | None = None,
-) -> Checkpoint:
-    """Create a checkpoint for the given channels.
+) -> CreateCheckpointResult:
+    """Build a new Checkpoint from the previous one and live channel state.
```
sydney-runkle, line 66: don't love changing return type here, as seen in tests this is technically breaking, but as far as i can tell this one is an internal util

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -905,37 +964,56 @@ def _first(
...
+        # `is` (object identity) — not `==`. Three of four call sites pass a
+        # fresh dict ({"source":"input"|"loop"|"fork"}); only
+        # `_suppress_interrupt`(will rename to _on_loop_exit soon)
+        # at exit reuses the existing `self.checkpoint_metadata` instance. So
+        # `metadata is self.checkpoint_metadata` is True only on the exit call,
+        # which is what we use to gate exit-only behaviour (skip count-bump,
+        # don't replace metadata). Could be replaced by an explicit
+        # `exiting: bool = False` parameter; left as-is to match the existing
+        # idiom in this file.
+        # TODO: replace with an explicit `exiting: bool = False` parameter.
```
sydney-runkle, line 976: should we address this?

sydney-runkle (dismissed):
can we add a test confirming that if a channel is at snapshot_frequency but the last checkpoint doesn't write to it, we don't snapshot?

generally seems like a clean solution given that it's isolated to delta channels w/ durability mode exit, and we definitely want snapshot_frequency to still have significance in this case

libs/langgraph/langgraph/pregel/_loop.py
longquanzheng: yes I wanted, that's why I left a TODO. but worried that you don't want to change anything unreleated to delta channel. 

You want me to do in this PR or an diff PR?

PR #7736, opened 2026-05-07
chore(langgraph): add guide/conformance for delta channel checkpointer

longquanzheng:
## Summary

Add a user-facing design doc and `get_delta_channel_keepset` helper for third-party `BaseCheckpointSaver` authors who need to support graphs using `DeltaChannel`.

**Deliverables:**

1. ~~**`docs/delta-channel-checkpointer-guide.md`** — comprehensive guide covering~~:
moved to docs repo

2. **`BaseCheckpointSaver.get_delta_channel_keepset` / `aget_delta_channel_keepset`** — returns the minimum set of ancestor `checkpoint_id`s that must survive deletion for a given head's `DeltaChannel` reconstruction to remain intact. Enables safe `prune` implementations without silently corrupting delta history.

3. **Docstring warnings** on `prune`, `aprune`, `delete_for_runs`, `adelete_for_runs`, `copy_thread`, `acopy_thread` explaining the DeltaChannel pitfall (silent data loss if ancestor writes/snapshots are deleted).

4. **Three new conformance capabilities** in `libs/checkpoint-conformance`:
   - `delta_channel_history` — validates the `aget_delta_channel_history` walk contract
   - `delta_channel_keepset` — validates the keep-set contract
   - `delta_channel_reconstruction` — end-to-end round-trip (aput + aput_writes + history + reconstruct)

## Test plan

- [x] `make format lint` passes in `libs/checkpoint`, `libs/checkpoint-conformance`
- [x] All three new conformance capabilities pass against `InMemorySaver`
- [x] Run conformance against SQLite saver 
- [x] Run conformance against Postgres saver


libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+from langgraph.checkpoint.conformance.spec._delta_fixtures import build_delta_chain
+
+
+def _reconstruct(seed: Any, writes: list) -> list:
+    """Minimal DeltaChannel reconstruction: list-append reducer.
+
+    Mirrors DeltaChannel.from_checkpoint(seed) + replay_writes(writes).
+    """
+    from langgraph.checkpoint.serde.types import _DeltaSnapshot
+
```
sydney-runkle, line 30: i feel like this is too mock-y? is it valuable for us to test this?

libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/__init__.py
sydney-runkle, line 1: nice

sydney-runkle, line 1: i think one test we're missing in this file is the case of a migration

old checkpoints has channel_values for the given channel
it gets swapped to delta channel
want to make sure we respect the "seed" as the raw channel value

libs/checkpoint-conformance/pyproject.toml
```diff
@@ -43,9 +43,14 @@ asyncio_mode = "auto"
...
 # at runtime via capability detection and may not exist on the installed
 # base class. Dict literal inference is also overly strict for RunnableConfig.
+# Delta-channel tests import from `langgraph` (not a declared dep of this
+# package — at test time it is installed alongside); private `_DeltaSnapshot`
+# imports are intentional (beta surface).
 unresolved-attribute = "ignore"
+unresolved-import = "ignore"
 invalid-argument-type = "ignore"
 invalid-return-type = "ignore"
+missing-typed-dict-key = "ignore"
```
sydney-runkle, line 53: what's this for?

libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -340,6 +348,17 @@ def copy_thread(
...
+        !!! warning "DeltaChannel"
+
+            Implementations must copy the **complete** parent chain (all
+            ancestor checkpoints and their `checkpoint_writes`) — copying
+            only the head checkpoint will leave the target thread with
+            `DeltaChannel` state that cannot be reconstructed (no path back
+            to a `_DeltaSnapshot` ancestor). Equivalently, the copy must
+            include enough ancestors that every `DeltaChannel`-backed key
+            has either a `_DeltaSnapshot` in `channel_values` somewhere in
+            the chain, or a complete write history back to the chain root.
```
sydney-runkle, line 361: should we add some sort of force snapshot mechanism on the checkpointer so implementations don't have to do this?

libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -612,6 +680,114 @@ async def aget_delta_channel_history(
             result[ch] = entry
         return result
 
+    def get_delta_channel_keepset(
+        self,
+        *,
+        config: RunnableConfig,
+        channels: Sequence[str],
+    ) -> set[str]:
```
sydney-runkle, line 688: custom implementations should implement a version of this that's efficient for their data model right?

sydney-runkle (commented):
an idea i need to flesh out more -- should we have a new `BaseDeltaCheckpointSaver` that checkpointers need to inherit to get these new methods?

just seems like we're complicating the interface of the base checkpoint saver a lot, and i wonder if a diff implementation would be cleaner

then in pregel we could guard by checking if the mixin was present

sydney-runkle:
ah one reason maybe not to do that is then i think inherently you'd have to make code changes to be compatible... so maybe ignore me. i'll dig in

libs/checkpoint-conformance/pyproject.toml
longquanzheng: All because of ty rules in the pkg.

The Checkpoint TypedDict gained updated_channels as a required key. 

When we construct Checkpoint(...) in the test fixtures, ty requires it even though it's None. Rather than making updated_channels NotRequired (which would be a broader change), we suppress the diagnostic here. 

The unresolved-import suppression is because the conformance tests import _DeltaSnapshot (private/beta) and langgraph.channels.delta (not a declared dep of this package — only available at test time).

libs/checkpoint/langgraph/checkpoint/base/__init__.py
longquanzheng: yeah unfortunately it won't work with LSD as you said

libs/checkpoint/langgraph/checkpoint/base/__init__.py
longquanzheng: yes, let me update the comment to make it clear this is an example as basic implementation.

Or maybe just remove it if you want? doesn't seem necessary to keep, but I am hesitating because it could be usefult to read as an example.

libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_reconstruction.py
longquanzheng: The main value is catching serialization round-trip bugs — if a saver fails to preserve _DeltaSnapshot through its blob codec. 

but yeah I feel like test_delta_channel_history.py should be enough. will remove it.

libs/checkpoint-conformance/pyproject.toml
sydney-runkle: > The Checkpoint TypedDict gained updated_channels as a required key.

This wasn't in this PR, so I don't think we should need?

PR #7738, opened 2026-05-07
fix(langgraph): join lifecycle.started cause on trigger_call_id, support list-shape input

nick-hollon-lc:
## Summary

- Adds a `metadata` field to `LifecyclePayload` carrying `tool_call_id` mined from a per-call dispatched task's `input` — anchors the lifecycle event back to the originating AI message tool call.
- `_TasksLifecycleBase` mines `tool_call_id` on every `tasks` start and joins child subgraphs to it via `parent_task_id` (the pregel task id of the dispatching task), which is unique per Send and disambiguates parallel tool calls dispatched in the same model turn.
- Recognises both per-call envelope shapes by structure (no internal prebuilt types named):
  - `[{"id": ..., "name": ..., "args": {...}}]` — single-element list shape `langchain.agents.create_agent` Send-fans out.
  - `{"tool_call": {"id": ..., ...}, ...}` — dict envelope older prebuilt agent paths emit.
- `lifecycle.started.metadata` is `{"type": "tool_call", "tool_call_id": ...}` (snake_case on the wire). Args (`subagent_type`, `description`, etc.) are deliberately NOT mined — they live on the AIMessage's `tool_calls[i].args` and consumers join via `tool_call_id` to read them, keeping one source of truth and decoupling the streaming layer from any specific tool's args schema.
- `metadata` is absent for structurally-triggered subgraphs and for tool dispatches whose envelope carried no `id`.
- `parent_task_id` is required and emitted on every event (started + every terminal path) so consumers can correlate `started` ↔ terminal without a namespace fallback.
- Cache of mined ids is popped on consumption — no unbounded per-stream-call growth on long subagent runs.
- Shape detection extracted to `_extract_dispatching_tool_call_id` helper.
- Renames the lifecycle event type alias from `SubgraphStatus` to `LifecycleEvent` (the values like `started` are transitions, not statuses), and adds inline docstrings on each `LifecyclePayload` field and each `LifecycleEvent` literal.

## `lifecycle` event reference

Pushed on the `lifecycle` projection (in-process: `run.lifecycle`; over the wire: protocol `method = "lifecycle"`). One payload per state transition of a subgraph the transformer is tracking.

### Payload fields

| Field             | When set                                              | Type               |
|-------------------|-------------------------------------------------------|--------------------|
| `event`           | always                                                | `LifecycleEvent`   |
| `namespace`       | always — checkpoint namespace of the subgraph         | `list[str]`        |
| `parent_task_id`  | always — pregel task id of the dispatching task in the parent scope | `str`  |
| `graph_name`      | on `started`, when the namespace segment carries one  | `str`              |
| `metadata`        | on `started`, when the dispatching task's `input` matched a per-call tool-dispatch shape with a string `id` | `dict[str, Any]`   |
| `error`           | on `failed`, carries the error string                 | `str`              |

### `event` values

`LifecycleEvent = Literal["started", "completed", "failed", "interrupted", "drained"]`

- `started` — first `tasks` event observed at a tracked namespace.
- `completed` — the dispatching task's `TaskResultPayload` arrived with neither error nor interrupts. Also emitted by `finalize` for any tracked namespace still open at run end (graph completed cleanly without us ever seeing the dispatching task's result).
- `failed` — the task that dispatched this subgraph errored, OR the run failed at top level with a non-interrupt / non-drain exception.
- `interrupted` — the dispatching task's result carried `interrupts` (takes precedence over `error` if both present), OR the run failed with `GraphInterrupt`. In standard graph wiring a subgraph node IS the body of its dispatching task, so this means the subgraph itself paused on `interrupt()`.
- `drained` — the run was cooperatively stopped at a superstep boundary via `RunControl.request_drain()` (e.g. SIGTERM). The checkpoint is saved and the run is resumable.

### `metadata`

Optional descriptor of *what triggered* the subgraph. Only set on `started`, and only when the dispatching task's `input` matched a recognised per-call tool-dispatch shape with a string `id`.

Currently the only `type` is `"tool_call"`:

```json
{
  "type": "tool_call",
  "tool_call_id": "call_xyz"
}
```

- `type` — always present, identifies the metadata variant.
- `tool_call_id` — model-side id of the originating tool call (matches `AIMessage.tool_calls[i].id`). Use it to anchor the lifecycle event back to the AI message that dispatched the subgraph; from there, args (e.g. `subagent_type`, `description`) are read off `tool_calls[i].args` directly.

The langgraph layer deliberately doesn't mine args. They have a single source of truth on the AIMessage; duplicating them on `metadata` would couple the streaming layer to specific tool schemas and create a second copy that can drift.

`metadata` is absent for structurally-triggered subgraphs (parallel branches via `Send` with non-tool-call payloads, nested `graph.invoke()`, etc.) and for tool dispatches whose envelope carried no `id`.

### Identity correlation

`parent_task_id` is the join key for correlating `started` ↔ terminal events for the same subgraph and for matching a `lifecycle.started` back to its `tasks` parent. Each Send produces its own pregel task with its own id, so the join is 1:1 even when a model dispatches multiple parallel tool calls in one turn. The field is required on every event.

## Test plan

- [x] `TEST=tests/test_stream_lifecycle_transformer.py make test` — 27 passed (includes `test_parallel_dispatches_attributed_to_correct_parent` and `test_parent_task_id_present_on_every_terminal_path`)
- [x] `TEST=tests/test_stream_subgraph_transformer.py make test` — 27 passed
- [x] broader `(stream or transformer or lifecycle) and not postgres` — 418 passed
- [x] `make lint` (ruff + mypy) — clean
- [x] `make format` — clean


libs/langgraph/tests/test_stream_lifecycle_transformer.py
```diff
@@ -35,8 +35,16 @@ def _tasks_start(
...
+    input: Any = None,
 ) -> dict[str, Any]:
-    """Build a `tasks` ProtocolEvent carrying a TaskPayload (start)."""
+    """Build a `tasks` ProtocolEvent carrying a TaskPayload (start).
+
+    Pass `input={"tool_call": {"args": {...}}}` (or any envelope with
+    that shape) to exercise the lifecycle transformer's input mining of
+    invocation-intent metadata (`subagent_type`, `description`) — this
+    mirrors the `ToolCallWithContext` payload `langgraph.prebuilt.ToolNode`
+    Send-fans out per tool call.
```
sydney-runkle: we actually moved away from this payload type in the latest release, we still send tool calls, but not the tool call w/ context

libs/langgraph/langgraph/stream/transformers.py
```diff
@@ -340,6 +340,41 @@ def _parse_ns_segment(segment: str) -> tuple[str, str | None]:
     return name, task_id if sep else None
 
 
+def _extract_per_call_args(payload: Any) -> tuple[dict[str, Any], str | None] | None:
+    """Pull `(args, tool_call_id)` out of a per-call dispatched task's `input`.
+
+    Two shapes are recognised; both are duck-typed so 3rd-party tool
+    runners that mimic the layout participate without importing prebuilt
+    types:
```
sydney-runkle: tool call w/ context was internal

nick-hollon-lc: langgraph doesn't import prebuilt types still right?

libs/langgraph/tests/test_stream_lifecycle_transformer.py
nick-hollon-lc: so just drop the comment and then check what the current payload is?

libs/langgraph/langgraph/stream/transformers.py
sydney-runkle: why is this typed as subgraph status? shouldn't it be like LifecycleEvent or something?

libs/langgraph/langgraph/stream/transformers.py
```diff
@@ -353,6 +389,30 @@ class LifecyclePayload(TypedDict, total=False):
     namespace: list[str]
     graph_name: NotRequired[str]
     trigger_call_id: NotRequired[str]
```
sydney-runkle: can we have inline docs for all of these?

libs/langgraph/langgraph/stream/transformers.py
sydney-runkle: inline docs?

a little confused about PR docs saying "parent" re failed...

libs/langgraph/langgraph/stream/transformers.py
```diff
@@ -430,21 +505,68 @@ def process(self, event: ProtocolEvent) -> bool:
...
+            return
+        extracted = _extract_per_call_args(data.get("input"))
+        if extracted is None:
+            return
+        args, tool_call_id = extracted
+        metadata: dict[str, str] = {}
+        for key in ("subagent_type", "description"):
+            value = args.get(key)
+            if isinstance(value, str):
+                metadata[key] = value
```
sydney-runkle: too tightly coupled w/ deepagents `task` implementation

libs/langgraph/langgraph/stream/transformers.py
sydney-runkle: trigger_call_id seems like not the right name here?

sydney-runkle: why not required?

nick-hollon-lc: what would be better? task_id? pregel_task_id? dispatch_task_id?

libs/langgraph/langgraph/stream/transformers.py
```diff
@@ -349,11 +399,55 @@ class LifecyclePayload(TypedDict, total=False):
...
+    """Checkpoint namespace of the subgraph the event is about. Always present.
+
+    A list of `node_name:task_id` segments, one per nesting level (root has
+    `[]`, a direct child of root has `["agent:abc123"]`, a grandchild has
+    `["agent:abc123", "tool:def456"]`, etc.). Stable identity across the
+    `started → terminal` pair for the same subgraph instance.
+    """
     graph_name: NotRequired[str]
-    trigger_call_id: NotRequired[str]
+    """Compiled name of the subgraph (e.g. the `add_node` name in the parent).
```
sydney-runkle: eh not necessarily if it's a graph called w/in a node right

sydney-runkle: like we do for subagents in tools

libs/langgraph/langgraph/stream/run_stream.py
```diff
@@ -524,7 +524,7 @@ class _SubgraphRunStreamMixin:
     path: tuple[str, ...]
     graph_name: str | None
     trigger_call_id: str | None
-    status: SubgraphStatus
+    status: LifecycleEvent
```
sydney-runkle, line 527: this is confusing now?

nick-hollon-lc: i think it's ok. a subgraph run stream has a lifecycle event but a lifecycle event doesn't have a subgraph status. better this way

sydney-runkle (approved):

PR #7743, opened 2026-05-08
fix(checkpoint): specify allowed_objects in Reviver

ccurme:
On langchain-core / master you will see warnings from importing things like
```python
```
or
```python
```
> LangChainPendingDeprecationWarning: The default value of `allowed_objects` will change in a future version. Pass an explicit value (e.g., allowed_objects='messages' or allowed_objects='core') to suppress this warning.

Here we address the warning.

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
         AllowedMsgpackModules,
     )
 
-LC_REVIVER = Reviver()
+LC_REVIVER = Reviver(allowed_objects="all")
 EMPTY_BYTES = b""
 logger = logging.getLogger(__name__)
 
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
         AllowedMsgpackModules,
     )
 
-LC_REVIVER = Reviver(allowed_objects="all")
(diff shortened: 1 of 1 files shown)


libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -44,7 +44,7 @@
         AllowedMsgpackModules,
     )
 
-LC_REVIVER = Reviver()
+LC_REVIVER = Reviver(allowed_objects="core")
```
ccurme, line 47: this has defaulted to `"core"` since late December: https://github.com/langchain-ai/langchain/pull/34455

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
```diff
@@ -44,7 +44,7 @@
         AllowedMsgpackModules,
     )
 
-LC_REVIVER = Reviver()
+LC_REVIVER = Reviver(allowed_objects="all")
 EMPTY_BYTES = b""
 logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
ccurme: Resolved by setting to `"core"` (default for last 5 months).

eyurtsev (approved):

libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
eyurtsev, line 47: Somewhere above that could you put a reference to the trust model for JSONPlus serializer

eyurtsev: ah it's fine in this context -- used within jsonplus serializer

PR #7746, opened 2026-05-08
feat(checkpoint): force delta channel snapshot after max supersteps since last snapshot

longquanzheng:
## Summary

Add a system-wide upper bound on supersteps-since-last-snapshot for `DeltaChannel`, preventing unbounded ancestor walks on long-lived threads where a delta channel stops receiving writes.

**Problem:** If a delta channel is written a few times (below `snapshot_frequency`) and then never written again, it is never snapshotted. Every subsequent run triggers an ancestor walk that grows linearly with thread length — on long threads this becomes catastrophic.

**Solution:** Track a second counter (total supersteps) per delta channel alongside the existing update count. Force a snapshot when EITHER `updates >= snapshot_frequency` OR `supersteps >= DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT` (default 5000, overridable via env `LANGGRAPH_DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT`).

### Changes

- **`checkpoint` lib**: Rename metadata field `delta_updates_since_snapshot: dict[str, int]` -> `counters_since_last_snapshot: dict[str, tuple[int, int]]` where index 0 = updates, index 1 = supersteps.
- **`langgraph/_internal/_config.py`**: Add `DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT` constant with env override.
- **`langgraph/pregel/_checkpoint.py`**: Update `delta_channels_to_snapshot()` predicate to fire on either threshold. Rename reader helper to `read_counters_since_last_snapshot()`.
- **`langgraph/pregel/_loop.py`**: Iterate all delta channels each superstep (not just updated ones) to bump the supersteps counter. Reset both counters to `(0, 0)` on snapshot.
- **Tests**: Updated existing exit-mode tests for new field shape. Added 4 new tests covering forced snapshot (single run + multi-run accumulation), predicate unit test, and counter reset.

## Test plan

- [x] `test_delta_channel_supersteps_bound.py` — 4 new tests all pass
- [x] `test_delta_channel_exit_mode.py` — 11 existing tests updated and pass
- [x] `test_delta_channel_migration.py` — 11 tests pass
- [x] `test_channels.py` — 29 tests pass
- [x] `test_pregel.py` — 457 tests pass
- [x] `libs/checkpoint` test suite — 151 pass, 16 skipped
- [x] `make lint` clean (langgraph + checkpoint)

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/base/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/base/__init__.py
@@ -60,7 +60,7 @@ class CheckpointMetadata(TypedDict, total=False):
     """
     run_id: str
     """The ID of the run that created this checkpoint."""
-    counters_since_last_snapshot: dict[str, tuple[int, int]]
+    counters_since_delta_snapshot: dict[str, tuple[int, int]]
     """Per-channel counters since the last `_DeltaSnapshot` was written.
 
     !!! warning "Beta"
--- a/libs/checkpoint/langgraph/checkpoint/base/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/base/__init__.py
@@ -60,20 +60,29 @@ class CheckpointMetadata(TypedDict, total=False):
     """
     run_id: str
     """The ID of the run that created this checkpoint."""
-    delta_updates_since_snapshot: dict[str, int]
-    """Per-channel update count since the last `_DeltaSnapshot` was written.
+    counters_since_last_snapshot: dict[str, tuple[int, int]]
+    """Per-channel counters since the last `_DeltaSnapshot` was written.
 
     !!! warning "Beta"
 
         This metadata field backs `DeltaChannel` (beta). The key name and
         contents may change while the delta-channel design stabilizes.
 
-    Maps channel name → number of supersteps that wrote to this channel
-    since its last snapshot blob. Used by `pregel.create_checkpoint` to
-    decide when to write the next snapshot (when the count reaches the
-    channel's `snapshot_frequency`, snapshot fires and the count resets
-    to 0). Absent on threads that don't use delta channels. Version-format
-    independent — works for int, float, and string version schemes.
+    Maps channel name -> `(updates, supersteps)`:
+
+    - index 0 (`updates`): number of supersteps that wrote to this channel
+      since its last snapshot blob.
+    - index 1 (`supersteps`): total supersteps elapsed since this channel's
+      last snapshot, regardless of whether the channel was written.
+
+    A snapshot fires when EITHER `updates >= ch.snapshot_frequency` OR
+    `supersteps >= DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT` (system-wide bound,
+    default 5000, env `LANGGRAPH_DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT`).
+    The supersteps bound prevents unbounded ancestor walks on threads where
+    a delta channel exists but is no longer being updated.
+
+    Absent on threads that don't use delta channels. Persisted as a
+    2-element list in JSON (no native tuple).
     """
 
 
--- a/libs/langgraph/langgraph/pregel/_checkpoint.py
+++ b/libs/langgraph/langgraph/pregel/_checkpoint.py
@@ -37,7 +37,7 @@ def empty_checkpoint() -> Checkpoint:
 
 def delta_channels_to_snapshot(
     channels: Mapping[str, BaseChannel],
-    counters_since_last_snapshot: Mapping[str, tuple[int, int]],
+    counters_since_delta_snapshot: Mapping[str, tuple[int, int]],
 ) -> set[str]:
     """Return the set of DeltaChannel names that should snapshot now.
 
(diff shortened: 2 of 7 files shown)


libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -60,20 +60,29 @@ class CheckpointMetadata(TypedDict, total=False):
     """
     run_id: str
     """The ID of the run that created this checkpoint."""
-    delta_updates_since_snapshot: dict[str, int]
-    """Per-channel update count since the last `_DeltaSnapshot` was written.
+    counters_since_last_snapshot: dict[str, tuple[int, int]]
+    """Per-channel counters since the last `_DeltaSnapshot` was written.
```
sydney-runkle: let's use a named tuple?

sydney-runkle: should still have delta in the name maybe?

libs/langgraph/langgraph/pregel/_checkpoint.py
```diff
@@ -233,15 +239,16 @@ def copy_checkpoint(checkpoint: Checkpoint) -> Checkpoint:
     )
 
 
-def read_delta_updates_since_snapshot(
+def read_counters_since_last_snapshot(
```
sydney-runkle: seems like overkill for a helper function

libs/langgraph/tests/test_delta_channel_supersteps_bound.py
```diff
@@ -0,0 +1,203 @@
...
+    ch.key = "x"
+    ch_instance = ch.from_checkpoint(None)
+
+    channels = {"x": ch_instance}
+    counters: dict[str, tuple[int, int]] = {"x": (0, 5000)}
+
+    with patch(
+        "langgraph.pregel._checkpoint.DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT",
+        5000,
+    ):
```
sydney-runkle: why do we have to patch this?

sydney-runkle (changes requested):
pattern seems reasonable, but some thoughts

i'm not super worried about this edge case tbh, idk why you'd have a delta channel that's not getting updates very frequently

longquanzheng:
> [libs/langgraph/tests/test_delta_channel_supersteps_bound.py](/langchain-ai/langgraph/pull/7746/files/317d04555caf7a13be0d793b37a9c99a2ba01f44#diff-2a70cf7b307cf1de62b868d6f1d519fb50d80c8ed1252b7cd4d5c1ead80c2158)

We never know. 
it could be read only after certain writes. 
and you don't want to fail in those cases as a bug which will be too late to fix later.

libs/langgraph/tests/test_delta_channel_supersteps_bound.py
longquanzheng: removed

libs/langgraph/langgraph/pregel/_checkpoint.py
longquanzheng: yeah but it's exsting code from your prev PR that I thought you have reasons to do. I can remove.

libs/checkpoint/langgraph/checkpoint/base/__init__.py
longquanzheng: I want to avoid super long name, maybe counters_since_delta_snapshot?

longquanzheng: landed onto `counters_since_delta_snapshot`

libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -60,20 +60,29 @@ class CheckpointMetadata(TypedDict, total=False):
     """
     run_id: str
     """The ID of the run that created this checkpoint."""
-    delta_updates_since_snapshot: dict[str, int]
-    """Per-channel update count since the last `_DeltaSnapshot` was written.
+    counters_since_delta_snapshot: dict[str, tuple[int, int]]
```
sydney-runkle: can we use a named tuple?

sydney-runkle (approved):
i think named tuple will clean up, otherwise LGTM

longquanzheng:
> i think named tuple will clean up, otherwise LGTM

reverted back to tuple bc our jsonplus component actually convert it to dict and add eiter complexity or storage overhead.

Sourav-Nandy-ai:
Hi Quanzheng,

This is exactly the type of state-boundary change where I wanted a real before-and-after test rather than just a read of the diff.

I built two isolated langgraph installs bracketing this PR's merge (1.2.0a7 pre-merge and 1.2.0 post-merge) and ran a real graph with a channel that is written once and never touched again for 10, 25, and 43 supersteps.

The Results
- Pre-fix: That channel is never force-snapshotted, no matter how long the thread runs (0/0/0 across all three lengths).
- Post-fix: The channel gets forced to a full snapshot every DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT supersteps regardless of writes (2, 5, and 9 snapshots respectively).
- This confirms the fix does exactly what it says for a long-running, sparsely updated thread.

I maintain agent-eval (agent_regress), an OSS stats and regression harness for agent behavior. You can find my full script and raw output here: https://github.com/RudrenduPaul/agent-eval.

Given you wrote the fix, I am curious whether this matches your own mental model of the bound, or if there is an edge case (such as multiple sparse channels with different frequencies) that I should test next.

Best,
Sourav

PR #7747, opened 2026-05-08
feat(langgraph): add set_node_defaults() to StateGraph

longquanzheng:
Add `StateGraph.set_node_defaults()` — a fluent builder method for setting graph-wide node policies in one place. Per-node values from `add_node()` always take precedence. Defaults are applied at `compile()` time.

```python
graph = (
    StateGraph(State)
    .set_node_defaults(
        retry_policy=RetryPolicy(max_attempts=3),
        error_handler=my_fallback_handler,
        timeout=TimeoutPolicy(run_timeout=30),
    )
    .add_node("a", node_a)
    .add_node("b", node_b, retry_policy=custom)  # overrides default
    .add_edge(START, "a")
    .compile()
)
```

Error handlers are never invoked on error-handler nodes themselves — handler failures fail the run. Not inherited by subgraphs.

## Supported defaults

| `set_node_defaults()` kwarg | Fallback for `add_node(...)` kwarg | Applies to error-handler nodes? |
|---|---|---|
| `retry_policy` | `retry_policy` | Yes |
| `cache_policy` | `cache_policy` | No — caching handler results is unsafe |
| `error_handler` | `error_handler` | No — handlers must never catch themselves |
| `timeout` | `timeout` | Yes |

## Changes

- `libs/langgraph/langgraph/graph/state.py` — new `_NodeDefaults` dataclass, `set_node_defaults()` method on `StateGraph`; `compile()` applies builder defaults to node specs with per-branch rules for which defaults apply to error-handler nodes.
- `libs/langgraph/tests/test_retry.py` — 14 new tests covering all four policy types, per-node override precedence, chaining, combined retry+handler, handler exclusion, `RunnableConfig` injection, and name collision.

## Verification

`make format`, `make lint`, `make test` all passing in `libs/langgraph`.

### DIFF
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -279,21 +279,30 @@ class StateGraph(Generic[StateT, ContextT, InputT, OutputT]):
         """Set default node policies that apply to every node in this graph.
 
         Per-node values passed to `add_node` always take precedence over these
-        defaults. Policies set here are **not** inherited by subgraphs.
+        defaults. Defaults are applied at `compile()` time. Policies set here
+        are **not** inherited by subgraphs.
+
+        `retry_policy` and `timeout` defaults apply to **all** nodes,
+        including error-handler nodes. `cache_policy` and `error_handler`
+        defaults only apply to regular nodes -- caching error-handler results
+        is unsafe, and handlers must never catch themselves.
 
         Args:
             retry_policy: Default retry policy for nodes that don't specify
-                their own via `add_node(..., retry_policy=...)`.
+                their own via `add_node(..., retry_policy=...)`. Also applies
+                to error-handler nodes.
             cache_policy: Default cache policy for nodes that don't specify
-                their own via `add_node(..., cache_policy=...)`.
+                their own via `add_node(..., cache_policy=...)`. Does **not**
+                apply to error-handler nodes.
             error_handler: Default error handler invoked when any regular node
                 raises and does not have its own `error_handler` set via
                 `add_node`. The handler is **not** invoked when an
                 error-handler node itself raises -- handler failures fail the
                 run.
             timeout: Default timeout policy for nodes that don't specify their
-                own via `add_node(..., timeout=...)`. Accepts a `TimeoutPolicy`,
-                a number of seconds (`float`), or a `timedelta`.
+                own via `add_node(..., timeout=...)`. Also applies to
+                error-handler nodes. Accepts a `TimeoutPolicy`, a number of
+                seconds (`float`), or a `timedelta`.
 
         Returns:
             Self: The builder instance, for chaining.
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/graph/state.py
```diff
@@ -1094,6 +1094,7 @@ def compile(
         debug: bool = False,
         name: str | None = None,
         transformers: Sequence[Callable[[tuple[str, ...]], Any]] | None = None,
+        default_error_handler: StateNode[Any, ContextT] | None = None,
```
sydney-runkle: ok actually, i think might be cleaner in `StateGraph.__init__`

that way we can just manage this in `add_node` :)

can we TAL at that alternative?

sydney-runkle (commented):
one other idea -- do we also want to support:
* default node retry policy?
* default node cache policy?

if we do, do we want a cleaner API (not just 3 new args) 

and another note, i've heard requests for graph level timeouts as well... that would be implemented differently though (it's not a node default)

libs/langgraph/langgraph/graph/state.py
longquanzheng: great feedback. change to `set_defaults` for now.

libs/langgraph/langgraph/graph/state.py
```diff
@@ -251,10 +251,71 @@ def __init__(
...
+        self._default_retry_policy: RetryPolicy | Sequence[RetryPolicy] | None = None
+        self._default_cache_policy: CachePolicy | None = None
+        self._default_error_handler: StateNode[Any, ContextT] | None = None
+        self._default_timeout: TimeoutPolicy | None = None
+
         self._add_schema(self.state_schema)
         self._add_schema(self.input_schema, allow_managed=False)
         self._add_schema(self.output_schema, allow_managed=False)
 
+    def set_defaults(
```
sydney-runkle: `set_node_defaults`?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -251,10 +251,71 @@ def __init__(
         self.output_schema = cast(type[OutputT], output_schema or state_schema)
         self.context_schema = context_schema
 
+        self._default_retry_policy: RetryPolicy | Sequence[RetryPolicy] | None = None
```
sydney-runkle: self._node_defaults.retry_policy? and similar

libs/langgraph/langgraph/graph/state.py
```diff
@@ -1094,6 +1155,7 @@ def compile(
         debug: bool = False,
         name: str | None = None,
         transformers: Sequence[Callable[[tuple[str, ...]], Any]] | None = None,
+        default_error_handler: StateNode[Any, ContextT] | None = None,
```
sydney-runkle: let's remove this arg to `compile` and just expose the `set_defaults` method

sydney-runkle (dismissed):

sydney-runkle:
question here is where we want to apply defaults
if we do it in compile, then the order of `set_node_defaults` and `add_node` doesn't matter
but it feels cleaner if when you add a node, the policy gets attached (so you could inspect the nodes and see)
but i don't really want to enforce this in 2 places

sydney-runkle:
if we supported on init as opposed to a method, we could enforce at node addition time

sydney-runkle (approved):
a few things that might be odd here -- nodes don't have the correct retry policy / error handler / timeout until compile time :/

PR #7773, opened 2026-05-12
feat(langgraph): durable error-handler resume across host crashes

longquanzheng:
## Summary

- **Consolidate error writes:** When a node fails and has an error handler, `commit()` now appends both `ERROR` and `ERROR_SOURCE_NODE` in a single `put_writes` call, eliminating the redundant overwrite that `schedule_error_handler` used to do.
- **Ensure durability before handler execution:** Reuses the `_delta_write_futs` pattern — a new `_error_handler_write_futs` list collects the persistence future from `put_writes` when `ERROR_SOURCE_NODE` is written, and `schedule_error_handler` / `aschedule_error_handler` drain it (sync: `concurrent.futures.wait`, async: `asyncio.gather`) before preparing the handler task.
- **Resume directly to error handler:** Adds `_resume_error_handlers_if_applicable()` to `PregelLoop`, called from `tick()` after `_reapply_writes_to_succeeded_nodes()`. On resume, it detects `ERROR_SOURCE_NODE` markers in `checkpoint_pending_writes`, marks the original task as done (so the runner skips it), and schedules a fresh handler task.
- **Rename internal methods for clarity:** `_match_writes` → `_reapply_writes_to_succeeded_nodes` (makes it clear that failed/interrupted tasks are skipped); `_resume_error_handlers` → `_resume_error_handlers_if_applicable`.

## Test plan

- [x] `test_error_handler_resumes_after_crash`: single node fails, handler crashes, resume re-runs the handler (not the original node). Verifies `NodeError.node` and error content survive checkpoint round-trip.
- [x] `test_error_handler_resumes_after_crash_multiple_nodes`: two nodes fail concurrently in the same superstep, each with its own handler. Verifies error handler starts while other nodes are still in-flight (via `threading.Event`), and on resume both handlers re-run with correct `NodeError.node` and error content.
- [x] All 101 tests in `test_retry.py` pass (including 19 error-handler tests).
- [x] `make format` + `make lint` clean.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -480,6 +480,9 @@ class PregelLoop:
                 isinstance(self.specs.get(c), DeltaChannel) for c, _ in writes_to_save
             ):
                 self._delta_write_futs.append(fut)
+            # ERROR_SOURCE_NODE is only appended by commit() when the task
+            # has an error handler (_should_route_to_error_handler), so this
+            # check naturally limits future collection to those tasks.
             if self._error_handler_write_futs is not None and any(
                 c == ERROR_SOURCE_NODE for c, _ in writes
             ):
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -203,6 +203,12 @@ class PregelLoop:
     # `__enter__`; stays `None` only when no checkpointer.
     _delta_write_futs: list[Any] | None = None
 
+    # Same pattern as `_delta_write_futs` but for error-handler writes.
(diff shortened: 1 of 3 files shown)


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -698,13 +709,88 @@ async def amatch_cached_writes(self) -> Sequence[PregelExecutableTask]:
 
     # private
 
-    def _match_writes(self, tasks: Mapping[str, PregelExecutableTask]) -> None:
+    def _reapply_writes_to_succeeded_nodes(
```
longquanzheng, line 715: @sydney-runkle just a rename! hope you are okay we I some minor cleanup/refactor as we touch the code (if we feel safe)

sydney-runkle (approved):

PR #7799, opened 2026-05-15
Cap jittered retry sleep at max interval

MukundaKatta:
## Summary

Cap jittered retry sleeps at the configured max interval.

When retry jitter pushes a computed delay over the max interval, the retry loop should still honor the configured cap. This patch applies the cap after jitter and adds regression coverage for the capped path.

## Validation

Attempted focused test:

- `NO_DOCKER=true PYTHONPATH=$PWD/libs/langgraph:$PWD/libs/checkpoint:$PWD/libs/checkpoint-sqlite uv run --with pytest --with redis --with ormsgpack --with pytest-mock --with langchain-core --with aiosqlite python -m pytest libs/langgraph/tests/test_retry.py -q`

This local checkout's test import path still requires additional optional checkpoint/postgres modules before test collection, so the focused test could not collect locally. The change itself is covered by the added unit test in `libs/langgraph/tests/test_retry.py`.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_retry.py
+++ b/libs/langgraph/langgraph/pregel/_retry.py
@@ -631,7 +631,9 @@ def run_with_retry(
 
             # Apply jitter if configured
             sleep_time = (
-                interval + random.uniform(0, 1) if matching_policy.jitter else interval
+                min(matching_policy.max_interval, interval + random.uniform(0, 1))
+                if matching_policy.jitter
+                else interval
             )
             time.sleep(sleep_time)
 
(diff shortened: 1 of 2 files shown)


salimassili62-afk:
This fix makes sense. The configured max interval should be a hard upper bound on sleep time, even after jitter is applied.

The order of operations is important here:

```python
delay = compute_backoff(...)
delay = apply_jitter(delay)
delay = min(delay, max_interval)
```

If the cap is applied before jitter, the effective sleep can still exceed the user’s configured maximum, which makes retry behavior harder to reason about in production.

The regression test is useful too, since this is exactly the kind of edge case that looks correct in normal retry paths but breaks when jitter lands near the upper bound.

PR #7818, opened 2026-05-15
feat(sdk-py): add v3 streaming primitives and SSE transport

nick-hollon-lc:
## Summary
- Add the initial v3 streaming package structure.
- Add subscription matching, replay buffer, fake streaming server helpers, and async SSE transport.

## Testing
- Not run while opening draft PR stack.

### DIFF
--- a/libs/langgraph/uv.lock
+++ b/libs/langgraph/uv.lock
@@ -1370,14 +1370,14 @@ wheels = [
 
 [[package]]
 name = "langchain-protocol"
-version = "0.0.14"
+version = "0.0.15"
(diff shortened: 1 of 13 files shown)


jdrogers940 (approved):

aswego123:
The max_queue_size fix bounds the queue, but what happens when the queue is full -does put_nowait raise QueueFull and kill the stream silently, or is there a backpressure signal surfaced to the caller? From SSE reconnect semantics, the client should ideally get a recoverable error rather than a hard crash.

PR #7823, opened 2026-05-15
feat(sdk-py): add messages and tool call projections

nick-hollon-lc:
## Summary
- Add root messages and tool call projections for async thread streams.
- Add protocol event builders and focused projection coverage.

## Testing
- Not run while opening draft PR stack.


jdrogers940 (approved):
Do we know why the test is failing?

nick-hollon-lc:
> Do we know why the test is failing?

Some weird flake I think we two PRs kick off checks at same time

PR #7882, opened 2026-05-21
feat(langgraph): add `before_builtins` opt-in for stream transformers

nick-hollon-lc:
Adds a `before_builtins: ClassVar[bool] = False` flag on `StreamTransformer`. When `True`, the mux registers the transformer ahead of the rest, preserving relative order within each lane.

## Motivation

Content-mutating transformers — PII redaction, content filters, profanity scrubbers — need to run before built-in transformers like `MessagesTransformer`, which eagerly snapshots text fields (`delta.text`, `delta.reasoning`) into its projection. Today there's no way to register a transformer ahead of the built-ins, so any mutation a user transformer makes to `delta.text` lands too late: `MessagesTransformer` has already pushed the original string into the `ChatModelStream` text accumulator.

Motivating use case: PII redaction middleware in langchain (see langchain-ai/langchain#37591). Stream-level redaction wants to mutate `delta.text` before any consumer projection captures it. Without `before_builtins`, the only workable Python-side approach is `wrap_model_call`, which is heavier and only catches model output (not tool deltas, custom events, or subgraph outputs).

## API

```python
class _PIIRedactor(StreamTransformer):
    before_builtins: ClassVar[bool] = True

    def process(self, event):
        # mutate delta.text in place; MessagesTransformer sees the redacted string
        ...
```

The flag is a class attribute so transformers carry the placement decision with them — callers wiring middleware don't need to know whether the transformer needs pre-lane placement; the transformer itself declares it.

## Ordering contract

Within each lane, the supplied order is preserved. The full registration order ends up as:

1. All `before_builtins=True` factories, in supplied order
2. All other factories, in supplied order  
3. Pre-built `transformers=` instances, partitioned the same way

Built-ins keep the default `before_builtins=False`, so registration order is identical to before for anyone who hasn't opted in. No backwards-compatibility break.

## Foot-gun (documented on the class)

Pre-lane transformers see `tasks` events before `LifecycleTransformer` and `SubgraphTransformer` consume them. Mutating `event["params"]["namespace"]` or the data dict's `id` / `result` / `error` / `interrupts` fields will desync their bookkeeping. Observe freely; mutate only fields no built-in reads.

The canonical safe mutation is `delta.text` (and `delta.reasoning`) on `messages` events — exactly the case content-filter transformers need.

## Tests

`tests/test_stream_before_builtins.py` covers:

- Pre-lane factories register before others, and run first on dispatch
- Within-lane order is preserved
- A pre-lane redactor is registered before `MessagesTransformer` even when supplied second
- A pre-lane observer doesn't break `LifecycleTransformer`'s bookkeeping (read-only observation is safe)
- Default is `False` on the base class and all built-ins
- **End-to-end**: a pre-lane redactor mutating `delta.text` results in the redacted string landing in `MessagesTransformer`'s text accumulator — the actual proof that pre-lane placement does what it's supposed to do

Existing stream-transformer tests (`test_stream_data_transformers`, `test_stream_messages_transformer`, `test_stream_lifecycle_transformer`, `test_stream_subgraph_transformer` — 133 tests) all still pass.

### DIFF
--- a/libs/langgraph/langgraph/stream/_mux.py
+++ b/libs/langgraph/langgraph/stream/_mux.py
@@ -23,6 +23,26 @@ transformers can close over their config:
 """
 
 
+def _partition_before_builtins(
+    transformers: list[StreamTransformer],
+) -> list[StreamTransformer]:
+    """Stable partition: `before_builtins = True` first, others after.
+
+    Preserves the relative order within each lane. Used during
+    registration so transformers that opt in via the
+    `before_builtins` class attribute see events ahead of built-in
+    transformers supplied later in the same factory list.
+    """
+    pre: list[StreamTransformer] = []
+    rest: list[StreamTransformer] = []
+    for transformer in transformers:
+        if getattr(transformer, "before_builtins", False):
+            pre.append(transformer)
+        else:
+            rest.append(transformer)
+    return [*pre, *rest]
+
+
 class StreamMux:
     """Central event dispatcher for the streaming infrastructure.
 
@@ -61,6 +81,14 @@ class StreamMux:
         transformer's `init()` is called, projections are merged into
         `extensions`, `_native` keys are recorded in `native_keys`, and
         any StreamChannel instances are bound and (if named) wired.
+
+        Transformers with `StreamTransformer.before_builtins = True` are
+        registered ahead of the rest, preserving relative order within
+        each lane. This lets content-mutating transformers (PII
+        redaction, content filters, etc.) run before built-ins like
+        `MessagesTransformer` that eagerly snapshot text fields into
+        their projections. See `StreamTransformer.before_builtins` for
+        the contract and foot-guns.
 
         Args:
             transformers: Already-built transformer instances. Registered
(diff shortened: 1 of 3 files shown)


libs/langgraph/langgraph/stream/_mux.py
```diff
@@ -23,6 +23,26 @@
...
+    Preserves the relative order within each lane. Used during
+    registration so transformers that opt in via the
+    `before_builtins` class attribute see events ahead of built-in
+    transformers supplied later in the same factory list.
+    """
+    pre: list[StreamTransformer] = []
+    rest: list[StreamTransformer] = []
+    for transformer in transformers:
+        if getattr(transformer, "before_builtins", False):
+            pre.append(transformer)
```
sydney-runkle: shouldn't we just have this?

sydney-runkle: could hopefully do inline if so?

sydney-runkle (approved):

libs/langgraph/langgraph/stream/_mux.py
nick-hollon-lc: done inline

PR #7913, opened 2026-05-26
fix(langgraph): assign stable IDs to id=None BaseMessages before DeltaChannel checkpoint writes

sydney-runkle:
## TL;DR

Applications depend on stable message IDs — LangSmith traces, message views, and `RemoveMessage` all break when the same message gets a different ID on every load.

ID assignment has historically lived in the `add_messages` reducer, but that's the wrong place: reducer logic runs *after* checkpoint serialisation has already started. The correct long-term fix is to assign IDs in `BaseMessage.__init__`, but that requires changes to `langchain-core` equality semantics first.

**This PR is the non-breaking interim fix:** assign IDs to `BaseMessage` objects in `put_writes()`, before the DeltaChannel write is handed to the background serialiser. This unblocks applications now without touching `BaseMessage`.

---

## Problem

In the default durability mode, `put_writes()` submits checkpoint writes to a background thread before `apply_writes()` runs. For DeltaChannel writes containing `BaseMessage` objects with `id=None`, the background thread may serialise `id=None` before the reducer ever sees them. Every `get_state()` call then replays `id=None` and assigns a fresh UUID — the same message has a different ID on every load and across every resumed invocation.

## Fix

Add `ensure_message_ids(value)` to `pregel/_messages.py` and call it in `put_writes()` for DeltaChannel writes **before** submitting to the background executor. This assigns UUIDs synchronously so the serialised bytes always carry a stable ID.

Both `SyncPregelLoop` and `AsyncPregelLoop` delegate to `super().put_writes()`, so both paths are covered with one change.

## Long-term

ID assignment belongs in `BaseMessage.__init__`. That requires first excluding `id` from `BaseMessage.__eq__` (currently `HumanMessage("x") != HumanMessage("x")` if both get different auto-assigned UUIDs). That is a `langchain-core` change tracked separately.

## Changes

- `pregel/_messages.py` — `ensure_message_ids(value)`: assigns a UUID to any `id=None` `BaseMessage`, handles single messages and lists
- `pregel/_loop.py` — call it for DeltaChannel writes in `put_writes()`
- `tests/test_delta_channel_id_stability.py` — two tests (sync + async) using a plain append reducer (no ID logic), confirmed to **fail on `main`** and pass with this fix

Related deepagents PR: https://github.com/langchain-ai/deepagents/pull/3590

### DIFF
--- a/libs/langgraph/langgraph/pregel/_messages.py
+++ b/libs/langgraph/langgraph/pregel/_messages.py
@@ -11,6 +11,7 @@ from uuid import UUID, uuid4
 
 from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.messages import BaseMessage, ToolMessage
+from langchain_core.messages.utils import convert_to_messages
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, LLMResult
 from pydantic import BaseModel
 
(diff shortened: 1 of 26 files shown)


libs/langgraph/langgraph/pregel/_messages.py
```diff
@@ -405,3 +405,24 @@ def on_stream_event(
...
+    Called in put_writes() before DeltaChannel writes are submitted to the
+    checkpointer. Without this, a reducer that assigns IDs inside apply_writes()
+    races with the background serialisation thread: the checkpoint may capture
+    id=None, and every get_state() replay produces a different UUID.
+
+    Mutating here (rather than in the reducer or after apply_writes) is safe
+    because put_writes() runs synchronously before the background thread
+    starts, so the serialised bytes always see the assigned ID.
+    """
+    if isinstance(value, BaseMessage):
```
eyurtsev, line 446: any other shapes we need to handle?

libs/langgraph/langgraph/pregel/_messages.py
```diff
@@ -405,3 +405,24 @@ def on_stream_event(
                     self.seen.add(msg_id)
             v2_meta = {**meta[1], "run_id": str(run_id)}
             self.stream((meta[0], "messages", (event, v2_meta)))
+
+
+def ensure_message_ids(value: Any) -> None:
```
eyurtsev, line 426: are values guarnatees to be BaseMessage we don't have to worry about dict like representation?

libs/langgraph/langgraph/pregel/_messages.py
```diff
@@ -405,3 +406,52 @@ def on_stream_event(
...
+    - Dicts with a known "role" (OpenAI-style) or "type" (LangChain format):
+      coerce to a typed BaseMessage via convert_to_messages, then assign a UUID
+      if needed. The list element is replaced in-place so the shared list
+      reference seen by checkpoint_pending_writes and the background thread both
+      get the typed message.
+
+    Mutating synchronously here (before the background thread is submitted) is
+    safe: the serialised bytes always reflect the post-coercion state.
+    """
+    if isinstance(value, BaseMessage):
```
eyurtsev, line 446: Can value only be BaseMessage or list at this level? Do we need to worry about dict-like Message for value or a dict for value with nested messages?

eyurtsev (approved):

PR #7926, opened 2026-05-27
fix(langgraph): merge instead of overwrite in `ensure_config` for callbacks, tags, metadata, configurable

### LINKED ISSUE #7441: bug: ensure_config mutates shared metadata when copiable keys passed by reference
### Checked other resources

### Related Issues / PRs

_No response_

```python
from langgraph._internal._config import ensure_config

original_config = {"metadata": {"ls_integration": "langchain_create_agent"}}
new_config = {"configurable": {"thread_id": "thread-1"}}
new_config = ensure_config(original_config, new_config)

# Bug: original_metadata is mutated — thread_id leaks from configurable into metadata
print("original_config after ensure_config:", original_config)
# Expected: {'metadata': {'ls_integration': 'langchain_create_agent'}}
# Actual:   {'metadata': {'ls_integration': 'langchain_create_agent'}, 'thread_id': 'thread-1'}
assert original_config == {"metadata": {"ls_integration": "langchain_create_agent"}}, (
    f"BUG: original config was mutated to {original_config}"
)
```

```shell

```

### Description

* I'm trying to use `langgraph` with `create_agent` from `langchain` >= 1.2.13
  (which introduced langchain-ai/langchain#35810). That PR sets a shared metadata
  dict `{"ls_integration": "langchain_create_agent"}` on the compiled graph's
  base config. When invoking the agent with different `thread_id` values via
  `configurable`, I expect each invocation to have its own isolated metadata.

* I expect to see that the base config's `metadata` dict remains
  `{"ls_integration": "langchain_create_agent"}` after each invocation,
  regardless of the `thread_id` passed.

* Instead, `ensure_config` assigns the `metadata` key by reference (without
  calling `.copy()`), so `configurable` values like `thread_id` get injected
  into the **shared** metadata dict of the base config. After the first
  invocation, subsequent calls see `{"ls_integration": "langchain_create_agent",
  "thread_id": "<previous_thread_id>"}` — data from a previous invocation leaks
  into the base config, contaminating all future calls.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.26200
> Python Version:  3.13.7 (tags/v3.13.7:bcee1c3, Aug 14 2025, 14:15:11) [MSC v.1944 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 1.2.27
> langchain: 1.2.15
> langsmith: 0.7.26
> langchain_mcp_adapters: 0.2.2
> langchain_openai: 1.1.12
> deepagents
> deepagents-cli

Other Dependencies
------------------
> httpx: 0.28.1
> jsonpatch: 1.33
> langgraph: 1.1.6
> mcp: 1.27.0
> openai: 2.30.0
> opentelemetry-api: 1.40.0
> opentelemetry-sdk: 1.40.0
> orjson: 3.11.8
> packaging: 26.0
> pydantic: 2.12.5
> pytest: 9.0.3
> pyyaml: 6.0.3
> requests: 2.33.1
> requests-toolbelt: 1.0.0
> tenacity: 9.1.4
> tiktoken: 0.12.0
> typing-extensions: 4.15.0
> uuid-utils: 0.14.1
> wrapt: 1.17.3
> xxhash: 3.6.0
> zstandard: 0.25.0

nick-hollon-lc:
## Summary

`ensure_config` did full overwrite for `callbacks`, `tags`, `metadata`, and `configurable` when merging multiple configs (e.g. `Pregel.stream` calls `ensure_config(self.config, config)`). This caused silent data loss: values bound via `.with_config({...})` were dropped whenever a later config supplied any other key in the same field.

`merge_configs` in the same file already does the correct per-key merge for these fields. This PR brings `ensure_config` into agreement, factoring out a small `_merge_callbacks` helper that both functions now share (consolidating the only piece of meaningful duplication between them).

## What this fixes

Concrete symptoms documented in this commit chain:
- `with_config({"callbacks": [cb]}) → astream_events(...)` silently drops `cb` (internal handler overwrites).
- `with_config({"configurable": {"ls_agent_type": "root"}}) → invoke({}, {"configurable": {"thread_id": "X"}})` silently drops `ls_agent_type` (so no root-level runs were tagged in LangSmith).
- Same shape for `metadata` and `tags`.
- The `metadata` shallow-merge also resolves the by-reference mutation bug where `ensure_config`'s metadata-propagation step would mutate the shared base config's metadata dict across invocations.

Downstream, this collapses the `_build_subagent_config` workaround in Deep Agents (which exists purely to defeat the `configurable` overwrite) and unblocks the long-standing "forward parent `metadata` into subagent invoke" request.

Closes #7441.

Related (likely benefit from this fix, worth verifying after merge):
- langchain-ai/langchain#37248 — `configurable` field lost in `ToolRuntime.config` when using tool_interceptors
- langchain-ai/deepagents#3634 — forward parent `metadata` into subagent invoke config (request that was blocked by this bug)
- langchain-ai/deepagents#2362 — `SubAgentMiddleware` runtime config not passed to subagent invocations

## Why this is safe

- **No API change.** `ensure_config` signature unchanged. No public types added or removed.
- **Behavior change is purely additive in correctness.** The fix preserves information that was previously being dropped. Anyone using `ensure_config` correctly sees identical results OR sees additional keys that should have been there.
- **All four behaviors already exist in `merge_configs`.** This closes a divergence between two functions in the same file, not introduces one.
- **Call-site audit:** 18 direct `ensure_config` call sites in `libs/` reviewed. 14 unaffected (single-config calls or only set keys we don't change), 4 were latent bugs now fixed (the `Pregel.stream`/`astream` and `bulk_update_state` paths), 0 depended on overwrite.
- **Test suites:** full langgraph monorepo (langgraph + checkpoint + checkpoint-sqlite + checkpoint-postgres + prebuilt + cli + sdk-py) — 3,257+ tests, all passing, zero adjustments required.
- **Downstream verification:** ran tests on the langchainplus sub-projects that depend on core `langgraph` (`smith-issues-agent`, `agent-builder-graphs`), comparing baseline (a worktree of `origin/main`) vs this branch. Verified the editable install pointed to the correct code in each case via `import langgraph; __path__`. Result: `smith-issues-agent` 246/246 passed identically on both; `agent-builder-graphs` 543/543 unit tests passed identically on both (335 pre-existing async-plugin failures present in both runs, not introduced here). No regressions.
- **Perf:** measured at ~461k calls/sec post-fix on a 100k-call synthetic harness with realistic inputs, matching pre-fix baseline (~461k). Within noise.

## Test plan

- [x] Regression tests in \`test_config_async.py\` — one per fixed key, in the \`with_config + invoke/astream_events\` shape (mirrors #7424's pattern).
- [x] Unit tests in \`test_utils.py\` — direct merge-semantics tests for each key.
- [x] Existing \`test_callback_manager_copies_*\` tests still pass (configurable→metadata propagation unaffected).
- [x] Full langgraph + downstream test suites pass.
- [x] Perf benchmark within noise.

### DIFF
--- a/libs/langgraph/langgraph/_internal/_config.py
+++ b/libs/langgraph/langgraph/_internal/_config.py
@@ -342,7 +342,8 @@ def ensure_config(*configs: RunnableConfig | None) -> RunnableConfig:
                     # bound via with_config(...) (e.g. ls_agent_type) are
                     # preserved when later configs (e.g. invoke-time) only
                     # specify a subset of keys like thread_id.
-                    empty[k] = {**cast(dict, empty.get(k) or {}), **cast(dict, v)}
+                    existing = empty.get(k)
+                    empty[k] = {**cast(dict, existing), **cast(dict, v)} if existing else cast(dict, v).copy()
                 elif k == "callbacks":
                     empty["callbacks"] = _merge_callbacks(
                         empty.get("callbacks"), cast(Callbacks, v)
@@ -351,15 +352,14 @@ def ensure_config(*configs: RunnableConfig | None) -> RunnableConfig:
                     # Shallow-merge metadata dicts across configs so values
                     # bound via with_config(...) (e.g. user_id) are preserved
                     # when later configs supply other metadata keys.
-                    empty["metadata"] = {
-                        **cast(dict, empty.get("metadata") or {}),
-                        **cast(dict, v),
-                    }
+                    existing = empty.get("metadata")
+                    empty["metadata"] = {**cast(dict, existing), **cast(dict, v)} if existing else cast(dict, v).copy()
                 elif k == "tags":
                     # Concatenate tags across configs so values bound via
                     # with_config(...) are preserved when later configs
                     # supply additional tags. Matches merge_configs.
-                    empty["tags"] = [*(empty.get("tags") or []), *cast(list, v)]
+                    existing = empty.get("tags")
+                    empty["tags"] = [*cast(list, existing), *cast(list, v)] if existing else list(cast(list, v))
                 else:
                     empty[k] = v  # type: ignore[literal-required]
         for k, v in config.items():
--- a/libs/langgraph/langgraph/_internal/_config.py
+++ b/libs/langgraph/langgraph/_internal/_config.py
@@ -79,6 +79,35 @@ def patch_checkpoint_map(
         return config
 
 
+def _merge_callbacks(base: Callbacks, new: Callbacks) -> Callbacks:
+    """Merge two callbacks values (None / list / BaseCallbackManager).
+
+    Six cases total (3 base types x 2 non-None new types). Mirrors the inline
+    callbacks merge in `merge_configs` and the logic in PR #7424.
+    """
+    if new is None:
+        return base
+    if base is None:
+        return new.copy() if isinstance(new, (list, BaseCallbackManager)) else new
+    if isinstance(new, list):
+        if isinstance(base, list):
+            return base + new
+        if isinstance(base, BaseCallbackManager):
+            mngr = base.copy()
+            for cb in new:
+                mngr.add_handler(cb, inherit=True)
+            return mngr
+    elif isinstance(new, BaseCallbackManager):
+        if isinstance(base, list):
+            mngr = new.copy()
+            for cb in base:
+                mngr.add_handler(cb, inherit=True)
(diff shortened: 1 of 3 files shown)


sydney-runkle (approved):
let's make sure we do a thorough issue scan + customer issue scan to see what this resolves

PR #7927, opened 2026-05-27
feat(langgraph): add v3 streaming support to RemoteGraph

nick-hollon-lc:
## Summary

- Adds `stream_events(version="v3")` and `astream_events(version="v3")` to `RemoteGraph`, matching the local `CompiledStateGraph` surface and unblocking polymorphic v3 streaming over `Graph | RemoteGraph`.
- Implementation is a thin adapter (`libs/langgraph/langgraph/pregel/_remote_run_stream.py`) that wraps the v3 SDK's `AsyncThreadStream` / `SyncThreadStream` and duck-types `GraphRunStream` / `AsyncGraphRunStream`. No coupling to local v3 mux internals.
- `v1` / `v2` paths unchanged. `astream_events(version='v1'|'v2')` still raises `NotImplementedError` (separate gap).

### Scope decisions baked into this PR

- Unsupported v3 kwargs hard-reject at dispatch with `NotImplementedError`: `control`, `transformers`, `interrupt_before`, `interrupt_after`, and any unknown `**kwargs`. Server / SDK don't plumb these through v3 yet; easy to lift later.
- Sync `interleave()` raises `NotImplementedError` pointing callers at `astream_events`. Real sync interleave would need drainer threads; deferred since most sync RemoteGraph callers just iterate raw events.
- Async `interleave()` is best-effort ordering (client receive order), documented as a divergence from local v3's monotonic stamp ordering.
- Adapter `interrupted` / `interrupts` properties are **non-blocking** snapshots of the SDK's current state. This differs from local `(Async)GraphRunStream.interrupted`, which pump-drives the run to terminal before returning. Callers needing a wait-for-interrupt pattern should drain a projection (e.g., `interleave('values')`) until the SDK's paused sentinel fires. Documented in the adapter docstrings.

### Audit of impact

Existing RemoteGraph callers in this org all use the v2 `.stream()` / `.astream()` path (deepagents production wrapper, langgraph-api test graphs, langgraph-supervisor TS type guard). **Zero callers** use `stream_events` / `astream_events` on RemoteGraph today, so the new v3 methods are net-new surface — no risk of breaking existing consumers.

### Out of scope (follow-ups)

- Bumping `libs/langgraph/pyproject.toml`'s `langgraph-sdk` constraint from `<0.4.0` to `<0.5.0`. Deferred until 0.4.0 publishes to PyPI; dev resolution unaffected via the editable workspace dep.
- Real `astream_events(version='v1'|'v2')` implementation.
- Server-side plumbing for `control` / `interrupt_before` / `interrupt_after` on v3 runs.
- Sync `interleave()` via drainer threads.

## Test plan

- [x] \`make test\` in \`libs/langgraph/\`: 1874 passed, 4 skipped (43 new in \`test_remote_graph_v3.py\`)
- [x] \`make lint\` in \`libs/langgraph/\`: ruff + mypy clean
- [x] \`pytest -m integration tests/integration/test_remote_graph_v3.py\` in \`libs/sdk-py/\` against the docker stack: 4/4 passed in 1.35s
- [x] Manual smoke: \`RemoteGraph('tools_agent', url='http://localhost:2024').astream_events(..., version='v3')\` end-to-end against the v3 integration api
- [x] Existing RemoteGraph v2 tests untouched (31 passed, 3 skipped with docker up)
- [x] Will need rebase after \`langgraph-sdk 0.4.0\` lands on PyPI and the version constraint is bumped in a separate PR


libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -996,21 +1037,102 @@ async def astream(
...
+        )
+        return _RemoteGraphRunStream(
+            sync_client=sync_client,
+            sdk_thread=sdk_thread,
+            input=_translate_command_input(input),
+            config=sanitized,
+            metadata=kwargs.get("metadata"),
+        )
+
+    def astream_events(
```
sydney-runkle: why not async?

sydney-runkle: also we dropped some args?

libs/langgraph/langgraph/pregel/remote.py
```diff
@@ -996,21 +1037,102 @@ async def astream(
...
+        )
+        sdk_thread = sync_client.threads.stream(
+            thread_id=thread_id,
+            assistant_id=self.assistant_id,
+            headers=merged_headers,
+        )
+        return _RemoteGraphRunStream(
+            sync_client=sync_client,
+            sdk_thread=sdk_thread,
+            input=_translate_command_input(input),
```
sydney-runkle: seems kinda odd that we need this, haven't thought deeply about it though

libs/langgraph/langgraph/pregel/remote.py
nick-hollon-lc: definitely should be async! args got dropped because the signature didn't match the local stream_events signature. it isn't implemented for v1/v2 anyway so ok if we make the signatures match

libs/langgraph/langgraph/pregel/remote.py
nick-hollon-lc: SDK takes a dict but the conversion should probably happen closer to where it's called and maybe even in SDK. i will check what the existing SDK methods do

nick-hollon-lc: fixed this so it takes Command

libs/langgraph/langgraph/pregel/remote.py
nick-hollon-lc: fixed this too

sydney-runkle (approved):
awesome, let's test with dcode to confirm it's working!

Issue #7910, opened 2026-05-26
v3 stream.subgraphs doesn't detect sub-agents invoked inside tool functions


sayanths05:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is.
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports.

### Related issues

- #5528 — same symptom (sub-agent streaming broken), workaround is `astream(subgraphs=True)` but doesn't apply to v3 `astream_events`
- #5249 — duplicate of #5528, fix shipped in v0.6.4 for the `astream` path only

### Description

The [event streaming docs](https://docs.langchain.com/oss/python/langchain/event-streaming#streaming-sub-agents) show a pattern where a sub-agent is invoked inside a tool function and `stream.subgraphs` yields a handle with `graph_name` matching the sub-agent's `name=` parameter. This doesn't work — `subagent.graph_name` is always `"tools"` (the internal ToolNode name), never the sub-agent's name.

**Root cause has two parts:**

**1. Detection failure — `find_subgraph_pregel` can't reach tool function closures**

`PregelNode.__init__` calls `find_subgraph_pregel(self.bound)` to detect subgraphs. For the `"tools"` node, `self.bound` is a `ToolNode` (`RunnableCallable`). The function inspects `ToolNode._func`'s non-locals via `get_function_nonlocals` — but `_func` is a class method whose non-locals are module globals from `langgraph/prebuilt/tool_node.py`, not the user's tool functions.

The sub-agent IS detectable — `get_function_nonlocals(tool.func)` on the individual `BaseTool` finds the `CompiledStateGraph` via the `invoke` bound method's `__self__`. The BFS just never gets there because it doesn't inspect `ToolNode._tools_by_name`:

```python
>>> outfit_agent.nodes["tools"].subgraphs
[]  # detection failed

>>> find_subgraph_pregel(RunnableCallable(call_weather.func))
<CompiledStateGraph name='weather_agent'>  # works when given the tool func directly
```

**2. Execution gap — nested `.invoke()` doesn't participate in parent's stream**

[...]

PR #7928, opened 2026-05-28
feat(langgraph): name tool-dispatched subagents via `lc_agent_name`

nick-hollon-lc:
Resolves the labeling half of #7910.

---

When a tool body invokes a named inner agent (`create_agent(name=...)`), the supervisor's `run.subgraphs` / `run.lifecycle` handle for that dispatch was named after the parent tool node (`tools`) instead of the agent. This wires the inner agent's own `lc_agent_name` — already bound into its config by `create_agent` — through to the lifecycle surfaces, with no tool annotation required.

`_TasksLifecycleBase` now tracks `lc_agent_name` per namespace and treats a nested task as a subagent when its `lc_agent_name` differs from its parent namespace's. `graph_name` becomes that name; the triggering `tool_call_id` (for `cause`) is recovered by joining the child subgraph's namespace task-id to the parent `tools` task's tool calls (handles the current `tool_call_with_context` push-task shape and the legacy list shape).

`TaskPayload.metadata` forwarding (from `task.config["metadata"]`) is retained as the transport. Plain non-agent subgraphs (which inherit the parent's `lc_agent_name`) and unnamed agents are correctly excluded.

This PR supersedes its own earlier approach, which added a `subagent_name` parameter to tools and stamped it via a `ToolNode`→Pregel scheduling hook. That machinery is removed, and the companion langchain-core PR (#37721) is closed — no langchain-core change is needed.

Builds on #7926 (now merged).


libs/langgraph/langgraph/_internal/_config.py
```diff
@@ -402,3 +402,17 @@ def _get_tracing_metadata_defaults(
         "graph_id",
     )
 )
+
+
+def filter_user_tags(tags: Sequence[str] | None) -> list[str] | None:
```
sydney-runkle: nit, this is really filtering out internal LG/LC tags right?

libs/langgraph/langgraph/pregel/_messages.py
```diff
@@ -143,9 +144,8 @@ def on_chat_model_start(
             ]
             if not self.subgraphs and len(ns) > 0 and ns != self.parent_ns:
                 return
-            if tags:
-                if filtered_tags := [t for t in tags if not t.startswith("seq:step")]:
-                    metadata["tags"] = filtered_tags
+            if (filtered_tags := filter_user_tags(tags)) is not None:
```
sydney-runkle: walruses make me happy

libs/langgraph/langgraph/pregel/debug.py
```diff
@@ -40,12 +40,25 @@ def map_debug_tasks(tasks: Iterable[PregelExecutableTask]) -> Iterator[TaskPaylo
...
             "input": task.input,
             "triggers": task.triggers,
         }
+        # Forward the task's metadata dict whole. Framework keys
+        # (langgraph_step, langgraph_node, lc_agent_name, ...) ride along —
+        # the same shape stream_mode="messages" consumers already receive —
+        # without needing to be enumerated here. Filtered config tags are
+        # folded in under `tags`, mirroring the messages stream handler.
+        if task.config is not None:
+            md = dict(task.config.get("metadata") or {})
```
sydney-runkle: why do we need to wrap in dict

libs/langgraph/langgraph/pregel/debug.py
```diff
@@ -40,12 +40,25 @@ def map_debug_tasks(tasks: Iterable[PregelExecutableTask]) -> Iterator[TaskPaylo
...
+            filtered_tags = filter_user_tags(task.config.get("tags"))
+            if filtered_tags is not None:
+                md["tags"] = filtered_tags
+            if md:
+                payload["metadata"] = md
```
sydney-runkle, line 70: for the other one (messages), we put the tags in the metadata, can we do the same here

libs/langgraph/tests/test_large_cases.py
```diff
@@ -4019,6 +4019,7 @@ def qa(data: State) -> State:
                     "name": "rewrite_query",
                     "input": {"query": "what is weather in sf", "docs": []},
                     "triggers": ("branch:to:rewrite_query",),
+                    "metadata": AnyObject(),
```
sydney-runkle: curious, what does this look like here?

libs/langgraph/langgraph/stream/transformers.py
```diff
@@ -9,7 +9,7 @@
     ChatModelStream,
 )
 from langchain_core.messages import AIMessageChunk, BaseMessage, ToolMessage
-from langchain_protocol.protocol import MessagesData
+from langchain_protocol.protocol import LifecycleCause, MessagesData
```
sydney-runkle, line 12: not blocking this PR, but i don't like the structure of `LifecycleCause`, the send + edge ones make total sense for langgraph workflows, tool call is a bit odd, would love to thread that info through in another way, but here we are

sydney-runkle (changes requested):

libs/langgraph/langgraph/stream/transformers.py
nick-hollon-lc: it would be nicer if lifecycle had metadata and we just put the tool call id in metadata instead of saying "cause" but i'd want to do that in a separate PR because it touches a lot more

libs/langgraph/langgraph/pregel/debug.py
nick-hollon-lc: i think it is in there md = metadata

libs/langgraph/langgraph/pregel/debug.py
nick-hollon-lc: defensive copy so we don't mutate it later on accident

libs/langgraph/tests/test_large_cases.py
nick-hollon-lc: {
    "langgraph_step": 1,
    "langgraph_node": "rewrite_query",
    "langgraph_triggers": ("branch:to:rewrite_query",),
    "langgraph_path": ("__pregel_pull", "rewrite_query"),
    "langgraph_checkpoint_ns": "rewrite_query:<uuid>",
    "checkpoint_ns": "",
    "thread_id": "<uuid>",
}

but asserted as any object because the values are non deterministic

libs/langgraph/langgraph/_internal/_config.py
nick-hollon-lc: renamed as filter_to_user_tags

sydney-runkle (approved):

PR #7951, opened 2026-05-31
Guard AsyncSqliteSaver sync writes in event loop

### LINKED ISSUE #7857: AsyncSqliteSaver.put() and put_writes() deadlock instead of raising when called synchronously from within the event loop
### Checked other resources

### Related Issues / PRs

_No response_

```import asyncio

from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver


async def main():
    async with AsyncSqliteSaver.from_conn_string(":memory:") as saver:
        await saver.setup()
        cfg = {"configurable": {"thread_id": "t1", "checkpoint_ns": ""}}

        # delete_thread guards against in-loop sync calls -> raises immediately
        try:
            saver.delete_thread("t1")
        except asyncio.InvalidStateError as e:
            print("delete_thread raised as expected:", e)

        # put_writes is NOT guarded -> deadlocks the event loop, no error
        saver.put_writes(cfg, [("channel", "value")], "task-1")
        print("this line is never reached")


asyncio.run(main())
```

```shell
No error is raised. The synchronous call to put_writes() (or put()) blocks
forever: run_coroutine_threadsafe(...).result() waits on a coroutine that
the current (blocked) event loop is the only one able to run.
The guarded methods, by contrast, raise:
asyncio.InvalidStateError: Synchronous calls to AsyncSqliteSaver are only
allowed from a different thread. From the main thread, use the async
interface. ...
```

### Description

AsyncSqliteSaver exposes synchronous bridge methods that wrap their async
counterparts with asyncio.run_coroutine_threadsafe(...).result(). Calling one
synchronously from within the saver's own event loop blocks the loop on a
coroutine only that loop can run --> a deadlock.
Four of the six sync-bridge wrappers guard against this and raise a
descriptive asyncio.InvalidStateError. Two do not — put and put_writes --> so
the same misuse hangs silently with no diagnostic.
Guard coverage on the current main branch:
get_tuple                            guarded
list                                      guarded
delete_thread                     guarded
get_delta_channel_history  guarded
put                                      NOT guarded
put_writes                           NOT guarded
Expected: put and put_writes should behave consistently with the other four
wrappers — an in-loop synchronous call should raise asyncio.InvalidStateError
with the same message, instead of deadlocking.
Suggested fix: add the same guard block already used by the four guarded
methods to put and put_writes. The existing inline comment notes the guard is
skipped elsewhere "to avoid the overhead"; put and put_writes are write paths
that serialize data and hit disk, so one get_running_loop() check is
negligible next to the work done, and converting a silent deadlock into a
fast, descriptive error is worth it.
I have a patch and a regression test ready (the test is verified to fail on
current main and pass with the fix). Could a maintainer assign this to me so I
can open the PR? Per the contributing flow, external PRs must link to an
assigned, approved issue.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.26200
> Python Version:  3.14.0 (tags/v3.14.0:ebf955d, Oct  7 2025, 10:15:03) [MSC v.1944 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 1.4.0
> langgraph: 1.2.0
> langgraph-checkpoint-sqlite: 3.1.0
> langgraph_sdk: 0.3.14
> langsmith: 0.8.5
> langchain_protocol: 0.0.15

Other Dependencies
------------------
> aiosqlite: (installed)
> httpx: 0.28.1
> orjson: 3.11.7
> pydantic: 2.12.5
> typing-extensions: 4.15.0

Oxygen56:
## Summary
- add event-loop thread guards to AsyncSqliteSaver.put and put_writes
- make those sync bridge methods raise InvalidStateError instead of deadlocking when called from the owning loop
- add regression tests for both sync write paths

Fixes #7857

## Testing
- python3 -m py_compile libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/aio.py libs/checkpoint-sqlite/tests/test_aiosqlite.py
- Not run: pytest, because this local Python environment is missing pytest and project test dependencies.

### DIFF
--- a/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/aio.py
+++ b/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/aio.py
@@ -239,6 +239,16 @@ class AsyncSqliteSaver(BaseCheckpointSaver[str]):
         Returns:
             RunnableConfig: Updated configuration after storing the checkpoint.
         """
+        try:
+            if asyncio.get_running_loop() is self.loop:
+                raise asyncio.InvalidStateError(
+                    "Synchronous calls to AsyncSqliteSaver are only allowed from a "
+                    "different thread. From the main thread, use the async interface. "
+                    "For example, use `await checkpointer.aput(...)` or `await "
+                    "graph.ainvoke(...)`."
+                )
+        except RuntimeError:
+            pass
         return asyncio.run_coroutine_threadsafe(
             self.aput(config, checkpoint, metadata, new_versions), self.loop
         ).result()
@@ -250,6 +260,16 @@ class AsyncSqliteSaver(BaseCheckpointSaver[str]):
         task_id: str,
         task_path: str = "",
     ) -> None:
+        try:
+            if asyncio.get_running_loop() is self.loop:
+                raise asyncio.InvalidStateError(
+                    "Synchronous calls to AsyncSqliteSaver are only allowed from a "
+                    "different thread. From the main thread, use the async interface. "
+                    "For example, use `await checkpointer.aput_writes(...)` or `await "
+                    "graph.ainvoke(...)`."
+                )
+        except RuntimeError:
+            pass
         return asyncio.run_coroutine_threadsafe(
             self.aput_writes(config, writes, task_id, task_path), self.loop
         ).result()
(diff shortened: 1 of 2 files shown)


Oxygen56:
Thanks for the auto-close notice. I've requested assignment on issue #7857 as required. This PR adds the same in-loop sync guard pattern already used by the other 4 sync bridge methods in AsyncSqliteSaver, with regression tests covering both put() and put_writes(). Ready to reopen once a maintainer assigns me to the issue.

PR #7952, opened 2026-05-31
Isolate nested v3 stream message callbacks

### LINKED ISSUE #7948: Nested astream_events(v3) inside a tool yields empty messages due to parent callback leaking via contextvar
### Checked other resources

- [x] I read what a minimal reproducible example is.
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports.

### Description

When calling `astream_events(version="v3")` on a sub-agent **inside a tool** that's already running within another `astream_events(v3)` call, the nested stream's `out.messages` projection yields zero items. The sub-agent's model output is instead captured by the **parent's** `stream.messages` projection.

**Root cause**: The parent's `StreamMessagesHandler` callback is inherited by the nested execution via `var_child_runnable_config` contextvar. The parent's handler captures the nested model's streaming events, leaving the nested stream's own handler with nothing. `await out.output()` then returns `None`, crashing on any attribute access.

**Workaround**: Clearing the contextvar before the nested call isolates the callback chain:

```python
from langchain_core.runnables.config import var_child_runnable_config

token = var_child_runnable_config.set(None)
try:
    out = await sub_agent.astream_events(input, version="v3")
    async for message in out.messages:
        async for delta in message.text:
            print(delta, end="", flush=True)
    full_out = await out.output()
finally:
    var_child_runnable_config.reset(token)
```

This is fragile (depends on a framework-internal contextvar) but confirms the root cause.

### Example Code

```python
import asyncio
from asyncio import Queue
from typing import Any

from langchain.agents import create_agent
from langgraph.prebuilt import ToolRuntime


@dataclass
class Context:
    queue: Queue


@tool("weather", description="Get weather for a city")
def weather(location: str) -> str:
    return f"Sunny in {location}"


weather_agent = create_agent(
    model="anthropic:claude-sonnet-4-5-20250929",
    tools=[weather],
    name="weather_agent",
)


@tool("call_weather", description="Query the weather agent")
async def call_weather(city: str, runtime: ToolRuntime[Context]) -> str:
    context = runtime.context

    # Nested astream_events(v3) inside a tool
    out = await weather_agent.astream_events(
        {"messages": [HumanMessage(f"Weather in {city}?")]},
        version="v3",
        context=context,
    )

    async for message in out.messages:
        # This never executes — parent's callbacks steal the events
        async for delta in message.text:
            await context.queue.put({"type": "agent_delta", "content": delta})

    full_out = await out.output()
    # TypeError: 'NoneType' object is not subscriptable
    # because full_out is None — no values events were observed
    return full_out["messages"][-1].text


supervisor = create_agent(
    model="anthropic:claude-sonnet-4-5-20250929",
    tools=[call_weather],
    name="supervisor",
)


async def main():
    ui_queue: Queue[Any] = Queue()
    stream = await supervisor.astream_events(
        {"messages": [HumanMessage("What's the weather in Boston?")]},
        version="v3",
        context=Context(queue=ui_queue),
    )
    async for message in stream.messages:
        text = await message.text
        if text:
            print(text)
    await stream.output()


asyncio.run(main())
```

```
  File "main.py", line 58, in call_weather_agent
    return full_out["messages"][-1].text
           ~~~~~~~~^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

### System Info

```
langchain: 1.3.1
langgraph: 1.2.1
langchain-core: 1.4.0
Python: 3.13
OS: macOS Darwin 25.2.0
```

Oxygen56:
## Summary
- strip inherited StreamMessagesHandlerV2 callbacks when v3 stream_events delegates through v2 streaming
- preserve unrelated callbacks, tags, and metadata while isolating nested graph message collection
- add an async regression test for nested astream_events(v3) inside a tool

## Tests
- uv run --project libs/langgraph pytest libs/langgraph/tests/test_stream_messages_transformer.py::TestEndToEnd::test_nested_astream_events_v3_isolates_messages_handler
- uv run --project libs/langgraph pytest libs/langgraph/tests/test_stream_messages_transformer.py
- uv run --project libs/langgraph ruff check libs/langgraph/langgraph/pregel/main.py libs/langgraph/tests/test_stream_messages_transformer.py

Fixes #7948

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -31,6 +31,7 @@ from typing import (
 from uuid import UUID, uuid5
 
 from langchain_core._api import beta
+from langchain_core.callbacks import BaseCallbackManager
 from langchain_core.globals import get_debug
 from langchain_core.runnables import (
     RunnableSequence,
@@ -40,6 +41,7 @@ from langchain_core.runnables.config import (
     RunnableConfig,
     get_async_callback_manager_for_config,
     get_callback_manager_for_config,
+    var_child_runnable_config,
 )
 from langchain_core.runnables.graph import Graph
 from langchain_core.runnables.schema import StreamEvent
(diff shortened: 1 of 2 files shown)


uditstocks:
Great fix! One thing worth considering: 
_strip_stream_messages_v2_callback currently handles 
list and BaseCallbackManager types. Are there other 
callback manager subclasses in langchain-core that 
might also need handling? For example, 
AsyncCallbackManager or similar types?

Also, does this fix cover the case where the nested 
call happens inside a sync tool rather than async? 
The _isolate_v3_graph_iter covers sync iteration, 
but the test only covers async path.

PR #7957, opened 2026-06-01
fix(checkpoint): replay migrated delta writes through a plain seed

christian-bromann:
## Summary
- Fix `InMemorySaver.get_delta_channel_history` dropping writes stored under a migration-boundary checkpoint. A plain (pre-delta) blob is the settled value AT that ancestor, so its pending writes produce the child and must be replayed on top of the seed — the guard that skipped writes for non-`_DeltaSnapshot` seeds made post-migration writes disappear on reload while remaining visible in the live invoke result.
- Removed the guard (and now-unused `_DeltaSnapshot` lazy import) so the optimized override matches the `BaseCheckpointSaver` reference path.
- Added two regression tests: `test_post_migration_write_survives_reload_through_plain_seed` and `test_post_migration_reload_base_matches_optimized_override`.

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/memory/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/memory/__init__.py
@@ -148,16 +148,16 @@ class InMemorySaver(
         whose stored blob is non-empty. Other channels keep walking until
         they find their own terminator or hit the root.
 
-        Pre-delta plain-value blobs subsume their ancestor's pending
-        writes (the value already includes them); `_DeltaSnapshot` blobs
-        do not (snapshot is the value AT that ancestor, prior to its own
-        pending writes that produce the child).
+        The seed value (whether a `_DeltaSnapshot` or a plain pre-delta
+        migration blob) is the value AT that ancestor, prior to its own
+        pending writes that produce the child. Those on-path writes —
+        including the ones stored on the terminating ancestor — are always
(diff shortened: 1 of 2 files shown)


ErenAta16:
Traced this independently before finding the PR, so I can confirm the mechanism from a fresh angle rather than just reading the diff.

The original guard (`if blob_value is not None and not isinstance(blob_value, _DeltaSnapshot): continue`) conflated two different situations that happen to look the same at the point of the check: a plain blob subsuming its own ancestor's writes (correct to skip), and a plain blob sitting at the migration boundary whose own outgoing pending write produced its child (wrong to skip, since that write hasn't been folded into any blob yet). Both show up as "plain blob encountered while walking channel history," so the old code treated them identically.

The `seen_delta_write` fix separates them correctly: a write only gets skipped once a delta-era ancestor has already contributed a write for that channel, meaning the walk has genuinely moved past the migration boundary into settled legacy history. Before that point, the plain blob's writes are the boundary's own outgoing writes and need to be collected. I reproduced the original bug with a two-step scenario (plain-channel write, then migrate to DeltaChannel and write again) and got the same failure mode described here: the post-migration write shows up in the live `invoke()` result but silently vanishes on reload.

The parity test between `InMemorySaver` and the base fallback path is a good addition, it directly guards against the optimized override diverging from the reference implementation again, which is exactly how this bug happened in the first place.

PR #7978, opened 2026-06-01
test(sdk-py): add factory-graph integration test exercising the server factory path

nick-hollon-lc:
## What

Adds the first sdk-py integration fixture+test that drives the **graph-factory** code path end to end, and makes the integration suite run the **local** langgraph core so it can catch core regressions pre-merge.

- New `factory_agent` fixture: `graph/factory_graph.py` — a `make_graph(config, runtime: ServerRuntime)` factory returning a compiled echo graph. All other integration graphs are pre-compiled, so this is the only one that exercises the server's factory path.
- New `tests/integration/test_factory_graph.py` — async + sync tests that execute a run (`runs.wait`) against the factory graph.
- Integration `Dockerfile` + `docker-compose.yml`: install this monorepo's local langgraph **core** over the published base image (via a `langgraph_src` additional build context), so the server runs the code under test (the base langgraph-api stack stays on `latest` for upstream-regression signal).
- `ci.yml`: widen the `sdk_py` path filter to `libs/langgraph/**` so core changes trigger the integration suite.

## Why

A 1.2.3 regression: langgraph-api's `get_graph` seeds an SDK `_ExecutionRuntime` into `__pregel_runtime` via the `var_child_runnable_config` contextvar, and #7926 changed `ensure_config` to shallow-merge `configurable` instead of overwrite — so the leaked runtime now survives into `astream`, which reads `parent_runtime.control` and raises:

```
AttributeError: '_ExecutionRuntime' object has no attribute 'control'
```

No integration suite caught it: langgraph-api's tests use a factory graph but only test *published* core; the monorepo suite runs on the PR but its server used published core and registered no factory graph.

## Status

**The integration test fails by design in this PR** — it reproduces the regression against local 1.2.3 core. The fix follows in a subsequent commit on this branch (the test should then go green).

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -4263,17 +4263,15 @@ def _coerce_parent_runtime(value: Any) -> Runtime[Any]:
     During a graph run this is always a `Runtime` the framework created and
     published for child tasks to inherit. Layers outside the run (for example a
     server's graph-factory path) may instead seed an object that only carries
-    `store`/`context`. Adopt those fields into a real `Runtime` so runtime
-    operations like `merge` and `control` work regardless of what populated the
-    slot. The run's own `context` is supplied separately and still takes
-    precedence via `merge`.
+    `context`/`store`. Adopt its `context` so context set at the graph level
+    plumbs through (`merge` lets the run's own `context` take precedence when
+    one is provided). `store` is resolved separately (passed to the graph
+    directly), so it is not read off here. `merge` then combines this with the
+    run's own runtime, including the framework's `control`.
     """
     if isinstance(value, Runtime):
         return value
-    return Runtime(
-        store=getattr(value, "store", None),
-        context=getattr(value, "context", None),
-    )
+    return Runtime(context=getattr(value, "context", None))
 
 
 def _build_server_info(
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -2835,7 +2835,9 @@ class Pregel(
                 config[CONF][CONFIG_KEY_DURABILITY] = durability_
 
             # build server_info from metadata + parent runtime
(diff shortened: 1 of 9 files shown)


nick-hollon-lc:
Note on `test_subgraphs_deep_agent_sync`: this PR makes the integration server run the local langgraph core, which exposes a **separate, pre-existing** sync-streaming bug in the SDK (`thread.subgraphs` drops direct-child handles that arrive after the root-terminal lifecycle event). That fix is split out into #7979.

Sequencing: **merge #7979 first**, then rebase this PR onto main and `test_subgraphs_deep_agent_sync` goes green. Until then, that one test is expected to be red here; the factory-graph tests (the focus of this PR) pass with the runtime fix.

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -4253,6 +4257,25 @@ def _resolve_parent_ns(
...
+    server's graph-factory path) may instead seed an object that only carries
+    `store`/`context`. Adopt those fields into a real `Runtime` so runtime
+    operations like `merge` and `control` work regardless of what populated the
+    slot. The run's own `context` is supplied separately and still takes
+    precedence via `merge`.
+    """
+    if isinstance(value, Runtime):
+        return value
+    return Runtime(
+        store=getattr(value, "store", None),
```
sydney-runkle: i don't think we need to take store off it, that should be passed to graph directly
i presume the rest of the merging happens appropriately then?
can we test that context set at the graph level (like LG API does) is plumbed through?

sydney-runkle: (the test doesn't check context plumbed through)

sydney-runkle (changes requested):
lgtm generally

sydney-runkle (approved):

sydney-runkle:
can we open an issue requesting that you can bind context at the graph level so we can make this less hacky for LG API

PR #7979, opened 2026-06-02
fix(sdk-py): deliver trailing child events after the root-terminal lifecycle in the sync stream

nick-hollon-lc:
## Bug

The sync stream controller (`SyncStreamController._fanout`) pushed the terminal `None` sentinel into every subscription queue the instant it saw a root-namespace `completed`/`failed` lifecycle event. Any event the server emits **after** that — e.g. a direct-child subgraph `lifecycle`/`tasks` event that feeds `thread.subgraphs` — was then queued *behind* the `None` and never consumed, since projection iterators return on the first `None`.

This surfaced as `thread.subgraphs` yielding **no direct-child handles** on the **sync** client for a factory / deep-agent graph, while the **async** client worked. The async controller already terminates only on stream-end (it never injects a mid-stream `None`), so it drains trailing events; the sync controller had an extra mid-stream terminal signal that the async one lacks.

## Fix

Align the sync controller with the async controller: terminate projection iterators when the shared stream **ends** (or on **interrupt**, which is signaled separately by the thread-stream via `_signal_paused`), not on a mid-stream root-terminal lifecycle event. Removes the now-dead `_is_root_terminal_lifecycle` helper.

## Tests

- New regression test `test_sync_controller_delivers_child_events_after_root_terminal` (a child event scripted after the root `completed` is still delivered).
- Full sdk-py unit suite passes (492 tests); no completion/interrupt-termination regressions.

Split out from the integration-test PR (#7978), which is what surfaced this on the sync `deep_agent` subgraphs test once its server runs current core.

### DIFF
--- a/libs/sdk-py/langgraph_sdk/stream/sync_controller.py
+++ b/libs/sdk-py/langgraph_sdk/stream/sync_controller.py
@@ -22,31 +22,6 @@ from langgraph_sdk.stream.transport import (
 _logger = logging.getLogger(__name__)
 
 
-_ROOT_TERMINAL_LIFECYCLE_EVENTS = frozenset({"completed", "failed"})
-
-
-def _is_root_terminal_lifecycle(event: Any) -> bool:
-    """Return True for a root-namespace lifecycle event marking run end.
-
-    Matches the wire shape ``{method: "lifecycle", params: {namespace: [],
-    data: {event: "completed" | "failed"}}}``. Subgraph lifecycle events
-    (non-empty namespace) do not terminate the parent run.
-    """
-    if not isinstance(event, dict):
-        return False
-    if event.get("method") != "lifecycle":
-        return False
-    params = event.get("params") or {}
-    if not isinstance(params, dict):
-        return False
-    if params.get("namespace") or []:
-        return False
-    data = params.get("data") or {}
-    if not isinstance(data, dict):
-        return False
-    return data.get("event") in _ROOT_TERMINAL_LIFECYCLE_EVENTS
-
-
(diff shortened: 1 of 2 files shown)


nick-hollon-lc:
Reverted the fix — **it was wrong and caused the sync integration tests to hang for ~6h in CI.**

Root cause of the regression: the mid-stream terminal `None` I removed is **load-bearing**. The real server keeps the SSE open after a run completes (for resume), so the sync client relies on the root-terminal lifecycle event to wake/terminate its blocking iterators. Removing it makes every sync iterator that runs to completion block forever. My unit test missed this because the fake server *ends* the stream (so termination happened via stream-end) — it never exercised the open-SSE case the real server produces.

Open question I haven't resolved: sync and async have **identical** terminal-signal logic, yet async both delivers the trailing child handle *and* terminates. Until I understand that (likely a transport or event-timing difference, not the terminal signal itself), I won't attempt another fix. Marking this draft.

nick-hollon-lc:
Closing — this was a misdiagnosis. There is no sync-streaming drop: when a subagent is dispatched, child events arrive before the terminal, so nothing is orphaned. The reverted change here also removed a load-bearing terminal signal (it hung sync integration tests ~6h).

The real bug was a cross-package break: langgraph #7928 added a `cause` kwarg to the `_on_started` extension point, which broke published deepagents' `SubagentTransformer` override → subagent runs errored → flaky `deep_agent_sync`. Fixed properly in #7987 (make `_on_started` backward-compatible). Verified against a local langgraph-api + published-deepagents stack.

PR #7996, opened 2026-06-02
fix(langgraph): honor per-node cache_policy=None with set_node_defaults

christian-bromann:
## Summary
- Fixes `StateGraph.set_node_defaults(cache_policy=...)` so a node can opt out by passing `cache_policy=None` to `add_node`.
- Distinguishes omitted `cache_policy` (inherit defaults) from explicit `None` (disable caching) via the `MISSING` sentinel and a `cache_policy_opt_out` flag on `StateNodeSpec`.
- Adds regression tests for default application, per-node override, and explicit opt-out with caching behavior.

### DIFF
--- a/libs/langgraph/langgraph/graph/state.py
+++ b/libs/langgraph/langgraph/graph/state.py
@@ -47,7 +47,12 @@ from langgraph._internal._fields import (
 from langgraph._internal._pydantic import create_model
 from langgraph._internal._runnable import coerce_to_runnable
 from langgraph._internal._timeout import coerce_timeout_policy
-from langgraph._internal._typing import EMPTY_SEQ, MISSING, DeprecatedKwargs
+from langgraph._internal._typing import (
+    EMPTY_SEQ,
+    MISSING,
+    DeprecatedKwargs,
+    UnsetSentinel,
+)
 from langgraph.channels.base import BaseChannel
 from langgraph.channels.binop import BinaryOperatorAggregate
 from langgraph.channels.delta import DeltaChannel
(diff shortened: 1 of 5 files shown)


libs/langgraph/langgraph/graph/state.py
```diff
@@ -789,6 +812,10 @@ def my_node(state: State, config: RunnableConfig) -> State:
                 )
         if action is None:
             raise RuntimeError
+        if explicit_cache_opt_out:
+            self._cache_policy_opt_out.add(node)
```
Sanjays2402, line 816: This mutates the opt-out set before the duplicate-name check. A failed second `add_node("x", ..., cache_policy=None)` leaves the existing `x` opted out, so catching the `ValueError` and compiling changes that node’s behavior. Can the validation happen before touching this set?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -789,6 +812,10 @@ def my_node(state: State, config: RunnableConfig) -> State:
                 )
         if action is None:
             raise RuntimeError
+        if explicit_cache_opt_out:
```
Sanjays2402, line 815: this mutates `_cache_policy_opt_out` before the `node in self.nodes` duplicate check, so `add_node("a", f)` followed by a failing `add_node("a", g, cache_policy=None)` raises but leaves "a" in the opt-out set and silently kills caching on the node that was already registered. moving it below the validation block (next to where the spec is built) would also make the `discard` branch unnecessary, since a name can never be re-added.

libs/langgraph/langgraph/graph/state.py
```diff
@@ -789,6 +812,10 @@ def my_node(state: State, config: RunnableConfig) -> State:
...
+            self._cache_policy_opt_out.add(node)
+        else:
+            self._cache_policy_opt_out.discard(node)
```
Sanjays2402, line 818: this mutates `_cache_policy_opt_out` before the `node in self.nodes` duplicate check three lines down, so a rejected re-add still has side effects: `add_node("x", f, cache_policy=None)` then `add_node("x", g)` raises "already present" but has already discarded "x" from the set, and the original node silently picks up the default policy at compile. moving the two lines below the duplicate/END/START guards fixes it.

PR #7999, opened 2026-06-03
chore: adopt mypy 2.1.0 across Python libs

jkennedyvz:
## Summary

Adopts **mypy 1.20.2 → 2.1.0** across all Python libs and fixes the type errors surfaced by the stricter checker. mypy 2.x landed via Dependabot's grouped `major` bumps but broke `make lint` in several libs; this PR adopts it cleanly in one place.

Only **mypy and its transitive deps** (`librt`, `ast-serialize`) change in the lockfiles — **no runtime dependencies** (redis, websockets, starlette, etc.) are touched. Every source change is type-only (annotations, `cast`, `# type: ignore` adjustments) with **no runtime behaviour change**.

## Changes by lib

| Lib | Fix |
|-----|-----|
| checkpoint-postgres | `kwargs: dict[str, Any]` before `Item(**kwargs)` |
| checkpoint-sqlite | `params: list[Any]` annotation |
| langgraph | `cast(...)` for the `retry` deprecation path in `state.py`/`func` (fixes 3 downstream `object` errors); cast into `chained_fut` in `_runner.py`; targeted `[arg-type]` ignores in `binop.py`/`func` for generic/ParamSpec limits; removed now-unused ignores in `delta.py`/`message.py` |
| prebuilt | `cast("list[Any]", ...)`; narrowed an over-broad `type: ignore` |
| checkpoint, checkpoint-sqlite | mypy override for the optional, untyped `langgraph.checkpoint.conformance.*` test module + return-type annotations on the conformance tests |
| cli, sdk-py | mypy bump (sdk-py pin `mypy==2.1.0`); no source changes needed |

## Verification

`make lint` passes in all eight affected libs. Changes are type-only, so the test matrix is exercised by CI.

## Relationship to open Dependabot PRs

- **Supersedes** #7969, #7970, #7974 (pure mypy bumps) — close once this merges.
- **Removes the mypy portion** of the grouped bumps #7976 (langgraph) and #7972 (sdk-py); their remaining redis 7→8 / websockets 15→16 / starlette 0.51→1.2 majors should be assessed separately.


sydney-runkle:
closing since we're migrating to ty

PR #8011, opened 2026-06-05
fix: updateState bug for deltaChannel on empty thread

longquanzheng:
Fixes langchain-ai/deepagents#3774

## Summary

`Pregel.update_state` / `aupdate_state` on a fresh thread silently dropped the first write to a `DeltaChannel`-backed channel (e.g. `DeepAgentState.messages`). This PR persists the first write under a lazily-created stub checkpoint so the read-path ancestor walk can replay it.

## Root cause

`DeltaChannel` reads its value back by walking ancestor checkpoints and replaying writes attached to them — non-snapshot steps don't store the value in `channel_values`. In `bulk_update_state` the channel writes were only persisted via `checkpointer.put_writes(...)` when a previous checkpoint existed:

```python
channel_writes = [w for w in task.writes if w[0] != PUSH]
if saved and channel_writes:
    checkpointer.put_writes(checkpoint_config, channel_writes, task_id)
```

On a fresh thread `saved is None`, so the `if saved` guard skipped persistence entirely. `create_checkpoint` then bumped the channel version but stored neither a value nor replayable writes, so reads returned `[]`.

## Fix

In both `bulk_update_state` (sync) and `abulk_update_state` (async), when the thread has no persisted parent **and** at least one write targets a `DeltaChannel`, lazily persist an empty stub checkpoint and use it as the parent for both the channel writes and the new update checkpoint. Mirrors the existing exit-mode pattern in `_loop._put_exit_delta_writes`.

The behavior for non-delta writes on a fresh thread is preserved (skip `put_writes` — values are stored directly in the new checkpoint's `channel_values`), so non-delta `update_state` paths add no extra checkpoint rows.

## Test coverage

New tests in `libs/langgraph/tests/test_delta_channel_update_state.py` (9 tests, sync + async):

- **Fresh-thread regression** (the bug): single `update_state` writes a message and reads back via `get_state`. Without the fix, both sync and async fail with `assert [] == ['hello']`.
- **`update_state` after `invoke`**: pins down the previously-working non-fresh-thread path so the lazy-stub change doesn't regress it.
- **Consecutive `update_state`s**: second call sees a real parent (`saved is not None`) and takes the original write path; both messages round-trip in chronological order.
- **Update-by-id end-to-end via `update_state`**: `_messages_delta_reducer`'s dedup-by-id semantics work through the `update_state` path, not just `invoke`.
- **`bulk_update_state` with multiple per-superstep updates**: locks in the per-task `put_writes` loop so all task writes persist (not just the last task's).
- **State-history chain shape**: validates the lazy stub via the public API — `get_state_history` returns `[update_checkpoint, stub]` where the stub has `source='update'`, `step=-1`, no parent, and the update checkpoint's `parent_config` points at the stub.

## Verification

- All 9 tests in `tests/test_delta_channel_update_state.py` pass.
- All 4 existing delta-channel suites pass (`test_delta_channel_exit_mode.py`, `test_delta_channel_migration.py`, `test_delta_channel_id_stability.py`, `test_delta_channel_supersteps_bound.py` — 30 tests, 39 total with the new file).
- All `update_state`-related tests across `test_pregel`, `test_pregel_async`, `test_time_travel`, `test_time_travel_async` pass (10 tests).
- `make format`, `make lint`, full `make test` pass locally in `libs/langgraph`.

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -2002,7 +2002,7 @@ class Pregel(
                 for task in run_tasks
                 for c, _ in task.writes
             )
-            can_put_writes = saved is not None or has_delta_writes
+            should_put_writes = saved is not None or has_delta_writes
 
             if saved is None and has_delta_writes:
                 # If there is no previous checkpoint, we need to create a stub checkpoint
@@ -2021,7 +2021,7 @@ class Pregel(
             for task_id, task in zip(run_task_ids, run_tasks):
                 # channel writes are saved to current checkpoint
                 channel_writes = [w for w in task.writes if w[0] != PUSH]
-                if can_put_writes and channel_writes:
+                if should_put_writes and channel_writes:
                     checkpointer.put_writes(checkpoint_config, channel_writes, task_id)
 
             # apply to checkpoint and save
@@ -2469,7 +2469,7 @@ class Pregel(
                 for task in run_tasks
                 for c, _ in task.writes
             )
-            can_put_writes = saved is not None or has_delta_writes
+            should_put_writes = saved is not None or has_delta_writes
 
             if saved is None and has_delta_writes:
                 # If there is no previous checkpoint, we need to create a stub checkpoint
@@ -2488,7 +2488,7 @@ class Pregel(
             for task_id, task in zip(run_task_ids, run_tasks):
                 # channel writes are saved to current checkpoint
                 channel_writes = [w for w in task.writes if w[0] != PUSH]
-                if can_put_writes and channel_writes:
+                if should_put_writes and channel_writes:
                     await checkpointer.aput_writes(
                         checkpoint_config, channel_writes, task_id
                     )
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -108,6 +108,7 @@ from langgraph.callbacks import (
     get_sync_graph_callback_manager_for_config,
 )
 from langgraph.channels.base import BaseChannel
+from langgraph.channels.delta import DeltaChannel
 from langgraph.channels.topic import Topic
 from langgraph.config import get_config
 from langgraph.constants import END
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/pregel/main.py
```diff
@@ -1996,11 +1997,33 @@ def perform_superstep(
                     ),
                 )
             # save task writes
+            has_delta_writes = any(
+                isinstance(channels.get(c), DeltaChannel)
+                for task in run_tasks
+                for c, _ in task.writes
+            )
+            can_put_writes = saved is not None or has_delta_writes
```
sydney-runkle: is this really like should_put_writes?

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -1996,11 +1997,33 @@ def perform_superstep(
...
+
+            if saved is None and has_delta_writes:
```
sydney-runkle, line 2007: when would saved not be `None` heree?

longquanzheng: updateState call could provide a checkpoint to load, so that updateState will use as parent checkpoint to write the next one. see line 1644 `saved = checkpointer.get_tuple(config)`.

libs/langgraph/langgraph/pregel/main.py
longquanzheng: yeah renamed

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -1996,11 +1997,33 @@ def perform_superstep(
...
+                # This is the model of DeltaChannel.
+                stub = empty_checkpoint()
+                checkpoint_config = checkpointer.put(
+                    patch_configurable(
+                        checkpoint_config, {CONFIG_KEY_CHECKPOINT_ID: None}
+                    ),
+                    stub,
+                    {"source": "update", "step": -1, "parents": {}},
+                    {},
+                )
```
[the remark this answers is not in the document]
longquanzheng: that's intentional.

sydney-runkle (approved):
can we add tests for a non-fresh thread? like updating state from a checkpoint?

PR #8014, opened 2026-06-06
fix: persist DeltaChannel writes on fresh threads

### LINKED ISSUE #8012: `update_state` on fresh `DeepAgentState` thread drops first `messages` write
### Submission checklist

- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox
- [ ] Other / not sure / general

### Related Issues / PRs

_No response_

```python
import asyncio

from deepagents.graph import DeepAgentState


async def model(state: DeepAgentState) -> dict:
    return {}


async def main() -> None:
    graph = StateGraph(DeepAgentState)
    graph.add_node("model", model)
    graph.set_entry_point("model")
    graph.set_finish_point("model")

    agent = graph.compile(checkpointer=InMemorySaver())

    config = {"configurable": {"thread_id": "t"}}
    message = HumanMessage(content="hello", id="m1")

    await agent.aupdate_state(
        config,
        {"messages": [message]},
        as_node="model",
    )

    snapshot = await agent.aget_state(config)
    print("snapshot.values:", snapshot.values)


asyncio.run(main())
```

```shell
snapshot.values: {'messages': []}
```

### Description

I am trying to write a message into a fresh `DeepAgentState` thread using LangGraph's `aupdate_state`.

Expected behavior:

```python
await agent.aupdate_state(
    config,
    {"messages": [HumanMessage(content="hello", id="m1")]},
    as_node="model",
)

snapshot = await agent.aget_state(config)
assert snapshot.values["messages"]
```

Actual behavior:

`aupdate_state` creates a checkpoint, but reading the state back returns:

```python
{"messages": []}
```

Control cases:

* `agent.ainvoke({"messages": [message]}, config)` persists the same message correctly.
* A normal `MessagesState` graph with `aupdate_state({"messages": [message]})` persists correctly.
* If a `DeepAgentState` thread is first seeded through normal `ainvoke`, later `aupdate_state` appends appear to work.

This seems specific to the first out-of-band write into `DeepAgentState.messages`, which is backed by `DeltaChannel`.

Possible cause:

In the `aupdate_state` path, channel writes appear to be persisted only when a previous checkpoint exists:

```python
if saved and channel_writes:
    await checkpointer.aput_writes(checkpoint_config, channel_writes, task_id)
```

On a fresh thread, `saved is None`, so the first `messages` write is applied to the live channel but not persisted as checkpoint writes. Then `create_checkpoint(...)` omits non-snapshot `DeltaChannel` values from `channel_values`, leaving a checkpoint with a bumped `messages` channel version but no readable value or replayable writes.

Question:

Is `update_state` / `aupdate_state` expected to support writing to `DeepAgentState.messages` on a fresh thread? If not, is there a recommended API for appending a message to a DeepAgents thread without invoking the model?

### Environment / System Info

OS: macOS 26.5.1 (25F80)
Python: 3.12.9
deepagents: 0.6.8
langgraph: 1.2.4
langgraph-checkpoint: 4.1.1
langgraph-checkpoint-sqlite: 3.1.0
langchain-core: 1.4.0

Oxygen56:
On fresh threads with no prior checkpoint, `update_state`/`aupdate_state` silently dropped the first write to `DeltaChannel`-backed state keys (e.g., `messages` in `DeepAgentState`).

## Root cause

In the `update_state` path, pending channel writes are saved via `checkpointer.put_writes` only when a prior checkpoint (`saved`) exists:
```python
if saved and channel_writes:
    checkpointer.put_writes(checkpoint_config, channel_writes, task_id)
```

On a fresh thread, `saved is None`, so writes are applied to the live channel but never checkpointed. `create_checkpoint` then calls `DeltaChannel.checkpoint()` which returns `MISSING` for non-snapshot steps, omitting the value from the checkpoint blob. With no stored writes and no snapshot in the checkpoint, the DeltaChannel value is lost on the next read.

## Fix

When no prior checkpoint exists, compute the set of available DeltaChannels and force-snapshot them via `create_checkpoint(channels_to_snapshot=...)`. This ensures the initial checkpoint includes a `_DeltaSnapshot` blob so the value survives the write-read roundtrip.

This is applied in both the sync (`bulk_update_state`) and async (`abulk_update_state`) paths.

## Testing

- DeltaChannel writes now persist on fresh threads via both `update_state` and `aupdate_state`
- Multiple writes accumulate correctly via the snapshot blob
- Non-DeltaChannel channels are unaffected
- Normal `ainvoke` flow is unchanged

Fixes #8012

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -108,6 +108,7 @@ from langgraph.callbacks import (
     get_sync_graph_callback_manager_for_config,
 )
 from langgraph.channels.base import BaseChannel
+from langgraph.channels.delta import DeltaChannel
 from langgraph.channels.topic import Topic
 from langgraph.config import get_config
 from langgraph.constants import END
@@ -2009,7 +2010,24 @@ class Pregel(
                 checkpointer.get_next_version,
                 self.trigger_to_nodes,
             )
-            checkpoint = create_checkpoint(checkpoint, channels, step + 1)
+            # On a fresh thread with no prior checkpoint, snapshot all
+            # available DeltaChannels because there are no ancestor writes
+            # to replay from (put_writes requires a checkpoint_id from
+            # a parent checkpoint that does not exist yet).
+            if saved is None:
+                channels_to_snapshot = {
+                    k
+                    for k, ch in channels.items()
+                    if isinstance(ch, DeltaChannel) and ch.is_available()
+                } or None
+            else:
+                channels_to_snapshot = None
+            checkpoint = create_checkpoint(
+                checkpoint,
+                channels,
+                step + 1,
+                channels_to_snapshot=channels_to_snapshot,
+            )
             next_config = checkpointer.put(
                 checkpoint_config,
                 checkpoint,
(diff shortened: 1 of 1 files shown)


Oxygen56:
Replacement PR opened at #8016 — the original was auto-closed because the linked issue was not assigned. I've requested assignment on the issue.

PR #8023, opened 2026-06-08
feat(cli): support compatible api version ranges

cwlbraa:
Adds `api_version` ranges for LangGraph API base image tags. `~=0.11.0.dev5` resolves to the newest matching compatible image while freezing later dev prereleases until rc/final/patch tags exist; `>~=0.11.0.dev5` uses the same floor but can also float to future stable releases while ignoring future prereleases.

Verified with `make format`, `make lint`, and `make test` from `libs/cli`.

### DIFF
--- a/libs/cli/langgraph_cli/cli.py
+++ b/libs/cli/langgraph_cli/cli.py
@@ -163,9 +163,9 @@ OPT_API_VERSION = click.option(
         "API server version to use for the base image. If unspecified, the "
         "latest stable version will be used. Compatible ranges like "
         "~=0.11.0.dev5 stay on 0.11.0.dev5 while only newer dev builds exist, "
-        "then resolve to the newest matching rc or stable release. "
-        "Stable-floating ranges like >~=0.11.0.dev5 can also pick up future "
-        "stable releases."
+        "then resolve to the newest matching rc or stable release, for example "
+        "0.11.0rc1 or 0.11.0. Stable-floating ranges like >~=0.11.0.dev5 "
+        "can also pick up future stable releases, for example 0.12.0."
     ),
 )
 
--- a/libs/cli/langgraph_cli/cli.py
+++ b/libs/cli/langgraph_cli/cli.py
@@ -159,7 +159,12 @@ OPT_POSTGRES_URI = click.option(
 OPT_API_VERSION = click.option(
     "--api-version",
     type=str,
-    help="API server version to use for the base image. If unspecified, the latest version will be used.",
+    help=(
+        "API server version to use for the base image. If unspecified, the "
+        "latest version will be used. Compatible ranges like ~=0.11.0.dev5 "
+        "resolve to the newest matching image tag. Stable-floating ranges "
+        "like >~=0.11.0.dev5 can also pick up future stable releases."
+    ),
 )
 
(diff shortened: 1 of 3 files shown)


libs/cli/langgraph_cli/cli.py
```diff
@@ -159,7 +159,12 @@
 OPT_API_VERSION = click.option(
     "--api-version",
     type=str,
-    help="API server version to use for the base image. If unspecified, the latest version will be used.",
+    help=(
+        "API server version to use for the base image. If unspecified, the "
+        "latest version will be used. Compatible ranges like ~=0.11.0.dev5 "
```
jdrogers940: nit: we should probably update to say latest stable version
```suggestion
        "latest stable version will be used. Compatible ranges like ~=0.11.0.dev5 "
```

libs/cli/langgraph_cli/cli.py
```diff
@@ -159,7 +159,12 @@
...
+        "resolve to the newest matching image tag. Stable-floating ranges "
```
jdrogers940: Maybe just clarify with an example? The first one would resolve to 0.11.0.dev6 when it comes out right? Whereas the second one would stay 0.11.0.dev5 until 0.11.0 comes out, then resolve to that?

libs/cli/langgraph_cli/config.py
```diff
@@ -41,6 +42,33 @@
...
+
+class _ApiVersionRange(NamedTuple):
+    floor: _ParsedApiVersion
+    allow_future_stable: bool
+
+
+_PRERELEASE_ORDER = {
+    "dev": 0,
+    "a": 1,
+    "b": 2,
```
jdrogers940: nit: drop these?

libs/cli/langgraph_cli/config.py
```diff
@@ -141,6 +169,149 @@ def _parse_api_version_parts(version_str: str) -> tuple[int, ...]:
...
+
+
+def _docker_hub_repository(base_image: str) -> str:
+    if ":" in base_image:
+        raise click.UsageError(
+            "Compatible api_version ranges cannot be used with a tagged base_image."
+        )
+    parts = base_image.split("/")
+    if len(parts) == 1:
+        return f"library/{base_image}"
```
jdrogers940: I think we can drop this

libs/cli/langgraph_cli/config.py
```diff
@@ -141,6 +169,149 @@ def _parse_api_version_parts(version_str: str) -> tuple[int, ...]:
...
+        ) from None
+
+    version_range = _ApiVersionRange(
+        floor=floor,
+        allow_future_stable=match.group("operator") == ">~=",
+    )
+    repository = _docker_hub_repository(base_image)
+    tag_suffix = f"-{version_distro_tag}"
+    tag_prefix = f"{floor.release[0]}."
+    tags = _get_docker_hub_tags(repository, tag_prefix)
```
jdrogers940: So right now we get these versions from pypi in the inmem CLI. Probably better to keep those in line? Would think also easier to parse cause it's just versions vs. the weird tag formats we've got.

jdrogers940 (approved):

PR #8030, opened 2026-06-09
feat(langgraph): add metadata parameter to @task decorator

### LINKED ISSUE #8042: Add `metadata` parameter to the functional API's `@task` decorator
### Feature request

Add a `metadata` parameter to `@task` in the functional API, mirroring `add_node(metadata=...)` in the graph API:

```python
@task(metadata={"priority": "high", "owner": "payments-team"})
def call_external_service(input: str) -> str:
    config = get_config()
    priority = config["metadata"]["priority"]
    ...
```

The metadata would be merged into the task's config metadata at run time (after the framework `langgraph_*` keys, same precedence as graph nodes), so it is:

- readable inside the task via `get_config()["metadata"]`
- visible to tracing callbacks and `debug` stream events

### Motivation / problem being solved

We attach custom per-task execution options to our nodes, which shared infrastructure code reads from config metadata at run time. With the graph API this is straightforward — `add_node(metadata=...)` already supports it. But the functional API has no equivalent, so when using `@task` we have to maintain a separate task-name-to-options mapping:

```python
TASK_OPTIONS = {
    "call_external_service": {"priority": "high", "owner": "payments-team"},
    ...
}
```

and do lookups by task name inside each task. This has several problems:

- The options live far away from the task definition, so they silently drift out of sync (e.g., a task rename breaks the lookup with no error).
- Unlike node metadata in the graph API, the options never reach tracing callbacks or `debug` stream events, so observability tooling can't see them.
- Code that mixes the graph API and the functional API has to handle metadata two different ways.

Since graph nodes already support this, it's a parity gap between the two APIs rather than a new concept.

Originally requested in https://forum.langchain.com/t/feature-request-task-metadata/3584. We've already made and tested this change on our fork.

### Proposed implementation

PR #8030 implements this: the `@task` decorator accepts `metadata`, which is threaded through `_call.py` / `_algo.py` / `_runner.py` and merged into the task config the same way node metadata is. Includes sync/async tests (`test_task_metadata_functional`).

brianstrauch:
Add a `metadata` parameter to the functional API's `@task` decorator, mirroring `add_node(metadata=...)` in the graph API. The metadata is merged into the task's config metadata at run time (after the framework `langgraph_*` keys, same precedence as graph nodes), so it is readable inside the task via `get_config()["metadata"]` and visible to tracing callbacks and `debug` stream events.

Requested in https://forum.langchain.com/t/feature-request-task-metadata/3584 — avoids keeping a separate task-name-to-options mapping when tasks need custom per-task options.

Verified with new sync/async tests (`test_task_metadata_functional`) and `make format`, `make lint`, `make test` in `libs/langgraph`.

Fixes: #8042

### DIFF
--- a/libs/langgraph/langgraph/func/__init__.py
+++ b/libs/langgraph/langgraph/func/__init__.py
@@ -65,6 +65,7 @@ class _TaskFunction(Generic[P, T]):
         cache_policy: CachePolicy[Callable[P, str | bytes]] | None = None,
         timeout: TimeoutPolicy | None = None,
         name: str | None = None,
+        metadata: dict[str, Any] | None = None,
     ) -> None:
         if name is not None:
             if hasattr(func, "__func__"):
@@ -81,6 +82,7 @@ class _TaskFunction(Generic[P, T]):
         self.retry_policy = retry_policy
         self.cache_policy = cache_policy
         self.timeout = timeout
+        self.metadata = metadata
         functools.update_wrapper(self, func)
 
     def __call__(self, *args: P.args, **kwargs: P.kwargs) -> SyncAsyncFuture[T]:
@@ -91,6 +93,7 @@ class _TaskFunction(Generic[P, T]):
             retry_policy=self.retry_policy,
             cache_policy=self.cache_policy,
             timeout=self.timeout,
+            metadata=self.metadata,
         )
 
     def clear_cache(self, cache: BaseCache) -> None:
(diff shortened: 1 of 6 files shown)


longquanzheng:
Hi @brianstrauch ,
could you share more context about why you need this task metadata (what problems are you trying to solve). 
So that we can have better understanding and prioritize the support. 

And like the GH actions said, it will be better to create an issue to associate with.

cc @sydney-runkle 

thanks!

brianstrauch:
Opened #8042. Looks like this PR was closed again...

PR #8043, opened 2026-06-10
fix(langgraph): require `langgraph-api>=0.10.0` for `DeltaChannel`

sydney-runkle:
A graph that uses `DeltaChannel` depends on server-side reconstruction support added in `langgraph-api>=0.10.0`. When such a graph is run under an older API server, delta channels fail at runtime. `StateGraph.compile` now reads `importlib.metadata.version("langgraph-api")` and raises a `RuntimeError` with an upgrade hint when it is `<0.10.0`; `packaging.version.Version` keeps release candidates such as `0.10.0rc1` below the final `0.10.0` boundary. The check is a no-op when `langgraph-api` is not installed (local execution) or when the graph has no delta channel.

Verified end-to-end against real `langgraph-api` releases: `0.9.0` raises at compile time, `0.10.0` compiles successfully.

## Self-Hosted Release Note
`StateGraph.compile` now raises a clear error when a graph uses `DeltaChannel` but the installed `langgraph-api` is `<0.10.0`, prompting an upgrade instead of failing later at runtime.

## Test Plan
- [ ] Compiling a `DeltaChannel` graph raises under `langgraph-api<0.10.0` (including `0.10.0rc1`), succeeds at `>=0.10.0`, and is silent when `langgraph-api` is absent or the graph has no delta channel.

Made by [Open SWE](https://openswe.vercel.app)


Shivani767:
I have a fix ready for this in PR #8044. Could you please assign this issue to me?

libs/langgraph/langgraph/graph/state.py
```diff
@@ -120,6 +122,34 @@ def _warn_invalid_state_schema(schema: type[Any] | Any) -> None:
...
+    `DeltaChannel` reconstruction depends on server-side support added in
+    `langgraph-api>0.9.0`. When running under an older API server, delta
+    channels fail at runtime, so we raise at compile time with an upgrade
+    hint. The check is skipped when `langgraph-api` is not installed (e.g.
+    local execution), since there is no API server to be incompatible.
+    """
+    if not any(isinstance(c, DeltaChannel) for c in channels.values()):
+        return
+    try:
+        api_version = metadata.version("langgraph-api")
```
longquanzheng: oh interesting. I was thinking of try import the version constant in langgraph-api, but I guess either way would work. 

not sure how to test this out though.

libs/langgraph/langgraph/graph/state.py
```diff
@@ -120,6 +122,34 @@ def _warn_invalid_state_schema(schema: type[Any] | Any) -> None:
...
+    except metadata.PackageNotFoundError:
+        return
+    if Version(api_version) > _MIN_DELTA_CHANNEL_API_VERSION:
+        return
+    raise RuntimeError(
```
longquanzheng: I thought we can just fallback to non-delta channel, but raising error may be even better since they explictly want to get the benefits of delta channel.

libs/langgraph/langgraph/graph/state.py
```diff
@@ -120,6 +122,34 @@ def _warn_invalid_state_schema(schema: type[Any] | Any) -> None:
...
+    hint. The check is skipped when `langgraph-api` is not installed (e.g.
+    local execution), since there is no API server to be incompatible.
+    """
+    if not any(isinstance(c, DeltaChannel) for c in channels.values()):
+        return
+    try:
+        api_version = metadata.version("langgraph-api")
+    except metadata.PackageNotFoundError:
+        return
+    if Version(api_version) > _MIN_DELTA_CHANNEL_API_VERSION:
```
longquanzheng: Would it include the 0.10.0rcs?
like 0.10.0rc3? 

maybe even better to just check on `>=0.10.0`. 
because all the RCs in 0.1.0 don't have prune support.

sydney-runkle:
@open-swe can you please do this

oh interesting. I was thinking of try import the version constant in langgraph-api, but I guess either way would work.also also let's just check on >=0.10.0, we don't want to include RCs for this particular condition

libs/langgraph/langgraph/graph/state.py
```diff
@@ -120,6 +122,27 @@ def _warn_invalid_state_schema(schema: type[Any] | Any) -> None:
...
+    server is installed. Skipped when `langgraph-api` is absent (local
+    execution) or the graph has no delta channel.
+    """
+    if not any(isinstance(c, DeltaChannel) for c in channels.values()):
+        return
+    try:
+        api_version = metadata.version("langgraph-api")
+    except metadata.PackageNotFoundError:
+        return
+    if Version(api_version) < Version("0.10.0"):
```
longquanzheng, line 139: we are going to release 0.10.1 to fix the read+prune bug found by Parker. 

maybe we can hold on with that(but not sure 0.10.1 will come tomorrow), or we can release this one, and then update this value to 0.10.1 in next release.

wdyt?

longquanzheng (approved):
if you can, or someone else (including me), create a deployment to use this code with pinned lg-api to 0.9.0 to see if it fail.

christian-bromann:
How do we guarantee this for JS deployments?

libs/langgraph/langgraph/graph/state.py
sydney-runkle: i think we want to raise bc delta channels have different types of reducers

libs/langgraph/langgraph/graph/state.py
sydney-runkle: yeah we can hold off on this one

PR #8077, opened 2026-06-14
prebuilt: add human_approval() ToolCallWrapper for HITL workflows (fixes #8026)

### LINKED ISSUE #8026: [Feature Request]: Add a high-level ApprovalNode for Human-in-the-Loop workflows
### Checked other resources

### Problem
Currently, building Human-in-the-Loop (HITL) workflows in LangGraph requires manual implementation of `interrupt()` and `Command(resume=...)` logic repeatedly. There is no standard, high-level component to handle common "approve/reject/modify" patterns.

### Proposed Solution
I propose adding a reusable `ApprovalNode` to `libs/prebuilt` that abstracts these patterns. Additionally, adding `pause()` and `resume()` convenience methods to the core `Pregel` class would simplify managing long-running agent threads.

### Benefits
- Standardizes HITL patterns across enterprise workflows.
- Reduces boilerplate code for developers.
- Provides structured event models for approval requests and outcomes.

```python
from langgraph.types import interrupt

def human_node(state):
    response = interrupt("Approve?")
    return {"status": response}

builder = StateGraph(dict)
builder.add_node("human", human_node)
builder.add_edge(START, "human")
graph = builder.compile()
```

```shell

```

### Description

I am proposing a standard 'ApprovalNode' for Human-in-the-Loop workflows. Currently, developers have to manually implement interrupt/resume logic. This feature request adds a reusable component and framework enhancements (pause/resume) to simplify this. I have already submitted a PR #8025 with the implementation.

### System Info

Siva141909:
## Summary

Implements the `human_approval()` factory requested in #8026 as a `ToolCallWrapper` for `ToolNode(wrap_tool_call=...)` — no new node class, no new graph topology.

### Design

Each tool call is classified against **deny** → **allow** glob patterns:
- **deny**: blocked immediately, `ToolMessage(status="error")` returned, no interrupt.
- **allow**: executed immediately, no pending record created.
- **unclassified**: a durable `PendingApproval` record is created and surfaced via `interrupt()`.

```python
from langgraph.prebuilt import ToolNode, human_approval

wrapper = human_approval(allow=["read_*", "list_*"], deny=["drop_*"])
node = ToolNode(tools, wrap_tool_call=wrapper)
```

### PendingApproval record (9 fields per reviewer spec)

| Field | Purpose |
|---|---|
| `thread_id` | LangGraph thread identifier |
| `node_name` | Owning ToolNode name (audit) |
| `tool_name` | Tool being gated |
| `tool_call_id` | Unique call ID from LLM message |
| `args_digest` | SHA-256 of canonical JSON args (sort_keys, no whitespace) |
| `policy_result` | Always `"requires_approval"` here; included for auditability |
| `decision_shape` | `"approve_or_reject"` or `"approve_reject_or_edit"` |
| `resume_token` | Deterministic SHA-256 derived from stable call identity¹ |
| `terminal_state` | `None` while pending; set to approved/rejected/edited/expired/cancelled/executed |
| `approved_args_digest` | SHA-256 of edited args when `terminal_state == "edited"` |

¹ `resume_token` is derived deterministically from `(thread_id, node_name, tool_name, tool_call_id, args_digest)` rather than randomly, because LangGraph re-executes the node from the top when resuming from `interrupt()` — a fresh random token would break the token-match validation on every resume.

### Resume protocol

```python
from langgraph.prebuilt.human_approval import ApprovalDecision

# interrupt value is the PendingApproval dict:
pending = state.tasks[0].interrupts[0].value

graph.invoke(
    Command(resume=ApprovalDecision(
        token=pending["resume_token"],       # must match
        tool_call_id=pending["tool_call_id"],# cross-call replay prevention
        action="approve",                    # or "reject" / "edit"
        edited_args={"key": "new_val"},      # required when action="edit"
    ).to_dict()),
    config,
)
```

### Validation (`_validate_decision`)

Four guards in order:
1. `token` must equal `pending.resume_token`
2. `tool_call_id` must equal `pending.tool_call_id` (cross-call replay)
3. `pending.terminal_state` must be `None` (terminal-state guard)
4. `edited_args` must be non-empty when `action == "edit"`

### Edit flow

When `action == "edit"`, the wrapper:
1. Computes `approved_args_digest = SHA-256(canonical(edited_args))`
2. Rebinds execution to the edited args via `request.override(tool_call={...})`
3. Stamps both `args_digest` and `approved_args_digest` on the `ToolMessage.additional_kwargs` for downstream audit

### Why `interrupt()` propagates through ToolNode

`interrupt()` raises `GraphInterrupt(GraphBubbleUp)`. The wrapper's try/except in `_run_one` calls `_handle_tool_error → _default_handle_tool_errors` which re-raises anything that is not `ToolInvocationError`, so `GraphInterrupt` bubbles up to the graph runtime unchanged.

## Files changed

- `libs/prebuilt/langgraph/prebuilt/human_approval.py` — new module
- `libs/prebuilt/langgraph/prebuilt/__init__.py` — exports `human_approval`, `async_human_approval`, `PendingApproval`, `ApprovalDecision`
- `libs/prebuilt/tests/test_human_approval.py` — 20 tests (7 pinned by reviewer)

## Test plan

All 7 regression tests pinned by @rpelevin pass locally:

- [x] allow-listed tool executes without a pending record
- [x] deny-listed tool returns terminal denial without interrupting
- [x] unclassified tool creates pending record; tool does not run before resume
- [x] resume token bound to `tool_call_id` — cross-call replay rejected
- [x] edited args produce distinct `approved_args_digest` visible in final state
- [x] terminal state guard — 6 terminal states each block replay
- [x] checkpoint/resume/reconnect preserves pending record and terminal outcome

Plus 13 additional unit tests (digest helpers, round-trip serialisation, async allow/deny paths, reject path).

```
20 passed in 0.88s
```

Fixes #8026

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/__init__.py
+++ b/libs/prebuilt/langgraph/prebuilt/__init__.py
@@ -2,6 +2,12 @@
 
 from langgraph.prebuilt._tool_call_transformer import ToolCallTransformer
 from langgraph.prebuilt.chat_agent_executor import create_react_agent
+from langgraph.prebuilt.human_approval import (
+    ApprovalDecision,
+    PendingApproval,
+    async_human_approval,
+    human_approval,
+)
 from langgraph.prebuilt.tool_node import (
     InjectedState,
     InjectedStore,
@@ -20,4 +26,8 @@ __all__ = [
     "InjectedState",
     "InjectedStore",
     "ToolRuntime",
+    "human_approval",
+    "async_human_approval",
+    "PendingApproval",
+    "ApprovalDecision",
 ]
--- a/libs/prebuilt/langgraph/prebuilt/human_approval.py
+++ b/libs/prebuilt/langgraph/prebuilt/human_approval.py
@@ -1,0 +1,462 @@
+"""human_approval — production-grade HITL ToolCallWrapper for ToolNode.
+
+Factory that returns a :data:`ToolCallWrapper` (or :data:`AsyncToolCallWrapper`)
+that intercepts every tool call, classifies it against a policy, and—for calls
+that require human oversight—:
+
+1. Creates a *durable* :class:`PendingApproval` record surfaced to the caller
+   via :func:`~langgraph.types.interrupt`.
+2. Waits for a ``Command(resume=<ApprovalDecision dict>)`` whose ``token`` is
+   cryptographically bound to that exact record.
+3. Validates the decision, marks the record terminal, then either executes the
+   tool (with original or edited args) or returns a denial
+   :class:`~langchain_core.messages.ToolMessage`.
+
+The record is intentionally self-contained: it can be serialised to JSON,
+stored in a database, and round-tripped through any LangGraph checkpointer
+without loss of information.
+
+Usage::
+
+    from langgraph.prebuilt import ToolNode, human_approval
+
+    wrapper = human_approval(
+        allow=["read_*", "list_*"],
+        deny=["drop_*"],
+        # everything else → requires_approval
+    )
+    node = ToolNode(tools, wrap_tool_call=wrapper)
+
+Resume protocol::
+
+    # The interrupt value is the PendingApproval serialised to a dict.
+    # Echo the token and tool_call_id unchanged to bind the decision.
+    from langgraph.prebuilt.human_approval import ApprovalDecision
+    from langgraph.types import Command
+
+    graph.invoke(
+        Command(resume=ApprovalDecision(
+            token=pending["resume_token"],
+            tool_call_id=pending["tool_call_id"],
+            action="approve",           # or "reject" / "edit"
+            edited_args={"key": "val"}, # required when action="edit"
+        ).to_dict()),
+        config,
+    )
+"""
+
+from __future__ import annotations
+
+import fnmatch
+import hashlib
+import json
+from collections.abc import Callable
+from dataclasses import asdict, dataclass, field
+from typing import Literal, Optional
+
+from langchain_core.messages import ToolMessage
+
+from langgraph.prebuilt.tool_node import AsyncToolCallWrapper, ToolCallRequest, ToolCallWrapper
+from langgraph.types import Command, interrupt
+
+__all__ = [
+    "human_approval",
+    "async_human_approval",
+    "PendingApproval",
+    "ApprovalDecision",
+]
+
+# ---------------------------------------------------------------------------
+# Type aliases
+# ---------------------------------------------------------------------------
+
+PolicyResult = Literal["allow", "deny", "requires_approval"]
+TerminalState = Literal["approved", "rejected", "edited", "expired", "cancelled", "executed"]
+
+
+# ---------------------------------------------------------------------------
+# Public dataclasses
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class PendingApproval:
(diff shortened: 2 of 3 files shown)


rpelevin:
Nice implementation direction. I would separate record identity from approval capability here.

The deterministic resume_token makes sense as a stable record key across interrupt replay, but I would not make it the capability that authorizes resume. If the token is derived only from thread, node, tool, call, and args fields, a caller that can reconstruct those fields can forge the resume value, and an error that echoes the expected token makes it easier.

The shape I would prefer:

- pending_approval_id can be deterministic from stable call identity;
- resume_secret or resume_nonce is unguessable, persisted with the pending record, and never echoed in errors;
- alternatively, resume_token is an HMAC over the stable preimage with a caller-supplied server secret;
- Command resume must match pending_approval_id plus the secret or HMAC before any execution path resumes;
- mismatch returns a terminal denial or retryable validation error without revealing the expected value.

I would also enforce decision_shape as part of validation: edit should be rejected when the record only allowed approve or reject. That deserves its own regression test because otherwise the recorded decision contract is not actually binding.

Two smaller determinism tests seem worth adding too: use case-sensitive policy matching across platforms, and copy allow and deny lists at wrapper construction so runtime policy cannot drift by mutating the caller's list after the wrapper is created.

That keeps the replay-safe resume behavior while making the approval credential non-forgeable.

Siva141909:
Thanks for the detailed breakdown — all four points are addressed in the latest push (fdf423b).

**HMAC-keyed `resume_token`**

Added a `secret: str | None = None` parameter to both `human_approval()` and `async_human_approval()`. When provided, the token is computed as:

```python
hmac.new(secret.encode(), preimage.encode(), hashlib.sha256).hexdigest()
```

This is non-forgeable without the secret while remaining fully deterministic across node re-executions (same secret + same stable preimage → same HMAC every time, which is the invariant LangGraph's replay model requires). Without a secret it falls back to plain SHA-256 of the same preimage (appropriate when the interrupt channel itself is access-controlled).

The token mismatch error no longer echoes the expected value. Token comparison now uses `hmac.compare_digest()` for timing safety.

The factory always pre-computes and injects the token into `PendingApproval(resume_token=...)` directly, so `__post_init__` only falls back to SHA-256 when the dataclass is constructed directly (e.g., in unit tests).

**`decision_shape` enforcement**

`_validate_decision()` now rejects `action='edit'` when `pending.decision_shape == 'approve_or_reject'` as check #4 (before the `edited_args` emptiness check). New regression test `test_decision_shape_blocks_edit_when_approve_or_reject_only` covers this and also asserts that `approve` is still accepted for the same record.

**Case-sensitive pattern matching**

`_match_patterns()` now uses `fnmatch.fnmatchcase` instead of `fnmatch.fnmatch`. New regression test `test_policy_matching_is_case_sensitive` exercises `_classify()` directly to confirm `Read_*` does not match `read_file` and `delete_*` does not match `Delete_file`.

**List copy at construction**

`allow or []` / `deny or []` replaced with `list(allow) if allow else []` / `list(deny) if deny else []`. New regression test `test_allow_deny_lists_are_copied_at_construction` appends `send_*` to the caller's `allow` list after wrapper construction and verifies the graph still interrupts for `send_email`.

**Test count: 23 passed in 0.90s** (was 20).

PR #8091, opened 2026-06-16
fix(langgraph): serialize `Overwrite` as canonical dict

### LINKED ISSUE #3789: `PatchToolCallsMiddleware` wedges threads: `Overwrite(...)` is type-erased to `{"value": [...]}` across any JSON boundary, breaking `_get_overwrite` and corrupting the messages channel permanently
### Submission checklist

- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)

### Related Issues / PRs

* langchain-ai/deepagents#2471 — same "Tool call ... was cancelled — another message came in" symptom; root cause not diagnosed in that thread. This issue identifies the mechanism that corrupts the thread *after* that marker is emitted.
* langchain-ai/deepagents#2703 — different failure in `PatchToolCallsMiddleware` (`invalid_tool_calls` handling); same `before_agent` path.
* langchain-ai/langgraph#8012 — on a fresh `DeepAgentState` thread, the first `update_state({"messages": ...})` is lost; same family of state-update breakage on the `messages` `DeltaChannel`.

The production symptom is straightforward: a multi-turn chat session that hits the cancel-mid-tool-call path eventually wedges its thread so that every `GET /threads/{id}/state` returns HTTP 400 with `Message dict must contain 'role' and 'content' keys, got {'value': [...]}`. No manual `update_state` calls are involved; this is a normal user-driven conversation against a deployed LangGraph Platform app.

I have not pinpointed the exact internal codepath in the deployed `langgraph-api 0.9.0` stack that performs the JSON roundtrip on the offending write (candidates: a JSON-backed checkpoint adapter, a cross-worker queue, or a trace-replay path — none of which were straightforward to confirm from a clean install). What I have proven is **a real, reproducible incompatibility between the `Overwrite` value `PatchToolCallsMiddleware` emits and any JSON serialization layer in the pipeline.** Given any such layer exists, the bug fires. The repro below uses the actual `langgraph_api.serde.json_dumpb` serializer to simulate the boundary in three lines:

```python
"""
deepagents==0.6.7, langgraph==1.2.4, langgraph-api==0.9.0, langchain-core==1.4.1
"""
import orjson
from langchain_core.messages import HumanMessage, AIMessage
from langgraph_api.serde import json_dumpb
from deepagents.graph import DeepAgentState

g = StateGraph(DeepAgentState)
g.add_node("noop", lambda s: {})
g.add_edge(START, "noop"); g.add_edge("noop", END)
graph = g.compile(checkpointer=MemorySaver())
cfg = {"configurable": {"thread_id": "t"}}
graph.invoke({"messages": [HumanMessage(content="hi", id="h1")]}, cfg)

# `PatchToolCallsMiddleware.before_agent` returns {"messages": Overwrite([...])}
# whenever there are dangling tool calls (cancelled tool message). Any layer in
# the pipeline that uses orjson/json on the value will collapse the dataclass:
ow = Overwrite(value=[
    HumanMessage(content="hi", id="h1"),
    AIMessage(content="patched", id="a1"),
])
erased = orjson.loads(json_dumpb(ow))
# erased == {'value': [
#   {'content': 'hi', 'additional_kwargs': {}, 'response_metadata': {},
#    'type': 'human', 'name': None, 'id': 'h1'},
#   {'content': 'patched', ..., 'type': 'ai', ...}
# ]}

# Once that erased shape reaches the channel reducer (here demonstrated via the
# closest accessible boundary, `graph.update_state` — but any internal path
# that feeds the erased value to the reducer produces the same failure):
graph.update_state(cfg, {"messages": erased})
# → ValueError: Message dict must contain 'role' and 'content' keys, got {'value': [...]}
```

The bug isolated to two lines, independent of any graph:

```python
from langgraph.channels.binop import _get_overwrite
import dataclasses

ow = Overwrite(value=[{"type": "human", "content": "hi"}])
erased = dataclasses.asdict(ow)            # {'value': [{'type': 'human', 'content': 'hi'}]}
print(_get_overwrite(erased))              # (False, None)  ← should be (True, [...])
```

I'd love help confirming the exact internal call site in `langgraph-api` / `langgraph-runtime-inmem` / `langgraph-checkpoint-postgres` that performs the roundtrip on the persisted write — but the producer/recognizer mismatch is the underlying defect either way, and the fix below closes it regardless of which layer triggers it.

```shell
ValueError: Message dict must contain 'role' and 'content' keys,
got {'value': [
  {'content': 'hi', 'additional_kwargs': {}, 'response_metadata': {},
   'type': 'human', 'name': None, 'id': 'h1'},
  {'content': 'patched', 'additional_kwargs': {}, 'response_metadata': {},
   'type': 'ai', 'name': None, 'id': 'a1', 'tool_calls': [],
   'invalid_tool_calls': [], 'usage_metadata': None}
]}
For troubleshooting, visit: https://docs.langchain.com/oss/python/langchain/errors/MESSAGE_COERCION_FAILURE
```

In production this surfaces as `HTTP 400` on `GET /threads/{thread_id}/state` against a deployed LangGraph Platform app. The thread is permanently wedged: every subsequent state read replays the same write through the reducer and 400s again.

## What happens

On a normal multi-turn chat session built with `create_deep_agent(...)`, an in-flight tool call gets cancelled because a new user message arrives before the tool result. `PatchToolCallsMiddleware` correctly inserts a tombstone `ToolMessage` and the conversation continues. At some later turn — and from that point on, forever — every `GET /threads/{id}/state` returns 400 with `Message dict must contain 'role' and 'content' keys, got {'value': [...]}`.

## Root cause (confirmed and reproduced)

There are three colluding facts:

1. **The producer.** [`deepagents/middleware/patch_tool_calls.py:46`](https://github.com/langchain-ai/deepagents/blob/main/src/deepagents/middleware/patch_tool_calls.py) returns:

   ```python
   return {"messages": Overwrite(patched_messages)}
   ```

   Same shape is used in [`deepagents/middleware/filesystem.py:1904`](https://github.com/langchain-ai/deepagents/blob/main/src/deepagents/middleware/filesystem.py):

   ```python
   Command(update={"messages": Overwrite([*messages[:-1], tagged])})
   ```

2. **The type-erasure point.** `langgraph.types.Overwrite` is a `@dataclass(slots=True)` with a single field `value: Any`. `orjson` serializes dataclasses natively. Therefore any code path that JSON-encodes a state update (`langgraph_api.serde.json_dumpb`, cross-worker queues, JSON-backed adapters, etc.) erases `Overwrite(value=[…])` into `{"value": […]}`. The contained `BaseMessage` objects are simultaneously erased to their `model_dump()` dict shape — `{"content": ..., "additional_kwargs": {}, "response_metadata": {}, "type": "human", "name": None, "id": ...}`. This is exactly the shape that appears in the 400.

3. **The vulnerable recognizer.** [`langgraph/channels/binop.py:31-37`](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/channels/binop.py):

   ```python
   def _get_overwrite(value):
       if isinstance(value, Overwrite):
           return True, value.value
       if isinstance(value, dict) and len(value) == 1 and OVERWRITE in value:   # OVERWRITE == "__overwrite__"
           return True, value[OVERWRITE]
       return False, None
   ```

   This recognizes the typed `Overwrite` instance and the dict form `{"__overwrite__": ...}`, but **not** the orjson-erased form `{"value": ...}`. The erased dict therefore returns `(False, None)`, falls through `DeltaChannel.replay_writes` / `BinaryOperatorAggregate.update`, and is passed verbatim to the channel's reducer (`deepagents._messages_reducer._messages_delta_reducer` for `DeepAgentState`). The reducer then calls `convert_to_messages([{"value": [...]}])`, which raises with the exact production error string.

Once persisted in `checkpoint_writes`, the corrupted write replays every time the state is read. The thread is permanently 400'ed and cannot be unwedged without bypassing the read path.

## What I expect

Either:

- `PatchToolCallsMiddleware` (and the eviction path in `FilesystemMiddleware`) should not rely on `Overwrite` for in-place rewrites of the `messages` channel — because the safety of that mechanism depends on the `Overwrite` dataclass surviving every serialization boundary in the LangGraph pipeline, which it does not.
- Alternatively, the dict form of `Overwrite` (post-orjson) should be recognized by `_get_overwrite` — but this is a `langgraph` fix, see the cross-filed issue (link below).

## Suggested fix (verified)

**Replace `Overwrite(patched_messages)` with `[RemoveMessage(REMOVE_ALL_MESSAGES), *patched_messages]`.**

The deepagents-local `_messages_delta_reducer` at `deepagents/_messages_reducer.py:63-69` already implements the reset semantics via `REMOVE_ALL_MESSAGES`. The change in `patch_tool_calls.py:46`:

```python
# before
return {"messages": Overwrite(patched_messages)}
# after
from langchain_core.messages import RemoveMessage
from langgraph.graph.message import REMOVE_ALL_MESSAGES
return {"messages": [RemoveMessage(id=REMOVE_ALL_MESSAGES), *patched_messages]}
```

Same shape applies to `filesystem.py:1904` (`Command(update={"messages": Overwrite([...])})`).

I verified this against two cases:

1. **Functional equivalence** — running the patched middleware on a realistic dangling-tool-call state produces the same final message list as the current code. The only visible difference is that the inserted `ToolMessage` tombstone receives a UUID via LangGraph's `ensure_message_ids` hook (the current code leaves it as `id=None`). Behavior downstream is unchanged.

2. **JSON-roundtrip immunity** — the same erased-via-`json_dumpb` payload that wedges the current code (`ValueError: Message dict must contain 'role' and 'content' keys, got {'value': [...]}`) passes cleanly with the fix:

   ```text
   === Current: Overwrite(patched) after JSON roundtrip ===
   update_state ERROR: ValueError: Message dict must contain 'role' and 'content' keys, got {'value': [...]}

   === Fix: [RemoveMessage(ALL), *patched] after JSON roundtrip ===
   update_state OK, final 2 messages:
     HumanMessage h1 'hi'
     AIMessage    a1 'patched'
   ```

   `RemoveMessage` is a pydantic-v2 `BaseMessage` subclass, so it serializes via `EXT_PYDANTIC_V2` in msgpack and as a typed dict (`{"type": "remove", "id": "__remove_all__"}`) in JSON — both survive every layer in the pipeline.

## Companion defense-in-depth fix in `langgraph`

I'll file a separate issue against `langchain-ai/langgraph` covering the broader hardening (either making `Overwrite` survive JSON roundtrip or having `_get_overwrite` recognize the erased shape). Linking here when filed. The deepagents fix above is independent of that and can land first.

### Environment / System Info

```
OS: macOS 14 (Darwin 25.5.0), production runtime: AWS Lambda (al2023)
Python: 3.11
deepagents: 0.6.7
langgraph: 1.2.4
langgraph-api: 0.9.0
langgraph-sdk: 0.4.2
langgraph-runtime-inmem: 0.29.0
langgraph-checkpoint: 4.1.1
langchain: 1.3.4
langchain-core: 1.4.1
```

No special env vars required to reproduce. Production deployment is a Lambda-backed LangGraph Platform image; the bug reproduces with `MemorySaver` in a single process once the JSON roundtrip is simulated (see repro).

sydney-runkle:
Fixes langchain-ai/deepagents#3789

---

`Overwrite` currently uses a dataclass shape with a single `value` field. Generic JSON serializers such as `orjson` encode that as `{"value": ...}`, but reducer channels only recognize `Overwrite(...)` instances or the canonical wire form `{"__overwrite__": ...}`. Once the dataclass-erased shape reaches a reducer channel, it is treated as a normal update and can corrupt message-channel writes.

This changes `Overwrite` to be a dict-like wrapper around the canonical `__overwrite__` key while preserving the `.value` accessor and `Overwrite(value=...)` construction. Generic JSON roundtrips now produce the shape reducer channels already recognize.

For `DeltaChannel`, a serialized overwrite can also contain reducer-native but unnormalized values, such as message dicts. `DeltaChannel` now runs the reducer on the overwritten base even when there are no subsequent writes, so reducers can normalize the base value after JSON boundaries while preserving overwrite-as-reset behavior.

Tested with:

- `make -C libs/langgraph format`
- `make -C libs/langgraph lint`
- `make -C libs/langgraph test TEST="tests/test_pregel.py -k overwrite"`
- `make -C libs/langgraph test TEST="tests/test_channels.py tests/test_pregel.py -k overwrite"`
- DeepAgents compatibility script against this branch via `PYTHONPATH`, covering serialized `PatchToolCallsMiddleware` overwrites and next-turn middleware access

Full `make -C libs/langgraph test` was also attempted, but exceeded 15 minutes locally before completing.

### DIFF
--- a/libs/langgraph/langgraph/channels/delta.py
+++ b/libs/langgraph/langgraph/channels/delta.py
@@ -154,7 +154,7 @@ class DeltaChannel(Generic[Value], BaseChannel[Any, Any, Any]):
                 base = _copy.copy(ow_value) if ow_value is not None else self.typ()
                 start = i + 1
         remaining = values[start:]
-        self.value = self.reducer(base, remaining) if remaining else base
+        self.value = self.reducer(base, remaining)
 
     def update(self, values: Sequence[Any]) -> bool:
         if not values:
@@ -178,7 +178,7 @@ class DeltaChannel(Generic[Value], BaseChannel[Any, Any, Any]):
                 else self.typ()
             )
             remaining = [v for i, v in enumerate(values) if i != overwrite_idx]
-            self.value = self.reducer(base, remaining) if remaining else base
+            self.value = self.reducer(base, remaining)
             return True
         base = self.typ() if self.value is MISSING else self.value
         self.value = self.reducer(base, list(values))
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -25,6 +25,7 @@ from xxhash import xxh3_128_hexdigest
 
 from langgraph._internal._cache import default_cache_key
 from langgraph._internal._constants import INTERRUPT as _INTERRUPT_KEY
+from langgraph._internal._constants import OVERWRITE
 from langgraph._internal._fields import get_cached_annotated_keys, get_update_as_tuples
 from langgraph._internal._retry import default_retry_on
 from langgraph._internal._typing import MISSING, DeprecatedKwargs
@@ -934,8 +935,7 @@ def interrupt(value: Any) -> Any:
     )
 
 
-@dataclass(slots=True)
-class Overwrite:
+class Overwrite(dict[str, Any]):
     """Bypass a reducer and write the wrapped value directly to a `BinaryOperatorAggregate` channel.
 
     Receiving multiple `Overwrite` values for the same channel in a single super-step
@@ -974,5 +974,16 @@ class Overwrite:
         ```
     """
 
-    value: Any
-    """The value to write directly to the channel, bypassing any reducer."""
+    __slots__ = ()
+
+    def __init__(self, value: Any) -> None:
+        super().__init__({OVERWRITE: value})
+
+    @property
+    def value(self) -> Any:
+        """The value to write directly to the channel, bypassing any reducer."""
+        return self[OVERWRITE]
+
+    @value.setter
+    def value(self, value: Any) -> None:
+        self[OVERWRITE] = value
--- a/libs/langgraph/tests/test_pregel.py
+++ b/libs/langgraph/tests/test_pregel.py
@@ -9336,6 +9336,47 @@ def test_overwrite_survives_json_roundtrip(
     assert result == {"messages": ["b"]}
 
 
+def test_delta_channel_overwrite_normalizes_json_messages(
+    sync_checkpointer: BaseCheckpointSaver,
+) -> None:
+    class State(TypedDict):
+        messages: Annotated[
+            list[AnyMessage],
+            DeltaChannel(_messages_delta_reducer),
+        ]
+
+    def node_a(state: State):
+        return {"messages": [HumanMessage(content="original", id="h0")]}
+
+    def node_b(state: State):
+        update = Overwrite([HumanMessage(content="replacement", id="h1")])
+        overwrite = json.loads(json.dumps(update, default=lambda obj: obj.model_dump()))
+        assert overwrite == {
+            "__overwrite__": [
+                {
+                    "content": "replacement",
+                    "additional_kwargs": {},
+                    "response_metadata": {},
+                    "type": "human",
+                    "name": None,
+                    "id": "h1",
+                }
+            ]
+        }
+        return {"messages": overwrite}
+
+    builder = StateGraph(State)
+    builder.add_node("node_a", node_a)
+    builder.add_node("node_b", node_b)
+    builder.add_edge(START, "node_a")
+    builder.add_edge("node_a", "node_b")
+
+    graph = builder.compile(checkpointer=sync_checkpointer)
+    result = graph.invoke({"messages": []}, {"configurable": {"thread_id": "1"}})
+
+    assert result == {"messages": [HumanMessage(content="replacement", id="h1")]}
+
+
 @pytest.mark.parametrize("as_json", [False, True])
 def test_overwrite_parallel(
     sync_checkpointer: BaseCheckpointSaver, as_json: bool
--- a/libs/langgraph/tests/test_pregel.py
+++ b/libs/langgraph/tests/test_pregel.py
@@ -9310,6 +9310,32 @@ def test_overwrite_sequential(
     assert result == {"messages": ["b"]}
 
 
+def test_overwrite_survives_json_roundtrip(
+    sync_checkpointer: BaseCheckpointSaver,
+) -> None:
+    class State(TypedDict):
+        messages: Annotated[list, operator.add]
+
+    def node_a(state: State):
+        return {"messages": ["a"]}
+
+    def node_b(state: State):
+        overwrite = json.loads(json.dumps(Overwrite(["b"])))
+        assert overwrite == {"__overwrite__": ["b"]}
+        return {"messages": overwrite}
+
+    builder = StateGraph(State)
+    builder.add_node("node_a", node_a)
+    builder.add_node("node_b", node_b)
+    builder.add_edge(START, "node_a")
+    builder.add_edge("node_a", "node_b")
+
+    graph = builder.compile(checkpointer=sync_checkpointer)
+    result = graph.invoke({"messages": ["START"]}, {"configurable": {"thread_id": "1"}})
+
+    assert result == {"messages": ["b"]}
+
+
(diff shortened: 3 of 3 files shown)


sydney-runkle:
don't like this approach

PR #8100, opened 2026-06-16
feat(cli): allow prebuild images for langgraph deploy

hari-dhanushkodi:
Fixes #

adds a --image flag to `langgraph deploy`, allows a user to `langgraph build` a custom image and then deploy this specific image with `langgraph deploy`.


phvash (approved):
Looks good. Non-blocking, but two ux improvement suggestions"
1. we should add existence check for the image, and 
2. We should also confirm the image's platform is `linux/amd64`  to avoid any gotchas during deployment.

PR #8109, opened 2026-06-16
feat: add delta-channel-dump recovery script as examples to dump from Postgres for deltaChannel rollback

longquanzheng:
Add a standalone recovery tool at `examples/delta-channel-dump/` for operators rolling back from langgraph >= 1.2 / deepagents 0.6.x to an older runtime that does not understand `EXT_DELTA_SNAPSHOT` msgpack blobs. Without this, channels like `messages` can appear empty after rollback because the old reducer does not decode delta snapshots.

`dump.py` connects directly to Postgres, walks the checkpoint parent chain (mirroring `aget_delta_channel_history`), decodes msgpack blobs via `ormsgpack` with a hand-rolled EXT 0–7 hook (no langgraph imports), and emits JSON with per-channel `seed` + oldest-to-newest `writes`. The operator reduces and re-applies via `update_state` manually.

## Scope

- Postgres checkpointer only
- Deps: `psycopg[binary]`, `ormsgpack` (not added to repo `pyproject.toml` — operator installs at runtime)
- No reducer application, no DB writes, no AES/custom encryption support (fails loudly on encrypted blob types)

## How verified

- E2E against local Postgres: 15 runs of `delta_channel_messages_freq` (`snapshot_frequency=10`)
- Asserted `delta_kind == "snapshot"`, `len(seed) == 10`, `len(writes) == 5`, message ids `ai-0`..`ai-14`
- Confirmed portability: fresh venv with only `pip install "psycopg[binary]" ormsgpack` produced identical output

### DIFF
--- a/examples/delta-channel-dump/README.md
+++ b/examples/delta-channel-dump/README.md
@@ -1,0 +1,125 @@
+# delta-channel-dump
+
+Recover messages (and other channels) from a Postgres-backed LangGraph thread
+written by **deepagents 0.6.x / langgraph >= 1.2** (DeltaChannel format) before
+rolling back to **deepagents 0.5.x / langgraph 0.5.x**.
+
+On the older runtime, `add_messages` does not understand the `EXT_DELTA_SNAPSHOT`
+msgpack ext code and silently returns an empty list for affected channels. This
+tool reads the raw checkpoint blobs from Postgres and emits JSON you can inspect
+and re-apply via `update_state`.
+
+## Install
+
+```bash
+pip install "psycopg[binary]" ormsgpack
+```
+
+## Run
+
+```bash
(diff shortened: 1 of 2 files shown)


examples/delta-channel-dump/README.md
```diff
@@ -0,0 +1,125 @@
+# delta-channel-dump
+
+Recover messages (and other channels) from a Postgres-backed LangGraph thread
+written by **deepagents 0.6.x / langgraph >= 1.2** (DeltaChannel format) before
+rolling back to **deepagents 0.5.x / langgraph 0.5.x**.
```
sydney-runkle: i think you mean langgraph <1.2?

sydney-runkle (approved):
fine w/ this as an example script, hopefully most folks won't have to use this

PR #8110, opened 2026-06-17
feat: support multiple extra LangSmith trace projects in SDK

mdrxy:
LangGraph Python SDK: `langsmith_tracing` now accepts `project_names` to trace a run to multiple extra LangSmith projects.

---

Adds an optional `project_names: list[str]` field to the `LangSmithTracing` TypedDict so callers can request multiple extra LangSmith trace replica projects on a run. The singular `project_name` field is unchanged and continues to work.

The SDK already forwards `langsmith_tracing` directly as `langsmith_tracer`, so this PR is a schema/documentation + test coverage change for the Python SDK pass-through behavior and is fully backward compatible.

Note: this SDK change depends on the corresponding server-side `langgraph-api` change to normalize/dedupe `project_name` and `project_names` and fan out traces to multiple projects. That server-side change, as well as JS LangSmith tracing support, lives outside this public repo.

### DIFF
--- a/libs/sdk-py/langgraph_sdk/schema.py
+++ b/libs/sdk-py/langgraph_sdk/schema.py
@@ -112,7 +112,13 @@ class LangSmithTracing(TypedDict, total=False):
     """Configuration for LangSmith tracing."""
 
     project_name: str
-    """The LangSmith project name to trace to."""
+    """A single extra LangSmith project name to trace to."""
+    project_names: list[str]
+    """Multiple extra LangSmith project names to trace to.
+
+    Combined with `project_name` (if both are provided) to form the full set of
+    extra projects. The server also traces to its primary tracing project.
+    """
     example_id: str
     """The LangSmith example/dataset ID to associate with the trace."""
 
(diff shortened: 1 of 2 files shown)


libs/sdk-py/langgraph_sdk/schema.py
```diff
@@ -112,7 +112,13 @@ class LangSmithTracing(TypedDict, total=False):
     """Configuration for LangSmith tracing."""
 
     project_name: str
```
sydney-runkle, line 114: str | list[str]?  so we don't break

mdrxy: I think keeping `project_name: str` unchanged is the backward-compatible path. The new multi-project field is `project_names: list[str]`; the server-side change normalizes/dedupes `project_name` and `project_names`. Typing `project_name` as `str | list[str]` would imply the server accepts a list under the existing singular key, which isn’t the intended API shape unless we also support that server-side.

sydney-runkle: fwiw i thought this was deleted i just can't see

paarth-a (approved):

PR #8125, opened 2026-06-17
fix(langgraph): snapshot `DeltaChannel` overwrite supersteps

sydney-runkle:
When a DeltaChannel receives an Overwrite, force that checkpoint to store a snapshot so sparse replay starts from the post-overwrite value. This also aligns live DeltaChannel overwrite handling with BinaryOperatorAggregate by letting Overwrite bypass other reducer writes in the same superstep.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -70,6 +70,7 @@ from langgraph.callbacks import (
     GraphResumeEvent,
 )
 from langgraph.channels.base import BaseChannel
+from langgraph.channels.binop import _get_overwrite
 from langgraph.channels.delta import DeltaChannel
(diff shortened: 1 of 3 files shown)


libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -1135,6 +1152,8 @@ def _put_checkpoint(self, metadata: CheckpointMetadata) -> None:
         )
         for k in channels_to_snapshot:
             new_counters[k] = (0, 0)
+        if do_checkpoint:
+            self._delta_channels_with_overwrite.difference_update(channels_to_snapshot)
```
longquanzheng, line 1156: I believe it's the same if we just clear it to empty set?
`self._delta_channels_with_overwrite = set()`?

longquanzheng (approved):

PR #8126, opened 2026-06-17
test(langgraph): cover JSON overwrite sentinel replay

sydney-runkle:
Refs #8125
Depends on #8125

Adds regression coverage for API-style JSON `Overwrite` sentinel updates on `DeltaChannel`, verifying the streamed update shape, snapshot boundary, and later checkpoint replay.

Not run; this test is expected to pass once #8125 lands.


sydney-runkle:
Closing this because the regression belongs in langgraph-api, not langgraph. The issue is API JSON serialization of Overwrite to the reserved __overwrite__ sentinel.

PR #8127, opened 2026-06-17
fix(langgraph): Make `Overwrite` survive JSON roundtrips

sydney-runkle:
Add a `type: Literal["__overwrite__"]` discriminator field to `Overwrite` and teach `_get_overwrite()` to recognise the dataclass-erased `{"value": ..., "type": "__overwrite__"}` form. This keeps `Overwrite` semantics intact across JSON boundaries that strip dataclass types (e.g. `langgraph-api.serde.json_dumpb`), without requiring callers to switch to the dict sentinel form.

This was an oversight in the initial `Overwrite` implementation, it should be json serializable out of the box.

Fixing langchain-ai/deepagents#3789

### DIFF
--- a/libs/langgraph/langgraph/channels/binop.py
+++ b/libs/langgraph/langgraph/channels/binop.py
@@ -29,11 +29,25 @@ def _strip_extras(t):  # type: ignore[no-untyped-def]
 
 
 def _get_overwrite(value: Any) -> tuple[bool, Any]:
-    """Inspects the given value and returns (is_overwrite, overwrite_value)."""
+    """Inspects the given value and returns (is_overwrite, overwrite_value).
+
+    Recognises three forms:
+
+    * The typed `Overwrite` dataclass instance.
+    * The sentinel-keyed `{"__overwrite__": value}` dict form.
+    * The dataclass-erased `{"value": ..., "type": "__overwrite__"}` form that
+      results from JSON-serialising an `Overwrite` (e.g. an `orjson`-encoded
+      state update routed through the LangGraph API server). This keeps the
+      `Overwrite` semantics intact across JSON boundaries that strip dataclass
+      types.
+    """
     if isinstance(value, Overwrite):
         return True, value.value
-    if isinstance(value, dict) and len(value) == 1 and OVERWRITE in value:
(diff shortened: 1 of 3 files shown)


libs/langgraph/langgraph/channels/binop.py
```diff
@@ -29,11 +29,25 @@ def _strip_extras(t):  # type: ignore[no-untyped-def]
...
+    """
     if isinstance(value, Overwrite):
         return True, value.value
-    if isinstance(value, dict) and len(value) == 1 and OVERWRITE in value:
-        return True, value[OVERWRITE]
+    if isinstance(value, dict):
+        if len(value) == 1 and OVERWRITE in value:
+            return True, value[OVERWRITE]
+        if value.get("type") == OVERWRITE and "value" in value:
+            return True, value["value"]
```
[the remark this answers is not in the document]
sydney-runkle: that's fine

libs/langgraph/langgraph/types.py
```diff
@@ -976,3 +976,9 @@ def node_b(state: State):
 
     value: Any
     """The value to write directly to the channel, bypassing any reducer."""
+
+    type: Literal["__overwrite__"] = "__overwrite__"
```
longquanzheng, line 980: cursor said we can use `type: Literal["__overwrite__"] = OVERWRITE`

not sure that is true

longquanzheng (approved):

libs/langgraph/langgraph/channels/binop.py
```diff
@@ -29,11 +29,25 @@ def _strip_extras(t):  # type: ignore[no-untyped-def]
 
 
 def _get_overwrite(value: Any) -> tuple[bool, Any]:
-    """Inspects the given value and returns (is_overwrite, overwrite_value)."""
+    """Inspects the given value and returns (is_overwrite, overwrite_value).
+
+    Recognises three forms:
+
+    * The typed `Overwrite` dataclass instance.
+    * The sentinel-keyed `{"__overwrite__": value}` dict form.
```
longquanzheng, line 37: maybe add more details where this is from -- seems like from user hand writing the return from a node?

PR #8242, opened 2026-07-01
fix: delta replay bug for channel migrated from non-delta channels

longquanzheng:
Fixes #

Read the full contributing guidelines: https://docs.langchain.com/oss/python/contributing/overview

> **All contributions must be in English.** See the [language policy](https://docs.langchain.com/oss/python/contributing/overview#language-policy).

If you paste a large clearly AI generated description here your PR may be IGNORED or CLOSED!

Thank you for contributing to LangGraph! Follow these steps to have your pull request considered as ready for review.

1. PR title: Should follow the format: TYPE(SCOPE): DESCRIPTION

    - feat(langgraph): add multi-tenant support
  - Allowed TYPE and SCOPE values: https://github.com/langchain-ai/langgraph/blob/main/.github/workflows/pr_lint.yml#L19-L43

2. PR description:

  - Write 1-2 sentences summarizing the change.
  - The `Fixes #xx` line at the top is **required** for external contributions — update the issue number and keep the keyword. This links your PR to the approved issue and auto-closes it on merge.
  - If there are any breaking changes, please clearly describe them.
  - If this PR depends on another PR being merged first, please include "Depends on #PR_NUMBER" in the description.

3. Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified.

  - We will not consider a PR unless these three are passing in CI.

4. How did you verify your code works?

Additional guidelines:

  - All external PRs must link to an issue or discussion where a solution has been approved by a maintainer, and you must be assigned to that issue. PRs without prior approval will be closed.
  - PRs should not touch more than one package unless absolutely necessary.
  - Do not update the `uv.lock` files or add dependencies to `pyproject.toml` files (even optional ones) unless you have explicit permission to do so by a maintainer.

## Social handles (optional)

Twitter: @
LinkedIn: https://linkedin.com/in/

### DIFF
--- a/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/_delta_fixtures.py
+++ b/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/_delta_fixtures.py
@@ -54,6 +54,12 @@ async def build_delta_chain(
     stored: list[RunnableConfig] = []
     parent_cfg: RunnableConfig | None = None
 
+    # Track supersteps since the channel's last snapshot, mirroring Pregel's
+    # `counters_since_delta_snapshot` bookkeeping: every step increments the
+    # counter; a snapshot step resets it (and carries no counter entry, so
+    # the walk locates the seed exactly `supersteps` hops back).
+    supersteps_since_snapshot = 0
+
     for step in range(total_steps):
         config: RunnableConfig = {
             "configurable": {
@@ -68,11 +74,19 @@ async def build_delta_chain(
 
         channel_values: dict[str, Any] = {}
         channel_versions: dict[str, int] = {}
+        supersteps_since_snapshot += 1
+        counters: dict[str, tuple[int, int]] = {}
         if step in snapshot_set:
             channel_values[channel] = _DeltaSnapshot(
                 write_value_fn(step),
             )
             channel_versions[channel] = step + 1
+            supersteps_since_snapshot = 0
+        else:
+            counters[channel] = (
+                supersteps_since_snapshot,
+                supersteps_since_snapshot,
+            )
 
         cp = Checkpoint(
             v=1,
@@ -84,9 +98,10 @@ async def build_delta_chain(
             updated_channels=None,
         )
         new_versions = dict(channel_versions)
-        parent_cfg = await saver.aput(
-            config, cp, generate_metadata(step=step), new_versions
-        )
+        md = generate_metadata(step=step)
+        if counters:
+            md["counters_since_delta_snapshot"] = counters
+        parent_cfg = await saver.aput(config, cp, md, new_versions)
         stored.append(parent_cfg)
 
         # Write a pending write for non-snapshot steps so the walk has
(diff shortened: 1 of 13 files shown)


Hotragn:
Your resolution hunk fixes a second bug that the PR title doesn't mention, and nothing in the test suite pins it — so it could easily be dropped in review as incidental. Measurements below in case they're useful as a regression test. **This is yours, I'm not claiming it**; I only went looking after finding the divergence independently and then finding your diff already fixed it.

### The second bug

`InMemorySaver.get_delta_channel_history` resolves the target with
`config["configurable"].get("checkpoint_id", "")` (`memory/__init__.py:163`) and looks that
value up directly. A config without `checkpoint_id` therefore matches nothing, the chain
comes back empty, and every requested channel gets `{"writes": []}` with **no** `seed`.
The consumer reads an absent seed as "start empty", so the channel silently hydrates empty —
no error.

Every other implementation resolves the omitted id to the latest checkpoint:

| implementation | resolves via |
|---|---|
| `BaseCheckpointSaver` (default) | `self.get_tuple(config)` — `base/__init__.py:625` |
| `SqliteSaver` / `AsyncSqliteSaver` | `sqlite/__init__.py:529`, `sqlite/aio.py:640` |
| `PostgresSaver` / `AsyncPostgresSaver` | `postgres/__init__.py:470`, `postgres/aio.py:420` |
| `InMemorySaver` | — (the line above) |

Repro — chain `cp1(acc="SEED") --w1--> cp2 --w2--> cp3`, asking for `acc` at the head:

```
=== InMemorySaver ===
  explicit checkpoint_id : seed='SEED'     writes=['w1', 'w2']
  omitted  checkpoint_id : seed='<ABSENT>' writes=[]        <-- silently empty
=== SqliteSaver (:memory:) ===
  explicit checkpoint_id : seed='SEED'     writes=['w1', 'w2']
  omitted  checkpoint_id : seed='SEED'     writes=['w1', 'w2']
```

Applying only your resolution hunk to `main` makes the omitted-id row match SqliteSaver exactly.

### Nothing covers it

Applying that hunk alone to `main`:

| suite | before | after |
|---|---|---|
| `libs/checkpoint` (excl. `test_redis_cache.py`) | 158 passed, 1 skipped | **158 passed, 1 skipped** |
| `libs/langgraph` `test_delta_channel_migration.py` + `test_channels.py` | 50 passed | **50 passed** |

Not one test moves, in either direction. Every existing call site — the 9 conformance specs in
`spec/test_delta_channel_history.py`, the 7 in `checkpoint-sqlite/tests/test_get_delta_channel_history.py`,
and the ones in `test_memory.py` — passes a config returned by `put()`, which always carries
`checkpoint_id`. The omitted-id path is never exercised.

### Scope, honestly

I could not reach this from the framework itself: `_loop.py` merges `saved.config` into
`checkpoint_config` before calling `channels_from_checkpoint`, and `main.py` passes
`saved.config` directly, so both arrive resolved. It bites direct callers of the public
(beta) saver API, and it's a contract divergence against the base class and the other four
implementations. Distinct from #8470, which is the same *symptom* from a different cause
(`saver=None` for nested subgraphs).

One note if you pick this back up: the branch is currently conflicted against `main`.

PR #8278, opened 2026-07-03
fix: check SKIP_RERAISE_SET in BackgroundExecutor.__exit__ to suppres…

### LINKED ISSUE #8277: error_handler: handled exception is re-raised anyway when the failing node runs in parallel with other tasks in the same superstep
### Checked other resources

### Related Issues / PRs

None found (searched for `error_handler`).

```python
import operator
from typing import Annotated

class S(TypedDict, total=False):
    xs: Annotated[list, operator.add]


def boom(state):
    raise RuntimeError("boom")


def handler(state, error: NodeError):
    print("HANDLER FIRED:", error.node, type(error.error).__name__)
    return {"xs": ["degraded"]}


# Single task in the superstep: works as documented.
single = (
    StateGraph(S)
    .add_node("boom", boom, error_handler=handler)
    .add_edge(START, "boom")
    .add_edge("boom", END)
    .compile()
)
print("single-node:", single.invoke({}))  # {'xs': ['degraded']}

# Two parallel tasks in the same superstep: the handler fires and its write is
# applied, but invoke() STILL raises the handled exception.
parallel = (
    StateGraph(S)
    .add_node("boom", boom, error_handler=handler)
    .add_node("ok", lambda s: {"xs": ["ok"]})
    .add_edge(START, "boom")
    .add_edge(START, "ok")
    .add_edge("boom", END)
    .add_edge("ok", END)
    .compile()
)
print("plain-parallel:", parallel.invoke({}))  # raises RuntimeError: boom
```

Output:

```
single-node: {'xs': ['degraded']}
HANDLER FIRED: boom RuntimeError
Traceback (most recent call last):
  ...
  File ".../langgraph/pregel/main.py", line 2914, in stream
    with SyncPregelLoop(
  File ".../langgraph/pregel/_loop.py", line 1719, in __exit__
    return self.stack.__exit__(exc_type, exc_value, traceback)
  ...
  File ".../langgraph/pregel/_executor.py", line 117, in __exit__
    task.result()
  ...
  File ".../langgraph/pregel/_retry.py", line 617, in run_with_retry
    return task.proc.invoke(task.input, config)
  ...
  File "repro.py", line 20, in boom
    raise RuntimeError("boom")
RuntimeError: boom
During task with name 'boom' and id '...'
```

See above. Notably, when streaming with a checkpointer, the events show that the
handler task ran and the graph **completed the whole run** — the spurious re-raise
happens only when the `SyncPregelLoop` context exits:

```python
# same graph plus a "join" node after both branches, compiled with InMemorySaver
for ev in g.stream({}, config=cfg, stream_mode="updates"):
    events.append(ev)   # raises RuntimeError: boom at generator exit

# events: [{'ok': {'xs': ['ok']}}, {'__error_handler__boom': {'xs': ['degraded']}}]
# get_state(cfg).values: {'xs': ['degraded', 'ok', 'join']}   <- run finished
# get_state(cfg).next:   ()
```

So the graph state is exactly what the documented error-handler semantics promise
("recover gracefully rather than abort the entire graph"), but the caller of
`invoke()`/`stream()` gets the original exception anyway. `ainvoke()` behaves the
same. Adding a `retry_policy` doesn't change the behavior (handler fires after
retries are exhausted, then the exception is still re-raised).

### Description

**What happens.** When a node with an `error_handler` fails while **other tasks run
in the same superstep** (plain parallel branches or a `Send` fan-out), the handler
fires and its `Command`/state update is applied, but the original exception is
re-raised to the caller at loop exit. With a single task in the superstep the
handler suppresses the exception as documented.

**Root cause (from reading the source).** In the multi-task path,
`PregelRunner.tick` submits tasks to the `BackgroundExecutor` and, when a future
fails and the task routes to an error handler, correctly records the handled
failure (`_handled_exception_ids.add(id(exc))`, `SKIP_RERAISE_SET.add(fut)` in
`pregel/_runner.py`), so `_panic_or_proceed` / `_should_stop_others` skip it.
However, the failed future was submitted with `__reraise_on_exit__=True` and stays
in `BackgroundExecutor.tasks` (its `done` callback keeps failed tasks). On loop
teardown, `BackgroundExecutor.__exit__` (`pregel/_executor.py`) calls
`task.result()` on every task with `reraise=True` **without consulting
`SKIP_RERAISE_SET`**, so the already-handled exception escapes.

The single-task fast path in `tick` (`len(tasks) == 1 and timeout is None and
get_waiter is None`) runs the task inline — no future is ever created — which is
why minimal single-node examples work and this slips through.

**Verified on** 1.2.5, 1.2.6 and 1.2.7 (sync `invoke`, `stream`, and async
`ainvoke`); the relevant code on `main` looks unchanged
(`_executor.py::BackgroundExecutor.__exit__` still re-raises unconditionally for
`reraise=True` tasks and has no reference to `SKIP_RERAISE_SET`).

### System Info

- langgraph 1.2.7 (also reproduced on 1.2.5, 1.2.6)
- langgraph-checkpoint 4.1.1
- langgraph-prebuilt 1.1.0
- langgraph-sdk 0.4.2
- langchain-core 1.4.8
- pydantic 2.13.4
- Python 3.12.12, Linux x86_64 (glibc 2.42)

UdayKiranJadi:
## What this fixes
Closes #8277

When a node with an `error_handler` fails while running in parallel with 
other tasks in the same superstep, the handler fired and its state update 
was applied correctly — but the original exception was still re-raised to 
the caller at loop exit.

## Root cause
`BackgroundExecutor.__exit__` (and its async counterpart) called 
`task.result()` on every failed future with `reraise=True`, without 
consulting `SKIP_RERAISE_SET` — the set that `_runner.py` uses to mark 
exceptions as already handled by an error handler.

The single-task fast path in `tick` runs inline (no future is created), 
so `SKIP_RERAISE_SET` was never hit and single-node examples worked fine. 
This masked the bug.

## Fix
- Moved `SKIP_RERAISE_SET` from `_runner.py` to a new shared module 
  `pregel/_constants.py` to avoid a circular import 
  (`_runner` → `_executor` → `_runner`)
- Both `BackgroundExecutor.__exit__` (sync) and 
  `AsyncBackgroundExecutor.__aexit__` (async) now skip re-raising if the 
  future is in `SKIP_RERAISE_SET`

## Reproduction
```python
# Before fix: raises RuntimeError: boom
# After fix: {'xs': ['degraded', 'ok']}
parallel.invoke({})
```

### DIFF
--- a/libs/langgraph/langgraph/pregel/_constants.py
+++ b/libs/langgraph/langgraph/pregel/_constants.py
@@ -1,0 +1,9 @@
+from __future__ import annotations
+
+import asyncio
+import concurrent.futures
+import weakref
+
+SKIP_RERAISE_SET: weakref.WeakSet[concurrent.futures.Future | asyncio.Future] = (
+    weakref.WeakSet()
+)
--- a/libs/langgraph/langgraph/pregel/_executor.py
+++ b/libs/langgraph/langgraph/pregel/_executor.py
@@ -19,6 +19,7 @@ from typing_extensions import ParamSpec
 
 from langgraph._internal._future import CONTEXT_NOT_SUPPORTED, run_coroutine_threadsafe
 from langgraph.errors import GraphBubbleUp
+from langgraph.pregel._constants import SKIP_RERAISE_SET
 
 P = ParamSpec("P")
 T = TypeVar("T")
@@ -113,6 +114,8 @@ class BackgroundExecutor(AbstractContextManager):
             for task, (_, reraise) in tasks.items():
                 if not reraise:
                     continue
+                if task in SKIP_RERAISE_SET:
+                    continue
                 try:
                     task.result()
                 except concurrent.futures.CancelledError:
@@ -203,6 +206,8 @@ class AsyncBackgroundExecutor(AbstractAsyncContextManager):
             # re-raise the first exception that occurred in a task
             for task, (_, reraise) in tasks.items():
                 if not reraise:
+                    continue
+                if task in SKIP_RERAISE_SET:
                     continue
                 try:
                     if exc := task.exception():
--- a/libs/langgraph/langgraph/pregel/_runner.py
+++ b/libs/langgraph/langgraph/pregel/_runner.py
@@ -67,9 +67,7 @@ EXCLUDED_FRAME_FNAMES = (
     "concurrent/futures/_base.py",
 )
 
-SKIP_RERAISE_SET: weakref.WeakSet[concurrent.futures.Future | asyncio.Future] = (
-    weakref.WeakSet()
-)
+from langgraph.pregel._constants import SKIP_RERAISE_SET
 
 
 class FuturesDict(Generic[F, E], dict[F, PregelExecutableTask | None]):


deuscapturus:
This PR is still needed to resolve a bug

deuscapturus:
https://github.com/langchain-ai/langgraph/commit/70153ceba22013d34b51317cd76c976a62af80cb

@nfcampos Can you help get this fix through as the creator of the SKIP_RERAISE_SET pattern

PR #8290, opened 2026-07-06
fix: delta channel bug with updateState on fresh thread will force snapshot instead of stub checkpoint

longquanzheng:
Fixes langchain-ai/deepagents#3774

Reworks the fresh-thread `update_state` fix for `DeltaChannel`: instead of creating stub checkpoint (#8011), force a new Snapshot into the first checkpoint so the value is stored inline and needs no ancestor replay.

## Background

`update_state` / `bulk_update_state` on a *fresh* thread silently dropped the first write to a `DeltaChannel`.

By design, a `DeltaChannel` reconstructs its value by walking ancestor checkpoints and replaying the writes attached to them. 
Checkpoint writes need a parent to persist. But on a fresh thread there is no ancestor, there is no parent to use.

#8011 fixed this by lazily persisting an empty stub checkpoint (step `-1`) to give the first write a parent to use. This PR reverts that and takes a simpler route.

## A better fix

On a fresh thread (`saved is None`), force a snapshot of every available `DeltaChannel` into the first checkpoint via `create_checkpoint(..., channels_to_snapshot=...)`. 

This way, the read/replay path is untouched and no stub is needed. 

## Behavior change

A fresh-thread `update_state` now produces a **single** self-contained checkpoint (step `0`, no parent, snapshot inline) instead of two (stub step `-1` + update step `0`). This is visible via `get_state_history`.

## Verify

`make format`, `make lint`, `make test` in `libs/langgraph`. `tests/test_delta_channel_update_state.py` is updated to assert the single-checkpoint shape on a fresh thread and pins the non-fresh paths (`update_state` after `invoke`, consecutive `update_state`, `bulk_update_state`) against regression.


libs/langgraph/langgraph/pregel/main.py
```diff
@@ -2500,7 +2471,21 @@ async def aperform_superstep(
...
+            # checkpoint instead of relying on ancestor write-replay.
+            delta_snapshot = (
+                {
+                    k
+                    for k, ch in channels.items()
+                    if isinstance(ch, DeltaChannel) and ch.is_available()
+                }
+                if saved is None
+                else None
+            )
```
syachamaneni-lc, line 2485: minor: Create a function since above is also same?

syachamaneni-lc (approved):

libs/langgraph/langgraph/pregel/main.py
longquanzheng: going to ignore this...it seems more python way to use this one liner and overkill to wrap a function for this.

PR #8311, opened 2026-07-09
fix: honor resource auth action filters

jkennedyvz:
## Description
Fixes resource-specific auth decorators so `@auth.on.threads(actions=[...])` registers only the requested actions instead of a wildcard handler. The change also rejects mismatched `resources` on resource-specific decorators to avoid silent authorization mismatches.

## Test Plan
- [x] Added and ran focused SDK auth decorator tests: `uv run --directory /workspace/langgraph/libs/sdk-py pytest tests/test_auth.py --no-header --no-summary --tb=short`
- [x] Ran `make -C /workspace/langgraph/libs/sdk-py format` and `make -C /workspace/langgraph/libs/sdk-py lint`

Made by [Open SWE](https://openswe.vercel.app/agents/dbd7ba1c-d790-3528-cf0d-20eef0bbb3d8)

### DIFF
--- a/libs/sdk-py/langgraph_sdk/auth/__init__.py
+++ b/libs/sdk-py/langgraph_sdk/auth/__init__.py
@@ -392,7 +392,7 @@ class _ResourceOn(typing.Generic[VCreate, VRead, VUpdate, VDelete, VSearch]):
     def __call__(
         self,
         *,
-        resources: str | Sequence[str],
+        resources: str | Sequence[str] | None = None,
         actions: str | Sequence[str] | None = None,
     ) -> Callable[
         [_ActionHandler[VCreate | VUpdate | VRead | VDelete | VSearch]],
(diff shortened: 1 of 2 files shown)


libs/sdk-py/langgraph_sdk/auth/__init__.py
```diff
@@ -416,25 +416,38 @@ def __call__(
...
+                    f"Resource-specific decorator for {self.resource!r} cannot "
+                    f"register handlers for {resource_list!r}. Use @auth.on(...) "
+                    "for multiple resources."
+                )
+            if isinstance(actions, str):
+                action_list = [actions]
+            else:
+                action_list = list(actions) if actions is not None else ["*"]
+            for action in action_list:
+                _register_handler(self.auth, self.resource, action, handler)
```
Sanjays2402, line 440: If `action_list` contains an action that already has a handler, earlier actions in the list have already been registered when `_register_handler` raises. That leaves auth state partially mutated even though decorator evaluation failed; can this validate all `(resource, action)` keys before registering any of them?

libs/sdk-py/langgraph_sdk/auth/__init__.py
```diff
@@ -416,25 +416,38 @@ def __call__(
...
+                raise ValueError(
+                    f"Resource-specific decorator for {self.resource!r} cannot "
+                    f"register handlers for {resource_list!r}. Use @auth.on(...) "
+                    "for multiple resources."
+                )
+            if isinstance(actions, str):
+                action_list = [actions]
+            else:
+                action_list = list(actions) if actions is not None else ["*"]
+            for action in action_list:
```
Sanjays2402, line 439: nothing validates the action strings here, so `auth.on.threads(actions="raed")` registers a `("threads", "raed")` key that never gets looked up and the handler silently never runs. before this change a typo still landed on the `"*"` key so the handler at least fired; the attribute form (`auth.on.threads.read`) also fails loudly with AttributeError. worth checking `action` against the resource's known actions and raising, since the failure mode is a permissive one.

Issue #8185, opened 2026-06-24
Checkpoint serialization rejects fractions.Fraction and complex (TypeError) though Decimal is supported


OmkumarSolanki:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph.
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Related Issues / PRs

`JsonPlusSerializer` already supports `Decimal`, `UUID`, `datetime`, `Path`, IP addresses, etc.; `Fraction` and `complex` are the gap.

### Reproduction Steps / Example Code (Python)

```python
from decimal import Decimal
from fractions import Fraction
from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer

s = JsonPlusSerializer()

s.loads_typed(s.dumps_typed(Decimal("1.5")))   # OK -> Decimal('1.5')
s.loads_typed(s.dumps_typed(Fraction(1, 3)))    # TypeError: Type is not msgpack serializable: Fraction
s.loads_typed(s.dumps_typed(complex(1, 2)))     # TypeError: Type is not msgpack serializable: complex
```

### Error Message and Stack Trace (if applicable)

```
TypeError: Type is not msgpack serializable: Fraction
TypeError: Type is not msgpack serializable: complex
```

### Description

`JsonPlusSerializer` special-cases many stdlib types in `_msgpack_default` (`libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py`) — including `Decimal` — but not `fractions.Fraction` or the builtin `complex`. Since neither is natively serializable by `ormsgpack`, both fall through to `default` and raise `TypeError`.

The effect: a graph whose checkpointed state contains a `Fraction` (exact rational arithmetic) or a `complex` number (scientific / signal-processing agents) cannot be checkpointed at all — `put`/`aput` raises. This is an inconsistency with the already-supported `Decimal`, and both have an obvious lossless representation.

[...]

PR #8338, opened 2026-07-14
fix(checkpoint): add range and PurePath serialization support

### LINKED ISSUE #8326: Checkpoint serialization rejects range and PurePath variants (TypeError) -- same gap as Fraction/complex in #8185
### Checked other resources

### Related Issues / PRs

Same root cause as #8185 (Fraction/complex) -- JsonPlusSerializer's _msgpack_default doesn't handle these types, and PurePath variants slip past the existing isinstance(obj, pathlib.Path) check since they aren't Path subclasses.

```python
import pathlib

s = JsonPlusSerializer()
s.loads_typed(s.dumps_typed(range(0, 10, 2)))
# TypeError: Type is not msgpack serializable: range

s.loads_typed(s.dumps_typed(pathlib.PurePosixPath("/a/b")))
# TypeError: Type is not msgpack serializable: PurePosixPath
```

```shell
Error Message
TypeError: Type is not msgpack serializable: range
TypeError: Type is not msgpack serializable: PurePosixPath
```

### Description

_msgpack_default in libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py special-cases many stdlib types (Decimal, Path, UUID, etc.) but not range, and the existing Path handler checks isinstance(obj, pathlib.Path), which excludes PurePath/PurePosixPath/PureWindowsPath since they aren't Path subclasses.

Proposed fix, happy to open a PR:
- range: encode (start, stop, step), reconstruct via range(start, stop, step)
- PurePath variants: broaden the isinstance check to pathlib.PurePath -- the  existing handler already encodes obj.__class__ and rebuilds from obj.parts, so no other logic needs to change

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 25.3.0: Wed Jan 28 20:49:24 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T8132
> Python Version:  3.13.14 (v3.13.14:fd17997c386, Jun 10 2026, 08:55:00) [Clang 21.0.0 (clang-2100.1.1.101)]

Mukller:
## Summary

Fixes #8326. Same root cause as #8185 (Fraction/complex missed types).

`JsonPlusSerializer` in `libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py`
does not handle two families of Python stdlib types:

1. **`range`** — no handler at all, raises `TypeError: Type is not msgpack serializable: range`.
2. **`PurePath` variants** — `PurePosixPath` / `PureWindowsPath` inherit from `pathlib.PurePath`
   but NOT from `pathlib.Path`, so the existing `isinstance(obj, pathlib.Path)` guard misses them.

These types are valid state values (e.g. `range(0, 10)` as a loop bound, `PurePosixPath`
as a portable file reference) and their presence in a checkpoint causes serialization to crash
instead of round-tripping cleanly.

### `jsonplus.py`

1. Broaden the `pathlib.Path` check to `pathlib.PurePath`:
   ```python
   # Before:
   elif isinstance(obj, pathlib.Path):
   # After:
   elif isinstance(obj, pathlib.PurePath):
   ```
   This handles `PurePosixPath`, `PureWindowsPath`, `PurePath`, **and** all concrete
   subclasses — the existing encode/decode logic (`obj.parts` → `classname(*parts)`)
   works identically for all of them.

2. Add a `range` handler before the `BaseException` catch-all:
   ```python
   elif isinstance(obj, range):
       return ormsgpack.Ext(
           EXT_CONSTRUCTOR_POS_ARGS,
           _msgpack_enc(
               ("builtins", "range", (obj.start, obj.stop, obj.step)),
           ),
       )
   ```
   Deserializes as `builtins.range(start, stop, step)`, which faithfully reproduces
   `range(0, 10, 2)` → `range(0, 10, 2)`.

### `_msgpack.py`

Add the new types to `SAFE_MSGPACK_TYPES` so they round-trip correctly when
`LANGGRAPH_STRICT_MSGPACK=true` is set:
- `("pathlib", "PurePath")`, `("pathlib", "PurePosixPath")`, `("pathlib", "PureWindowsPath")`
- Same three entries for `pathlib._local` (Python 3.13+)
- `("builtins", "range")`

## Reproduction (from the issue)

```python
import pathlib

s = JsonPlusSerializer()

# Previously: TypeError: Type is not msgpack serializable: range
assert s.loads_typed(s.dumps_typed(range(0, 10, 2))) == range(0, 10, 2)

# Previously: TypeError: Type is not msgpack serializable: PurePosixPath
p = pathlib.PurePosixPath("/a/b")
assert s.loads_typed(s.dumps_typed(p)) == p
```

Closes #8326.

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py
@@ -41,14 +41,22 @@ SAFE_MSGPACK_TYPES: frozenset[tuple[str, ...]] = frozenset(
         ("ipaddress", "IPv6Address"),
         ("ipaddress", "IPv6Interface"),
         ("ipaddress", "IPv6Network"),
-        # pathlib
+        # pathlib (concrete + pure variants)
         ("pathlib", "Path"),
         ("pathlib", "PosixPath"),
         ("pathlib", "WindowsPath"),
+        ("pathlib", "PurePath"),
+        ("pathlib", "PurePosixPath"),
+        ("pathlib", "PureWindowsPath"),
         # pathlib in Python 3.13+
         ("pathlib._local", "Path"),
         ("pathlib._local", "PosixPath"),
         ("pathlib._local", "WindowsPath"),
+        ("pathlib._local", "PurePath"),
+        ("pathlib._local", "PurePosixPath"),
+        ("pathlib._local", "PureWindowsPath"),
+        # builtins
+        ("builtins", "range"),
         # zoneinfo
         ("zoneinfo", "ZoneInfo"),
         # regex
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -350,7 +350,7 @@ def _msgpack_default(obj: Any) -> str | ormsgpack.Ext:
                 ),
             ),
         )
-    elif isinstance(obj, pathlib.Path):
+    elif isinstance(obj, pathlib.PurePath):
         return ormsgpack.Ext(
             EXT_CONSTRUCTOR_POS_ARGS,
             _msgpack_enc(
@@ -528,6 +528,13 @@ def _msgpack_default(obj: Any) -> str | ormsgpack.Ext:
             meta = (obj.dtype.str, obj.shape, order, buf)
             return ormsgpack.Ext(EXT_NUMPY_ARRAY, _msgpack_enc(meta))
 
+    elif isinstance(obj, range):
+        return ormsgpack.Ext(
+            EXT_CONSTRUCTOR_POS_ARGS,
+            _msgpack_enc(
+                ("builtins", "range", (obj.start, obj.stop, obj.step)),
+            ),
+        )
     elif isinstance(obj, BaseException):
         return repr(obj)
     else:


Mukller (commented):
## Self-Review

### Root Cause

Two gaps in `JsonPlusSerializer._msgpack_default`:

1. `range` has no handler — falls through to the `TypeError` raise at the end of the function.
2. The `pathlib.Path` isinstance check excludes `PurePath` and its direct subclasses
   (`PurePosixPath`, `PureWindowsPath`) because they don't inherit from `Path`.

Both were called out in issue #8326 as the same structural gap as the Fraction/complex
miss fixed in #8185.

### Fix

- **`PurePath`**: one-word change — `pathlib.Path` → `pathlib.PurePath`. Since
  `Path` is a subclass of `PurePath`, all concrete paths still match. The encode/decode
  logic (`obj.parts` → `classname(*parts)`) works identically for all path variants.

- **`range`**: new `elif isinstance(obj, range)` branch using the existing
  `EXT_CONSTRUCTOR_POS_ARGS` mechanism. Serializes `(start, stop, step)` and
  reconstructs via `builtins.range(start, stop, step)`.

- **`SAFE_MSGPACK_TYPES`**: added the three `PurePath` variants (plus Python 3.13
  `pathlib._local` equivalents) and `builtins.range` so strict mode doesn't break.

### Correctness

- Concrete `Path`/`PosixPath`/`WindowsPath` instances: unaffected — still match
  the broadened `PurePath` check.
- `range` is immutable and equality-comparable by value — round-trips exactly.
- No change to any existing code path for types already handled.

### Scope

Two files, four hunks. Zero new imports (pathlib already imported; builtins never
needs importing).

PR #8343, opened 2026-07-15
fix(prebuilt): allow return_direct tools to run at remaining_steps=1

### LINKED ISSUE #8204: create_react_agent aborts return_direct=True tools with "need more steps" when remaining_steps == 1
### Checked other resources

### Related Issues / PRs

_No response_

```python
import asyncio

@tool(return_direct=True)
def my_direct_tool() -> str:
    """Returns answer directly without going back to LLM."""
    return "direct answer"

# Minimal fake model that always calls my_direct_tool
from unittest.mock import MagicMock
from langchain_core.messages import AIMessage

model = MagicMock()
model.invoke.return_value = AIMessage(
    content="",
    tool_calls=[{"name": "my_direct_tool", "args": {}, "id": "1", "type": "tool_call"}]
)
model.bind_tools.return_value = model

agent = create_react_agent(model, [my_direct_tool])

# remaining_steps=1 should be enough for a return_direct tool
# (it exits immediately — no second LLM call needed)
result = agent.invoke({
    "messages": [HumanMessage("run tool")],
    "remaining_steps": 1
})

print(result["messages"][-1].content)
# Expected: "direct answer"
# Actual:   "Sorry, need more steps to process this request."
```

```shell
No exception raised. Wrong output:

AIMessage(content='Sorry, need more steps to process this request.')

Expected:
ToolMessage(content='direct answer', name='my_direct_tool')
```

### Description

I'm using create_react_agent with a tool decorated with return_direct=True.

I expect the tool to execute and return its result directly even when
remaining_steps == 1, because return_direct tools exit the agent immediately
without making another LLM call — they only need 1 step, not 2.

Instead, the agent aborts with "Sorry, need more steps" even though 1 step
is sufficient.

Root cause: In libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py,
_are_more_steps_needed() enforces remaining_steps < 2 as a hard abort for
ANY tool call, without checking whether all requested tools are
return_direct=True.

Buggy code (~line 629):
    if remaining_steps < 1 and all_tools_return_direct:
        return True
    elif remaining_steps < 2 and has_tool_calls:   # <-- blindly aborts direct tools too
        return True

The logic is inverted — the return_direct check was placed on the wrong
branch. The fix is to only abort on remaining_steps < 2 if the tools are
NOT all return_direct, since direct tools only consume 1 step.

### System Info

System Information
------------------
> OS:  Linux (WSL2)
> Python Version:  3.14.6 (main, Jun 23 2026, 15:18:23) [Clang 22.1.3]

Package Information
-------------------
> langchain_core: 1.4.7
> langsmith: 0.8.18
> langchain_protocol: 0.0.16
> langgraph_sdk: 0.4.2

Other Dependencies
------------------
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.9
> packaging: 25.0
> pydantic: 2.13.4
> pytest: 9.0.3
> pyyaml: 6.0.3
> requests: 2.33.1
> tenacity: 9.1.2
> typing-extensions: 4.15.0
> websockets: 15.0.1

Mukller:
## Summary

Fixes #8204.

`create_react_agent` uses `_are_more_steps_needed` to decide whether to abort with
"need more steps" instead of running a tool call. The logic was:

```python
# Before (buggy):
if remaining_steps < 1 and all_tools_return_direct:
    return True
elif remaining_steps < 2 and has_tool_calls:   # catches direct tools too!
    return True
```

The second branch fires for **any** tool call when `remaining_steps == 1`, including
`return_direct=True` tools that only need **one** step (tool execute → done, no
second LLM call). This causes agents using `return_direct=True` tools to fail with
the abort message instead of returning the tool result when `remaining_steps=1`.

## Root Cause

`return_direct` tools exit the agent immediately after execution — they don't
make a follow-up LLM call. They therefore need only 1 remaining step to complete,
not 2. The old code didn't distinguish between the two cases in the `< 2` guard.

## Fix

Compute the minimum steps needed based on whether all requested tools are
`return_direct`:

```python
# After (fixed):
if remaining_steps is not None and has_tool_calls:
    min_steps_needed = 1 if all_tools_return_direct else 2
    if remaining_steps < min_steps_needed:
        return True
```

This correctly allows `remaining_steps=1` to proceed when all tools are
`return_direct`, and still aborts at `remaining_steps=1` for normal (non-direct)
tools that need an extra LLM step.

## Reproduction (from the issue)

```python
@tool(return_direct=True)
def my_direct_tool() -> str:
    return "direct answer"

agent = create_react_agent(model, [my_direct_tool])
result = agent.invoke({
    "messages": [HumanMessage("run tool")],
    "remaining_steps": 1
})
# Before: AIMessage(content='Sorry, need more steps to process this request.')
# After:  ToolMessage(content='direct answer', name='my_direct_tool')
```

Closes #8204.

### DIFF
--- a/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
+++ b/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py
@@ -625,10 +625,12 @@ def create_react_agent(
             else False
         )
         remaining_steps = _get_state_value(state, "remaining_steps", None)
-        if remaining_steps is not None:
-            if remaining_steps < 1 and all_tools_return_direct:
-                return True
-            elif remaining_steps < 2 and has_tool_calls:
+        if remaining_steps is not None and has_tool_calls:
+            # return_direct tools only consume 1 step (tool call → done, no LLM re-entry).
+            # Non-direct tools need at least 2 steps: one for tool execution and one
+            # for the subsequent LLM call to process the tool result.
+            min_steps_needed = 1 if all_tools_return_direct else 2
+            if remaining_steps < min_steps_needed:
                 return True
 
         return False


Mukller (commented):
## Self-Review

### Root Cause
`_are_more_steps_needed` used `remaining_steps < 2` as a hard cutoff for
**all** tool calls. `return_direct=True` tools need only 1 step (execute → done,
no re-entry into the LLM), so `remaining_steps == 1` is sufficient. The old code
incorrectly fired the abort for these tools as well.

The first branch (`remaining_steps < 1 and all_tools_return_direct`) was dead in
practice — reaching `remaining_steps < 1` while `has_tool_calls` would mean the
LLM made tool calls with zero budget remaining, which the graph should never allow.

### Fix
Compute `min_steps_needed` from whether all tools are `return_direct`:
- `1` if every requested tool has `return_direct=True` (skip re-entry)
- `2` otherwise (need tool execution + one more LLM call)

The guard is simplified: a single `if remaining_steps < min_steps_needed` replaces
the two disconnected `if`/`elif` branches.

### Correctness
| `remaining_steps` | `all_return_direct` | `has_tool_calls` | Before | After |
|---|---|---|---|---|
| 1 | True | True | abort ❌ | proceed ✓ |
| 1 | False | True | abort ✓ | abort ✓ |
| 0 | True | True | proceed ❌ | abort ✓ |
| 2 | any | True | proceed ✓ | proceed ✓ |
| any | any | False | proceed ✓ | proceed ✓ |

The old code had a logic bug at row 3 (0 steps, direct tools): `0 < 1` is False
so the first branch didn't fire, `0 < 2` is True so the second branch fired — but
only because it coincidentally caught the case. The new logic handles row 3 explicitly.

### Scope
One function, no API change, no new state, no new imports.

PR #8344, opened 2026-07-15
fix(config): fix silenced async guard in get_config() on Python < 3.11

### LINKED ISSUE #8203: get_config() async guard silenced on Python < 3.11 — RuntimeError never raised in async context
### Checked other resources

### Related Issues / PRs

_No response_

```python
import asyncio
import sys

print(sys.version_info)  # e.g. (3, 10, ...)

# Simulate being on Python 3.10 by confirming behaviour:
# This should raise RuntimeError but silently does nothing instead.
async def main():
    from langgraph.config import get_config
    # Expected: RuntimeError("Python 3.11 or later required to use this in an async context")
    # Actual: no error raised — guard is silently swallowed
    get_config()

asyncio.run(main())
# No exception raised on Python < 3.11 — the guard is broken
```

```shell
No error raised. Expected:
RuntimeError: Python 3.11 or later required to use this in an async context
```

### Description

I'm using get_config() from within an async context on Python 3.10.

I expect a RuntimeError to be raised warning that Python 3.11+ is required for async use.

Instead, no error is raised at all. The guard is completely silent.

Root cause: In langgraph/config.py, the except RuntimeError: pass block wraps the
intentional `raise RuntimeError("Python 3.11 or later required...")`, causing
the except clause to catch and silently discard its own warning.

Buggy code (libs/langgraph/langgraph/config.py ~line 19):
    try:
        if asyncio.current_task():
            raise RuntimeError(...)  # <-- this gets caught below!
    except RuntimeError:
        pass   # silences both the env error AND the intentional raise

### System Info

System Information
------------------
> OS:  Linux (WSL2)
> Python Version:  3.14.6 (main, Jun 23 2026, 15:18:23) [Clang 22.1.3]

Package Information
-------------------
> langchain_core: 1.4.7
> langsmith: 0.8.18
> langchain_protocol: 0.0.16
> langgraph_sdk: 0.4.2

Other Dependencies
------------------
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.9
> packaging: 25.0
> pydantic: 2.13.4
> pytest: 9.0.3
> pyyaml: 6.0.3
> requests: 2.33.1
> tenacity: 9.1.2
> typing-extensions: 4.15.0
> websockets: 15.0.1

Mukller:
## Summary

Fixes #8203.

`get_config()` in `libs/langgraph/langgraph/config.py` has an async-context guard
for Python < 3.11. The guard was meant to raise `RuntimeError` when called from an
async context on Python versions that don't propagate contextvars through
`asyncio.create_task`. Instead, it silently swallowed its own error.

## Root Cause

```python
# Before (buggy):
try:
    if asyncio.current_task():
        raise RuntimeError(           # raised here...
            "Python 3.11 or later required..."
        )
except RuntimeError:
    pass                              # ...silently swallowed here
```

The `except RuntimeError: pass` block was intended to handle the case where
`asyncio.current_task()` itself raises `RuntimeError` (Python < 3.10 with no
running event loop). It also catches the intentional raise, neutralising the guard.

## Fix

Store the result of `asyncio.current_task()` in a boolean inside the `try` block,
then raise the intentional error **outside** the `try` block:

```python
# After (fixed):
if sys.version_info < (3, 11):
    _in_async = False
    try:
        _in_async = asyncio.current_task() is not None
    except RuntimeError:
        pass                          # only catches errors from current_task()
    if _in_async:
        raise RuntimeError(           # raised outside try — cannot be caught
            "Python 3.11 or later required..."
        )
```

Closes #8203.

### DIFF
--- a/libs/langgraph/langgraph/config.py
+++ b/libs/langgraph/langgraph/config.py
@@ -16,13 +16,15 @@ def _no_op_stream_writer(c: Any) -> None:
 
 def get_config() -> RunnableConfig:
     if sys.version_info < (3, 11):
+        _in_async = False
         try:
-            if asyncio.current_task():
-                raise RuntimeError(
-                    "Python 3.11 or later required to use this in an async context"
-                )
+            _in_async = asyncio.current_task() is not None
         except RuntimeError:
             pass
+        if _in_async:
+            raise RuntimeError(
+                "Python 3.11 or later required to use this in an async context"
+            )
     if var_config := var_child_runnable_config.get():
         return var_config
     else:


Mukller (commented):
## Self-Review

### Root Cause
The `except RuntimeError: pass` block in `get_config()` was written to silence errors from `asyncio.current_task()` when there is no running event loop (Python < 3.10). The bug: the `raise RuntimeError(...)` guard was inside the same `try` block, so the `except` clause immediately caught and discarded it — on any Python < 3.11 version.

### Fix
Separate the two concerns:
1. `_in_async = asyncio.current_task() is not None` — inside try, catches env errors from `current_task()`.
2. `if _in_async: raise RuntimeError(...)` — outside try, propagates to the caller.

### Correctness

| Python | Async context | Before | After |
|--------|---------------|--------|-------|
| < 3.11 | yes | silent (bug) | RuntimeError raised correctly |
| < 3.11 | no loop | silent | silent (expected) |
| < 3.11 | no | silent | silent |
| >= 3.11 | any | guard skipped | guard skipped |

### Scope
Seven-line change inside one function. No API change, no new imports.

PR #8362, opened 2026-07-17
feat(langgraph): expose `trace_policy` on `add_node`

ccurme:
Adds `TracePolicy`, allowing specification of callables to process tracing inputs, as well as tags:
```python
@dataclass(**_DC_KWARGS)
class TracePolicy:

    process_inputs: Callable[[Any], Any] | None = None

    tags: list[str] = field(default_factory=list)
```

We thread the `trace_inputs` param of a node's `RunnableSeq` through `PregelNode`, `StateNodeSpec`, and expose in `add_node` via `TracePolicy`.

Motivation is to allow transforms on tracing from nodes. LangChain middleware node-type hooks (`before/after_*`) currently trace state; for longish middleware stacks and long message histories this adds considerable latency (500-1,000ms per middleware in deepagents eval suite).

### DIFF
--- a/libs/langgraph/langgraph/_internal/_runnable.py
+++ b/libs/langgraph/langgraph/_internal/_runnable.py
@@ -572,7 +572,6 @@ class RunnableSeq(Runnable):
         *steps: RunnableLike,
         name: str | None = None,
         trace_inputs: Callable[[Any], Any] | None = None,
-        trace_outputs: Callable[[Any], Any] | None = None,
     ) -> None:
         """Create a new RunnableSeq.
 
@@ -598,7 +597,6 @@ class RunnableSeq(Runnable):
         self.steps = steps_flat
         self.name = name
         self.trace_inputs = trace_inputs
-        self.trace_outputs = trace_outputs
 
     def __or__(
         self,
@@ -691,9 +689,7 @@ class RunnableSeq(Runnable):
             run_manager.on_chain_error(e)
             raise
         else:
-            run_manager.on_chain_end(
-                self.trace_outputs(input) if self.trace_outputs is not None else input
-            )
+            run_manager.on_chain_end(input)
             return input
 
     async def ainvoke(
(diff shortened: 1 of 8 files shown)


libs/langgraph/langgraph/graph/state.py
```diff
@@ -672,6 +672,7 @@ def add_node(
         error_handler: StateNode[Any, ContextT] | None = None,
         destinations: dict[str, str] | tuple[str, ...] | None = None,
         timeout: float | timedelta | TimeoutPolicy | None = None,
+        trace_inputs: Callable[[Any], Any] | None = None,
```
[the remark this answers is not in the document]
ccurme: can do.

libs/langgraph/langgraph/pregel/_read.py
```diff
@@ -227,9 +233,9 @@ def node(self) -> Runnable[Any, Any] | None:
         elif self.bound is DEFAULT_BOUND and len(writers) == 1:
             return writers[0]
         elif self.bound is DEFAULT_BOUND:
-            return RunnableSeq(*writers)
+            return RunnableSeq(*writers, trace_inputs=self.trace_inputs)
         elif writers:
-            return RunnableSeq(self.bound, *writers)
+            return RunnableSeq(self.bound, *writers, trace_inputs=self.trace_inputs)
```
Sanjays2402: When `bound` is a `RunnableCallable`, `RunnableSeq` starts the transformed parent run but then the first step starts a child run with the original input. The payload is still recorded on that child trace, so this does not actually omit message history or secrets from LangSmith.

ccurme: Thanks for flagging! Updated the ref docs to clarify. For redacting across all runs, use the LangSmith client's `hide_inputs` / `hide_outputs` / `anonymizer`.

libs/langgraph/langgraph/_internal/_runnable.py
```diff
@@ -689,7 +691,9 @@ def invoke(
             run_manager.on_chain_error(e)
             raise
         else:
-            run_manager.on_chain_end(input)
+            run_manager.on_chain_end(
+                self.trace_outputs(input) if self.trace_outputs is not None else input
```
Sanjays2402: `process_outputs` runs after the `try` block, so if the processor raises, the node invocation fails without `on_chain_error` and leaves this trace run unfinished. Can the transform and `on_chain_end` call stay inside the exception-handled path?

Issue #8157, opened 2026-06-21
Checkpoint serialization drops deque maxlen: a bounded deque becomes unbounded after a round-trip


wilmerdooley:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Related Issues / PRs

I could not find an existing report. `deque` support was added in #1311, which did not carry `maxlen` through.

### Reproduction Steps / Example Code (Python)

```python
from collections import deque
from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer

s = JsonPlusSerializer()
original = deque([1, 2, 3], maxlen=3)

restored = s.loads_typed(s.dumps_typed(original))

print(original.maxlen)   # 3
print(restored.maxlen)   # None  <- the bound is gone

restored.append(99)
restored.append(100)
print(list(restored))    # [1, 2, 3, 99, 100]  grows without bound
                         # the original would have evicted to [3, 99, 100]
```

### Error Message and Stack Trace (if applicable)

No error. The failure is silent: the restored deque keeps its contents but loses its `maxlen`, so it stops evicting and grows without bound.

### Description

A `deque` created with a `maxlen` loses that bound when it is serialized and restored by the checkpoint serializer. A bounded deque held in graph state therefore becomes effectively unbounded after the first checkpoint round-trip, which defeats its purpose (capping memory / keeping a sliding window) and changes its eviction behavior.

Root cause is in `libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py`: `deque` is encoded in the same branch as `set` and `frozenset`, carrying only `tuple(obj)` (the elements). On decode the constructor is called with that single argument, `deque(elements)`, so `maxlen` is never passed and comes back as `None`. Single-argument construction is correct for `set` and `frozenset`, but `deque` also needs its `maxlen`.

I have a small, backward-compatible fix (encode `deque` so its `maxlen` round-trips, leaving `set`/`frozenset` unchanged; checkpoints written before the fix still load) plus a unit test. Happy to open a PR if that is welcome.

### System Info

[...]

PR #8380, opened 2026-07-19
fix(checkpoint): preserve Counter and OrderedDict type through msgpack round-trip

### LINKED ISSUE #8184: Checkpoint serialization downcasts dict subclasses (defaultdict/Counter/OrderedDict) to plain dict, losing default_factory
### Checked other resources

- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph.

### Related Issues / PRs

Same class of silent round-trip state loss as #8157 (`deque` drops `maxlen`), but for `dict` subclasses.

```python
from collections import defaultdict

s = JsonPlusSerializer()
original = defaultdict(list)
original["a"].append(1)

restored = s.loads_typed(s.dumps_typed(original))

print(type(restored).__name__)   # 'dict'        <- was defaultdict
restored["new"].append(2)        # KeyError: 'new'  <- no longer auto-creates
```

`Counter` and `OrderedDict` are affected the same way (both come back as plain `dict`): a restored `Counter` loses `.most_common()` and its count semantics.

```
KeyError: 'new'
```

No error on the round-trip itself — the failure is silent. The restored object is a plain `dict`, so the missing-key behaviour only surfaces later when the code relies on `default_factory` (or `Counter`/`OrderedDict` behaviour).

### Description

`JsonPlusSerializer` special-cases `set`, `frozenset`, and `deque` in `_msgpack_default` (`libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py`) so they survive a round-trip, but `dict` subclasses (`defaultdict`, `Counter`, `OrderedDict`) are serialized through `ormsgpack`'s native map path and never reach `_msgpack_default`. On decode they come back as a plain `dict`, dropping both the subclass type and its state (`default_factory` for `defaultdict`).

A `defaultdict` held in graph state therefore becomes a plain `dict` after the first checkpoint round-trip, so code that relies on auto-vivification (`state["x"]["k"].append(...)`) starts raising `KeyError` after a resume/replay — the same defeats-its-purpose silent failure as the bounded-`deque` case in #8157.

Note the fix is not a drop-in addition to `_msgpack_default`: because `ormsgpack` handles dict subclasses natively, intercepting them requires `OPT_PASSTHROUGH_SUBCLASS`, which also reroutes other native-base subclasses (e.g. numpy scalars) to the encoder. I'm happy to work on a fix and would appreciate a maintainer's steer on the preferred approach (and on whether `Counter`/`OrderedDict` should be preserved too) before opening a PR.

### System Info

`langgraph-checkpoint` on `main`; default `JsonPlusSerializer` (msgpack path). The encode path has special-cased `set`/`frozenset`/`deque` but not `dict` subclasses since that handling was added, so released versions are affected too. Python 3.12.

Varshith-Kali:
Fixes #8184

`JsonPlusSerializer` was silently downcasting dict subclasses (`Counter`, `OrderedDict`) to plain `dict` on checkpoint round-trip. A `Counter` held in graph state would lose `.most_common()` / count semantics, and an `OrderedDict` would lose `move_to_end()` / explicit-order semantics. The failure was silent at write time and only surfaced later when resumed code relied on the subclass behaviour — classic "works until restore" failure mode (same shape as the bounded-`deque` issue #8157).

Scoped to `Counter` + `OrderedDict`; `defaultdict` deferred to a follow-up — see "Out of scope" below.

## Root cause

`ormsgpack` encodes `dict` subclasses through its native map path, which discards the subclass type. The `_msgpack_default` fallback that already handles `set`/`frozenset`/`deque` was never reached for `dict` subclasses because `ormsgpack` short-circuited them before calling `default`.

## What changed

| File | Change |
|------|--------|
| `libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py` | 1) Add `ormsgpack.OPT_PASSTHROUGH_SUBCLASS` to the encoder option bitmask so `dict` subclasses (and other native-base subclasses) are routed through `_msgpack_default` instead of being encoded natively. 2) Add explicit `Counter` and `OrderedDict` branches in `_msgpack_default` that encode via `EXT_CONSTRUCTOR_SINGLE_ARG` — `Counter(dict(items))` and `OrderedDict(list_of_pairs)` both reconstruct correctly from a single positional arg. 3) Add a fallback branch that converts any *other* native-base subclass (`str`/`int`/`float`/`bool`/`list`/`dict`/`bytes`/`bytearray` subclasses, including numpy scalars like `np.int64`) to its base type, restoring the pre-passthrough behaviour so the `OPT_PASSTHROUGH_SUBCLASS` change doesn't regress them. |
| `libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py` | Add `("collections", "Counter")` and `("collections", "OrderedDict")` to `SAFE_MSGPACK_TYPES` so they pass the `_check_allowed` gate under strict mode (`LANGGRAPH_STRICT_MSGPACK=true`). |
| `libs/checkpoint/tests/test_jsonplus.py` | 7 new round-trip tests (see below). |

**No decoder changes needed** — the existing `EXT_CONSTRUCTOR_SINGLE_ARG` decoder branch already does `getattr(import_module(module), name)(arg)`, so `Counter(dict)` and `OrderedDict(list_of_pairs)` both reconstruct correctly.

## Why this approach

Building on @ly-wang19's analysis in the issue thread:

- **`OPT_PASSTHROUGH_SUBCLASS` + explicit branches** is the cleanest interception point. The alternative (intercepting only via `isinstance` checks without the option) doesn't work because ormsgpack never calls `default` for native-type subclasses unless `OPT_PASSTHROUGH_SUBCLASS` is set.
- **`EXT_CONSTRUCTOR_SINGLE_ARG` for both** (not `POS_ARGS` for `OrderedDict`) because `OrderedDict.__init__` takes a single iterable of pairs, not multiple positional args. I initially used `POS_ARGS` and the decoder returned `None` because `OrderedDict(*[(k,v), ...])` raises `TypeError: expected at most 1 arguments, got N`. Fixed by switching to `SINGLE_ARG`.
- **Fallback branch for unhandled subclasses** is required to avoid regressing numpy scalars (`np.int64` etc.) and any other native-base subclass that ormsgpack previously encoded directly. The fallback converts the instance to its base type via `base_type(obj)`, mirroring what ormsgpack does internally without `OPT_PASSTHROUGH_SUBCLASS`. (Note: `np.int64` was *already* broken on `main` — it raised `TypeError: Object of type int64 is not serializable` — so this PR preserves that pre-existing behaviour rather than making it worse. A proper numpy-scalar fix is a separate concern.)
- **`defaultdict` deferred**: its `default_factory` can be a lambda or custom callable that is not safely encodable without a pickle fallback. That needs a separate design call (pickle fallback vs. warn-and-downcast) and is better handled in a follow-up PR. The issue thread explicitly flagged this as the awkward case.

## Backward compatibility

Pre-fix checkpoints deserialize unchanged (as plain `dict`), so the change is purely additive on the encode side — no migration debt, no risk of breaking existing sessions.

## Tests and Documentation

7 new round-trip tests in `libs/checkpoint/tests/test_jsonplus.py`:

- `test_serde_jsonplus_counter_roundtrip` — type preservation + `.most_common()` survival
- `test_serde_jsonplus_ordereddict_roundtrip` — type preservation + order preservation + `.move_to_end()` survival
- `test_serde_jsonplus_counter_in_nested_state` — nested in a larger state dict
- `test_serde_jsonplus_ordereddict_in_nested_state` — nested in a larger state dict
- `test_serde_jsonplus_empty_counter_roundtrip` — empty-instance edge case
- `test_serde_jsonplus_empty_ordereddict_roundtrip` — empty-instance edge case
- `test_serde_jsonplus_plain_dict_still_round_trips_as_dict` — backward-compat: plain dict still round-trips as plain dict (not upcast)

### Test commands run and results

```bash
# 1. New tests only
$ python -m pytest libs/checkpoint/tests/test_jsonplus.py -v -k "counter or ordereddict or plain_dict_still"
======================= 7 passed, 99 deselected in 0.41s =======================

# 2. Full test_jsonplus.py suite (no regressions)
$ python -m pytest libs/checkpoint/tests/test_jsonplus.py
============================= 106 passed in 0.49s =============================

# 3. Broader checkpoint test suite (test_jsonplus + test_memory + test_store)
$ python -m pytest libs/checkpoint/tests/test_jsonplus.py libs/checkpoint/tests/test_memory.py libs/checkpoint/tests/test_store.py
============================= 146 passed in 0.77s =============================

# 4. Lint + format
$ ruff check libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py libs/checkpoint/tests/test_jsonplus.py
All checks passed!
$ black --check libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py libs/checkpoint/tests/test_jsonplus.py
All done. ✨ 🍰 ✨
```

## Out of scope (follow-up)

`defaultdict` is intentionally NOT covered in this PR. Its `default_factory` can be:
- a builtin (`list`, `int`, `set`) — encodable as `(module, name)` and reconstructed via import
- `None` — trivially encodable
- a lambda / custom callable — not safely encodable without pickle fallback

The first two cover the overwhelmingly common case. The third needs a design call (pickle fallback vs. warn-and-downcast) and is better handled in a follow-up PR. Happy to open that follow-up if maintainers confirm the preferred approach.

## AI assistance disclosure

Drafted with AI assistance (Claude), reviewed and tested by me before submission. I have read every line of the diff, run the full test suite locally (106 + 146 tests pass), and verified the fix end-to-end against the reproduction case in the issue body. The scoping decision (Counter + OrderedDict only, defaultdict deferred) was mine, based on @ly-wang19's analysis in the issue thread and the encodability tradeoff for `default_factory`.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py
@@ -34,6 +34,8 @@ SAFE_MSGPACK_TYPES: frozenset[tuple[str, ...]] = frozenset(
         ("builtins", "set"),
         ("builtins", "frozenset"),
         ("collections", "deque"),
+        ("collections", "Counter"),
+        ("collections", "OrderedDict"),
         # ip addresses
         ("ipaddress", "IPv4Address"),
         ("ipaddress", "IPv4Interface"),
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -10,7 +10,7 @@ import pathlib
 import pickle
 import re
 import sys
-from collections import deque
+from collections import Counter, OrderedDict, deque
 from collections.abc import Callable, Iterable, Sequence
 from datetime import date, datetime, time, timedelta, timezone
 from enum import Enum
@@ -378,6 +378,34 @@ def _msgpack_default(obj: Any) -> str | ormsgpack.Ext:
                 (obj.__class__.__module__, obj.__class__.__name__, str(obj)),
             ),
         )
+    elif isinstance(obj, Counter):
+        # Counter is a dict subclass; without OPT_PASSTHROUGH_SUBCLASS ormsgpack
+        # would encode it as a plain map and lose the type. Reconstruct via
+        # Counter(dict(items)) — Counter's __init__ accepts a dict and preserves
+        # count semantics (.most_common(), arithmetic, etc.).
+        return ormsgpack.Ext(
+            EXT_CONSTRUCTOR_SINGLE_ARG,
+            _msgpack_enc(
+                (obj.__class__.__module__, obj.__class__.__name__, dict(obj)),
+            ),
+        )
+    elif isinstance(obj, OrderedDict):
+        # OrderedDict is a dict subclass; encode as list of pairs so
+        # OrderedDict([(k, v), ...]) round-trips and order is explicit
+        # (matters for code that relies on move_to_end / iteration order
+        # semantics that differ from plain dict in some edge cases).
+        # Use EXT_CONSTRUCTOR_SINGLE_ARG (not POS_ARGS) because OrderedDict's
+        # __init__ takes a single iterable of pairs, not multiple positional args.
+        return ormsgpack.Ext(
+            EXT_CONSTRUCTOR_SINGLE_ARG,
+            _msgpack_enc(
+                (
+                    obj.__class__.__module__,
+                    obj.__class__.__name__,
+                    list(obj.items()),
+                ),
+            ),
+        )
     elif isinstance(obj, (set, frozenset, deque)):
         return ormsgpack.Ext(
             EXT_CONSTRUCTOR_SINGLE_ARG,
@@ -531,6 +559,20 @@ def _msgpack_default(obj: Any) -> str | ormsgpack.Ext:
     elif isinstance(obj, BaseException):
         return repr(obj)
     else:
+        # OPT_PASSTHROUGH_SUBCLASS routes all native-base subclasses (str/int/
+        # float/list/dict/bytes subclasses, including numpy scalars like np.int64)
+        # through this default function rather than letting ormsgpack encode them
+        # natively. The branches above handle the subclasses we explicitly care
+        # about (Counter, OrderedDict, set/frozenset/deque subclasses, Enum, etc.);
+        # for any *other* native-base subclass, fall back to encoding the instance
+        # as its base type so we don't regress on types that ormsgpack previously
+        # encoded directly. This mirrors what ormsgpack does internally without
+        # OPT_PASSTHROUGH_SUBCLASS, restoring the pre-passthrough behavior for
+        # unregistered subclasses while still letting our explicit branches above
+        # intercept the ones we want to round-trip with type preservation.
+        for base_type in (str, int, float, bool, list, dict, bytes, bytearray):
+            if isinstance(obj, base_type):
+                return base_type(obj)
         raise TypeError(f"Object of type {obj.__class__.__name__} is not serializable")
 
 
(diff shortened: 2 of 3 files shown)


Varshith-Kali:
Reopen requested — I've updated the PR body to put `Fixes #8184` at the top (was previously `Closes #8184` in the middle, which the `check-issue-link` workflow didn't pick up). The remaining gate is the assignee check: I've commented on #8184 requesting assignment. Maintainers — please reopen and assign, or let me know if a different process is preferred.

PR #8430, opened 2026-07-24
fix(sdk-py): unblock subscribers on stream close

### LINKED ISSUE #8429: AsyncThreadStream.close() does not unblock active subscribe() iterators
### Checked other resources

### Related Issues / PRs

I did not find an existing report after searching open and closed issues and
pull requests for combinations of `AsyncThreadStream`, `subscribe`, `close`,
`fanout`, `CancelledError`, and subscription shutdown.

```python
import asyncio

import httpx

from langgraph_sdk._async.http import HttpClient
from langgraph_sdk._async.stream import AsyncThreadStream


class EndlessSSE(httpx.AsyncByteStream):
    async def __aiter__(self):
        while True:
            yield b": keepalive\n\n"
            await asyncio.sleep(0.05)


async def serve_sse(_: httpx.Request) -> httpx.Response:
    return httpx.Response(
        200,
        headers={"content-type": "text/event-stream"},
        stream=EndlessSSE(),
    )


async def consume(stream: AsyncThreadStream) -> None:
    async for _ in stream.subscribe(["messages"]):
        pass


async def main() -> None:
    transport = httpx.MockTransport(serve_sse)
    async with httpx.AsyncClient(
        transport=transport,
        base_url="http://test",
    ) as raw:
        stream = AsyncThreadStream(
            http=HttpClient(raw),
            thread_id="thread-1",
            assistant_id="agent",
            explicit_thread_id=True,
        )
        async with stream:
            consumer = asyncio.create_task(consume(stream))
            # Wait only until the consumer has registered its subscription.
            while not stream._subscriptions:
                await asyncio.sleep(0)
            await asyncio.sleep(0.1)

        # The context has closed, so the active iterator should now terminate.
        # Actual on langgraph-sdk 0.4.2: TimeoutError.
        await asyncio.wait_for(consumer, timeout=0.25)


asyncio.run(main())
```

```shell
TimeoutError

The consumer remains suspended here until `asyncio.wait_for()` cancels it:

File "langgraph_sdk/_async/stream.py", in _subscription_iter
    item = await sub.queue.get()
```

### Description

`AsyncThreadStream` supports concurrent subscriptions and typed projections.
If its context closes while one of those consumer tasks is waiting on an empty
subscription queue, `close()` returns but that iterator does not terminate.
Code that subsequently waits for the consumer can therefore wait indefinitely
unless it explicitly cancels the task.

One realistic trigger is concurrent projection consumption: if one sibling
consumer raises and the stream context exits while another sibling is still
waiting, the remaining iterator is left blocked. Direct cancellation of the
consumer task itself is not affected and remains a valid workaround.

`AsyncThreadStream.close()` cancels and awaits `_fanout_task`. Cancellation
raises `CancelledError` inside `_fanout()`, so execution never reaches the
subscription-sentinel loop placed after the `while` loop. The parallel
`StreamController` has the same ordering.

This appears to be an unintended shutdown-ordering gap:

- The SDK documentation demonstrates concurrent projection consumers with
  `asyncio.gather()`.
- `_fanout()` says it terminates consumers cleanly on shutdown or stream end,
  but cancellation skips that code on the shutdown path.
- `SyncStreamController.close()` explicitly sends `None` to every active
  subscription.
- `test_close_unblocks_active_subscription_before_lifecycle_join` already
  requires the corresponding synchronous behavior.

The shutdown path needs to make every active subscription iterator terminal
before or while cancelling the fanout task. I verified locally that delivering
the shutdown sentinel before fanout cancellation fixes the public
`client.threads.stream()` path; the SDK suite still passes.

### System Info

OS: Darwin
Python: 3.12.10
langgraph_sdk: 0.4.2
langchain_core: 1.5.0
langchain_protocol: 0.0.18
httpx: 0.28.1
orjson: 3.11.9
websockets: 15.0.1

donggyun112:
Fixes #8429

## Summary

- mark active subscriptions terminal before fanout cancellation
- apply one termination path to explicit close, run pause, and natural EOF
- preserve buffered events when a bounded subscription queue is saturated
- cover both `AsyncThreadStream` and `StreamController`

## Why

`close()` cancels `_fanout_task`, so `CancelledError` bypasses the subscription
sentinel delivery after the fanout loop. Consumers waiting on an empty
subscription queue therefore remain blocked after the stream context exits.

The existing terminal sentinel also cannot be enqueued when a subscription
queue is full. This affects explicit close, natural fanout completion, and run
pause: `QueueFull` either escapes or the sentinel is silently lost.

Each async subscription now records terminal state separately from its bounded
queue. A sentinel still wakes an iterator already blocked on an empty queue.
When the queue is full, the iterator drains buffered events and then observes
the terminal state, so shutdown neither blocks nor drops an event.

## Why safe

- no public API changes
- no change to ordinary event overflow or transport backpressure policy
- terminal subscriptions stop accepting new fanout events
- empty, partially buffered, and saturated queues retain FIFO behavior

## Tests

- `make format`
- `make lint`
- `make test` (`499 passed, 55 deselected`)

### DIFF
--- a/libs/sdk-py/langgraph_sdk/_async/stream.py
+++ b/libs/sdk-py/langgraph_sdk/_async/stream.py
@@ -1290,6 +1290,9 @@ class AsyncThreadStream:
         if self._closed:
             return
         self._closed = True
+        # Cancelling fanout skips its post-loop subscription sentinels.
+        for sub in self._subscriptions.values():
+            sub.queue.put_nowait(None)
         for handle in self._open_handles:
             await handle.close()
         # Cancel _run_done so thread.output doesn't wait forever on close.
--- a/libs/sdk-py/langgraph_sdk/stream/controller.py
+++ b/libs/sdk-py/langgraph_sdk/stream/controller.py
@@ -166,6 +166,9 @@ class StreamController:
         if self._closed:
             return
         self._closed = True
+        # Cancelling fanout skips its post-loop subscription sentinels.
+        for sub in self._subscriptions.values():
+            sub.queue.put_nowait(None)
         if self._fanout_task is not None:
             self._fanout_task.cancel()
             with contextlib.suppress(Exception, asyncio.CancelledError):
--- a/libs/sdk-py/tests/streaming/test_shared_stream.py
+++ b/libs/sdk-py/tests/streaming/test_shared_stream.py
@@ -134,6 +134,43 @@ async def test_two_concurrent_subscribes_share_one_stream():
     assert len(fake.stream_request_bodies) == 2
 
 
+async def test_thread_stream_close_unblocks_active_subscription():
+    """Exiting the public stream context terminates active subscribers."""
+
+    class EndlessSSE(httpx.AsyncByteStream):
+        async def __aiter__(self):
+            while True:
+                yield b": keepalive\n\n"
+                await asyncio.sleep(0.01)
+
+    async def serve_sse(_: httpx.Request) -> httpx.Response:
+        return httpx.Response(
+            200,
+            headers={"content-type": "text/event-stream"},
+            stream=EndlessSSE(),
+        )
+
+    async with httpx.AsyncClient(
+        transport=httpx.MockTransport(serve_sse),
+        base_url="http://test",
+    ) as raw:
+        threads = ThreadsClient(HttpClient(raw))
+        async with threads.stream(
+            thread_id="t-1",
+            assistant_id="agent",
+        ) as thread:
+
+            async def drain() -> None:
+                async for _ in thread.subscribe(["messages"]):
+                    pass
+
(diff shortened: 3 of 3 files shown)


donggyun112:
The implementation has been updated on the fork branch at donggyun112/langgraph@a043f28f. It now covers explicit close, natural EOF, and run pause without dropping buffered events. `make format`, `make lint`, and `make test` (`499 passed, 55 deselected`) all pass.

GitHub currently keeps this auto-closed PR pinned to its original head commit. The latest branch head will be attached when the maintainer assignment workflow reopens the PR.

donggyun112:
Superseded by #8436, which is pinned to the latest implementation at `a043f28f`. Keeping this PR closed to avoid duplicate review.

Issue #8300, opened 2026-07-08
PostgresStore.search() matches namespace_prefix with an unescaped SQL LIKE, returning rows from foreign namespaces


alex-feel:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Related Issues / PRs

None found (searched issues and PRs for "LIKE", "escape", "namespace prefix", and "wildcard").

### Reproduction Steps / Example Code (Python)

```python
# pip install langgraph-checkpoint-postgres "psycopg[binary]" "testcontainers[postgres]"
# (Docker is required for the throwaway PostgreSQL container.)
from langgraph.store.memory import InMemoryStore
from langgraph.store.postgres import PostgresStore
from testcontainers.postgres import PostgresContainer

NS_A = ("users", "john-doe", "memories")  # user A
NS_B = ("users", "john_doe", "memories")  # user B ("_" is a LIKE single-character wildcard)

with PostgresContainer("postgres:16") as pg:
    url = pg.get_connection_url().replace("postgresql+psycopg2://", "postgresql://")
    with PostgresStore.from_conn_string(url) as store:
        store.setup()
        store.put(NS_A, "note", {"text": "user A private data"})
        store.put(NS_B, "note", {"text": "user B private data"})

        # 1) Wildcard over-match: a search scoped to user B returns user A's row.
        print("PostgresStore, search(NS_B):")
        for item in store.search(NS_B):
            print("   ", item.namespace, item.value)

        # 2) Boundary over-match (no special characters involved): the prefix
        #    ("users", "a") must not match the namespace ("users", "ab", ...).
        store.put(("users", "ab", "memories"), "note", {"text": "user ab data"})
        print("PostgresStore, search(('users', 'a')):")
        for item in store.search(("users", "a")):
            print("   ", item.namespace, item.value)

# InMemoryStore contrast: exact tuple-prefix semantics, both cases behave correctly.
mem = InMemoryStore()
mem.put(NS_A, "note", {"text": "user A private data"})
[...]

PR #8478, opened 2026-07-30
fix(checkpoint-postgres,checkpoint-sqlite): scope namespace matching to segment boundaries

eliornl:
## Summary

Namespace scoping in the Postgres and SQLite stores matched the dot-joined prefix with `LIKE '<path>%'`, which does not respect the `.` separator — a search scoped to `("foo",)` also returned rows under `("foobar",)`. Scoping now matches the namespace exactly or requires the separator before any remainder, and pattern metacharacters in labels are escaped.

`list_namespaces` moves to segment-aware matching for prefix and suffix conditions, since neither `LIKE` nor `GLOB` can express "any character except the separator".

Per-package reasoning is in the commit message.

## Compatibility

`*` in a `list_namespaces` match path now spans exactly one segment, restoring the documented behavior (`NamespacePath` documents `("cache", "*", "v1")` as "any cache category with v1 version") and matching `InMemoryStore`. To match at any depth, combine both conditions, which are ANDed: `list_namespaces(prefix=["uid"], suffix=["alice"])`.

## Test plan

- [x] `make format` / `make lint` / `make test` from `libs/checkpoint-postgres` (224 passed) and `libs/checkpoint-sqlite` (112 passed, 3 skipped)

### DIFF
--- a/libs/checkpoint-sqlite/langgraph/store/sqlite/base.py
+++ b/libs/checkpoint-sqlite/langgraph/store/sqlite/base.py
@@ -149,14 +149,14 @@ def _namespace_match_pattern(path: tuple[str, ...], match_type: str) -> str:
     on a separator; suffix matches anchor at the end and begin on one.
 
     Examples:
-        prefix ("uid", "*", "alice") -> ^uid\\.[^.]+\\.alice(\\.|$)
-        suffix ("alice",)            -> (^|\\.)alice$
+        prefix ("uid", "*", "alice") -> ^uid\\.[^.]+\\.alice(\\.|\\Z)
+        suffix ("alice",)            -> (^|\\.)alice\\Z
     """
     segments = ("[^.]+" if part == "*" else re.escape(part) for part in path)
     body = r"\.".join(segments)
     if match_type == "suffix":
-        return rf"(^|\.){body}$"
-    return rf"^{body}(\.|$)"
+        return rf"(^|\.){body}\Z"
+    return rf"^{body}(\.|\Z)"
 
 
 def _namespace_to_text(namespace: tuple[str, ...]) -> str:
--- a/libs/checkpoint-sqlite/langgraph/store/sqlite/base.py
+++ b/libs/checkpoint-sqlite/langgraph/store/sqlite/base.py
@@ -623,10 +623,22 @@ class BaseSqliteStore:
                             # no-op rather than emitting a pattern that matches no
                             # namespace at all.
                             continue
-                        where_clauses.append(f"{NS_MATCH_FUNCTION}(prefix, ?) = 1")
-                        params.append(
-                            _namespace_match_pattern(cond.path, cond.match_type)
-                        )
+                        if cond.match_type == "prefix" and "*" not in cond.path:
+                            # Equivalent to the anchored pattern, but SQLite can
+                            # satisfy `=` and a trailing-wildcard GLOB from
+                            # store_prefix_idx. The user function is opaque to the
+                            # planner, so it would scan every row and call back
+                            # into Python for each one.
+                            condition, args = _namespace_prefix_condition(
+                                tuple(cond.path)
+                            )
+                            where_clauses.append(condition)
+                            params.extend(args)
+                        else:
+                            where_clauses.append(f"{NS_MATCH_FUNCTION}(prefix, ?) = 1")
+                            params.append(
+                                _namespace_match_pattern(cond.path, cond.match_type)
+                            )
                     else:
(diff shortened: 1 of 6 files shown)


libs/checkpoint-sqlite/langgraph/store/sqlite/base.py
```diff
@@ -93,12 +93,74 @@ class SqliteIndexConfig(IndexConfig):
...
+
+    Examples:
+        prefix ("uid", "*", "alice") -> ^uid\\.[^.]+\\.alice(\\.|$)
+        suffix ("alice",)            -> (^|\\.)alice$
+    """
+    segments = ("[^.]+" if part == "*" else re.escape(part) for part in path)
+    body = r"\.".join(segments)
+    if match_type == "suffix":
+        return rf"(^|\.){body}$"
+    return rf"^{body}(\.|$)"
```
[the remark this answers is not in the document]
eliornl: Confirmed and fixed in `ca279d17`. Good catch — this reproduced exactly as described:

```
suffix alice   -> [('users', 'alice'), ('users', 'alice\n')]
prefix u,alice -> [('users', 'alice'), ('users', 'alice\n')]
```

Checked both engines, since the two packages evaluate these patterns differently:

| | `users.alice\n` |
|---|---|
| Python `$` | matches — affected |
| Postgres `$` | no match — unaffected |

So this was SQLite-only, as its matching runs through `re.search`. Postgres treats `$` as end-of-string already.

Switched to `\Z` in both packages anyway — verified equivalent on the Postgres side, and it avoids leaving a subtle difference between two otherwise identical helpers.

Added a trailing-newline test to each package and updated the pattern assertions that pinned `$`.

nick-hollon-lc (approved):

libs/checkpoint-sqlite/langgraph/store/sqlite/base.py
```diff
@@ -549,15 +617,15 @@ def _get_batch_list_namespaces_queries(
...
-                        )
-                    elif cond.match_type == "suffix":
-                        where_clauses.append("prefix LIKE ?")
+                    if cond.match_type in ("prefix", "suffix"):
+                        if not cond.path:
+                            # An empty path constrains nothing; skipping keeps it a
+                            # no-op rather than emitting a pattern that matches no
+                            # namespace at all.
+                            continue
+                        where_clauses.append(f"{NS_MATCH_FUNCTION}(prefix, ?) = 1")
```
Sanjays2402: this drops the index for the common case: `prefix LIKE ?` was satisfiable from `store_prefix_idx`, but a user function forces a full scan of `store` plus a Python call per row. for `match_type == "prefix"` with no `*` in the path the pattern is exactly what `_namespace_prefix_condition` already builds, so you could reuse `(prefix = ? OR prefix GLOB ?)` there (GLOB with a bound param still gets the LIKE optimization on a BINARY-collated indexed column) and only fall through to `_langgraph_namespace_match` for suffix or wildcard paths.

eliornl: Good catch, thanks — confirmed and fixed in `74e60c95`.

The user function is opaque to the planner, so it scanned the whole index with a Python call per row. On 50k rows, prefix-scoped `list_namespaces`:

```
before  10.69 ms   SCAN store USING COVERING INDEX store_prefix_idx
after    0.01 ms   MULTI-INDEX OR
```

Took your suggestion as-is: a prefix path with no wildcard now reuses `_namespace_prefix_condition`, and only suffix or wildcard paths fall through to the function.

Verified equivalence two ways before landing it, since this sits on top of a correctness fix and a silent divergence would be worse than the slow query:

- 60 differential comparisons of `SqliteStore` against `InMemoryStore`, over labels containing GLOB, regex and LIKE metacharacters, case variants, a trailing newline, and non-ASCII text
- a direct A/B of the two SQL forms over 37 prefix paths

No divergence in either. Added tests pinning both the condition chosen per match type and the metacharacter behavior — including that a label merely *containing* `*` is not treated as the wildcard.

alex-feel:
Heads-up: this PR also fixed #8300 (`PostgresStore.search()` matching `namespace_prefix` with an unescaped SQL `LIKE`, filed before this PR but never linked to it). I re-ran that issue's reproduction against `langgraph-checkpoint-postgres` 3.1.1: both over-match modes — the `_`/`%` wildcard match and the mid-component prefix match — are gone, and `PostgresStore` now agrees with `InMemoryStore`. Closed the issue as resolved by this PR. Thanks!

PR #8482, opened 2026-07-30
feat(cli): add --image-uri flag for self-hosted deployments

l2and:
Adds `--image-uri <uri>` to `langgraph deploy` so self-hosted LangSmith customers can build, push, and deploy in one command without needing to script the three steps manually.

When `--image-uri` is provided the CLI:
- Builds the image tagged to the provided URI (same Docker build path as the local build flow)
- Pushes using whatever Docker credentials are already in the environment (e.g. WIF, `aws ecr get-login-password`) — no auth handling in the CLI
- PATCHes the deployment with `source_revision_config.image_uri` (no `revision_source`, which the self-hosted control plane rejects for `external_docker` deployments)

Also fixes two self-hosted-specific issues uncovered during testing:
- `LANGSMITH_ENDPOINT` is now used as a fallback when `LANGGRAPH_HOST_URL` isn't set — the CLI strips the `/api/v1` path and appends `/api-host` to reach the control plane
- The httpx client now builds full URLs via string concatenation rather than relying on httpx base_url merging, which silently dropped the `/api-host` path prefix when paths started with `/`
- The "Check status at" URL after a deploy now correctly points to the self-hosted LangSmith UI instead of `smith.langchain.com`

**How did you verify your code works?**
Tested end-to-end against a self-hosted LangSmith instance using ECR as the registry. `langgraph deploy --image-uri <ecr-uri>` successfully built, pushed, and triggered a deployment revision. Confirmed the existing SaaS flow (`langgraph deploy` without `--image-uri`) is unaffected — the new flag is opt-in and the `LANGSMITH_ENDPOINT` fallback only activates when `LANGGRAPH_HOST_URL` is unset and `LANGSMITH_ENDPOINT` is present.

### DIFF
--- a/libs/cli/langgraph_cli/host_backend.py
+++ b/libs/cli/langgraph_cli/host_backend.py
@@ -36,11 +36,14 @@ class HostBackendClient:
             headers["X-Tenant-ID"] = tenant_id
         self._base_url = base_url.rstrip("/")
         self._client = httpx.Client(
-            base_url=self._base_url,
             headers=headers,
             transport=transport,
             timeout=30,
         )
+
+    @property
+    def base_url(self) -> str:
+        return self._base_url
 
     def _request(
         self,
@@ -50,7 +53,8 @@ class HostBackendClient:
         params: dict[str, Any] | None = None,
     ) -> Any:
         try:
-            resp = self._client.request(method, path, json=payload, params=params)
+            full_url = self._base_url + path
+            resp = self._client.request(method, full_url, json=payload, params=params)
             resp.raise_for_status()
         except httpx.HTTPStatusError as err:
             detail = err.response.text or str(err.response.status_code)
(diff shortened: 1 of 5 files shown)


libs/cli/langgraph_cli/host_backend.py
```diff
@@ -2,11 +2,86 @@
...
 import click
 import httpx
 
+CLOUD_CONTROL_PLANE_URL = "https://api.host.langchain.com"
+CLOUD_DASHBOARD_URL = "https://smith.langchain.com"
+CLOUD_DOMAIN = "langchain.com"
+CLOUD_API_HOST = "api.smith.langchain.com"
+CLOUD_CONTROL_PLANE_HOST = "api.host.langchain.com"
+CLOUD_DASHBOARD_HOST = "smith.langchain.com"
+CONTROL_PLANE_PATH = "/api-host"
```
hari-dhanushkodi, line 18: would maybe rename this one; this is the case for self hosted only

hari-dhanushkodi (approved):

PR #8523, opened 2026-08-04
feat(langgraph): expose `trace_policy` on `add_node`

ccurme:
Adds `TracePolicy`, allowing specification of callables to process tracing inputs and outputs:
```python
@dataclass(**_DC_KWARGS)
class TracePolicy:

    process_inputs: Callable[[Any], Any] | None = None
    process_outputs: Callable[[Any], Any] | None = None
```

We thread the `trace_inputs` param of a node's `RunnableSeq` through `PregelNode`, `StateNodeSpec`, and expose in `add_node` via `TracePolicy`. We also implement `trace_outputs` on `RunnableSeq` and thread it through the same way.

We also implement a small helper function `omit_payload` that can be specified in `process_(inputs/outputs)` to scrub tracing payloads.

Motivation is to allow transforms on tracing from nodes. LangChain middleware node-type hooks (`before/after_*`) currently trace state; for longish middleware stacks and long message histories this adds considerable latency (500-1,000ms per middleware in deepagents eval suite).


libs/langgraph/langgraph/_internal/_runnable.py
```diff
@@ -63,6 +65,30 @@
...
+        return value
+    if not any(isinstance(h, BaseTracer) for h in manager.handlers):
+        return value
+    try:
+        return transform(value)
+    except Exception:
+        logger.exception(
+            "TracePolicy processor raised; recording untransformed payload"
+        )
+        return value
```
sydney-runkle, line 85: is there precedent for using logging like this w/ tracing? i'm not familiar

libs/langgraph/langgraph/_internal/_runnable.py
```diff
@@ -63,6 +65,30 @@
 except ImportError:
     _StreamingCallbackHandler = None  # type: ignore
 
+logger = logging.getLogger(__name__)
+
+
+def _trace_payload(
+    manager: Any, value: Any, transform: Callable[[Any], Any] | None
```
sydney-runkle: what are manager and value?

sydney-runkle: i think if we documented this a bit better (types and kwarg only) the following code would read better

libs/langgraph/langgraph/pregel/_read.py
```diff
@@ -222,14 +227,23 @@ def flat_writers(self) -> list[Runnable]:
...
         writers = self.flat_writers
+        trace_inputs = self.trace_policy.process_inputs if self.trace_policy else None
+        trace_outputs = self.trace_policy.process_outputs if self.trace_policy else None
         if self.bound is DEFAULT_BOUND and not writers:
             return None
         elif self.bound is DEFAULT_BOUND and len(writers) == 1:
             return writers[0]
         elif self.bound is DEFAULT_BOUND:
-            return RunnableSeq(*writers)
+            return RunnableSeq(
```
sydney-runkle, line 237: i wonder if it makes more sense for RunnableSeq to carry a trace policy, if we are gonna extend
this is fine though, can adjust (i think this is private)

libs/langgraph/langgraph/types.py
```diff
@@ -527,6 +529,41 @@ class CachePolicy(Generic[KeyFuncT]):
...
+    Scope: this only transforms what the node's *own* run records. Child runs created
+    by a traced `bound` runnable and the root graph run are not affected. Plain
+    function nodes are traced with `trace=False`, so they have no such child runs.
+
+    Not intended to redact secrets. To redact inputs/outputs across all runs
+    (children included), use the LangSmith client's
+    `hide_inputs`/`hide_outputs`/`anonymizer` instead.
+    """
+
+    process_inputs: Callable[[Any], Any] | None = None
```
sydney-runkle, line 548: is this how it's typed in the sdk?

sydney-runkle (approved):
A few nits and qs, LGTM

libs/langgraph/langgraph/_internal/_runnable.py
ccurme: yes, this was done to conform to norms around exception handling in the tracer.

libs/langgraph/langgraph/_internal/_runnable.py
ccurme: updated types here.

libs/langgraph/langgraph/types.py
ccurme: sdk takes `dict`, but here the inputs are the inputs to the runnable. the tracer will coerce to a dict if it needs (e.g., by wrapping `{"inputs": inputs}`)

PR #8526, opened 2026-08-04
fix(checkpoint): collect writes at plain-value seed in delta channel history

### LINKED ISSUE #8384: InMemorySaver silently and permanently drops the first write after migrating a channel to DeltaChannel
### Checked other resources

### Related Issues / PRs

None found. I searched open issues for `DeltaChannel`, `BinaryOperatorAggregate`, `snapshot_frequency`, `get_delta_channel_history`, and `migration` and didn't see this reported.

```python
from typing import Annotated

def add(a, b):
    return a + b


def noop(_state):
    return {}


saver = InMemorySaver()
config = {"configurable": {"thread_id": "t1"}}


# 1. Build up state under the "old" channel type (this is what a real
#    conversation looks like before opting into DeltaChannel).
class OldState(TypedDict):
    items: Annotated[list, BinaryOperatorAggregate(list, add)]


old_graph = (
    StateGraph(OldState)
    .add_node("noop", noop)
    .add_edge(START, "noop")
    .add_edge("noop", END)
    .compile(checkpointer=saver)
)
old_graph.invoke({"items": ["a"]}, config)


# 2. Migrate the SAME thread to DeltaChannel (documented/supported path,
#    see libs/langgraph/tests/test_delta_channel_migration.py).
def reducer(state, writes):
    result = list(state)
    for w in writes:
        result.extend(w)
    return result


class NewState(TypedDict):
    items: Annotated[list, DeltaChannel(reducer)]


new_graph = (
    StateGraph(NewState)
    .add_node("noop", noop)
    .add_edge(START, "noop")
    .add_edge("noop", END)
    .compile(checkpointer=saver)
)
new_graph.invoke({"items": ["b"]}, config)  # first post-migration write

print("Actual:  ", new_graph.get_state(config).values["items"])
print("Expected:", ["a", "b"])

#Output:
#Actual:   ['a']
#Expected: ['a', 'b']

# "b" - the first write made after migrating the thread to DeltaChannel - is silently dropped. It never reappears on any later read or turn, either.
```

```shell
N/A - no exception is raised and nothing is logged. That's the core of the bug: the data loss is completely silent.
```

### Description

I'm migrating an existing thread's channel from `BinaryOperatorAggregate` (the same pattern `add_messages` uses under the hood) to the new `DeltaChannel`, using the documented migration path that's already covered by `libs/langgraph/tests/test_delta_channel_migration.py`, on `InMemorySaver`.

I expected the state to keep accumulating normally across the migration. Instead, the very first write made right after switching a thread's channel to `DeltaChannel` is silently and permanently lost the moment the state is read back (`get_state()`, a fresh `.invoke()` on the same thread, a process restart, etc.) unless that write happens to land on a snapshot boundary. Since the default `snapshot_frequency` is `1000`, the common case is that the first post-migration write is non-snapshotting and therefore exposed to this bug.

In the attached example, `new_graph.invoke({"items": ["b"]})` returns `["a", "b"]` from the live call (this part looks correct), but `new_graph.get_state(config)` right afterward returns only `["a"]`.

**Root cause**

`InMemorySaver.get_delta_channel_history()` (`libs/checkpoint/langgraph/checkpoint/memory/__init__.py`, ~lines 202-215) walks the ancestor chain to find a channel's `seed` value, then returns the writes that must be replayed on top of that seed. The optimized `InMemorySaver` override currently diverges from the reference replay contract by conditionally skipping pending writes for a visited ancestor when that ancestor's stored blob is a *plain* value (rather than a `_DeltaSnapshot`) -- exactly what a checkpoint written under the pre-migration channel type looks like:

```python
blob_value = blob_value_by_ch.get(ch)
if blob_value is not None and not isinstance(blob_value, _DeltaSnapshot):
    continue
```

That special case is incorrect for migrated threads. A plain-value seed checkpoint does not necessarily include that same checkpoint's own pending writes. In the failing case, the stored plain value is `["a"]`, while the pending writes attached to that same checkpoint still contain `["b"]`. Those writes are exactly what produced the seed checkpoint's *child* -- i.e. the first write made after migration -- so skipping them discards required replay data and permanently loses the first post-migration write.

Neither the reference implementation (`BaseCheckpointSaver.get_delta_channel_history` in `libs/checkpoint/langgraph/checkpoint/base/__init__.py`) nor the optimized `SqliteSaver` / `AsyncSqliteSaver` override (`libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/_delta.py`) has this special case -- they collect writes at every visited ancestor until replay terminates. `InMemorySaver` is the outlier, and its behavior diverges from the replay contract used by the other implementations.

**I'd like to work on this**

I've already isolated the root cause and verified a local fix: removing the `if blob_value is not None and not isinstance(blob_value, _DeltaSnapshot): continue` special case so writes are always collected for visited ancestors, matching the reference/base implementation and the SQLite optimized path. I also compared the patched behavior against both the base fallback and `SqliteSaver` across the failing repro and nearby edge cases (multiple post-migration writes, `update_state()` after migration, fresh delta-only threads, `add_messages` -> `DeltaChannel` migration, and fork-from-`update_state`), and the patched `InMemorySaver` matches them. I also re-ran the existing delta-channel suite (`test_delta_channel_migration.py`, `test_delta_channel_exit_mode.py`, `test_delta_channel_update_state.py`, `test_delta_channel_id_stability.py`, `test_delta_channel_supersteps_bound.py` -- 42 tests) with that change and all of them still pass.

My plan, if the approach is approved:
1. Remove the special case in `InMemorySaver.get_delta_channel_history` (`libs/checkpoint/langgraph/checkpoint/memory/__init__.py`) so the optimized path matches the reference replay contract.
2. Tighten the nearby comment/docstring to clarify that a pre-delta plain-value seed may subsume older ancestor writes, but not necessarily the seed checkpoint's own pending writes.
3. Add regression coverage in `libs/langgraph/tests/test_delta_channel_migration.py` for the exact failing scenario: migrate, perform one non-snapshotting post-migration write, then cold-read via `get_state()`.
4. Add a parity-style assertion that the optimized `InMemorySaver` matches the `BaseCheckpointSaver` fallback behavior for this migration scenario.
5. Run `make format`, `make lint`, `make test` in `libs/checkpoint` and `libs/langgraph`.

Happy to open a PR for this once a maintainer can confirm the approach and assign me to the issue.

### System Info

```
System Information
------------------
> OS:  Darwin
> Python Version:  3.13.5

Package Information
-------------------
> langgraph: 1.2.9
> langgraph-checkpoint: 4.1.1
> langchain_core: 1.4.8
```

eliornl:
Fixes langchain-ai/langgraph#8384

Reported by @Aarkin7, with two self-contained runnable repros and a search of the existing issues that turned up nothing. The report also pinned the shape of the bug precisely: the first write after migrating a thread to `DeltaChannel` is dropped, silently and permanently.

`InMemorySaver.get_delta_channel_history` skipped the writes stored at the ancestor it seeded from whenever that ancestor's blob was a plain value rather than a `_DeltaSnapshot`, silently dropping the first write made after migrating a thread to `DeltaChannel`.

### Why the old rule was wrong

A stored blob is the value *entering* its checkpoint; the writes stored under that same checkpoint are what produce its child. That's true for `_DeltaSnapshot` blobs and pre-delta plain values alike, so there was never a reason to treat them differently.

Writes at ancestors *older* than the seed genuinely are subsumed by the seed value — but that's already guaranteed by terminating the walk, since the channel leaves `remaining` once its seed is found. The removed check re-solved that and overreached by one checkpoint.

`BaseCheckpointSaver`, `SqliteSaver` and `PostgresSaver` never had this check. `InMemorySaver` was the only outlier.

### How I verified it

Built a differential harness running the same migration scenarios through `InMemorySaver`, the `BaseCheckpointSaver` reference walk, and `SqliteSaver`. **4 of 11 scenarios agreed before this change; 11 of 11 after.** The loss is wider than one write — on the `add_messages` → `DeltaChannel` path it drops a real user message.

Suites: `libs/checkpoint` 156 passed, `libs/langgraph` 1972 passed, `libs/checkpoint-sqlite` 117 passed, `libs/checkpoint-postgres` passed against PG 16. `make format`, `make lint` clean in each.

### Two things worth a closer look in review

**1. I inverted two existing assertions** in `TestPreDeltaBlobTerminator` (`libs/checkpoint/tests/test_memory.py`). They encoded the old rule. Their fixture is the real migration shape — a plain-value blob carrying pending writes, with a delta-era child — which I confirmed against a dumped checkpoint chain from the issue's repro, so the assertions were wrong rather than the fixture being unrealistic. I added an ancestor *older* than the seed so the terminator still guards what it legitimately should: older writes stay excluded, the seed's own writes replay.

**2. The new conformance test fails against Postgres**, for a reason unrelated to this change. Postgres `aput` leaves an inline `True` marker in `channel_values` only for `_DeltaSnapshot`; plain non-primitive values are popped with no marker, and seed detection is `(checkpoint -> 'channel_values' -> ch) IS NOT NULL`. So Postgres can't locate a plain-value seed at all:

```
seed stored as plain list:      InMemorySaver -> [10, 20]    AsyncPostgresSaver -> no seed key
seed stored as _DeltaSnapshot:  InMemorySaver -> found       AsyncPostgresSaver -> found
```

The pre-existing `test_history_migration_plain_value_as_seed` already fails there too — conformance CI only validates `InMemorySaver`, so nobody was watching. Values still come out correct today (with no seed the walk runs to the root and replays everything), but early termination is lost: 1 write replayed on `InMemorySaver` vs 7 on Postgres for the same 6-turn thread. Filing separately rather than folding a write-path/format decision into this PR.

### Note on scope

This touches three packages: the fix in `libs/checkpoint`, graph-level regression tests in `libs/langgraph` (the bug is only observable through a graph read), and the contract test in `libs/checkpoint-conformance` so third-party savers are covered too.

### DIFF
--- a/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
+++ b/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
@@ -208,6 +208,73 @@ async def test_history_migration_plain_value_as_seed(
     assert values == [2], f"Expected [2], got {values}"
 
 
+async def test_history_seed_ancestor_own_writes_are_replayed(
+    saver: BaseCheckpointSaver,
+) -> None:
+    """Writes stored AT the seed ancestor must be included in `writes`.
+
+    A stored value is the state ENTERING its checkpoint; the writes stored
+    under that same checkpoint are what produced its child and are therefore
+    NOT subsumed by it. Only writes at ancestors OLDER than the seed are
+    subsumed, and the walk terminates before reaching them.
+
+    This holds for plain-value seeds (migration from a pre-delta channel type)
+    exactly as it does for `_DeltaSnapshot` seeds. Skipping the seed
+    ancestor's own writes silently drops the first post-migration write.
+    """
+    from langgraph.checkpoint.base import Checkpoint
+    from langgraph.checkpoint.base.id import uuid6
+
+    from langgraph.checkpoint.conformance.test_utils import generate_metadata
+
+    tid = str(uuid4())
+    configs: list = []
+    parent_cfg = None
+
+    # step 0: older pre-delta value; its write is subsumed by step 1's value.
+    # step 1: the seed — a plain value AND a pending write that produced step 2.
+    # step 2: delta-era, no stored value, one write.
+    # step 3: head — its own write is pending for the next step, excluded.
+    for step in range(4):
+        config = {"configurable": {"thread_id": tid, "checkpoint_ns": ""}}
+        if parent_cfg:
+            config["configurable"]["checkpoint_id"] = parent_cfg["configurable"][
+                "checkpoint_id"
+            ]
+        cv: dict = {}
+        cvs: dict = {}
+        if step == 0:
+            cv["ch"] = [10]
+            cvs["ch"] = step + 1
+        elif step == 1:
+            cv["ch"] = [10, 20]
+            cvs["ch"] = step + 1
+        cp = Checkpoint(
+            v=1,
+            id=str(uuid6(clock_seq=-1)),
+            ts="",
+            channel_values=cv,
+            channel_versions=cvs,
+            versions_seen={},
+            updated_channels=None,
+        )
+        parent_cfg = await saver.aput(config, cp, generate_metadata(step=step), cvs)
+        configs.append(parent_cfg)
+        await saver.aput_writes(parent_cfg, [("ch", step)], str(uuid4()))
+
+    head = configs[-1]
+    result = await saver.aget_delta_channel_history(config=head, channels=["ch"])
+
+    assert "seed" in result["ch"], "Expected seed from plain value at step 1"
+    assert result["ch"]["seed"] == [10, 20], (
+        f"Expected nearest plain value [10, 20], got {result['ch']['seed']}"
+    )
+    values = [w[2] for w in result["ch"]["writes"]]
+    # 1 = the seed ancestor's own write (not subsumed), 2 = delta-era write.
+    # 0 is older than the seed (subsumed); 3 is pending at the head.
+    assert values == [1, 2], f"Expected [1, 2], got {values}"
+
+
 ALL_DELTA_CHANNEL_HISTORY_TESTS = [
     test_history_returns_writes_oldest_first,
     test_history_seed_is_nearest_snapshot,
@@ -216,6 +283,7 @@ ALL_DELTA_CHANNEL_HISTORY_TESTS = [
     test_history_empty_channels_returns_empty,
     test_history_walk_to_root_no_seed,
     test_history_migration_plain_value_as_seed,
+    test_history_seed_ancestor_own_writes_are_replayed,
 ]
 
 
--- a/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
+++ b/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
@@ -6,9 +6,12 @@ import traceback
 from collections.abc import Callable
 from uuid import uuid4
 
-from langgraph.checkpoint.base import BaseCheckpointSaver
+from langgraph.checkpoint.base import BaseCheckpointSaver, Checkpoint
+from langgraph.checkpoint.base.id import uuid6
+from langgraph.checkpoint.serde.types import _DeltaSnapshot
 
 from langgraph.checkpoint.conformance.spec._delta_fixtures import build_delta_chain
+from langgraph.checkpoint.conformance.test_utils import generate_metadata
 
 
 async def test_history_returns_writes_oldest_first(
@@ -48,8 +51,6 @@ async def test_history_seed_is_nearest_snapshot(
     result = await saver.aget_delta_channel_history(config=head, channels=["ch"])
     assert "seed" in result["ch"], "Expected seed from snapshot at step 3"
     seed = result["ch"]["seed"]
-    from langgraph.checkpoint.serde.types import _DeltaSnapshot
-
     actual_value = seed.value if isinstance(seed, _DeltaSnapshot) else seed
     assert actual_value == 3, f"Expected seed value 3 (step 3), got {actual_value}"
     writes = result["ch"]["writes"]
@@ -81,11 +82,6 @@ async def test_history_multi_channel(
     tid = str(uuid4())
     configs: list = []
     parent_cfg = None
-    from langgraph.checkpoint.base import Checkpoint
-    from langgraph.checkpoint.base.id import uuid6
-    from langgraph.checkpoint.serde.types import _DeltaSnapshot
-
-    from langgraph.checkpoint.conformance.test_utils import generate_metadata
 
     for step in range(5):
         config = {"configurable": {"thread_id": tid, "checkpoint_ns": ""}}
@@ -161,11 +157,6 @@ async def test_history_migration_plain_value_as_seed(
     channel_values[ch] (not a _DeltaSnapshot). The walk should treat it as the
     seed and terminate there.
     """
-    from langgraph.checkpoint.base import Checkpoint
-    from langgraph.checkpoint.base.id import uuid6
-
-    from langgraph.checkpoint.conformance.test_utils import generate_metadata
-
     tid = str(uuid4())
     configs: list = []
     parent_cfg = None
@@ -222,19 +213,21 @@ async def test_history_seed_ancestor_own_writes_are_replayed(
     exactly as it does for `_DeltaSnapshot` seeds. Skipping the seed
     ancestor's own writes silently drops the first post-migration write.
     """
-    from langgraph.checkpoint.base import Checkpoint
-    from langgraph.checkpoint.base.id import uuid6
-
-    from langgraph.checkpoint.conformance.test_utils import generate_metadata
-
     tid = str(uuid4())
     configs: list = []
     parent_cfg = None
 
-    # step 0: older pre-delta value; its write is subsumed by step 1's value.
-    # step 1: the seed — a plain value AND a pending write that produced step 2.
-    # step 2: delta-era, no stored value, one write.
-    # step 3: head — its own write is pending for the next step, excluded.
+    # Each step's write is labelled by the role it plays, so the assertion
+    # below reads directly rather than by step index.
+    writes_by_step = {
+        0: "older-than-seed",  # subsumed by the value stored at step 1
+        1: "at-seed",  # the seed's own write, produced step 2
+        2: "after-seed",  # delta-era write on the path to the head
+        3: "pending-at-head",  # pending for the next step, never replayed
+    }
+    # Steps 0 and 1 store a plain value; 1 is the nearest, so it is the seed.
+    values_by_step = {0: [10], 1: [10, 20]}
+
     for step in range(4):
         config = {"configurable": {"thread_id": tid, "checkpoint_ns": ""}}
         if parent_cfg:
@@ -243,11 +236,8 @@ async def test_history_seed_ancestor_own_writes_are_replayed(
             ]
         cv: dict = {}
         cvs: dict = {}
-        if step == 0:
-            cv["ch"] = [10]
-            cvs["ch"] = step + 1
-        elif step == 1:
-            cv["ch"] = [10, 20]
+        if step in values_by_step:
+            cv["ch"] = values_by_step[step]
             cvs["ch"] = step + 1
         cp = Checkpoint(
             v=1,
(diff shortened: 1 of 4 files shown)


libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
```diff
@@ -208,6 +208,73 @@ async def test_history_migration_plain_value_as_seed(
...
+    under that same checkpoint are what produced its child and are therefore
+    NOT subsumed by it. Only writes at ancestors OLDER than the seed are
+    subsumed, and the walk terminates before reaching them.
+
+    This holds for plain-value seeds (migration from a pre-delta channel type)
+    exactly as it does for `_DeltaSnapshot` seeds. Skipping the seed
+    ancestor's own writes silently drops the first post-migration write.
+    """
+    from langgraph.checkpoint.base import Checkpoint
+    from langgraph.checkpoint.base.id import uuid6
```
sydney-runkle: let's do imports at the top level of the file not within the function

side note, we should have a lint rule that enforces this, but doesn't need to be added in this PR

libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
```diff
@@ -208,6 +208,73 @@ async def test_history_migration_plain_value_as_seed(
...
+    result = await saver.aget_delta_channel_history(config=head, channels=["ch"])
+
+    assert "seed" in result["ch"], "Expected seed from plain value at step 1"
+    assert result["ch"]["seed"] == [10, 20], (
+        f"Expected nearest plain value [10, 20], got {result['ch']['seed']}"
+    )
+    values = [w[2] for w in result["ch"]["writes"]]
+    # 1 = the seed ancestor's own write (not subsumed), 2 = delta-era write.
+    # 0 is older than the seed (subsumed); 3 is pending at the head.
+    assert values == [1, 2], f"Expected [1, 2], got {values}"
```
sydney-runkle: a little confused by this, where do the 1 and 2 come from?

sydney-runkle: a little confused by this, where do the 1 and 2 come from?

libs/checkpoint/langgraph/checkpoint/memory/__init__.py
```diff
@@ -205,11 +204,6 @@ def get_delta_channel_history(
             ):
                 if ch not in remaining:
                     continue
-                blob_value = blob_value_by_ch.get(ch)
```
sydney-runkle, line 208: was there any indication as to why this was broken in in memory but not any of the others?

libs/checkpoint/tests/test_memory.py
```diff
@@ -651,15 +674,19 @@ def test_seed_from_pre_delta_ancestor_blob(self) -> None:
...
-        # must be excluded.
+        # The seed ancestor's own write and the delta-era write from cp2 both
+        # replay through the reducer on top of the seed, oldest first. cp3 is
+        # the target — its own write is pending for the NEXT step and must be
+        # excluded.
         values = [v for _, _, v in result["writes"]]
-        assert values == ["B"]
+        assert values == ["PRE-DELTA-WRITE", "B"]
 
     def test_pre_delta_blob_terminates_walk_before_older_writes(self) -> None:
```
sydney-runkle: should we change this function name?

libs/langgraph/tests/test_delta_channel_migration.py
```diff
@@ -616,3 +616,115 @@ async def test_add_messages_to_delta_migration_preserves_message_history_async()
...
+        "optimized override diverges from the base-saver fallback once "
+        f"post-migration writes exist; fast={fast}, slow={slow}"
+    )
+    # Guard the assertion above against both paths being wrong in the same way.
+    assert fast[0][1] == ["u0", "u1", "d0", "d1", "d2"], (
+        f"unexpected accumulated state: {fast[0][1]}"
+    )
+
+
+def test_add_messages_migration_keeps_first_post_migration_message() -> None:
```
sydney-runkle, line 715: and this is failing on main?
it would only drop the singular write at the boundary right?

sydney-runkle (commented):
great digging here, this is super tricky!
will give a final approval after comments / questions addressed

libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
eliornl: Done, moved to the top. Some other tests here had the same imports inside them, so I removed those too since they're duplicates now.

libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
eliornl: They were the step number being used as the write value, so it read badly. Changed them to names, now it says `assert values == ["at-seed", "after-seed"]`.

libs/checkpoint/langgraph/checkpoint/memory/__init__.py
eliornl: It was there before #7699 with a comment saying the same thing, and the refactor carried it over. The other savers were written against the contract directly, so they never had it.

libs/checkpoint/tests/test_memory.py
eliornl: Agreed, the old name only covered half of what it checks now. Renamed to `test_seed_bounds_walk_without_dropping_its_own_writes`.

libs/langgraph/tests/test_delta_channel_migration.py
eliornl: Yes to both. I checked by putting the old code back and rerunning. It only loses the write on the row the walk stops at - with 4 writes after the migration, only the first one goes missing.

libs/langgraph/tests/test_delta_channel_migration.py
```diff
@@ -616,3 +616,115 @@ async def test_add_messages_to_delta_migration_preserves_message_history_async()
...
+    config = {"configurable": {"thread_id": "add-messages-first-write"}}
+
+    pre_graph = _add_messages_graph(checkpointer)
+    pre_graph.invoke({"messages": [HumanMessage(content="hello", id="h1")]}, config)
+
+    delta_graph = _delta_messages_graph(checkpointer)
+    delta_graph.invoke({"messages": [HumanMessage(content="second", id="h2")]}, config)
+
+    ids = [m.id for m in delta_graph.get_state(config).values["messages"]]
+    assert "h2" in ids, f"first post-migration message dropped: {ids}"
```
sydney-runkle: should we also assert h1 in it?

libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
```diff
@@ -208,6 +199,74 @@ async def test_history_migration_plain_value_as_seed(
...
+    parent_cfg = None
+
+    # Each step's write is labelled by the role it plays, so the assertion
+    # below reads directly rather than by step index.
+    writes_by_step = {
+        0: "older-than-seed",  # subsumed by the value stored at step 1
+        1: "at-seed",  # the seed's own write, produced step 2
+        2: "after-seed",  # delta-era write on the path to the head
+        3: "pending-at-head",  # pending for the next step, never replayed
+    }
```
sydney-runkle, line 227: nice this is much cleaner

sydney-runkle (approved):
Excellent!

One final question -- the reducer format is different for binary operator channels vs delta channels

Do we think the write will still be applied correctly assuming the developer designs an appropriate reducer?

libs/langgraph/tests/test_delta_channel_migration.py
eliornl: Good call — now `assert ids == ["h1", "h2"]`, so the seed message and the dropped write are both pinned, and in order. `2f791d70`.

eliornl:
Yes, and worth being precise about what does and doesn't change across the boundary.

The write **values** are identical either way: a write is whatever the node returned for that channel, `["c"]` or similar, and that shape doesn't depend on the channel type. What differs is only how they get combined. Pre-migration writes are replayed through the **new** reducer, never the old one.

So it comes down to the `DeltaChannel` contract you already documented — batching-invariance:

```
reducer(reducer(state, xs), ys) == reducer(state, xs + ys)
```

Replay folds every on-path write into one call, so a reducer satisfying that gives the same answer as the pairwise binop folds did. Checked it:

```
pre-migration ['a','b'] under binop, then 3 writes under delta

  batching-invariant       live=['a','b','n0','n1','n2']  cold=['a','b','n0','n1','n2']
  NOT batching-invariant   live=['a','b','n1','n2']       cold=['a','b','n2']
```

And the reason I don't think migration adds risk here: that second reducer is already broken without any migration at all.

```
fresh delta-only thread, never migrated
  batching-invariant       -> ['n0','n1','n2']
  NOT batching-invariant   -> ['n2']
```

So a reducer that violates the contract loses writes on a plain delta thread too. Migration doesn't introduce the failure, it just inherits it. An appropriate reducer: the batching-invariant equivalent of the old binary operator -  replays the pre-migration writes correctly, which is what `test_first_post_migration_write_survives_cold_read` and the `add_messages` test cover.

One caveat worth naming: nothing validates batching-invariance at runtime, so a subtly wrong reducer fails silently. That's pre-existing and out of scope here, but it might be worth a note in the `DeltaChannel` docstring pointing at the failure mode.

PR #8535, opened 2026-08-05
fix(checkpoint-postgres): find plain-value seeds when walking delta history

### LINKED ISSUE #8534: PostgresSaver cannot locate a plain-value seed, so migrated DeltaChannel threads walk to the root on every read
`PostgresSaver` cannot locate a plain-value seed, so migrated `DeltaChannel` threads walk to the root on every read.

Values still come out correct, so this is a performance and contract problem rather than data loss. Found while working on #8384 / #8526.

### Cause

`aput` leaves an inline `True` marker in `channel_values` only for `_DeltaSnapshot`. Plain non-primitive values are popped with no marker:

```python
for k, v in checkpoint["channel_values"].items():
    if isinstance(v, _DeltaSnapshot):
        blob_values[k] = copy["channel_values"].pop(k)
        copy["channel_values"][k] = True          # marker kept
    elif v is None or isinstance(v, (str, int, float, bool)):
        pass
    else:
        blob_values[k] = copy["channel_values"].pop(k)   # popped, no marker
```

Stage-1 seed detection tests for that marker:

```sql
(checkpoint -> 'channel_values' -> %s) IS NOT NULL AS hs_0
```

So a pre-delta plain value (what a thread migrated from `BinaryOperatorAggregate` leaves behind) is invisible to the walk.

### Repro

Same logical chain, both savers, public APIs only:

```
seed stored as PLAIN list:       InMemorySaver -> [10, 20]                 AsyncPostgresSaver -> <<NO SEED KEY>>
seed stored as _DeltaSnapshot:   InMemorySaver -> _DeltaSnapshot([10,20])  AsyncPostgresSaver -> _DeltaSnapshot([10,20])
```

### Impact

With no seed, the walk runs to the root and replays every write. For an additive reducer that rebuilds the correct value by coincidence, which is why nothing looks wrong. What is lost is early termination:

```
                 2 turns   6 turns     (writes replayed)
InMemorySaver        1         1
Postgres             3         7
```

That is O(history) work per read on exactly the threads that opted into `DeltaChannel` to avoid it. Correctness risk remains where the writes are not all present — `prune`, TTL, or `copy_thread`, which the `BaseCheckpointSaver` docstrings already warn about. Untested.

### Why it went unnoticed

The existing conformance test `test_history_migration_plain_value_as_seed` already fails against Postgres and has for some time. `libs/checkpoint-conformance/tests/` contains only `test_validate_memory.py`, so conformance CI validates `InMemorySaver` alone. `checkpoint-sqlite` has its own `tests/test_conformance_delta.py`; `checkpoint-postgres` has no equivalent.

Running the conformance suite against Postgres (and Sqlite) in CI would have caught this, and is probably worth doing independently of the fix.

### Fix options

**A — write the marker for plain values too** (~2 lines). Does not help any thread already on disk, since existing checkpoints have no marker. Also changes the stored shape for every non-primitive channel, not just delta ones.

**B — detect the seed by blob existence instead of the marker**, as `InMemorySaver` does (`channel_versions[ch]` → blob lookup). Fixes existing data, but stage 1 is currently a pure JSONB scan on `checkpoints` with paging, deliberately tuned to avoid touching the blobs table until stage 2. Would need a join or a second query per page.

B is the one that actually fixes affected threads; A is cheap but leaves them broken. Happy to implement either — flagging the tradeoff since it is a stored-format vs query-shape decision.

For reference, `langgraph-api`'s Go/Python walks resolve the seed by version rather than by a `channel_values` marker, so they do find plain-value seeds.

eliornl:
Fixes langchain-ai/langgraph#8534

`put` splits stored values in two: primitives stay inline in the checkpoint's `channel_values`, everything else moves to `checkpoint_blobs`, and only `_DeltaSnapshot` leaves an inline marker behind when it moves. Stage-1 seed detection tested for that marker, so a plain value — what a thread migrated from `BinaryOperatorAggregate` leaves behind — was invisible to the walk.

### Effect

Migrated threads found no seed, walked to the root, and replayed every write on every read. Values still came out correct, because replaying an additive reducer from empty rebuilds the same list, which is why nothing looked wrong. What was lost is early termination — the entire point of `DeltaChannel`:

| thread length | writes replayed, before | after |
| -- | -- | -- |
| 2 turns | 3 | 1 |
| 6 turns | 7 | 1 |
| 20 turns | 21 | 1 |

Read latency is flat at \~0.6ms across all three after the change.

### Approach

Stage 1 now checks both places a value can live rather than trusting the marker. It probes `checkpoint_blobs`:

```sql
EXISTS (SELECT 1 FROM checkpoint_blobs b0
        WHERE b0.thread_id     = checkpoints.thread_id
          AND b0.checkpoint_ns = checkpoints.checkpoint_ns
          AND b0.channel       = %s
          AND b0.version       = checkpoint -> 'channel_versions' ->> %s
          AND b0.type         <> 'empty') AS hb_0
```

and selects the inline value alongside it, since `None`, `str`, `int`, `float` and `bool` stay in `channel_values` with no blob row:

```sql
checkpoint -> 'channel_values' -> %s AS inline_0
```

The blob predicate matches `checkpoint_blobs`' primary key `(thread_id, checkpoint_ns, channel, version)` exactly, so it is one index lookup per row per channel, bounded by the 1024-row page.

I picked reading storage over the cheaper alternative — also writing the marker for plain values — because **that would not fix any thread already on disk.** Existing checkpoints have no marker and there is nowhere to add one retroactively.

The seed resolves to the blob when one exists and the inline value otherwise. That ordering is also what keeps a genuine inline `true` — a `bool` channel holding `True` — distinguishable from the literal `true` marker `put` inlines for a `_DeltaSnapshot`: only the snapshot has a blob.

`None` is deliberately not treated as a seed; a JSON null is indistinguishable from "nothing stored" at this layer, so the walk continues and replay from empty is correct.

Params go from two to four per channel; both callers updated.

The inline half came out of review on this PR — a blob-only probe would have left scalar-aggregate migrations (an integer sum, say) still replaying their full history.

### On the `type <> 'empty'` predicate

Being upfront since it isn't demonstrable with a test: `put` does not currently produce `empty` rows on this path — `blob_versions` is filtered to keys present in `channel_values`, so `_dump_blobs`' empty branch is unreachable from it. I confirmed there are no `empty` rows in a populated test database.

I kept it because stage 2 already applies the same check when resolving the seed blob. Without it the two stages could disagree: stage 1 terminates the walk on a row stage 2 then discards, producing no seed *and* a truncated write chain — the same failure shape this function exists to avoid. Rationale is in the docstring so the next reader doesn't have to ask. Happy to drop it if you'd rather not carry an unexercised predicate.

### Tests

`libs/checkpoint-postgres/tests/test_delta_plain_value_seed.py` — blob-stored plain-value seed, `_DeltaSnapshot` seed, a version bump with nothing stored (which must not stop the walk short of an older real value), inline primitives (`int`, `str`, `float`, `None`), and inline `True` versus the snapshot marker. Each fails against the behaviour it fixes.

Verified: postgres suite 269 passed on PG 15 and 16; delta-channel conformance against `AsyncPostgresSaver` went from 6 of 8 to 8 of 8, including the pre-existing `test_history_migration_plain_value_as_seed` failure this was causing; `make lint` clean.

### Not included

I wanted a Postgres conformance runner alongside `checkpoint-sqlite`'s, but it needs `langgraph-checkpoint-conformance` as a dev dependency and the contributing guide asks for maintainer sign-off before adding one. The direct tests above cover the same ground without it.

Worth flagging separately: **conformance effectively runs against** `InMemorySaver` **only today.** `libs/checkpoint-conformance/tests/` contains just `test_validate_memory.py`, and `checkpoint-sqlite`'s `test_conformance_delta.py` silently skips because the package isn't installed in its test environment (`importorskip`). Wiring it up for sqlite and postgres is what would have caught this bug, and langchain-ai/langgraph#8534 notes it.

Sqlite is unaffected by the bug itself — it stores `channel_values` inline and inspects them directly. `langgraph-api` already resolves seeds by version rather than by marker.

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -210,23 +210,39 @@ def _build_delta_stage1_sql(channels: Sequence[str], *, paged: bool) -> str:
                          AND b0.checkpoint_ns = checkpoints.checkpoint_ns
                          AND b0.channel = %s
                          AND b0.version = checkpoint -> 'channel_versions' ->> %s
-                         AND b0.type <> 'empty') AS hs_0,
+                         AND b0.type <> 'empty') AS hb_0,
+               checkpoint -> 'channel_values' -> %s AS inline_0,
                checkpoint -> 'channel_versions' ->> %s AS ver_1,
-               EXISTS (...) AS hs_1
+               EXISTS (...) AS hb_1,
+               checkpoint -> 'channel_values' -> %s AS inline_1
         FROM checkpoints
         WHERE thread_id = %s AND checkpoint_ns = %s
           AND (%s::text IS NULL OR checkpoint_id < %s)
         ORDER BY checkpoint_id DESC
         LIMIT %s
 
-    `hs_i` ("has seed") asks whether a stored value exists for the channel at
-    this checkpoint. It probes `checkpoint_blobs` rather than testing for a key
-    in `checkpoint_values`, because `put` only leaves an inline marker there for
-    `_DeltaSnapshot` values — a plain value (what a thread migrated from a
-    pre-delta channel type leaves behind) is moved to the blobs table with no
-    marker, and was therefore invisible to the walk. The probe hits
-    `checkpoint_blobs`' primary key `(thread_id, checkpoint_ns, channel,
-    version)` exactly, so it is an index lookup per row per channel.
+    A stored value for a channel lives in one of two places, because `put`
+    splits them:
+
+    * **blob** — non-primitive values (and `_DeltaSnapshot`) are moved to
+      `checkpoint_blobs`. `hb_i` ("has blob") probes for one. The probe hits
+      that table's primary key `(thread_id, checkpoint_ns, channel, version)`
+      exactly, so it is an index lookup per row per channel.
+    * **inline** — `None`, `str`, `int`, `float` and `bool` stay in the
+      checkpoint's own `channel_values` and get no blob row at all. `inline_i`
+      returns that value.
+
+    Testing only for a key in `channel_values` (the previous approach) missed
+    blob-stored plain values, since `put` leaves an inline marker there for
+    `_DeltaSnapshot` but not for a plain value — which is what a thread
+    migrated from a pre-delta channel type leaves behind. Probing only the
+    blobs table would conversely miss inline primitives. Both are needed, and
+    the caller treats "either present" as the seed.
+
+    `hb_i` also disambiguates the two: for a `_DeltaSnapshot`, `inline_i` is the
+    literal `true` marker rather than the value, so a blob must win over an
+    inline reading whenever one exists. That ordering is what makes a genuine
+    inline `true` (a bool channel) distinguishable from the marker.
 
     The `type <> 'empty'` predicate mirrors the check stage 2 already applies
     when resolving the seed blob. `put` does not currently produce `empty` rows
@@ -237,14 +253,14 @@ def _build_delta_stage1_sql(channels: Sequence[str], *, paged: bool) -> str:
     truncated write chain, which is the failure this function exists to avoid.
 
     Channel names are passed as `%s` parameters (safe from SQL injection).
-    Only the column aliases `ver_i` / `hs_i` and the subquery alias `b{i}` are
-    interpolated into the SQL string (i is bounded by len(channels) and uses
-    safe identifiers).
+    Only the column aliases `ver_i` / `hb_i` / `inline_i` and the subquery alias
+    `b{i}` are interpolated into the SQL string (i is bounded by len(channels)
+    and uses safe identifiers).
 
-    Caller must extend params with `[ch_0, ch_0, ch_0, ch_1, ch_1, ch_1, ...,
-    thread_id, ns, cursor, cursor, page_size]` when `paged=True` — three per
-    channel: the version lookup, the blob's channel, and the version the blob
-    must match.
+    Caller must extend params with `[ch_0 x4, ch_1 x4, ..., thread_id, ns,
+    cursor, cursor, page_size]` when `paged=True` — four per channel: the
+    version lookup, the blob's channel, the version the blob must match, and the
+    inline lookup.
 
     When `paged=False`, the WHERE has no cursor predicate and there's no
     LIMIT/ORDER BY — kept as a non-public helper for tests/diagnostics.
@@ -258,7 +274,8 @@ def _build_delta_stage1_sql(channels: Sequence[str], *, paged: bool) -> str:
             f"AND b{i}.checkpoint_ns = checkpoints.checkpoint_ns "
             f"AND b{i}.channel = %s "
             f"AND b{i}.version = checkpoint -> 'channel_versions' ->> %s "
-            f"AND b{i}.type <> 'empty') AS hs_{i}"
+            f"AND b{i}.type <> 'empty') AS hb_{i}, "
+            f"checkpoint -> 'channel_values' -> %s AS inline_{i}"
         )
     sql = (
         "SELECT checkpoint_id, parent_checkpoint_id, "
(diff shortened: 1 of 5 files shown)


libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
```diff
@@ -228,7 +253,12 @@ def _build_delta_stage1_sql(channels: Sequence[str], *, paged: bool) -> str:
     for i in range(len(channels)):
         cols.append(
             f"checkpoint -> 'channel_versions' ->> %s AS ver_{i}, "
-            f"(checkpoint -> 'channel_values' -> %s) IS NOT NULL AS hs_{i}"
+            f"EXISTS (SELECT 1 FROM checkpoint_blobs b{i} "
+            f"WHERE b{i}.thread_id = checkpoints.thread_id "
+            f"AND b{i}.checkpoint_ns = checkpoints.checkpoint_ns "
+            f"AND b{i}.channel = %s "
+            f"AND b{i}.version = checkpoint -> 'channel_versions' ->> %s "
+            f"AND b{i}.type <> 'empty') AS hs_{i}"
```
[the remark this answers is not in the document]
eliornl: Good catch that inline primitives aren't detected by the blob probe — that's true. But the conclusion doesn't hold: this PR improves that case rather than regressing it.

Measured both, seed at step 1, writes at every step:

```
                            main (pre-fix)              this PR
int 42 (inline)      no seed, writes [w1,w2]     no seed, writes [w0,w1,w2]
list [10,20]         no seed, writes [w0,w1,w2]  seed=[10,20], writes [w1,w2]
```

`main` didn't resolve inline primitive seeds either. Stage 1 detected the inline key and terminated the walk, but stage 2 fetches the seed from `checkpoint_blobs` by version and there is no row for an inline value — so `seed` was omitted anyway, and the chain had already been truncated. That drops the value *and* the earlier writes. This PR walks past it and replays the full history instead, which reconstructs correctly for an additive reducer.

The same applies to the pruning scenario: if pre-seed writes are gone, both versions are equally stuck, since neither returns a seed.

So restoring the inline key check as an `OR` would put stage 1 back to claiming a seed that stage 2 cannot fetch — worse than either current behaviour.

The real fix is to carry the inline value through rather than just detect it: select `checkpoint -> 'channel_values' -> %s` in stage 1 alongside the blob probe, and in assembly prefer the blob when one exists, else fall back to the inline value. The blob check is what disambiguates a genuine inline `true` from the `_DeltaSnapshot` marker.

That touches the walk plumbing in about six places (stage-1 SQL, page ingest, walk advance, history assembly, and both callers' state), so I'd rather not fold it into a three-file fix. Happy to do it here if a maintainer prefers, otherwise I'll open it as a follow-up carrying the measurement above.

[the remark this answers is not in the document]
eliornl: Implemented it here rather than deferring — `83ffb168`.

Stage 1 now selects `checkpoint -> 'channel_values' -> %s` alongside the blob probe, and the seed resolves to the blob when one exists and the inline value otherwise. That ordering is what keeps a genuine inline `true` distinguishable from the `_DeltaSnapshot` marker, since only the snapshot has a blob.

`None` is deliberately not a seed: a JSON null is indistinguishable from "nothing stored" at this layer, so the walk continues and replay from empty is correct.

Tests cover inline `int`, `str`, `float`, `None`, and inline `True` versus the marker. Both new cases fail without the change. Postgres suite 269 passed on PG 15 and 16.

libs/checkpoint-postgres/tests/test_delta_plain_value_seed.py
```diff
@@ -0,0 +1,204 @@
...
+
+@pytest.mark.asyncio
+async def test_version_bump_without_a_value_does_not_hide_an_older_seed() -> None:
+    """A delta-era step bumps `channel_versions` without storing a value, so no
+    blob exists for that version. The probe must report no seed there and keep
+    walking rather than stopping at a version it cannot resolve.
+
+    Step 0 holds the real value; step 1 bumps the version with nothing stored.
+    Walking back from the head has to pass step 1 to reach step 0.
+    """
```
pjrule, line 106: to make sure I'm following properly: is this a red/green test? (that is, was it failing before?)

eliornl: yes. reverted the three source files to main and reran: 4 of 5 fail, including this one. pre-fix it gives no seed and writes [w0, w1, w2], since stage 1 stops at step 1's bumped version where no blob exists; with the fix it reaches step 0, seed=[10, 20] and writes [w1, w2]. the only one that passes both ways is test_delta_snapshot_seed_is_found, which is the regression guard for the existing path.

pjrule: gotcha, thx!

pjrule (approved):

PR #8537, opened 2026-08-05
test(checkpoint-postgres,checkpoint-sqlite): run the conformance suite

eliornl:
Depends on #8535

`libs/checkpoint-conformance/tests/` only validates `InMemorySaver`. `checkpoint-sqlite` has had a `test_conformance_delta.py` for a while, but it guards on `importorskip("langgraph.checkpoint.conformance")` and the package was never in its test environment — so it has been skipping silently every run. `checkpoint-postgres` had no runner at all.

Net effect: the shared checkpointer contract was effectively unenforced everywhere except in-memory.

### Change

Adds `langgraph-checkpoint-conformance` to the `test` dependency group of both packages, with a path source like the existing `langgraph-checkpoint` entry. That alone is what makes sqlite's runner start executing. Postgres gets the equivalent runner.

Both pass the `delta_channel_history` capability.

### Why it's stacked

Against `main`'s Postgres, the new runner fails:

```
Capability delta_channel_history failed:
  test_history_migration_plain_value_as_seed
```

That is exactly the bug #8535 fixes, and it had been failing unnoticed precisely because nothing ran the suite there. So this is based on that branch rather than `main` — the diff here is the one conformance commit, and it will retarget once #8535 lands.

Reasonable to read that as the change justifying itself: the first thing turning the suite on did was catch a real bug that had been sitting in `main`.

### Verified

`checkpoint-postgres` 270 passed on PG 15 and 16, `checkpoint-sqlite` 118 passed, lint and `ty` clean in both. The `uv.lock` updates are the conformance package entry only.

### Note

The sync `PostgresSaver` and `SqliteSaver` aren't covered — the conformance harness reports every capability as `detected=False` for them, so only the async savers are exercised. Pre-existing and not addressed here, but worth knowing the coverage isn't total.

### DIFF
--- a/libs/checkpoint-postgres/pyproject.toml
+++ b/libs/checkpoint-postgres/pyproject.toml
@@ -32,6 +32,7 @@ test = [
   "pytest-mock",
   "psycopg[binary]",
   "langgraph-checkpoint",
+  "langgraph-checkpoint-conformance",
   "pytest-watcher",
 ]
 lint = [
@@ -49,6 +50,7 @@ default-groups = ['dev']
 
 [tool.uv.sources]
 langgraph-checkpoint = { path = "../checkpoint", editable = true }
+langgraph-checkpoint-conformance = { path = "../checkpoint-conformance", editable = true }
 
 [tool.hatch.build.targets.wheel]
 include = ["langgraph"]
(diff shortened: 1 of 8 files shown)


sydney-runkle (commented):
nice, do we need to add a test file to the sqlite saver as well? 
also, i'm seeing more local imports in tests -- can we a) fix that in this PR and b) in a different PR, update lint rules across packages so that this doesn't happen?

eliornl:
sqlite already has one - it was just silently skipping, since importorskip guarded on a package that was never installed. adding the dev dep makes it actually run.

imports fixed in 32a5ae83. they were only inside the function to get past that same guard, so with the dep declared both guards came out and the imports moved to the top.

happy to take the lint rule as a follow-up

sydney-runkle (approved):
nice!

Issue #7127, opened 2026-03-12
Can not get sub graph state when using Per-thread


a171232886:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Reproduction Steps / Example Code (Python)

```python
from langgraph.graph import START, StateGraph, MessagesState
from langgraph.checkpoint.memory import MemorySaver

class SubState(MessagesState):
    pass

def node(state: SubState):
    print("Node received messages:", state["messages"])
    return {"messages": state["messages"] + [{"role": "assistant", "content": "Hello!"}]}

# Subgraph with its own persistent state
subgraph_builder = StateGraph(SubState)
subgraph_builder.add_node("sub_node", node)
subgraph_builder.add_edge(START, "sub_node")
subgraph = subgraph_builder.compile(checkpointer=True)

# Parent graph
builder = StateGraph(MessagesState)
builder.add_node("agent", subgraph)
builder.add_edge(START, "agent")

checkpointer = MemorySaver()
graph = builder.compile(checkpointer=checkpointer)

config = {"configurable": {"thread_id": "1"}}

graph.invoke({"messages": [{"role": "user", "content": "hi"}]}, config)
graph.invoke({"messages": [{"role": "user", "content": "what did I say?"}]}, config)

# View accumulated subgraph state (includes messages from both invocations)
subgraph_state = graph.get_state(config, subgraphs=True).tasks[0].state
```

### Error Message and Stack Trace (if applicable)

```shell
Node received messages: [HumanMessage(content='hi', additional_kwargs={}, response_metadata={}, id='838756a8-a649-448b-8ad6-9b40b739d01e')]
[...]

PR #8538, opened 2026-08-05
fix(langgraph): hydrate subgraph delta channels with the caller-resolved saver

### LINKED ISSUE #8470: Reading a nested subgraph's state hydrates every DeltaChannel empty
### Summary

Reading a **nested subgraph's** state returns every `DeltaChannel` **empty**, silently. Non-delta channels in the same namespace hydrate correctly, so nothing errors and nothing looks wrong — the caller just sees an empty channel and cannot tell it from a subgraph that genuinely never wrote one.

This bites anyone inspecting run history per subgraph: `messages` on an agent that opts into delta storage is exactly the channel you want, and it always comes back `[]`.

### Repro (self-contained, `InMemorySaver`)

```python
from typing import Annotated
from langgraph.graph import START, END, StateGraph

def _append(state, writes):
    out = list(state or [])
    for w in writes:
        out.extend(w if isinstance(w, list) else [w])
    return out

class S(TypedDict, total=False):
    msgs: Annotated[list, DeltaChannel(_append)]

child = StateGraph(S)
child.add_node("a", lambda s: {"msgs": ["a1"]})
child.add_node("b", lambda s: {"msgs": ["b1", "b2"]})
child.add_edge(START, "a"); child.add_edge("a", "b"); child.add_edge("b", END)

parent = StateGraph(S)
parent.add_node("child", child.compile())
parent.add_edge(START, "child"); parent.add_edge("child", END)

app = parent.compile(checkpointer=InMemorySaver())
cfg = {"configurable": {"thread_id": "t1"}}
app.invoke({}, cfg)

child_ns = next(
    t.state["configurable"]["checkpoint_ns"]
    for snap in app.get_state_history(cfg)
    for t in snap.tasks
    if t.name == "child" and isinstance(t.state, dict)
)

print(app.get_state(cfg).values["msgs"])                      # ['a1', 'b1', 'b2']  ✅
print(app.get_state({"configurable": {"thread_id": "t1",
                     "checkpoint_ns": child_ns}}).values)      # {'msgs': []}        ❌
```

**Expected:** the child namespace reports `['a1', 'b1', 'b2']`.
**Actual:** `[]`.

Affects `get_state`, `aget_state`, `get_state_history` and `aget_state_history`.

### Cause

A subgraph obtained through `get_subgraphs()` was compiled **without** a checkpointer — the parent supplies one via `CONFIG_KEY_CHECKPOINTER` at read time, which is how the public readers already resolve it (`pregel/main.py`):

```python
checkpointer = ensure_config(config)[CONF].get(CONFIG_KEY_CHECKPOINTER, self.checkpointer)
```

But `_prepare_state_snapshot` then hydrates channels using `self.checkpointer` alone, so `saver` is `None` for any nested subgraph:

```python
channels, managed = channels_from_checkpoint(
    self.channels, saved.checkpoint,
    saver=self.checkpointer if isinstance(self.checkpointer, BaseCheckpointSaver) else None,
    config=saved.config,
)
```

`channels_from_checkpoint` needs that saver to call `get_delta_channel_history` and replay a `DeltaChannel`'s ancestor writes. Without it, the channel falls through to `from_checkpoint(MISSING)` — empty, no error (`pregel/_checkpoint.py`):

```python
if delta_channels and saver is not None and config is not None:
    histories = saver.get_delta_channel_history(config=config, channels=delta_channels)
...
ch = spec.from_checkpoint(checkpoint["channel_values"].get(k, MISSING))
```

Recovering it from the config isn't possible either: `get_state_history` passes `checkpoint_tuple.config` — the checkpointer's *stored* config, which never carries `CONFIG_KEY_CHECKPOINTER`.

### Suggested fix

Pass the already-resolved saver into `_prepare_state_snapshot` / `_aprepare_state_snapshot` from the four public readers, preferring it over `self.checkpointer`. +26/−4, read path only.

I have this working with regression tests (subgraph history, subgraph `get_state`, async history, plus a root-graph control) — the subgraph tests fail on `main` and pass with the change, and the full suite shows a byte-for-byte identical failure set before and after. Happy to open the PR once this is approved; it was auto-closed for not linking an issue.

### How it was found

Building a run-inspection UI on `POST /threads/{id}/history`. Every nested deep agent reported an empty `messages` channel while its tasks had demonstrably run many model turns and tool calls — with 1452 kB of message blobs sitting in `checkpoint_blobs` for one such namespace. Plain agents, whose `messages` is a `BinaryOperatorAggregate` rather than a `DeltaChannel`, were unaffected.

### System info

langgraph `main` @ 1.2.10 (also reproduced on 1.2.1), Python 3.13, macOS.

eliornl:
Fixes langchain-ai/langgraph#8470

Reported by @gururafiki, with a self-contained `InMemorySaver` repro and the observation that non-delta channels in the same namespace hydrate fine, which is what makes the failure silent.

`_prepare_state_snapshot` re-derived the saver from `self.checkpointer`, which is `None` for a subgraph, so every `DeltaChannel` in a nested subgraph hydrated empty on read, silently.

### Why the old resolution was wrong

A `DeltaChannel` stores no value in `channel_values`, so reconstructing it means walking ancestors and replaying their writes, which needs a saver. A subgraph has none of its own: it borrows the parent's through `CONFIG_KEY_CHECKPOINTER` at read time, exactly as the four public readers already resolve it a few lines above the call.

`_prepare_state_snapshot` ignored that and went back to `self.checkpointer`. With no saver the walk never ran, and the channel fell through to `from_checkpoint(MISSING)`: an empty value, indistinguishable from a channel that was never written, raising nothing. Non-delta channels in the same namespace hydrate fine, so the payload looks healthy.

Recovering it inside the function isn't possible: `get_state_history` passes `checkpoint_tuple.config`, the checkpointer's *stored* config, which never carries a live saver. The caller is the only place the resolved saver exists, so the caller has to pass it.

### How I verified it

`tests/test_delta_channel_subgraph.py` adds 12 tests. **8 fail on main, 4 pass.** The 4 that pass are controls for modes that stay unchanged (the two root-graph cases, a stateless subgraph, a completed subgraph). The 8 failures, by surface:

| surface | on main |
| -- | -- |
| `get_state` / `aget_state` on a subgraph ns | `{'msgs': []}` |
| `get_state_history` / `aget_state_history` | every snapshot empty |
| two levels of nesting | `{'msgs': []}` |
| `get_state(subgraphs=True)` task state, subgraph interrupted | `{'msgs': []}` |
| `update_state` / `aupdate_state` | `['manual']`, prior history destroyed |

Tests use `InMemorySaver`: the defect is in `Pregel`'s saver resolution rather than in any checkpointer implementation, so the storage backend is irrelevant to the reproduction.

Full suite `1968 -> 1980 passed, 4 skipped`, exactly the 12 new tests. `make format`, `lint_package`, `lint_tests` clean; `check_sdk_methods.py` passes.

### Two things worth a closer look in review

**1. This fixes a write path as well as a read path, and both are load-bearing.** `bulk_update_state` / `abulk_update_state` carried the same expression, and that half is worse than a misleading read: when the update reaches the channel's snapshot cadence, `create_checkpoint` persists a `_DeltaSnapshot` built from the empty-hydrated channel, losing history **on disk**. I checked whether it could be split: reverting only the two `bulk_update_state` hunks leaves `test_subgraph_update_state_preserves_delta_channel_history` and its async twin failing while the other 10 pass. Splitting would ship a fix that still corrupts history on write.

The write-path tests use `snapshot_frequency=2`, which is what makes the update reach the cadence and force a snapshot blob. At the default 1000 nothing is written, the loss stays latent, and the assertion would hold either way.

**2.** `prepare_next_tasks` **in the same two methods still receives** `self.checkpointer`**, and I left it.** It is inert on the read path: `_algo.py:915` builds the task config as `checkpointer or configurable.get(CONFIG_KEY_CHECKPOINTER)`, so a subgraph's `None` already falls back to the config value, and `PregelTask` carries no `config` field, so that config never reaches a reader. Changing it would be cosmetic, but I'll make it symmetric if you'd rather not leave two different expressions side by side.

### Note on the signature

`saver` is a **required** keyword argument rather than a defaulted one. A silent fallback is what caused this bug, and both methods are private with no other callers, so there is nothing to stay compatible with. `recurse` stays defaulted because `None` is meaningful there (do not resolve subgraph task states); for `saver`, `None` is only ever a bug.

Net -6 lines of logic: passing an already-narrowed saver removes the `isinstance` dance at every site.

### DIFF
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -1146,9 +1146,26 @@ class Pregel(
         self,
         config: RunnableConfig,
         saved: CheckpointTuple | None,
+        *,
+        saver: BaseCheckpointSaver | None,
         recurse: BaseCheckpointSaver | None = None,
         apply_pending_writes: bool = False,
     ) -> StateSnapshot:
+        """Assemble a `StateSnapshot` from a saved checkpoint.
+
+        Args:
+            config: Config identifying the checkpoint being read.
+            saved: The checkpoint tuple to render, or `None` for an empty snapshot.
+            saver: Checkpointer to read with, as resolved by the caller from
+                `CONFIG_KEY_CHECKPOINTER` before falling back to `self.checkpointer`.
+                Required rather than defaulted because `self.checkpointer` is `None`
+                for a subgraph — it borrows the parent's saver through the config —
+                and a `DeltaChannel` silently hydrates empty without one.
+            recurse: When set, resolve subgraph task states with this checkpointer
+                instead of returning a config that merely signals they exist.
+            apply_pending_writes: Apply this checkpoint's pending writes to the
+                returned values.
+        """
         if not saved:
             return StateSnapshot(
                 values={},
@@ -1169,9 +1186,7 @@ class Pregel(
         channels, managed = channels_from_checkpoint(
             self.channels,
             saved.checkpoint,
-            saver=self.checkpointer
-            if isinstance(self.checkpointer, BaseCheckpointSaver)
-            else None,
+            saver=saver,
             config=saved.config,
         )
         # tasks for this checkpoint
@@ -1269,9 +1284,12 @@ class Pregel(
         self,
         config: RunnableConfig,
         saved: CheckpointTuple | None,
+        *,
+        saver: BaseCheckpointSaver | None,
         recurse: BaseCheckpointSaver | None = None,
         apply_pending_writes: bool = False,
     ) -> StateSnapshot:
+        """Async version of `_prepare_state_snapshot`. See docstring there."""
         if not saved:
             return StateSnapshot(
                 values={},
@@ -1292,9 +1310,7 @@ class Pregel(
         channels, managed = await achannels_from_checkpoint(
             self.channels,
             saved.checkpoint,
-            saver=self.checkpointer
-            if isinstance(self.checkpointer, BaseCheckpointSaver)
-            else None,
+            saver=saver,
             config=saved.config,
         )
         # tasks for this checkpoint
@@ -1429,6 +1445,7 @@ class Pregel(
         return self._prepare_state_snapshot(
             config,
             saved,
+            saver=checkpointer,
             recurse=checkpointer if subgraphs else None,
             apply_pending_writes=CONFIG_KEY_CHECKPOINT_ID not in config[CONF],
         )
@@ -1473,6 +1490,7 @@ class Pregel(
         return await self._aprepare_state_snapshot(
             config,
             saved,
+            saver=checkpointer,
             recurse=checkpointer if subgraphs else None,
             apply_pending_writes=CONFIG_KEY_CHECKPOINT_ID not in config[CONF],
         )
@@ -1527,7 +1545,7 @@ class Pregel(
             checkpointer.list(config, before=before, limit=limit, filter=filter)
         ):
             yield self._prepare_state_snapshot(
-                checkpoint_tuple.config, checkpoint_tuple
+                checkpoint_tuple.config, checkpoint_tuple, saver=checkpointer
             )
 
     async def aget_state_history(
@@ -1584,7 +1602,7 @@ class Pregel(
             )
         ]:
             yield await self._aprepare_state_snapshot(
-                checkpoint_tuple.config, checkpoint_tuple
+                checkpoint_tuple.config, checkpoint_tuple, saver=checkpointer
             )
 
     def bulk_update_state(
@@ -1666,10 +1684,7 @@ class Pregel(
             channels, managed = channels_from_checkpoint(
                 self.channels,
                 checkpoint,
-                saver=self.checkpointer
-                if saved is not None
-                and isinstance(self.checkpointer, BaseCheckpointSaver)
-                else None,
+                saver=checkpointer if saved is not None else None,
                 config=saved.config if saved is not None else None,
             )
             values, as_node = updates[0][:2]
@@ -2132,10 +2147,7 @@ class Pregel(
             channels, managed = await achannels_from_checkpoint(
                 self.channels,
                 checkpoint,
-                saver=self.checkpointer
-                if saved is not None
-                and isinstance(self.checkpointer, BaseCheckpointSaver)
-                else None,
+                saver=checkpointer if saved is not None else None,
                 config=saved.config if saved is not None else None,
             )
             values, as_node = updates[0][:2]
--- a/libs/langgraph/langgraph/pregel/main.py
+++ b/libs/langgraph/langgraph/pregel/main.py
@@ -1159,7 +1159,7 @@ class Pregel(
             saver: Checkpointer to read with, as resolved by the caller from
                 `CONFIG_KEY_CHECKPOINTER` before falling back to `self.checkpointer`.
                 Required rather than defaulted because `self.checkpointer` is `None`
-                for a subgraph — it borrows the parent's saver through the config —
+                for a subgraph, which borrows the parent's saver through the config,
                 and a `DeltaChannel` silently hydrates empty without one.
             recurse: When set, resolve subgraph task states with this checkpointer
                 instead of returning a config that merely signals they exist.
--- a/libs/langgraph/tests/test_delta_channel_subgraph.py
+++ b/libs/langgraph/tests/test_delta_channel_subgraph.py
@@ -2,7 +2,7 @@
 
 Regression suite for #8470.
 
-A subgraph is compiled without a checkpointer of its own — the parent lends it
+A subgraph is compiled without a checkpointer of its own. The parent lends it
 one through `CONFIG_KEY_CHECKPOINTER` at read time. `DeltaChannel` stores no
 value in `channel_values`, so hydrating it requires that saver to walk ancestors
 and replay their writes. Reading with `self.checkpointer` instead of the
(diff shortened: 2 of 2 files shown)


sydney-runkle (commented):
a few questions -- can we find any related bugs in deepagents?
is a checkpointer always none for subgraph? i think there are many options for how we can add: https://docs.langchain.com/oss/python/langgraph/use-subgraphs#subgraph-persistence...

sydney-runkle:
basically i'm not actually sure this is incorrect, i think it might just be a feature of how subgraphs manage state, worth reading the above

eliornl:
Good questions, and the doc changed my framing, so let me redo this properly.

**On whether it's a feature of subgraph state.** I went through the three modes the doc describes, using the read path it documents (`get_state(config, subgraphs=True).tasks[0].state`) rather than the raw `checkpoint_ns` read the issue uses. Setup is one subgraph state holding two channels written by the same nodes, one a `DeltaChannel` and one a plain `operator.add` list.

Stateless (`checkpointer=False`) gives nothing at all, and after a run completes there's no task state for either channel. You're right about both, they're subgraph behaviour rather than delta replay, and this PR leaves them alone. I've added tests pinning them so they stay that way.

The two that concern me are per-invocation and per-thread, during an interrupt. Both give:

```
plain (operator.add) : ['a1']
delta (DeltaChannel) : []
```

Same call, same instant, same state object. A subgraph-state rule would empty both. It's only the delta channel because that's the one type which stores no value in `channel_values` at all, so it has to walk ancestors and replay writes, and that needs a saver. `_prepare_state_snapshot` was passing `self.checkpointer`.

**On whether a checkpointer is always `None`.** It isn't, and per-thread is the case I'd most like your read on. `checkpointer=True` is an explicit opt-in to use the parent's checkpointer, and it also resolved to no saver, because `isinstance(True, BaseCheckpointSaver)` is `False`.

**On deepagents.** I cloned it at 0.7.4, built a parent with one subagent using their own fake chat model from the unit tests so there's no network, ran a turn, and dumped every namespace. Subagents do checkpoint under `tools:<task_id>`, and the bug is real in those bytes: `files` reads back `{}` while `messages` in the same read comes back with all four messages, and a control read by a graph that owns the saver returns the file contents fine.

But nothing that ships reaches that read. A subagent lives in the `task` tool's closure rather than as a node, so `get_subgraphs(recurse=True)` is empty and `get_state(checkpoint_ns="tools:...")` raises `Subgraph tools not found`, identically before and after this change. I had to hold the subagent `Pregel` directly to trigger anything. So on deepagents this is latent, and this PR moves nothing user-visible there.

Related issues point the same way. #4818 is a user doing exactly that read, closed by mdrxy with "the `tools:<uuid>` checkpoints are internal". #2629 closed by ccurme on the same grounds. And your #777 treats fetchable subagent history as something we haven't built.

**Where I land.** The documented limitation covers subgraphs reached through indirection, and its symptom is a hard error. This PR's case is a subgraph added as a node, which the docs list as supported, where you get a well-formed snapshot with one channel silently empty next to a correct one. Those seem like different things to me. I also couldn't find anywhere we'd decided the empty case is fine: no test in `libs/langgraph`, the conformance spec, or langgraph-api ever puts a `DeltaChannel` outside `checkpoint_ns: ""`, so it's untested rather than asserted. And #7127/#7128 are open reports of subgraph state being unreadable in per-thread mode.

So I'd still merge, but deepagents isn't the argument for it. If that were the only case I'd close this. What keeps me on it is the node-added subgraph plus the write path, where `update_state` persists the empty hydration and loses history on disk.

gururafiki:
Hi! Thanks for opening PR and for review. I have experienced related problem in deepagents. Here is the opened issue for it: https://github.com/langchain-ai/deepagents/issues/5136

eliornl:
Thanks, and good to see #5136 written up. Worth separating the two though, since they look alike from the outside.

#5136 is the tool-indirection case: the subagent lives in the `task` tool's closure, so LangGraph can't statically discover it and the read raises `Subgraph tools not found`. That's documented as unsupported, and this PR doesn't change it. I checked separately while looking into deepagents, and the error is identical before and after this change.

This PR is the case where the subgraph *is* statically discoverable, added as a node, which the docs list as supported. There you get no error at all, just a well-formed snapshot with a `DeltaChannel` silently empty next to a correct plain channel. Same asymmetry you saw in the stored deepagents data, but reachable through a supported path.

dineshyadav03 (commented):
Read through the diff and the linked repro. Confirmed the root cause: `get_state`/`aget_state` already resolve `checkpointer` correctly via `CONFIG_KEY_CHECKPOINTER` (falling back to `self.checkpointer`), but `_prepare_state_snapshot`/`_aprepare_state_snapshot` discarded that local and re-read `self.checkpointer` directly, which is `None` for any subgraph. Since `DeltaChannel` has no value in `channel_values` and needs a saver to replay ancestor writes, that produced an empty-but-valid-looking snapshot instead of an error. Good catch pulling in `bulk_update_state`/`abulk_update_state` too — that one persists the empty-hydrated channel as a `_DeltaSnapshot` blob once `snapshot_frequency` cadence hits, i.e. actual on-disk history loss.

Question: `prepare_next_tasks` in the same two methods still gets `self.checkpointer` and is called inert because `_algo.py` falls back to `CONFIG_KEY_CHECKPOINTER` internally. Is that fallback covered by a test, or only reasoned about? Given this whole bug was exactly this kind of "should be equivalent" assumption, I'd rather see the two expressions unified even if cosmetic today.

Making `saver` a required kwarg rather than defaulting to `self.checkpointer` is the right call. Tests are thorough. Approve.

PR #8540, opened 2026-08-05
chore(checkpoint-postgres,checkpoint-sqlite): enable PLC0415 lint rule

eliornl:
Follow-up to review on #8537: turn on ruff's `PLC0415` (`import-outside-top-level`) so deferred imports in tests stop accumulating.

Scoped to `checkpoint-postgres` and `checkpoint-sqlite` rather than repo-wide, because the sweep turns up three different things and only one of them is a style problem.

### What the rule finds today

```
package                 tests   src    files
checkpoint                13     10      11
checkpoint-conformance     0     10       4
checkpoint-postgres        6      0       2
checkpoint-sqlite          9      0       3
langgraph                130     23      32
prebuilt                  14      3       7
cli                        9     14      10
sdk-py                   189     23      38
                        ────────────────────
                         370     83     107
```

453 violations across 107 files, and ruff has no autofix for this rule.

### Three categories, not one

**Style — hoist.** `checkpoint-sqlite/tests/test_store.py` deferred `math`, `random`, `time`, `Counter` and `defaultdict` inside methods for no reason.

**Deliberate — keep, annotate.** `checkpoint-postgres/tests/test_async.py` defers behind `pytest.importorskip("langgraph.channels.delta")` because langgraph core is *not* a test dependency of that package. Hoisting would break the skip. Those get `# noqa: PLC0415` and a comment.

**Redundant guard — hoist.** `checkpoint-sqlite/tests/test_conformance_delta.py` deferred imports only to get past its own `importorskip`. Imports move up; the `aiosqlite` guard stays, since that dependency genuinely can be absent.

The second category is why I did not enable this everywhere in one go. Most of the 83 source-level violations look like the same pattern — optional-dependency handling and circular-import avoidance in `jsonplus.py`, `embed.py`, `encrypted.py` and friends. Blanket-enabling would mean `# noqa` on a lot of correct code, and each one wants an owner's eye rather than a mechanical pass.

These two packages are clean to enforce today because both have **zero** source-level violations.

### Suggested rollout for the rest

Either extend package by package as owners confirm which deferrals are intentional, or enable everywhere at once with `per-file-ignores` grandfathering the current 107 files so new code is blocked immediately and the debt burns down. Happy to do either — the second is a smaller diff but leaves a long ignore list.

### Verified

`checkpoint-sqlite` 118 passed, `checkpoint-postgres` 264 passed on PG 15 and 16, `make lint` clean in both.

One overlap worth flagging: `checkpoint-sqlite/tests/test_conformance_delta.py` is also touched by #8537. The change is identical in both, so it should merge cleanly either way.

### DIFF
--- a/libs/checkpoint-sqlite/tests/test_conformance_delta.py
+++ b/libs/checkpoint-sqlite/tests/test_conformance_delta.py
@@ -3,10 +3,17 @@
 from __future__ import annotations
 
 import pytest
-from langgraph.checkpoint.conformance import validate
-from langgraph.checkpoint.conformance.initializer import checkpointer_test
 
-from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
+pytest.importorskip(
+    "langgraph.checkpoint.conformance",
+    reason="langgraph-checkpoint-conformance not installed",
+)
+pytest.importorskip("aiosqlite", reason="aiosqlite not installed")
+
+from langgraph.checkpoint.conformance import validate  # noqa: E402
+from langgraph.checkpoint.conformance.initializer import checkpointer_test  # noqa: E402
+
+from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver  # noqa: E402
 
 
 @pytest.mark.asyncio
--- a/libs/checkpoint-sqlite/tests/test_conformance_delta.py
+++ b/libs/checkpoint-sqlite/tests/test_conformance_delta.py
@@ -10,14 +10,14 @@ pytest.importorskip(
 )
 pytest.importorskip("aiosqlite", reason="aiosqlite not installed")
 
+from langgraph.checkpoint.conformance import validate  # noqa: E402
+from langgraph.checkpoint.conformance.initializer import checkpointer_test  # noqa: E402
+
+from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver  # noqa: E402
+
 
 @pytest.mark.asyncio
 async def test_delta_channel_conformance():
-    from langgraph.checkpoint.conformance import validate
-    from langgraph.checkpoint.conformance.initializer import checkpointer_test
-
-    from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
-
     @checkpointer_test(name="AsyncSqliteSaver")
     async def sqlite_saver():
         async with AsyncSqliteSaver.from_conn_string(":memory:") as saver:
--- a/libs/checkpoint-sqlite/tests/test_conformance_delta.py
+++ b/libs/checkpoint-sqlite/tests/test_conformance_delta.py
@@ -3,17 +3,10 @@
 from __future__ import annotations
 
 import pytest
+from langgraph.checkpoint.conformance import validate
+from langgraph.checkpoint.conformance.initializer import checkpointer_test
 
-pytest.importorskip(
-    "langgraph.checkpoint.conformance",
-    reason="langgraph-checkpoint-conformance not installed",
-)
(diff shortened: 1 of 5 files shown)


libs/checkpoint-sqlite/tests/test_conformance_delta.py
```diff
@@ -10,14 +10,14 @@
 )
 pytest.importorskip("aiosqlite", reason="aiosqlite not installed")
 
+from langgraph.checkpoint.conformance import validate  # noqa: E402
```
sydney-runkle: what's e402

sydney-runkle (approved):
nice

let's do for other packages too!

libs/checkpoint-sqlite/tests/test_conformance_delta.py
eliornl: `E402` is "module level import not at top of file". Ruff flags imports that sit below other statements, and the `importorskip` calls above them are statements, so the imports below trip it.

Good question though, because it turns out I did not need it. Both guards are dead:

- `aiosqlite` is a hard runtime dependency of this package (`dependencies` in pyproject), so that guard can never skip.
- `langgraph-checkpoint-conformance` became a declared test dependency in this PR, so that one cannot either.

Dropped both in `a2007a0a`. The imports now sit at the top and the `noqa: E402` suppressions go with them. Test still passes.

sydney-runkle: can we have a lint rule that errors if an unused rule is being used? i think that exists

libs/checkpoint-sqlite/tests/test_conformance_delta.py
```diff
@@ -3,21 +3,14 @@
 from __future__ import annotations
 
 import pytest
+from langgraph.checkpoint.conformance import validate
+from langgraph.checkpoint.conformance.initializer import checkpointer_test
```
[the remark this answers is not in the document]
eliornl: Correct, and thanks. I removed the guards on the assumption the conformance package was a declared test dependency, but that declaration is in #8537, and this branch came off `main`. Reverted in `4cbcb526`, so the guards and the `noqa: E402` lines are back and sqlite is green again (118 passed).

The guard removal already exists on #8537 where the dependency is actually present, so it lands with that PR instead.

libs/checkpoint-sqlite/tests/test_conformance_delta.py
eliornl: Correcting myself here: the second bullet was wrong for this PR. `langgraph-checkpoint-conformance` becomes a declared test dependency in #8537, not this one, and this branch is off `main`, so removing the guard broke sqlite test collection. Reverted in `4cbcb526`.

So the answer is just the first part: `E402` is "module level import not at top of file", and the imports trip it because the `importorskip` calls above them are statements. The `noqa` lines stay until #8537 lands, at which point the conformance guard really is dead and both can go.

eliornl: Yes, `RUF100`. It flags a `noqa` that suppresses nothing, and it is autofixable.

Ran it across the packages with `--extend-select` so ruff still knows what else is enabled, which matters — running `--select RUF100` alone makes every `noqa` for another rule look unused:

```
checkpoint-sqlite    15
langgraph             9
prebuilt              5
checkpoint-postgres   4
others                0
                     ──
                     33
```

Two flavours in there: `unused: F821` for suppressing something that no longer fires, and `non-enabled: FBT001` for suppressing a rule that is not even turned on.

Fitting given the thread above, since it would have caught my dead `E402` lines. Taking it as its own PR as you suggested.

PR #8542, opened 2026-08-05
fix(cli): reject credential-bearing Git dependencies

jkennedyvz:
## Description
Reject Git HTTP dependency URLs containing userinfo before Docker generation so credentials cannot persist in Dockerfiles or image layers. Validation covers direct Python dependencies, local Python metadata, uv project and lock inputs, and common Node.js package and lock files at both workspace and application roots. Errors identify the inspected file path without echoing dependency contents.

Detection intentionally scans raw text conservatively, so placeholders such as `${GIT_TOKEN}@...` and commented examples are rejected. Coverage remains best-effort because arbitrary files, nested requirements includes, and constraints can still be copied into a build context; the README now describes that boundary and warns against credentials in copied pip configuration.

## Test Plan
- [x] Validate credential-bearing direct, local-manifest, uv-managed, Node.js, and workspace Git URLs are rejected without echoing credentials
- [x] Validate credential-free HTTPS and SSH Git URLs remain supported
- [x] Validate direct dependency errors identify the source `langgraph.json`
- [x] `make format`; `make lint`; focused config suite: 132 passed

Made by [Open SWE](https://openswe.vercel.app/agents/1b3b604f-1ee4-57de-8bf3-4defc911054b)


anujbolewar (commented):
Good hardening — rejecting credential-bearing Git URLs at config-parse time is the correct place to stop them before they reach a generated Dockerfile or image layer. Since the check is regex based, cover the obvious bypasses in tests: percent-encoded userinfo and alternative schemes such as git+ssh where a username may be embedded differently. Confirming the error message points users at the secret-backed credential-helper flow keeps the escape hatch discoverable.

eliornl (commented):
Superseded, see the inline comments on the diff.

libs/cli/README.md
```diff
@@ -103,6 +103,8 @@ The CLI uses a `langgraph.json` configuration file with these key settings:
 }
 ```
 
+Git dependencies must use credential-free URLs. The CLI rejects HTTP Git URLs with userinfo because generated Dockerfiles and image layers can retain embedded usernames or tokens. For private dependencies, provide short-lived credentials through your build environment's secret-backed Git credential helper. Do not store credentials in `langgraph.json`, requirement or lock files, or a `pip_config_file` copied into the image.
```
eliornl: This sentence is unconditional, but the check only runs on the Python path. `node_config_to_docker` does `ADD . /deps/<name>`, so `package.json` and the lockfile land in a layer verbatim, and it never calls the new validator.

Verified against this branch: a JS config with `"priv": "git+https://user:token@github.com/org/private.git"` in `package.json` generates a Dockerfile with no error and the token copied in.

Either extend `_validate_git_http_url_userinfo_files` over `package.json` / `package-lock.json` / `yarn.lock` / `pnpm-lock.yaml` in `node_config_to_docker`, or scope this sentence to Python. An unqualified safety claim that is false for JS users is worse than no claim.

libs/cli/langgraph_cli/config.py
```diff
@@ -78,6 +83,57 @@ def has_disallowed_build_command_content(command: str) -> bool:
...
+def _validate_local_dependency_files(config_path: pathlib.Path, config: Config) -> None:
+    """Validate dependency files copied into a non-uv Python image."""
+    paths: list[pathlib.Path] = []
+    for dependency in config["dependencies"]:
+        if not isinstance(dependency, str) or not dependency.startswith("."):
+            continue
+        root = (config_path.parent / dependency).resolve()
+        paths.extend(
+            root / name
+            for name in ("requirements.txt", "pyproject.toml", "setup.py", "setup.cfg")
```
eliornl, line 137: Four fixed filenames, but the Dockerfile copies the whole directory. Two cases I ran against this branch that pass validation with the credential still reaching the image:

* `requirements.txt` containing `-r requirements-private.txt`, credential in the included file. Includes are not followed.
* `constraints.txt` with a credentialed Git URL. Not in this list.

I do not think chasing every filename is the right call, partial coverage is genuinely better than none. But it argues for describing what is actually enforced in the README rather than promising a guarantee.

libs/cli/langgraph_cli/config.py
```diff
@@ -78,6 +83,57 @@ def has_disallowed_build_command_content(command: str) -> bool:
...
+    )
+
+
+def _validate_git_http_url_userinfo(values: Iterable[str]) -> None:
+    """Reject credential-bearing Git HTTP URLs without echoing their values."""
+    if not any(_has_git_http_url_userinfo(value) for value in values):
+        return
+    raise click.UsageError(
+        "Git dependency URLs must not contain credentials or other URL "
+        "userinfo because generated Dockerfiles and image layers can retain "
```
eliornl, line 103: Hiding the value is right. The path is not a secret though, and with several local deps there is nothing to go on.

`_validate_git_http_url_userinfo_files` just below already puts `{path}` in its `OSError` message, so it reads as inconsistent within the same module. Suggest threading the path through and appending `Found in: {path}`.

libs/cli/langgraph_cli/config.py
```diff
@@ -36,6 +37,10 @@
 # This blocks background execution (cmd &) while allowing command
 # chaining (cmd1 && cmd2) which is common in build commands.
 _SINGLE_AMPERSAND_RE = re.compile(r"(?<!&)&(?:&&)*(?!&)")
+_GIT_HTTP_AUTHORITY_RES = (
+    re.compile(r"git\+https?://(?P<authority>[^/\s\"']+)", re.I),
```
eliornl, line 41: This tests for `@` before the first slash rather than for a credential, so these are rejected too:

* `git+https://${GIT_TOKEN}@github.com/org/private.git`
* a commented-out example line, `# example: git+https://user:tok@host/x`

I lean towards the `$` case being fine as-is: dependency strings get interpolated unquoted into a `RUN pip install` line, and `has_disallowed_build_command_content` already blocks `${...}` in build commands as an injection vector. Worth a sentence in the description saying it is deliberate, so nobody has to ask.

libs/cli/langgraph_cli/schemas.py
```diff
@@ -689,6 +689,9 @@ class Config(TypedDict, total=False):
       - "." or "./src" if you have a local Python package
       - str (aka "anthropic") for a PyPI package
       - "git+https://github.com/org/repo.git@main" for a Git-based package
+    Git HTTP URLs must not contain userinfo such as a username or token. For private
```
eliornl, line 693: While you are in here: the `pip_config_file` docstring above still advertises that file for "custom indices, credentials, etc.", and `config.py` does `ADD {pip_config_file} /pipconfig.txt` straight into the image.

So the new README line tells users not to store credentials there while this schema doc still points them at it. Cheap to align in the same PR.

eliornl:
Title: `fix(cli): reject credential-bearing Git dependencies`. `cli` is an allowed scope and is the convention in this repo.

Gate is clean on my end: `lint_package`, `lint_tests`, 345 tests passing in `libs/cli`, and `generate_schema.py` produces no drift.

Specifics are inline on the diff.

eliornl (approved):
Thanks, this covers everything I raised. Verified against `62ecd541`: Node package and lock files are now scanned, the README describes what is actually covered, and errors name the file without echoing the value. `libs/cli` clean locally, 352 passed.

One nit, not blocking: the `validate_config` call site still passes no `source`, so a credential in `langgraph.json` itself gets the generic message. Passing `config_path` would close it.

PR #8543, opened 2026-08-06
fix(cli): remove discontinued local debugger image

jkennedyvz:
## Description

Removes the discontinued local debugger image path from `langgraph up` while retaining configurable hosted Studio links.

- Removes the EOL debugger Compose service and `--debugger-port`.
- Adds `--studio-url` for the hosted/self-hosted Studio origin.
- Adds `--api-url` for the API URL embedded in Studio's `baseUrl` query parameter.
- Keeps `--debugger-base-url` as a hidden deprecated alias for `--api-url` during the compatibility window.
- Extends the API healthcheck startup window so slower graph imports keep using one-second probes.
- Updates the integration runner to match the Compose helper signature.
- Removes the unrelated `uv export` workspace-metadata change from this PR; it can return separately with a reproducible failure and executable export test.

### Compatibility note

`langgraph up --debugger-port ...` now returns `No such option` because the local debugger image is discontinued. Hosted Studio remains available through the printed link and the new URL overrides above.

Of the removed debugger-named Compose tests, only `test_compose_with_api_version_and_debugger` exercised the debugger service. The others did not pass `debugger_port` and duplicated no-debugger coverage.

Documentation follow-ups:

- langchain-ai/docs#5454
- langchain-ai/langchain-skills#59

## Test Plan

- [x] `make format`
- [x] `make lint`
- [x] Studio URL tests: 10 passed
- [x] uv-lock config regression test: 1 passed
- [x] Healthcheck Compose generation: 21 passed
- [ ] Full local `make test`: 342 passed; two unchanged Docker-command tests failed in this local environment (`test_dockerfile_command_with_docker_compose`, `test_build_generate_proper_build_context`)

Made by [Open SWE](https://openswe.vercel.app/agents/3eb05644-444c-5065-83fb-988489ffecd3)


libs/cli/langgraph_cli/uv_lock.py
```diff
@@ -970,6 +970,18 @@ def copy_from_project_root(
             f"{uv_export_project_dir}/uv.lock",
         )
     )
+    for package in plan.install_order:
```
eliornl: What `uv export` failure does this fix?

I tried to reproduce it and couldn't. With only root `pyproject.toml` + `uv.lock` in the export dir, `uv export --package agent --frozen --no-hashes --no-emit-project --no-emit-workspace` succeeds, and the output is byte-identical with and without the member `pyproject.toml` files — on uv 0.5.31, 0.7.22, 0.9.7 and 0.11.32, with both `members = ["apps/*", "libs/*"]` and an explicit member list that included a member outside the install closure.

That isn't proof it's unnecessary: my repro is a 3-member synthetic workspace, and I didn't test path sources pointing outside the workspace, members with `[tool.uv] package = false`, or the uv bundled in the base image. But the description says this metadata is "required for package discovery", and that's the specific claim I can't confirm — could you paste the error you hit?

Related: the new assertions only check Dockerfile text, so they'd pass even if these `COPY` lines were a no-op. If there's a real failure, a test that actually runs `uv export` against a fixture workspace would pin it.

Separately — this is an independent fix from the debugger removal, and the PR title only covers the latter. Probably cleaner as its own PR.

libs/cli/langgraph_cli/cli.py
```diff
@@ -337,20 +320,12 @@ def on_stdout(line: str):
...
-                debugger_base_url_query = (
-                    debugger_base_url or f"http://127.0.0.1:{port}"
-                )
                 set("")
                 sys.stdout.write(
                     f"""Ready!
 - API: http://localhost:{port}
 - Docs: http://localhost:{port}/docs
-- LangGraph Studio: {debugger_origin}/studio/?baseUrl={debugger_base_url_query}
+- LangGraph Studio: https://smith.langchain.com/studio/?baseUrl=http://127.0.0.1:{port}
```
eliornl: With `--debugger-base-url` gone, `up` has no way to influence this link: origin is pinned to `https://smith.langchain.com` and `baseUrl` to `http://127.0.0.1:{port}`.

Both siblings have an override — `dev` takes `--studio-url` (line 684), and `deploy` derives the host from `LANGSMITH_ENDPOINT` via `_smith_dashboard_base_url()` (`deploy.py:661`), including regional prefixes. `up` is now the only one without one.

Two cases that regress: EU / self-hosted LangSmith, where the printed link points at the wrong host; and `langgraph up` on a remote box, where `127.0.0.1` isn't the user's machine.

Is adding `--studio-url` to `up` for parity in scope here? If not, a line in the description saying `up` deliberately has no override would head off the question.

libs/cli/langgraph_cli/cli.py
```diff
@@ -140,17 +140,6 @@
     help="Show more output from the server logs",
 )
 OPT_WATCH = click.option("--watch", is_flag=True, help="Restart on file changes")
-OPT_DEBUGGER_PORT = click.option(
```
eliornl, line 143: These are documented options, so removing them turns `langgraph up --debugger-port 8124` from "ignored" into `Error: No such option: --debugger-port`.

No objection if that's intended — the CLI is pre-1.0 and the image is EOL — but it's worth calling out in the description as breaking, and two first-party places still tell people to pass it:

- `langchain-ai/docs` → `src/langsmith/cli.mdx` lines 805-806, the `langgraph up` options table. (The `--debug-port` / `--wait-for-client` rows at 527/553 are `dev`'s remote-debug flags — unrelated, leave those.)
- `langchain-ai/langchain-skills` → `config/skills/langgraph-cli/SKILL.md` line 99, `langgraph up --debugger-port 8124  # serve debugger UI`

Could you link follow-ups for both?

libs/cli/tests/unit_tests/test_docker.py
```diff
@@ -75,35 +75,7 @@ def test_compose_with_no_debugger_and_custom_db_with_healthcheck():
     assert clean_empty_lines(actual_compose_str) == expected_compose_str
 
 
-def test_compose_with_debugger_and_custom_db():
```
eliornl, line 78: For anyone reading the -100 lines: `test_compose_with_debugger_and_custom_db` never passed `debugger_port`, so it was byte-identical to `test_compose_with_no_debugger_and_custom_db` directly above it. Only `test_compose_with_api_version_and_debugger` (old line 305) actually exercised the debugger path.

So the real coverage loss is one test, not four — worth a line in the description, since the diff looks alarming otherwise.

eliornl (commented):
Three questions and one note. Local gate is clean and all 72 checks are green, so nothing here blocks — but I couldn't reproduce the failure the `uv export` change is described as fixing, and the two removed flags are still documented in two first-party places.

Main ask: either paste the `uv export` error, or split that commit into its own PR — the title only covers the debugger removal.

libs/cli/tests/unit_tests/test_docker.py
jkennedyvz: Added this clarification to the PR description: only `test_compose_with_api_version_and_debugger` exercised the debugger service; the other debugger-named tests duplicated no-debugger coverage.

libs/cli/langgraph_cli/cli.py
jkennedyvz: Called out the breaking `--debugger-port` removal in the PR description and opened the two first-party follow-ups: langchain-ai/docs#5454 and langchain-ai/langchain-skills#59. `--debugger-base-url` remains as a hidden deprecated alias for the new `--api-url` during the compatibility window.

libs/cli/langgraph_cli/cli.py
jkennedyvz: Addressed in 5f479f1: `up` now supports `--studio-url` for the Studio origin and `--api-url` for the API URL embedded in `baseUrl`. The old `--debugger-base-url` remains as a hidden deprecated alias. Values are validated as HTTP(S) URLs and the query parameter is encoded.

libs/cli/langgraph_cli/uv_lock.py
jkennedyvz: Removed the unrelated uv export change from this PR in 33c3edf. I will keep it out unless there is a reproducible failure and an executable `uv export` test for a separate PR.

eliornl: Scratch my earlier comment here, this was load-bearing and I was wrong to ask you to drop it. Sorry for the churn.

The 3.14 leg A integration test passed with these COPY lines (292fd578) and fails without them (5f479f1a). main is green on the same job and only #8546 / #8547 landed there since, both lint-only, so it isn't the base. The failure is this exact command:

```
error: The workspace does not have a member agent: /tmp/uv_export/project
uv export --package agent --frozen ...  exit code: 2
```

PR #8544, opened 2026-08-06
fix: order delta channel replay by task path

### LINKED ISSUE #8382: DeltaChannel replay order diverges from live execution order for parallel-superstep writes, corrupting continued-thread state
### Checked other resources

- [x] I used the GitHub search to find a similar issue and didn't find one (searched "DeltaChannel", "get_delta_channel_history", "delta channel reconstruct", "superstep order", "apply_writes" across open and closed issues/PRs; closest hits were #8039/#8234, which are about `durability="sync"` crash-recovery atomicity, a different bug with a different trigger — no crash is needed to reproduce this one).
- [x] I am sure this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved on the latest commit on `main`.

### Description

`DeltaChannel` (beta, added in #7586) stores only a sentinel in checkpoint blobs and reconstructs its value by replaying ancestor writes through the reducer, via `BaseCheckpointSaver.get_delta_channel_history()`. When more than one parallel task writes to the same `DeltaChannel` in a single superstep, the order those writes get replayed in in `get_delta_channel_history()` does not match the order `apply_writes()` used to produce the value during live execution. The result is silently reordered channel state, both when inspecting `get_state()` and — more seriously — when continuing (re-`invoke()`-ing) the same thread, where the wrongly-reordered replay becomes the new base that subsequent writes build on.

Reducers are documented as needing to be "batching-invariant" (`reducer(reducer(state, xs), ys) == reducer(state, xs + ys)`), but nothing requires them to be *order*-invariant, and the built-in example in the docstring / tests (list accumulation via `extend`/`append`) is manifestly order-sensitive. So this isn't a reducer-authoring mistake, it's the stored replay order not matching the live order.

### Root cause

Live execution deliberately makes superstep write order deterministic and independent of task completion order:

```python
# libs/langgraph/langgraph/pregel/_algo.py:256
# sort tasks on path, to ensure deterministic order for update application
tasks = sorted(tasks, key=lambda t: task_path_str(t.path[:3]))
```

Each task's writes are handed to `apply_writes()` here after this path-sort, so `channel.update(vals)` always sees the same order for a given graph regardless of which parallel node happens to finish first (`libs/langgraph/langgraph/pregel/_runner.py:583/591/603/613` calls `put_writes()` per task as it completes — completion order is not deterministic for genuinely parallel tasks).

But `get_delta_channel_history()`, which reconstructs a `DeltaChannel`'s value for `get_state()` and for resuming a thread, sorts each checkpoint's writes by `(task_id, idx)` instead of path — and this is the documented cross-backend contract, not a one-off slip:

```python
# libs/checkpoint/langgraph/checkpoint/memory/__init__.py:203-205 (InMemorySaver)
for (_task_id, _idx), (tid, ch, serialized, _) in sorted(
    step_writes.items(), reverse=True
):
```

```python
# libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/_delta.py:139-140,155
# Final write order is oldest->newest globally and (task_id, idx) within
# a checkpoint, matching the contract on DeltaChannelHistory.writes
...
ws.sort(key=lambda w: (w[2], w[3]))
```

`task_id` is a hash of the task's path (`_xxhash_str`/`_uuid5_str`, `libs/langgraph/langgraph/pregel/_algo.py:550,616`), not the path itself, so sorting by it bears no relationship to the path-sorted order `apply_writes()` used live. For a single writer per superstep this is invisible (there's nothing to reorder against), which is presumably why the conformance suite (`libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py`) doesn't catch it — every test there does one `aput_writes()` call per step, never two-plus tasks writing the same channel in one step.

### Reproduction

Against `langchain-ai/langgraph @ 49ae27c` (`main`, langgraph==1.2.9, langgraph-checkpoint==4.1.1):

```python
from typing import Annotated


def reducer(current, updates):
    result = list(current)
    for u in updates:
        result.extend(u) if isinstance(u, list) else result.append(u)
    return result


NAMES = ["a", "b", "c", "d", "e", "f", "g", "h"]


def build_graph():
    ch = DeltaChannel(reducer, list, snapshot_frequency=10_000)

    class State(TypedDict):
        items: Annotated[list, ch]

    def make_node(label):
        def node(state):
            return {"items": [label]}
        return node

    builder = StateGraph(State)
    for n in NAMES:
        builder.add_node(n, make_node(n))
        builder.add_edge(START, n)
        builder.add_edge(n, END)
    return builder.compile(checkpointer=InMemorySaver())


# Part 1: get_state() disagrees with the value invoke() actually returned
graph = build_graph()
config = {"configurable": {"thread_id": "t1"}}
live_order = graph.invoke({"items": []}, config)["items"]
reconstructed_order = graph.get_state(config).values["items"]
print("live invoke() order:", live_order)
print("get_state() order:  ", reconstructed_order)

# Part 2: continuing the thread corrupts the already-committed prefix
graph2 = build_graph()
config2 = {"configurable": {"thread_id": "t2"}}
run1 = graph2.invoke({"items": []}, config2)["items"]
run2 = graph2.invoke({"items": []}, config2)["items"]
print("run 1:", run1)
print("run 2:", run2, "  (expected run1 + run1, unchanged)")
```

**Actual output** (captured, unedited — 8 nodes fan out from `START` in parallel, one superstep, each appending its own label):

```
live invoke() order: ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
get_state() order:   ['d', 'a', 'g', 'e', 'c', 'h', 'f', 'b']
run 1: ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
run 2: ['c', 'g', 'f', 'd', 'b', 'a', 'e', 'h', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']   (expected run1 + run1, unchanged)
```

**Expected:** `get_state()` should report the same order `invoke()` actually produced, and continuing a thread should never reorder items that were already committed in a prior run. **Actual:** `get_state()` returns a different, effectively arbitrary permutation each time (it changes between runs of the same script, since `task_id` is a hash that depends on inputs including the checkpoint id), and re-invoking the same thread silently rewrites the first 8 elements of `items` into a different order before appending the new run's 8 elements.

Ran the script 5 times in a row; every run reproduces a mismatch in both parts (the specific permutation varies run to run, the mismatch does not).

### Impact

This only affects `DeltaChannel`, which is explicitly marked beta, but the failure mode is a silent data-integrity issue, not a crash or a warning: any application relying on write order within a `DeltaChannel`-backed field (chat history, audit trail, ordered event log, etc.) will see values reshuffled the moment two or more nodes write to that field in the same superstep and the thread is later inspected via `get_state()`/`get_state_history()` or continued with another `invoke()`/`stream()` call. Sequential graphs (one writer per superstep) are unaffected, which is likely why this hasn't surfaced yet.

### Suggested fix direction

`get_delta_channel_history()` implementations need to recover the same order `apply_writes()` used live, not `(task_id, idx)`. Since `task_id` is currently a one-way hash of the path, either:

- store the task's path (or a path-derived sort key) alongside each write so replay can sort by it the same way `apply_writes()` does, or
- persist an explicit per-checkpoint sequence number for writes at `put_writes()` time that reflects `apply_writes()`'s path-sorted order rather than task completion order.

Both `InMemorySaver.get_delta_channel_history()` (`libs/checkpoint/langgraph/checkpoint/memory/__init__.py`) and the sqlite/postgres `_delta.py` helpers implement the same `(task_id, idx)` contract, so whichever fix is chosen needs to land in all three (and in `checkpoint-conformance`'s spec, which should gain a two-tasks-one-channel-one-superstep test so this can't regress silently again).

### System Info

```
langgraph==1.2.9
langgraph-checkpoint==4.1.1
langgraph-checkpoint-sqlite==3.1.0
langgraph-prebuilt==1.1.0
Python 3.12
Windows 11 (should be OS-independent — the bug is in write ordering logic, not I/O)
Commit under test: 49ae27c2ae983cfb92091b0dea9f7bc37a716479 (main)
```

eliornl:
Fixes langchain-ai/langgraph#8382

`DeltaChannel` rebuilds its value by replaying ancestor writes through the reducer. Every saver ordered a checkpoint's writes by `(task_id, idx)`, but live execution applies them in task-path order: `apply_writes` sorts a super-step's tasks by `task_path_str(task.path[:3])` before calling `channel.update`. `task_id` is a hash of the path, so the two orders are unrelated, and two or more tasks writing one `DeltaChannel` in a single super-step replayed in an arbitrary permutation.

Reducers are only required to be batching-invariant, not order-invariant, so that permutation changes the value. `get_state` disagreed with what `invoke` had returned, and continuing the thread persisted the reordered replay as the base for later writes.

Replay now orders by `(task_path, task_id, idx)`, the same ordering `SELECT_PENDING_SENDS_SQL` already uses for the Send channel.

## Per-backend

| saver | state before | change |
| -- | -- | -- |
| `InMemorySaver` | stored `task_path` | sort key only |
| postgres | stored `task_path` | sort key, plus `task_path` in the stage-2 select |
| sqlite | accepted `task_path` on `put_writes` and dropped it | `writes` gains the column, plus a probe-guarded `ALTER` for existing databases |

Writes stored without a `task_path` sort first within their checkpoint. That is not only a legacy case: a task-less write (graph input) persists `""` today, and first is where live execution applies it.

## Compatibility

Worth calling out because it is asymmetric and not visible in the diff:

| backend | existing threads with parallel writes |
| -- | -- |
| memory, postgres | `task_path` was already on disk, so old threads replay in the corrected order after upgrade |
| sqlite | rows predating the column backfill to `''` and sort equal, so old threads keep their existing order |

That is the intended outcome. Sqlite genuinely has no recoverable order for those rows, so nothing is reinterpreted under a rule that never applied to them. New sqlite writes are ordered correctly from the migration onward.

Downgrade is safe in both directions: older code's `INSERT` omits the column and it carries a `DEFAULT`.

Sqlite has no `ADD COLUMN IF NOT EXISTS`, so `setup()` probes `pragma_table_info` first and treats `duplicate column name` as success, since two connections opening the same file can both pass the probe and only one can win the `ALTER`.

## How it was verified

Against `origin/main` in a clean worktree, copying only the new test files in:

* `tests/test_delta_channel_parallel_order.py` runs on the full `async_checkpointer` matrix (memory, sqlite, postgres in three pool modes). **20 of 25 cases fail on main**; the 5 that pass are the sequential control, which is there to localise this to parallel writers rather than to delta replay in general.
* The two new conformance tests fail on all three backends on main and pass with the fix.
* The sqlite migration is exercised against a database built on the pre-`task_path` schema: the column is added, old rows backfill to `''`, `setup()` is repeatable, and losing the migration race no longer raises.
* Delta read latency is unchanged: 1.157 / 1.105 / 1.127 ms on this branch against 1.101 / 1.128 / 1.137 ms on main, at 1280 replayed writes. Within run-to-run noise.

Suites run locally with Docker: `checkpoint` 156, `checkpoint-sqlite` 124, `checkpoint-postgres` 264, `langgraph` 3493, `prebuilt` 228 (downstream). The only failures are the two `test_remote_graph` cases that need a live server and fail identically on main.

## Notes for review

The pre-existing function-level imports in `test_delta_channel_history.py` are left alone here; they belong with the PLC0415 rollout rather than this fix.

Thanks to @ErenAta16 for the report, the reproduction, and for tracing `task_path` through all three backends' storage, and to @ragnarok268 for the conformance-test design that assigns task ids in the opposite order to their task paths so the test cannot pass by coincidence.

### DIFF
--- a/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
+++ b/libs/checkpoint-conformance/langgraph/checkpoint/conformance/spec/test_delta_channel_history.py
@@ -208,6 +208,93 @@ async def test_history_migration_plain_value_as_seed(
     assert values == [2], f"Expected [2], got {values}"
 
 
+# Task ids used by the ordering tests below. `build_delta_chain` tags its own
+# writes with a `uuid4`, whose hex digits are all <= "f", so "aaaa..." sorts
+# before every fixture task id and "zzzz..." sorts after every one of them.
+# That makes the expected order fully determined rather than dependent on which
+# uuid4 the fixture happened to draw.
+TASK_ID_SORTS_FIRST = "aaaaaaaa-0000-0000-0000-000000000000"
+TASK_ID_SORTS_LAST = "zzzzzzzz-0000-0000-0000-000000000000"
+
+
+async def test_history_orders_parallel_writes_by_task_path(
+    saver: BaseCheckpointSaver,
+) -> None:
+    """Writes from several tasks in one super-step replay in task_path order.
+
+    Live execution sorts a super-step's tasks by `task_path_str(path[:3])`
+    before applying their values, so replay has to recover that order rather
+    than `task_id` order — `task_id` is a hash of the path, so the two
+    disagree, and reducers are only required to be batching-invariant, not
+    order-invariant.
+
+    The two task_ids are assigned so they sort in the *opposite* order from
+    their task_paths. A saver ordering by `(task_id, idx)` therefore returns
+    these writes reversed, rather than passing by happening to agree.
+    """
+    configs = await build_delta_chain(
+        saver,
+        thread_id=str(uuid4()),
+        channel="ch",
+        snapshots_at_steps=[0],
+        total_steps=3,
+    )
+    # The chain is: step 0 snapshot (seed), step 1 write, step 2 write.
+    # `aget_delta_channel_history` walks from the head's parent back to the
+    # seed, so it collects step 1's writes only — step 0 terminates the walk
+    # and step 2 is the head, whose own writes are pending for the next
+    # super-step and excluded. So step 1 is where these writes have to go.
+    step_1, head = configs[1], configs[2]
+    await saver.aput_writes(
+        step_1, [("ch", "second")], TASK_ID_SORTS_FIRST, "~pull, 02"
+    )
+    await saver.aput_writes(step_1, [("ch", "first")], TASK_ID_SORTS_LAST, "~pull, 01")
+
+    result = await saver.aget_delta_channel_history(config=head, channels=["ch"])
+    values = [w[2] for w in result["ch"]["writes"]]
+    # 1 is the fixture's own write at step 1. It carries no task_path, so it
+    # sorts ahead of both writes added above.
+    assert values == [1, "first", "second"], (
+        f"Expected task_path order [1, 'first', 'second'], got {values}. "
+        "Ordering by (task_id, idx) alone yields [1, 'second', 'first']."
+    )
+
+
+async def test_history_orders_pathless_writes_first(
+    saver: BaseCheckpointSaver,
+) -> None:
+    """Writes stored without a task_path sort ahead of path-carrying ones.
+
+    A task-less write (graph input) persists `task_path=""`, as does any row
+    written before a saver recorded the column. `""` precedes every
+    `task_path_str` output because that function prefixes tuples with `~`, so
+    those writes replay first — where live execution applies graph input.
+    """
+    configs = await build_delta_chain(
+        saver,
+        thread_id=str(uuid4()),
+        channel="ch",
+        snapshots_at_steps=[0],
+        total_steps=3,
+    )
+    # Same chain shape as above: step 1 is the only step the walk collects.
+    step_1, head = configs[1], configs[2]
+    # Committed in the opposite order to the one they must replay in, so the
+    # assertion cannot pass on insertion order alone.
+    await saver.aput_writes(
+        step_1, [("ch", "from_node")], TASK_ID_SORTS_FIRST, "~pull, a"
+    )
+    await saver.aput_writes(step_1, [("ch", "from_input")], TASK_ID_SORTS_LAST)
+
+    result = await saver.aget_delta_channel_history(config=head, channels=["ch"])
+    values = [w[2] for w in result["ch"]["writes"]]
+    # Both 1 (the fixture's write) and "from_input" are pathless, so they sort
+    # by task_id among themselves and both precede the path-carrying write.
+    assert values == [1, "from_input", "from_node"], (
+        f"Expected pathless writes first, got {values}"
+    )
+
+
 ALL_DELTA_CHANNEL_HISTORY_TESTS = [
     test_history_returns_writes_oldest_first,
     test_history_seed_is_nearest_snapshot,
@@ -216,6 +303,8 @@ ALL_DELTA_CHANNEL_HISTORY_TESTS = [
     test_history_empty_channels_returns_empty,
     test_history_walk_to_root_no_seed,
     test_history_migration_plain_value_as_seed,
+    test_history_orders_parallel_writes_by_task_path,
+    test_history_orders_pathless_writes_first,
 ]
 
 
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -168,6 +168,7 @@ class _DeltaStage2Row(TypedDict, total=False):
     type: str | None
     blob: bytes | None
     task_id: str | None  # "w" rows only
+    task_path: str | None  # "w" rows only
     idx: int | None  # "w" rows only
     version: str | None  # "b" rows only
 
@@ -272,7 +273,7 @@ def _build_delta_stage2_sql(
         branches.append(
             "SELECT 'w'::text AS _kind, "
             "checkpoint_id, channel, "
-            "type, blob, task_id, idx, NULL::text AS version "
+            "type, blob, task_id, task_path, idx, NULL::text AS version "
             "FROM checkpoint_writes "
             "WHERE thread_id = %s AND checkpoint_ns = %s AND channel = %s "
             "AND checkpoint_id = ANY(%s)"
@@ -280,7 +281,8 @@ def _build_delta_stage2_sql(
     for _ in channels_with_seed:
         branches.append(
             "SELECT 'b'::text AS _kind, NULL::text AS checkpoint_id, channel, "
-            "type, blob, NULL::text AS task_id, NULL::int AS idx, version "
+            "type, blob, NULL::text AS task_id, NULL::text AS task_path, "
+            "NULL::int AS idx, version "
             "FROM checkpoint_blobs "
             "WHERE thread_id = %s AND checkpoint_ns = %s AND channel = %s "
             "AND version = %s"
@@ -426,10 +428,11 @@ class BasePostgresSaver(BaseCheckpointSaver[str]):
         seed blob is sentinel "empty" — in both cases the consumer treats
         absence as "start empty".
         """
-        # writes_by_ch_by_cid[channel][cid] = list of (type, blob, task_id, idx)
-        writes_by_ch_by_cid: dict[str, dict[str, list[tuple[str, bytes, str, int]]]] = {
-            ch: {} for ch in channels
-        }
+        # writes_by_ch_by_cid[channel][cid] = list of
+        # (type, blob, task_id, idx, task_path)
+        writes_by_ch_by_cid: dict[
+            str, dict[str, list[tuple[str, bytes, str, int, str]]]
+        ] = {ch: {} for ch in channels}
         # seed_blob_by_ver[(channel, version)] = (type, blob)
         seed_blob_by_ver: dict[tuple[str, str], tuple[str, bytes]] = {}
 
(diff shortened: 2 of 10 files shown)


ErenAta16 (commented):
I filed #8382, so not a neutral reviewer. Checked the premise against `_algo.py` rather than the description, and both halves hold.

**Live execution really does sort by path, not by id** (`_algo.py:255-256`):

```python
# (we use them for eg. task ids which aren't good for sorting)
tasks = sorted(tasks, key=lambda t: task_path_str(t.path[:3]))
```

That comment is the strongest argument in the PR and it is already in the tree. The codebase knew task ids are not sortable, `apply_writes` acted on it, and the delta-channel replay did not.

**And the two orders genuinely disagree**, because `task_id` is a digest of the path rather than a transform of it (`_algo.py:834-842`):

```python
task_id = task_id_func(          # _xxhash_str, or _uuid5_str for v1 checkpoints
    checkpoint_id_bytes, checkpoint_ns, str(step),
    PUSH, task_path_str(task_path[1]), str(task_path[2]),
)
```

so sorting by `task_id` sorts by hash output. It is not "a different but stable order", it is unrelated to the order the reducer was fed during execution, which is what makes batching-invariance insufficient.

**The tests are the part I would call out.** They cannot pass by coincidence:

> The two task_ids are assigned so they sort in the *opposite* order from their task_paths. A saver ordering by `(task_id, idx)` therefore returns these writes reversed, rather than passing by happening to agree.

That is the difference between a test that pins the fix and one that pins whatever the fixture happened to produce. Pinning the constants so the fixture's own `uuid4` always sorts between them (`aaaa…` / `zzzz…`, since uuid hex digits are all `<= f`) removes the last source of flake, and it is the kind of thing that is obvious only after it bites someone.

Putting it in `checkpoint-conformance` rather than in one saver's test file is also the right call, since the bug is a contract violation and every backend has to satisfy it, not just the one that got patched.

I also verified the ordering claim the second test rests on: `task_path_str` prefixes tuples with `~` (`_algo.py:1414-1415`), and `""` precedes `~`, so pathless writes really do sort ahead of every path-carrying one, which is where live execution applies graph input.

**One question about historical data, and it is the only thing I would want answered before this merges.** Rows written before the `task_path` column existed get `""`, which means they all sort first, among themselves by `task_id`. For a thread checkpointed by an older version, replay therefore keeps the old, wrong order for those writes and only becomes correct for writes recorded after the upgrade.

That may well be the right call, since there is no way to recover a path that was never stored, and inventing one would be worse. But it means a continued thread can replay half in the old order and half in the new, and the failure mode I described in #8382 (the reordered replay becoming the base that later writes build on) still applies to that prefix. If that is the intended semantics, it is worth a sentence in the migration notes, because "upgrade fixes it" and "upgrade fixes it for new writes" are very different promises for anyone with long-lived threads.

Nothing else from me. The diagnosis matches what I reported and the fix is at the right layer.

VANDRANKI (commented):
Community review, not a merge-gate approval. I spent real time on this one since it touches DeltaChannel replay ordering, an area I've dug into before on this repo.

I verified the core premise directly against the live runtime rather than trusting the PR's own comments: `libs/langgraph/langgraph/pregel/_algo.py`'s `apply_writes()` does `tasks = sorted(tasks, key=lambda t: task_path_str(t.path[:3]))` at line 256 before handing each task's writes to `channel.update`. That confirms live execution really does apply a super-step's writes in `task_path` order, not `task_id` order, and `task_id` is a hash of the path so the two orders genuinely disagree for parallel tasks. Reducers are only required to be batching-invariant (per `DeltaChannel`'s own contract), so an order-sensitive reducer (like the PR's own `_append_reducer` test fixture) can reconstruct a different value if replay recovers the wrong order.

I also checked what was and wasn't already fixed. Postgres's `checkpoint_writes` table already has `task_path` (migration at `checkpoint-postgres/.../base.py` line ~90, `ALTER TABLE checkpoint_writes ADD COLUMN IF NOT EXISTS task_path ...`), and the main pending-writes query used for normal checkpoint loading already orders by `task_path, task_id, idx` (line ~123). So the general checkpoint-writes path for postgres was already correct, presumably from an earlier PR. What this PR fixes is narrower and separate: the DeltaChannel-specific history-reconstruction path (`_build_delta_stage2_sql` / `_build_delta_channels_writes_history` in the same file) had not been updated to select or sort by `task_path` at all, so replaying a DeltaChannel's history specifically could still return writes in the wrong order even though the general write-ordering machinery next to it was already fixed. The PR closes that gap consistently: postgres's delta-specific SQL and sort key, sqlite's schema (a new additive migration in `_schema.py`, applied by both the sync and async savers, with a `pragma_table_info` probe plus a caught `duplicate column name` OperationalError to handle two connections racing the same `ALTER`, since sqlite has no `ADD COLUMN IF NOT EXISTS`), and `InMemorySaver`'s in-process sort key all move from `(task_id, idx)` to `(task_path, task_id, idx)`.

I traced the sqlite migration race test (`test_setup_survives_losing_the_migration_race`) by hand: it stubs the "has column" probe to always report false, pre-applies the `ALTER` from a second connection to simulate the winner, then confirms the saver under test survives the resulting `duplicate column name` error rather than raising. That matches the guard in `_schema.py`'s `setup()`, which catches `OperationalError` and only re-raises when the message isn't that specific string.

The parallel-order regression suite (`test_delta_channel_parallel_order.py`) fans 8 named nodes out from `START` into one `DeltaChannel`-backed `items` field in a single super-step, so `task_id` (a uuid4 per task, effectively random relative to node name) and `task_path` (alphabetical by node name here) are very likely to disagree, and asserts both a cold `aget_state` read and multi-run continuation match what `ainvoke` actually returned. The included `sequential=True` control test (one writer per super-step) documents that it already passes on main, which correctly localizes the bug to the parallel-write-to-one-channel case rather than delta replay in general. It runs against the full `async_checkpointer` matrix (memory, sqlite, postgres in three pool modes) per the file's own docstring, which is the right scope for a fix touching three separate saver implementations.

I did not run the suite myself, this is from reading the runtime code and the tests together, but the sort key claim is independently confirmed against `_algo.py` rather than taken on faith, and I did not find a saver where the sort key was updated inconsistently with the others.

PR #8547, opened 2026-08-06
chore: enforce PLC0415 in tests for the remaining packages

eliornl:
Follow-up to #8540, which turned on `PLC0415` (import-outside-top-level) for checkpoint-postgres and checkpoint-sqlite. This does the remaining six packages: checkpoint, checkpoint-conformance, langgraph, prebuilt, cli, sdk-py.

Scoped to tests, per @sydney-runkle's call on #8540: library code is exempted with `per-file-ignores`, since it still has deferred imports nobody has reviewed and mixing that in would make this hard to read.

## What changed

Function-level imports across 56 test files moved to module level. Nine could not move and carry an explicit `# noqa: PLC0415` with a reason:

| File | Why it stays local |
|---|---|
| `libs/langgraph/tests/test_deprecation.py` (4) | the import has to run inside `pytest.warns` for the warning to be observed |
| `libs/langgraph/tests/test_serde_allowlist.py` | try/except guard, skips when langchain_core is absent |
| `libs/langgraph/tests/test_delta_channel_benchmark.py` | optional psycopg probe |
| `libs/checkpoint/tests/test_conformance_delta.py` (3) | protected by a module-level `pytest.importorskip`; hoisting past the guard turns a skip into a collection error |

That last one is the trap: an import moved above `pytest.importorskip` silently defeats the guard. I hit it locally and it turned the skip into a `ModuleNotFoundError` at collection. Every file with an `importorskip` or `except ImportError` was checked by hand for this.

## Verification

`make lint` and `make test` in each of the six:

| Package | Tests |
|---|---|
| checkpoint | 156 passed, 17 skipped |
| checkpoint-conformance | 1 passed |
| langgraph | 1968 passed, 4 skipped |
| prebuilt | 284 passed |
| cli | 336 passed |
| sdk-py | 493 passed |

Also confirmed the rule actually fires: a throwaway test file with a function-level import is flagged in all six packages, and the source exemption holds.

### DIFF
--- a/libs/checkpoint/README.md
+++ b/libs/checkpoint/README.md
@@ -39,12 +39,7 @@ You must pass these when invoking the graph as part of the configurable part of 
 
 ```python
 {"configurable": {"thread_id": "1"}}  # valid config
-{
-    "configurable": {
-        "thread_id": "1",
-        "checkpoint_id": "0c62ca34-ac19-445d-bbb0-5b4984975b2a",
-    }
-}  # also valid config
+{"configurable": {"thread_id": "1", "checkpoint_id": "0c62ca34-ac19-445d-bbb0-5b4984975b2a"}}  # also valid config
 ```
 
 ### Serde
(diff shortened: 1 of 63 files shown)


libs/checkpoint/README.md
```diff
@@ -39,7 +39,12 @@ You must pass these when invoking the graph as part of the configurable part of
 
 ```python
 {"configurable": {"thread_id": "1"}}  # valid config
-{"configurable": {"thread_id": "1", "checkpoint_id": "0c62ca34-ac19-445d-bbb0-5b4984975b2a"}}  # also valid config
+{
```
sydney-runkle: just formatting changes? why?

sydney-runkle (approved):

libs/checkpoint/README.md
eliornl: accidental, `ruff format` rewrites the python blocks in README.md so it got swept in. reverted, it's out of the PR now.

Issue #8550, opened 2026-08-06
SQLite delta history skips parent checkpoints with non-monotonic IDs


lylelllll:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Related Issues / PRs

#6821 covers latest-checkpoint selection in InMemorySaver. #8448 covers a Postgres pagination bug. Neither covers the SQLite delta-history parent walk.

### Reproduction Steps / Example Code (Python)

```python
import asyncio

from langgraph.checkpoint.base import empty_checkpoint
from langgraph.checkpoint.sqlite import SqliteSaver
from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver


def checkpoint(checkpoint_id, values):
    value = empty_checkpoint()
    value["id"] = checkpoint_id
    value["channel_values"] = values
    return value


def run_sync():
    with SqliteSaver.from_conn_string(":memory:") as saver:
        config = {"configurable": {"thread_id": "t", "checkpoint_ns": ""}}
        root = saver.put(config, checkpoint("z-older", {"ch": "seed"}), {}, {})
        saver.put_writes(root, [("ch", "write-root")], "task")
        child = saver.put(root, checkpoint("a-newer", {}), {}, {})
        print(saver.get_delta_channel_history(config=child, channels=["ch"]))


async def run_async():
    async with AsyncSqliteSaver.from_conn_string(":memory:") as saver:
        config = {"configurable": {"thread_id": "t", "checkpoint_ns": ""}}
        root = await saver.aput(
            config, checkpoint("z-older", {"ch": "seed"}), {}, {}
        )
        await saver.aput_writes(root, [("ch", "write-root")], "task")
        child = await saver.aput(root, checkpoint("a-newer", {}), {}, {})
        print(await saver.aget_delta_channel_history(config=child, channels=["ch"]))


run_sync()
asyncio.run(run_async())
```

### Error Message and Stack Trace (if applicable)

```shell
No exception is raised. Both calls silently return:

{'ch': {'writes': []}}

The expected result is:

[...]

PR #8556, opened 2026-08-06
fix(checkpoint-postgres): derive the delta walk cursor once the target loads

### LINKED ISSUE #8448: PostgresSaver: get_delta_channel_history permanently poisons walk cursor when target checkpoint isn't in the first pagination page, silently dropping DeltaChannel history
### Description

`BasePostgresSaver._try_advance_walks` (and its async twin in `aio.py`) permanently poisons a channel's walk cursor to `None` when the target checkpoint hasn't been loaded yet by the current pagination page, causing `get_delta_channel_history` to silently return an empty history (`{"writes": []}`, no `"seed"`) for `DeltaChannel`s on checkpoints that aren't within the first `_DELTA_PAGE_SIZE` (1024) rows of the thread.

### Where

`libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py`, `_try_advance_walks` (~line 392-394):

```python
# First-time entry: cursor starts at the target's parent.
if ch not in walk_cursor_by_ch:
    walk_cursor_by_ch[ch] = parent_of.get(target_id)
```

`get_delta_channel_history` (`libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py`, ~line 476-522) pages the `checkpoints` table **newest-first** in chunks of `_DELTA_PAGE_SIZE = 1024`, starting from `cursor=None`, and calls `_try_advance_walks(checkpoint_id, ...)` after each page, where `checkpoint_id` is the *target* checkpoint being hydrated (which can be any checkpoint in the thread's history, not just the latest — e.g. via `get_state({"configurable": {"checkpoint_id": X}})`, `update_state`, or `get_state_history`).

### Why it's wrong

`walk_cursor_by_ch[ch]` is initialized exactly once, guarded by `if ch not in walk_cursor_by_ch`. On the first page, it's seeded from `parent_of.get(target_id)` — but `parent_of` only contains rows fetched *so far*. If `target_id` is older than the newest 1024 checkpoints in the thread, it isn't in `parent_of` yet, so `.get(target_id)` returns `None`, and the code stores `walk_cursor_by_ch[ch] = None` **as if the target had no parent** (i.e. as if it were the root).

On every subsequent page, the guard `ch not in walk_cursor_by_ch` is now `False` (the key already exists, mapped to `None`), so the cursor is never re-derived even after `target_id`'s real row — and its real parent — is loaded on a later page. The walk for that channel is stuck permanently at `cur_cid = None`; `chain_by_ch[ch]` stays `[]` and the channel is never added to `seeded`. Pagination stops once it reaches the true end of the thread's checkpoints, and the method returns with no `"seed"` and empty `"writes"` for that channel — no exception, no warning.

Downstream, `channels_from_checkpoint` (`libs/langgraph/langgraph/pregel/_checkpoint.py`) does:

```python
history = histories[k]
replay_ch = delta_spec.from_checkpoint(history.get("seed", MISSING))
replay_ch.replay_writes(history["writes"])
```

With `seed` absent and `writes` empty, the channel silently hydrates as empty — e.g. a `messages` `DeltaChannel` reconstructs to `[]` even though the real thread has hundreds of real messages.

### Repro scenario

1. Use `PostgresSaver`/`AsyncPostgresSaver` with a graph whose reducer uses a `DeltaChannel` (e.g. `messages`), on a thread long enough to have accumulated more than 1024 checkpoints after some earlier checkpoint `X` (common in long-running agent loops / supervisor patterns).
2. Call anything that hydrates channels from that older checkpoint — `graph.get_state({"configurable": {"thread_id": t, "checkpoint_id": X}})`, `get_state_history`, or `update_state` targeting `X`.
3. Because `X` isn't within the first paginated 1024-row window, its `DeltaChannel` value (e.g. `messages`) comes back empty instead of the real accumulated history — silent, incorrect state with no error raised.

This class of bug is avoided in the sqlite counterpart (`_delta.py`'s `step_walk_with_row`), which streams starting from `target_id` itself (`checkpoint_id <= ?`) and only marks a channel "started" once the target row is actually observed, rather than overloading `None` to mean both "target not loaded yet" and "target has no parent". The postgres fast path should adopt the same distinction, e.g.:

```python
if ch not in walk_cursor_by_ch:
    if target_id not in parent_of:
        continue  # target row not loaded yet; retry on next page
    walk_cursor_by_ch[ch] = parent_of.get(target_id)
```

### Notes

- Confirmed by reading the current code on `main` (`libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py` / `__init__.py` / `aio.py`).
- `libs/langgraph/tests/test_delta_channel_migration.py` has no test coverage that exercises pagination across multiple stage-1 pages, so this gap wouldn't be caught by the existing test suite.
- Distinct from the other currently-open `DeltaChannel` issues (#8382 replay order, #8384 InMemorySaver first-write drop, #8443 fork replay) — this one is specific to the Postgres paginated fast-path cursor logic.

eliornl:
## Summary

`get_delta_channel_history` on Postgres returns an empty history for any `DeltaChannel` on a
target checkpoint that is not within the first stage-1 pagination page (1024 rows) of the thread.
No exception, no warning: the channel just hydrates empty.

Fixes #8448

## Problem

Stage 1 pages `checkpoints` newest-first from the head of the thread, and after each page
`_try_advance_walks` tries to move every not-yet-seeded channel's walk along the partial
`parent_of` map accumulated so far. The walk starts at the target's parent:

```python
if ch not in walk_cursor_by_ch:
    walk_cursor_by_ch[ch] = parent_of.get(target_id)
```

The target can be any checkpoint in the thread, not just the head, so on the first page
`parent_of` frequently has no row for it yet. `.get` then returns `None`, which is also what a
target with no parent returns, and the two are stored identically. Because the initialisation is
guarded by `ch not in walk_cursor_by_ch`, it never runs again: once the walk is parked at `None`
it stays there even after the target's real row and real parent load on a later page.

The result is an empty chain and no seed. Downstream `channels_from_checkpoint` does

```python
replay_ch = delta_spec.from_checkpoint(history.get("seed", MISSING))
replay_ch.replay_writes(history["writes"])
```

so `get_state`, `get_state_history` and `update_state` against an older checkpoint reconstruct a
`messages` channel as `[]` on a thread with hundreds of real messages.

## Fix

Start the walk only once `target_id` is actually present in `parent_of`, so "the target has not
loaded yet" stops sharing a representation with "the target is a root":

```python
if ch not in walk_cursor_by_ch:
    if target_id not in parent_of:
        continue
    walk_cursor_by_ch[ch] = parent_of[target_id]
```

`_try_advance_walks` is a static method on `BasePostgresSaver`, so `PostgresSaver` and
`AsyncPostgresSaver` are both covered by the one change.

## Why it's safe

`continue` leaves the channel exactly as it was, so a later page retries. The three existing
stop conditions are untouched: a channel that finds its seed still seeds, one that reaches a real
root still parks at `None`, and one waiting on an ancestor still keeps its cursor. Paging still
terminates on a short page, which is what ends the run for a target that really is a root.

## Long-term

The sibling sqlite implementation avoids this class of bug differently, by starting its stage-1
scan at the target (`checkpoint_id <= ?`) instead of at the head. Postgres could adopt the same
bound and would then never fetch a checkpoint newer than the target at all, which looks like the
bigger win on a long thread. It makes the read path depend on ancestors always sorting below their
descendants, though, which sqlite already assumes but the Postgres fast path currently does not.
#8550 now reports that assumption as a bug in sqlite, on the grounds that ancestry is defined by
`parent_checkpoint_id` and the contract does not require ids to be monotonic, so the bound is the
wrong direction to move Postgres in. Paging the full thread and following parent pointers is what
keeps this path correct when ids are not monotonic, and with this fix Postgres returns the right
history for #8550's scenario at every page size.

## Test plan

New `libs/checkpoint-postgres/tests/test_delta_pagination.py`. Page size is monkeypatched rather
than writing 1024+ real checkpoints per case, since the only thing that decides the behaviour is
which page the target lands on.

- [x] `test_async_target_older_than_the_first_page` and its sync twin, parametrised over page
      sizes `[_DELTA_PAGE_SIZE, 3, 2, 1]`. The thread has 8 checkpoints with a snapshot at step 1
      and the target at step 4, so every size at or below 3 leaves the target off the first page.
      The real page size is the control.
- [x] `test_root_target_has_no_history_and_still_terminates` covers the case where a `None` cursor
      is the correct answer, at page size 1 so the paging loop runs the length of the thread.
- [x] 6 of the 9 fail on `main` (`expected a snapshot seed, got '<missing>'`); the 3 that pass are
      the two controls and the root case.
- [x] `make format`, `make lint_package`, `make lint_tests` clean.
- [x] Full `libs/checkpoint-postgres` suite: 273 passed, 3 skipped on both Postgres 15 and 16.

Thanks to @Navneet-Scaler for the report, the mechanism write-up, and the fix in #8453, which this
matches.

### DIFF
--- a/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
+++ b/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py
@@ -435,15 +435,26 @@ class BasePostgresSaver(BaseCheckpointSaver[str]):
           (c) the next ancestor cid isn't in `parent_of` yet (waiting for
               a later page; the cursor stays put).
 
+        A walk that hasn't started yet is a fourth state, distinct from all
+        three: pages arrive newest-first from the head of the thread, so a
+        target deeper than one page isn't in `parent_of` on the early pages
+        and the walk can only begin once its own row lands.
+
         Mutates `chain_by_ch`, `seed_ver_by_ch`, `seed_inline_by_ch`,
         `walk_cursor_by_ch`, and `seeded` in place.
         """
         for i, ch in enumerate(channels):
             if ch in seeded:
                 continue
-            # First-time entry: cursor starts at the target's parent.
+            # First-time entry: cursor starts at the target's parent, but
+            # only once the target's own row has loaded. Reading it earlier
+            # would record `None` for "target not seen yet" the same way it
+            # records `None` for "target is a root", and this guard fires
+            # only once, so the walk would stay stranded for good.
             if ch not in walk_cursor_by_ch:
-                walk_cursor_by_ch[ch] = parent_of.get(target_id)
+                if target_id not in parent_of:
+                    continue
+                walk_cursor_by_ch[ch] = parent_of[target_id]
             cur_cid = walk_cursor_by_ch[ch]
             ch_chain = chain_by_ch[ch]
             hb_i = hb_by_i_by_cid[i]
--- a/libs/checkpoint-postgres/tests/test_delta_pagination.py
+++ b/libs/checkpoint-postgres/tests/test_delta_pagination.py
@@ -1,0 +1,172 @@
+"""Stage-1 pagination for `DeltaChannel` histories on Postgres.
+
+`get_delta_channel_history` pages the `checkpoints` table newest-first in
+chunks of `_DELTA_PAGE_SIZE`, starting from the head of the thread rather than
+from the target checkpoint. The target's own row therefore only lands once
+paging has reached back to it, which for a long thread can be several pages in.
+
+Until then `parent_of` has no entry for it, and reading the walk cursor out of
+that map records `None`, the same value that means "the target is a root". The
+cursor is derived once, so a target older than the first page kept that `None`
+forever: no seed, no writes, and the channel hydrated empty with no error.
+See #8448.
+
+These tests shrink the page size instead of writing 1024+ real checkpoints per
+case. The behaviour under test is "the target is not on the first page", and
+the page the target lands on is the only thing that decides it.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+from uuid import uuid4
+
+import pytest
+from langgraph.checkpoint.base import (
+    Checkpoint,
+    DeltaChannelHistory,
+    empty_checkpoint,
+)
+from langgraph.checkpoint.base.id import uuid6
+from langgraph.checkpoint.serde.types import _DeltaSnapshot
+
+from langgraph.checkpoint.postgres import PostgresSaver
+from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
+from langgraph.checkpoint.postgres.base import _DELTA_PAGE_SIZE
+from tests.conftest import DEFAULT_URI
+
+CHANNEL = "items"
+STEPS = 8
+SEED_STEP = 1
+SEED_VALUE = [10, 20]
+# Deep enough that the smaller page sizes below have to page past it.
+TARGET_STEP = 4
+
+# The real page size holds this whole thread on one page, so it is the control.
+# The rest each leave the target off the first page: there are three
+# checkpoints newer than `TARGET_STEP`, so any page size at or below 3 does it.
+PAGE_SIZES = [_DELTA_PAGE_SIZE, 3, 2, 1]
+
+
+def _step_args(
+    thread_id: str, step: int, parent: dict | None
+) -> tuple[dict, Checkpoint, dict[str, Any]]:
+    """Return the `(config, checkpoint, new_versions)` triple for one step.
+
+    Step `SEED_STEP` stores a snapshot for `CHANNEL`; the rest only bump the
+    channel version, which is what a delta channel does between snapshots.
+    """
+    config: dict = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    if parent is not None:
+        config["configurable"]["checkpoint_id"] = parent["configurable"][
+            "checkpoint_id"
+        ]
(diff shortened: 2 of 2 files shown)


VANDRANKI (commented):

Traced the logic in `libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py`'s `_try_advance_walks`. The bug: `walk_cursor_by_ch[ch] = parent_of.get(target_id)` runs once per channel, guarded by `if ch not in walk_cursor_by_ch`. Pages arrive newest-first from the head of the thread, so if the target checkpoint is deeper than the first page, `target_id` isn't in `parent_of` yet on that first call, and `.get()` silently returns `None`. That `None` is indistinguishable from the legitimate "target is a root, no parent" case, and since the guard only fires once, the cursor gets permanently stuck at `None` and the channel hydrates empty with no error, matching the referenced #8448.

The fix (`if target_id not in parent_of: continue`) correctly defers setting the cursor until the target's real parent pointer is known, and only then distinguishes "not yet seen" from "confirmed root." This is the right fix for the described class of bug.

The new test file (`test_delta_pagination.py`) is well constructed: it shrinks `_DELTA_PAGE_SIZE` via monkeypatch instead of writing 1000+ real checkpoints, parametrizes over page sizes that put the target on different pages, and includes a dedicated root-target case (`test_root_target_has_no_history_and_still_terminates`) to confirm the loop still terminates correctly when `None` really is the right answer.

I traced the logic and the tests by hand but don't have a Postgres instance running to actually execute this test suite, so I can't confirm it passes as written.

PR #8557, opened 2026-08-06
fix(checkpoint-sqlite): walk delta ancestors by parent pointer

### LINKED ISSUE #8550: SQLite delta history skips parent checkpoints with non-monotonic IDs

eliornl:
## Summary

The sqlite delta history silently drops a parent checkpoint whose id sorts above its child's,
losing that parent's stored value and its pending writes. The channel hydrates short with no error.

Fixes #8550

## Problem

Stage 1 walked ancestors with:

```sql
WHERE thread_id = ? AND checkpoint_ns = ? AND checkpoint_id <= ?
ORDER BY checkpoint_id DESC
```

Ancestry is defined by the `parent_checkpoint_id` column. These two predicates add a second,
unstated requirement: that every child's id sorts above its parent's. Real ids are `uuid6` and
happen to satisfy it, but nothing in the contract promises it, and a caller supplying its own ids
or two processes writing one thread under clock skew will not.

When the requirement is violated the parent is excluded from the stream and its seed and writes go
with it. Dropping the range filter alone does not fix it: in `checkpoint_id DESC` order that parent
arrives *before* the target, so the walk streams past it before it has started. A single-pass
ordered stream cannot express this walk.

## Fix

A recursive CTE anchored at the target, following `parent_checkpoint_id`:

```sql
WITH RECURSIVE ancestors(checkpoint_id, parent_checkpoint_id, type, checkpoint) AS (
    SELECT ... FROM checkpoints
    WHERE thread_id = ? AND checkpoint_ns = ? AND checkpoint_id = ?
    UNION ALL
    SELECT c.... FROM checkpoints c JOIN ancestors a
      ON c.checkpoint_id = a.parent_checkpoint_id
    WHERE c.thread_id = ? AND c.checkpoint_ns = ?
)
SELECT checkpoint_id, parent_checkpoint_id, type, checkpoint FROM ancestors
```

Row order no longer carries meaning, so both assumptions go away at once.

## Why it's safe

- **`step_walk_with_row` keeps its shape.** The CTE yields the target first, then its parent, then
  its grandparent, which is exactly the sequence the existing walk already consumed. Its off-path
  and not-yet-started branches stay in place and stay correct.
- **Strictly less IO than before.** The query reads only true ancestors, where the old one read
  every row at or below the target, including every sibling branch. No second query was added.
- **The fast path is asserted equal to `BaseCheckpointSaver`** on the same rows, so the optimised
  and reference implementations are pinned together rather than just both being green.

## The cycle guard

Following pointers can loop where a bounded id scan could not, and a loop is reachable through
`put` alone rather than only by corruption: `put` writes with `INSERT OR REPLACE`, so re-putting an
existing checkpoint id under a descendant's config repoints that checkpoint at its own descendant.
The walk therefore stops on a repeated `checkpoint_id`.

Cycle detection is exact rather than a depth ceiling on purpose. A ceiling has to come from
somewhere: a constant is a magic number that silently truncates a long migrated thread if it is too
low, and counting the thread's checkpoints puts an `O(thread length)` query on every delta read,
which is the opposite of what this path is for. Tracking walked ids costs one set insert per row.

sqlite yields recursive rows lazily, so abandoning the cursor ends the recursion rather than
waiting on it. Verified against a live cycle: the guard fires after 3 rows.

## Note for review

`test_cyclic_parent_chain_terminates` fails by **hanging**, not by asserting, because a regression
means the row stream never ends. The package has no `pytest-timeout` and I did not want to add a
dependency for one test, so the CI job timeout is the backstop. Happy to swap it for a direct unit
test on `step_walk_with_row` if you would rather the failure be loud.

## Postgres

No equivalent change needed. It pages the whole thread and follows parent pointers already, so it
returns the correct history for this scenario. I ran #8550's case against it to confirm. This is
also why #8556 deliberately does not adopt sqlite's old `checkpoint_id <= target` bound.

## Test plan

New `libs/checkpoint-sqlite/tests/test_delta_parent_walk.py`.

- [x] Sync and async, parametrised over both id orders. `parent_id_sorts_above_child` is the bug;
      `parent_id_sorts_below_child` is the control that already passed.
- [x] `test_long_chain_walks_the_whole_thread`: 40 checkpoints with the only stored value at the
      root and ids descending as the chain grows, so id order fights the walk at every step.
- [x] `test_cyclic_parent_chain_terminates`: cycle built through `put` alone.
- [x] 3 of the 6 fail on `main`. The two controls pass, and the cycle test passes on `main` as well
      since the old bounded scan could not loop, so it is a regression guard for the new path
      rather than a bug reproducer.
- [x] `make format`, `make lint_package`, `make lint_tests` clean.
- [x] `libs/checkpoint-sqlite`: 123 passed, 3 skipped, including the delta conformance runner.
- [x] `libs/langgraph -k sqlite`: 634 passed.

Thanks to @lylelllll for the report, the minimal repro, the base-saver comparison that isolated it
to the fast path, and for suggesting the recursive CTE.

### DIFF
--- a/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
+++ b/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/__init__.py
@@ -538,7 +538,10 @@ class SqliteSaver(BaseCheckpointSaver[str]):
         seeded: set[str] = set()
 
         with self.cursor(transaction=False) as cur:
-            cur.execute(DELTA_STAGE1_SQL, (thread_id, checkpoint_ns, checkpoint_id))
+            cur.execute(
+                DELTA_STAGE1_SQL,
+                (thread_id, checkpoint_ns, checkpoint_id, thread_id, checkpoint_ns),
+            )
             for row in cur:
                 cid, parent_cid, type_tag, blob = row
                 if step_walk_with_row(
--- a/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/_delta.py
+++ b/libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/_delta.py
@@ -26,16 +26,38 @@ from typing import Any
 
 from langgraph.checkpoint.base import DeltaChannelHistory, PendingWrite
 
-# Stage 1 streams ancestors of `target_cid` newest-first. The `<=`
-# predicate keeps target itself in the stream so we can read its
-# `parent_checkpoint_id` from the first row without a separate lookup;
-# the caller skips target's own writes/seed (matches the
-# `BaseCheckpointSaver` contract).
+# Stage 1 streams target followed by its ancestors, oldest step last, by
+# following `parent_checkpoint_id` in a recursive CTE. Target is the anchor
+# row, so the caller reads its `parent_checkpoint_id` without a separate
+# lookup and skips its own writes/seed (matches the `BaseCheckpointSaver`
+# contract).
+#
+# Ancestry is defined by `parent_checkpoint_id` alone. An earlier form
+# filtered `checkpoint_id <= target` and ordered by `checkpoint_id DESC`,
+# which additionally required every child's id to sort above its parent's.
+# Nothing in the contract promises that, and a parent sorting above its
+# child was dropped from the stream entirely, silently costing that
+# parent's seed and writes. See #8550.
+#
+# A parent chain can cycle: `put` writes with `INSERT OR REPLACE`, so
+# re-putting an existing checkpoint id under a descendant's config rewrites
+# its parent. The old id-range scan read a finite row set and could not
+# loop; this one can, so `step_walk_with_row` stops on a repeated
+# checkpoint_id. sqlite yields recursive rows lazily, so abandoning the
+# cursor ends the recursion rather than waiting on it.
 DELTA_STAGE1_SQL = (
+    "WITH RECURSIVE ancestors(checkpoint_id, parent_checkpoint_id, type, "
+    "checkpoint) AS ("
     "SELECT checkpoint_id, parent_checkpoint_id, type, checkpoint "
     "FROM checkpoints "
-    "WHERE thread_id = ? AND checkpoint_ns = ? AND checkpoint_id <= ? "
-    "ORDER BY checkpoint_id DESC"
+    "WHERE thread_id = ? AND checkpoint_ns = ? AND checkpoint_id = ? "
+    "UNION ALL "
+    "SELECT c.checkpoint_id, c.parent_checkpoint_id, c.type, c.checkpoint "
+    "FROM checkpoints c JOIN ancestors a "
+    "ON c.checkpoint_id = a.parent_checkpoint_id "
+    "WHERE c.thread_id = ? AND c.checkpoint_ns = ?"
+    ") "
+    "SELECT checkpoint_id, parent_checkpoint_id, type, checkpoint FROM ancestors"
 )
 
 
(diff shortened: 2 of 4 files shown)


VANDRANKI (commented):

Traced the SQL change in libs/checkpoint-sqlite/langgraph/checkpoint/sqlite/_delta.py and the two call sites in __init__.py (sync) and aio.py (async). The old DELTA_STAGE1_SQL filtered checkpoint_id <= target ORDER BY checkpoint_id DESC over a single table scan. The PR's stated bug (#8550, referenced in the comment) is real given that predicate: it defines ancestry by id sort order, not by parent_checkpoint_id, so a parent whose id happens to sort above its child's would never appear in the stream, silently dropping that parent's seed/writes. Real ids are uuid6 and monotonic in practice, but nothing enforces that.

The replacement is a recursive CTE that follows parent_checkpoint_id explicitly: anchor row is the target (WHERE checkpoint_id = ?), recursive step joins c.checkpoint_id = a.parent_checkpoint_id. I checked the placeholder count and order against the Python call: (thread_id, checkpoint_ns, checkpoint_id, thread_id, checkpoint_ns) maps to the anchor's three `?` marks then the recursive step's two - correct.

The new cycle guard in step_walk_with_row (a walked set that returns True on a repeated cid) is necessary and its justification checks out: put() uses INSERT OR REPLACE, so re-putting an existing checkpoint_id under a new parent can point a checkpoint at its own descendant, and a recursive parent-pointer CTE (unlike the old bounded id-range scan) can loop on that. sqlite yields recursive CTE rows lazily, so abandoning the cursor on cycle detection does stop the recursion rather than exhausting it first.

I hand-traced the CTE's row order for a linear chain: each recursive step derives exactly one new row (the parent of the previous step's row) via the join, so for a non-branching ancestor chain the CTE naturally yields target, then parent, then grandparent, etc. - the order step_walk_with_row's cur_cid tracking assumes. I did not run this against an actual sqlite instance, so I'm trusting my read of CTE evaluation order rather than an observed run.

The test file is a good sign: it tests both id-sort directions explicitly (parent id below and above child's, parametrized), a 40-step chain, and a cycle that would hang instead of assert-fail if the guard regressed (relying on the CI job timeout as backstop, which the test's own docstring is upfront about). That's a reasonable trade given sqlite has no per-test timeout plugin here.

PR #8569, opened 2026-08-08
fix(langgraph): detect subgraphs from bytecode instead of source

### LINKED ISSUE #8559: Unecessary source parsing for subgraph detection dominates graph build time
### Checked other resources

### Related Issues / PRs

_No response_

```python
# see also https://github.com/soarez/langgraph-build-bench

import time

import langgraph.pregel._read as _read


class State(TypedDict):
n: int


def node(state):  # a plain node: no subgraph anywhere in sight
return {"n": state["n"] + 1}


def build():
g = StateGraph(State)
for i in range(300):
    g.add_node(f"n{i}", node)
g.add_edge(START, "n0")
for i in range(299):
    g.add_edge(f"n{i}", f"n{i + 1}")
g.add_edge("n299", END)
return g.compile()


def best_of(n=3):
build()
return min((lambda t=time.perf_counter(): (build(), time.perf_counter() - t)[-1])()
           for _ in range(n)) * 1e3


full = best_of()
original = _read.find_subgraph_pregel
_read.find_subgraph_pregel = lambda bound: None   # skip detection entirely
stubbed = best_of()
_read.find_subgraph_pregel = original

print(f"compile(), 300 nodes : {full:7.1f} ms")
print(f"detection stubbed out: {stubbed:7.1f} ms")
print(f"attributable to it   : {100 * (full - stubbed) / full:5.1f}%")
```

```shell

```

### Description

The bulk of the cost of `StateGraph.compile()` is in subgraph detection via
reading source code from disk and parsing it.

`PregelNode.__init__` calls `find_subgraph_pregel` once per node
(`libs/langgraph/langgraph/pregel/_read.py`). For a function-backed node that
reaches `get_function_nonlocals` (`libs/langgraph/langgraph/pregel/_utils.py`),
which runs `inspect.getsource` → `ast.parse` → `inspect.getclosurevars`. Nothing
is cached, so a graph with N nodes does this N times on every build, and
`inspect.getsource` tokenizes past the end of the function until the next
statement at column 0 — so the cost tracks the size of the source file rather
than anything about the graph.

The information we get from parsing the source code is already in-memory
in the function itself:  `__closure__` holds the captured values, `__code__` the
referenced names and the compiled instructions. 

So graph build time could be dramatically reduced by replacing file system access
 and source parsing with checks on objects python already holds available.

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.6.0: Tue Apr 21 20:18:13 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T8132
> Python Version:  3.12.8 (main, Dec  6 2024, 19:42:06) [Clang 18.1.8 ]

Package Information
-------------------
> langchain_core: 1.5.3
> langsmith: 0.10.15
> langchain_protocol: 0.0.18
> langgraph_cli: 0.4.31
> langgraph_sdk: 0.4.2

> deepagents
> deepagents-cli

Other Dependencies
------------------
> anyio: 4.14.2
> click: 8.4.2
> distro: 1.9.0
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.9
> packaging: 26.2
> pathspec: 1.1.1
> pydantic: 2.13.4
> python-dotenv: 1.2.2
> pyyaml: 6.0.3
> requests: 2.34.2
> requests-toolbelt: 1.0.0
> sniffio: 1.3.1
> tenacity: 9.1.4
> typing-extensions: 4.16.0
> uuid-utils: 0.17.0
> websockets: 15.0.1
> xxhash: 3.8.1
> zstandard: 0.25.0

soarez:
Fixes langchain-ai/langgraph#8559

`find_subgraph_pregel` runs once per node at build time and recovers each node function's reachable values with `inspect.getsource` + `ast.parse`; langchain-ai/langgraph#8559 measures that source parsing at 80% of `StateGraph.compile()`. This replaces it with a `dis` walk over `func.__code__`, which reads the same information already in memory.

Closure cells and the globals named in `co_names` supply the values directly, and a walk over the instruction stream recovers the attribute paths the function actually takes. The closure alone cannot express those: a captured `holder` whose graph lives at `holder.graph` is reachable only if something records that `graph` is loaded off `holder`.

Both implementations over-declare — a node can reference a graph it never invokes — and this one over-declares a different set. It no longer reports a graph named only along an attribute path in code the compiler removed; such a path cannot execute, so that entry was always a phantom. Nothing that can actually run stopped being detected. `find_subgraph_pregel` returns the first `PregelProtocol` it finds, so the remaining extra candidates only widen detection.

## Release note

Subgraph auto-detection now reads node functions' bytecode instead of parsing their source. Graph builds with many function-backed nodes are substantially faster, and subgraphs are now detected inside functions with no retrievable source — defined in a REPL or notebook cell, or via `exec` — where detection previously failed silently and returned nothing.

## Performance

Against [langgraph-build-bench](<https://github.com/soarez/langgraph-build-bench>) — 713 nodes, 500 state fields, 264 tools; CPython 3.12.8, Apple silicon:

```
before   build 1412 ms   subgraph detection 1132 ms  (80%)
after    build  286 ms   subgraph detection    4 ms   (2%)
```

**4.9x on total build, \~266x on detection.** The bench prints whole ms; at full precision detection is 1139 ms -> 4.23 ms.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_utils.py
+++ b/libs/langgraph/langgraph/pregel/_utils.py
@@ -1,11 +1,10 @@
 from __future__ import annotations
 
-import ast
-import inspect
+import dis
 import re
-import textwrap
 from collections.abc import Callable, Sequence
 from functools import partial
+from types import CodeType, FunctionType
 from typing import Any
 
 from langchain_core.runnables import (
@@ -17,7 +16,6 @@ from langchain_core.runnables import (
 from langchain_core.runnables.base import RunnableBindingBase
 from langchain_core.runnables.config import run_in_executor
 from langgraph.checkpoint.base import ChannelVersions
-from typing_extensions import override
 
 from langgraph._internal._runnable import RunnableCallable, RunnableSeq
 from langgraph._internal._timeout import sync_timeout_unsupported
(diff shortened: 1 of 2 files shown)


libs/langgraph/langgraph/pregel/_utils.py
```diff
@@ -146,146 +167,52 @@ def get_function_nonlocals(func: Callable) -> list[Any]:
...
+    cells: dict[str, Any] = {}
+    for name, cell in zip(code.co_freevars, func.__closure__ or ()):
+        try:
+            cells[name] = cell.cell_contents
+        except ValueError:
+            continue  # empty cell: a recursive def not yet bound
+
+    # Captured values count whether or not the bytecode still names them: the
+    # compiler drops instructions for unreachable code but keeps the closure,
+    # so `if False: sub.invoke()` is visible here and nowhere else.
```
eliornl: This holds when the captured value *is* the graph, but not when the graph sits one attribute off it. The closure root survives; the `LOAD_ATTR` chain doesn't, and `co_names` keeps the names without their order or depth, so there's nothing left to reconstruct from.

```python
holder = Box(child_graph)

def node(state):
    if False:
        holder.payload.invoke(state)   # the only place `.payload` appears
    return {"log": []}
```

Reverting only this file to `origin/main` and rerunning the same shapes: main resolves it, this branch returns `None`. Same for code after `return`, for `while False:`, and at any chain depth. Checked on 3.10, 3.12, 3.13 and 3.14. A dead-code *root* (`if False: sub.invoke()`, where `sub` is the graph) does still resolve, so this is narrower than it sounds.

To be clear about how much this matters: I ran both implementations end to end through `find_subgraph_pregel` over every live function object after importing `langgraph`, `prebuilt`, `langchain_core` and the test suite, 19,981 to 31,093 functions per version across 3.10-3.14, and got **zero** detection differences. So nothing real is hitting this.

Given that, I don't think it's worth chasing in code. Recovering the path means guessing name sequences, which is exponential in depth, and a source-parse fallback would fire on 54% of the functions I swept, which is exactly the cost this PR removes. But could we drop "superset" from the description and name the limit instead? Something like: *an attribute path that appears only in code the compiler removes is not recoverable from bytecode; a captured value that is itself a graph still resolves, which is the common form.* Better a documented boundary than one someone finds later.

libs/langgraph/tests/test_subgraph_detection.py
```diff
@@ -0,0 +1,266 @@
...
+    def node(state: State) -> Any:
+        def inner() -> Any:
+            return holder.payload.invoke(state)
+
+        return inner()
+
+    return node
+
+
+def unreachable_branch() -> Any:
```
eliornl, line 110: `unreachable_branch` covers the dead-code case where the captured value is the graph, and that one genuinely works. Could we add the counter-case beside it so the boundary is pinned rather than implied?

```python
def unreachable_attribute_chain() -> Any:
    """A dead-code attribute path: unrecoverable, unlike a dead-code root."""
    holder = _Box(_leaf("unreachable_attr"))

    def node(state: State) -> Any:
        if False:
            holder.payload.invoke(state)
        return {"log": []}

    return node
```

listed as `(unreachable_attribute_chain, None)`. As it stands, `unreachable_branch` passing reads as coverage of the whole dead-code class, and it's half of it.

libs/langgraph/langgraph/pregel/_utils.py
```diff
@@ -146,146 +167,52 @@ def get_function_nonlocals(func: Callable) -> list[Any]:
     Returns:
         List[Any]: The nonlocal variables accessed by the function.
```
eliornl: The docstring didn't move with the body. It says "the nonlocal variables accessed by a function", but what comes back now is captured cell values whether or not the bytecode names them, plus globals from `co_names`, plus every intermediate along an attribute path. `Returns: List[Any]:` also restates the annotation, in the old capital-`List` spelling.

The wording predates this PR so it's fair to leave it, but the body underneath is entirely new, so it seemed worth raising while we're here.

libs/langgraph/tests/test_subgraph_detection.py
```diff
@@ -0,0 +1,266 @@
...
+        (no_subgraph, None),
+        (recombined_names, None),
+        (broken_attribute_chain, None),
+        (nested_def_global, None),
+        (graph_in_plain_list, None),
+        (bound_method_self, None),
+    ],
+    ids=lambda value: value.__name__ if callable(value) else str(value),
+)
+def test_detects_subgraph(factory: Any, expected: str | None) -> None:
```
eliornl: nit: `test_detects_subgraph` asserts *non*-detection for 6 of its 16 parameters. `test_subgraph_detection` would cover both halves.

libs/langgraph/langgraph/pregel/_utils.py
soarez: Good catch, reworded.

libs/langgraph/langgraph/pregel/_utils.py
soarez: Thanks for running that sweep, that's useful! I agree that "Superset" is strictly incorrect. Reworded. I must point out though that these subgraphs no longer being identified, when they could not be invoked anyway (since the compiler drops the code path) is actually an improvement, i.e. identifying less of these non-invokable subgraphs (false-positives). Added your test case.

libs/langgraph/tests/test_subgraph_detection.py
soarez: Yup, added.

libs/langgraph/tests/test_subgraph_detection.py
soarez: Renamed to test_subgraph_detection. 17 parameters now, 7 asserting non-detection.

soarez:
🙌 Thanks for the review @eliornl
Comments addressed, PTAL

eliornl:
Thanks for this one, genuinely nice work: clear problem statement, a benchmark to back it, and your reworded code comment is better than what I suggested.

Reopened and added `bypass-issue-check`, so it won't close on you again. Cause was the link format: your description says `Fixes langchain-ai/langgraph#8559` and the check only matches the bare `Fixes #8559`. Worth fixing in the body, it tripped three times today.

One thing still outstanding from [that thread](https://github.com/langchain-ai/langgraph/pull/8569#discussion_r3752288009): the description is unchanged and still says "superset". I think you reworded the code comment instead. Agreed on your phantom point.

soarez:
@eliornl Thank you 🙏  You were right, I have now fixed the link format, and dropped "superset" from the description.

eliornl (approved):
Thanks for the quick turnaround, all four addressed.

Verified the two new detections fail on `main` and pass here, and swept every live function object across 3.10 to 3.14 with no unintended detection changes. The description reads accurately now, and your phantom framing is a better explanation of the change than mine was.

Nice work on this one.

libs/langgraph/langgraph/pregel/_utils.py
```diff
@@ -137,155 +135,87 @@ def validate_timeout_supported(runnable: Runnable, *, name: str) -> None:
...
     Args:
         func: The function to check.
 
     Returns:
-        List[Any]: The nonlocal variables accessed by the function.
+        Every captured cell value, the globals the function names, and each
+        value along an attribute path it loads. Over-approximates: a value can
+        come back without the function reaching it at runtime.
     """
-    try:
```
sydney-runkle, line 149: seems like we used to be pretty defensive here and return `[]` if we hit a broad swath of errors -- should we be doing the same wrap here? if not, why not?

sydney-runkle (approved):
one question -- what happens when multiple subgraphs are used within a single node?

libs/langgraph/langgraph/pregel/_utils.py
soarez: My understanding is that this catch-all was there to catch OSError from getsource, SyntaxError from ast.parse, and maybe also TypeError from getclosurevars.

Looking again, under scrutiny, there is one case that this changes: `getattr` can be called on a class that intercepts attribute access (and runs custom logic), if those interceptors raise anything other than `AttributeError`, then `getattr` does not swallow. `_read.py:191` catches bare `Exception` when calling `find_subgraph_pregel`, that's where errors end up.

If we hit that case, the subgraph search now stops early without a result (potentially failing to find a subgraph in a later candidate). If we want strict equivalence, we need to add the try-except block.
This is how I'd do it: in `libs/langgraph/langgraph/pregel/_utils.py`, inside `find_subgraph_pregel`, replacing the final `elif` branch of the worklist loop:

```python
elif isinstance(c, RunnableCallable):
    fn = c.func if c.func is not None else c.afunc
    if fn is not None:
        # Reading a value's attributes runs user code, which can raise.
        # Let one hostile value end its own branch, not the whole search.
        try:
            candidates.extend(
                nl.__self__ if hasattr(nl, "__self__") else nl
                for nl in get_function_nonlocals(fn)
            )
        except Exception:
            continue
```

It seems like a fringe edge case, but it's not a lot of lines to address it, so I'm happy to add this if you agree.

soarez:
@sydney-runkle thanks for taking a look!

PregelNode supports a list of subgraphs, but that list is used more like an optional than an actual list. I can't trace uses of that subgraph list being specified in the PregelNode constructor, the value is taken from find_subgraph_pregel(self.bound), with at most one subgraph.

When Pregel.get_subgraphs reads PregelNode.subgraphs it also takes at most one subgraph. It is unclear to me why PregelNode.subgraphs is a list - maybe there was an intention to support multiple subgraphs at some point. 
Subgraph state gets addressed in a way that breaks get_state/update_state if there's more than one.

In a comment in #8587 , Elior is suggesting we support this properly in a new separate issue. https://github.com/langchain-ai/langgraph/issues/8587#issuecomment-5257787839 seems there's quite the gap in supporting subgraphs properly. The problem is already in `main`, this PR does not address it.

PR #8596, opened 2026-08-11
fix(langgraph): type undeclared v3 stream projections

nick-hollon-lc:
`stream_events(version="v3")` returns `GraphRunStream`, which declares only the four always-registered native projections. Native projections are attached from a registry, so any other one is invisible to type checkers and raises `attr-defined`. These overloads returned `Any` before 1.2.10 (#8389), so the errors are new.

- Declare the opt-in native projections this package ships (`updates`, `custom`, `checkpoints`, `debug`, `tasks`).
- Add `__getattr__ -> StreamChannel[Any]` to both run streams for projections declared outside this package (`tool_calls` from langgraph-prebuilt, `subagents` from langchain, user transformers). The body only raises `AttributeError`, now listing the registered projections; it reads the mux from `__dict__` so a miss before `__init__` can't recurse.

Projections owned by other distributions are deliberately not annotated here — that would bind this package's typing surface to another package's private module. Precise types for those need a downstream subclass, which composes with this change since declared attributes take precedence over `__getattr__`.

**Tradeoff:** a misspelled projection name now type-checks and fails at runtime instead of at type-check time.

## Verified downstream

mypy, local langgraph swapped in:

| | released 1.2.11 | this branch |
|---|---|---|
| langchain `test_agent_streaming.py` | 9 × `tool_calls` | clean |
| langchain `test_subagent_transformer.py` | 5 × `subagents` | clean |
| langchain `test_transformers.py` | 4 × `tool_calls` | clean |
| deepagents `test_deep_agent_streaming.py` | 6 × `subagents` | clean |

Test suites against this branch: langchain 22 passed, deepagents 8 passed.

Note: langchain's `test_transformers.py` has 4 stale `# type: ignore[attr-defined]` on `run._mux` lines that are already unused before this change — a separate cleanup on that side.

### DIFF
--- a/libs/langgraph/langgraph/stream/run_stream.py
+++ b/libs/langgraph/langgraph/stream/run_stream.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterator, Mapping
 from types import MappingProxyType, TracebackType
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, NoReturn
 
 from langchain_core._api import beta
 
@@ -33,6 +33,21 @@ async def _adrive_until_done(pump: Callable[[], Awaitable[bool]]) -> None:
         pass
 
 
+def _missing_projection(run: object, name: str) -> NoReturn:
+    """Raise for a projection name that was never registered.
+
+    Reads the mux out of `__dict__` because `__getattr__` also fires before
+    `__init__` assigns `_mux` and for dunder probes, where `run._mux` would
+    recurse until the stack is exhausted.
+    """
+    mux = run.__dict__.get("_mux")
+    registered = sorted(mux.native_keys) if mux is not None else []
+    raise AttributeError(
+        f"{type(run).__name__!r} object has no attribute {name!r} "
+        f"(registered projections: {', '.join(registered) or 'none'})"
+    )
+
+
 @beta(message="The v3 streaming protocol on Pregel is experimental.")
 class GraphRunStream:
     """Sync run stream with caller-driven pumping.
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):

Traced `_missing_projection` and the new `__getattr__` on both `GraphRunStream` and `AsyncGraphRunStream`. The design is careful about a real footgun: `__getattr__` reads `run.__dict__.get("_mux")` instead of `run._mux`, specifically to avoid infinite recursion. If it used `self._mux` directly and `_mux` weren't set yet (e.g. a partially constructed instance from `__new__`, or a `hasattr`/`copy`/pickle probe before `__init__` runs), that attribute access would itself trigger `__getattr__` again, recursing until the stack overflows. Reading straight from `__dict__` sidesteps that entirely, and there's a dedicated test for exactly this scenario (`test_getattr_fallback_does_not_recurse_before_init`, using `GraphRunStream.__new__` directly and checking `copy.copy` doesn't blow up).

The other test, `test_unregistered_projection_raises_attribute_error`, confirms `__getattr__` doesn't shadow the real `setattr` loop that attaches actually-registered native projections, and that the error message lists what is registered so a typo is diagnosable from the traceback alone. That's a nice touch for debuggability.

The typing rationale in the comments is honest about its own limitation: `__getattr__` type-checks any attribute name, including a genuine typo, so the cost of getting IDE/mypy support for dynamically-registered projections is that a misspelling won't be caught until runtime. That's a real tradeoff, but it's the same tradeoff every dynamic-attribute pattern makes, and the runtime error message mitigates it reasonably well.

I did not dig into whether other consumers of `GraphRunStream`/`AsyncGraphRunStream` rely on `AttributeError` vs some other exception type for missing projections elsewhere in the codebase, but within the diff itself this looks correct and well tested.

ccurme (approved):

PR #8624, opened 2026-08-15
fix(checkpoint): include state path in serialization error

### LINKED ISSUE #8606: Improve checkpoint serialization errors by including the state path of unserializable values
### Checked other resources

### Related Issues / PRs

#5054  , #8522 , #4956 

```python
from typing import Any, Optional


class ErrorClass:
    def __init__(self):
        self.mssg = "An error occurred"


class dummyState(BaseModel):
    text: str
    error_var: Optional[Any] = None

    class ConfigDict:
        arbitrary_types_allowed = True


state = dummyState(
    text="Hello, world!",
    error_var=ErrorClass()
)

serializer = JsonPlusSerializer()

try:
    serializer.dumps_typed(state)
    print("Serialization successful")
except Exception as e:
    print(f"Serialization failed: {type(e).__name__}: {e}")
```

```shell
Serialization failed: TypeError: Type is not msgpack serializable: dummyState
```

## Problem

When LangGraph checkpoint serialization encounters a value that cannot be serialized by the configured serializer, the resulting error identifies the type of the unsupported value, but not where that value exists within the graph state.

```text
TypeError: Type is not msgpack serializable: {state_name}
```

This is hard to debug in practice for a specific reason: node execution,completes and logs normally, so there's nothing in the immediate output pointing back to the cause. The failure only surfaces later, when the checkpointer tries to persist the resulting state — by which point the connection between "this node ran" and "this field is the problem" isn't visible anywhere. With deeply nested state (dicts of Pydantic models, lists of dataclasses, etc.), tracking down the offending field means manually inspecting the entire state object by hand.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.26200
> Python Version:  3.11.7 | packaged by Anaconda, Inc. | (main, Dec 15 2023, 18:05:47) [MSC v.1916 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 1.5.3
> langsmith: 0.10.17
> langchain_protocol: 0.0.18
> langgraph_sdk: 0.4.2

> deepagents
> deepagents-cli

Other Dependencies
------------------
> anyio: 4.14.2
> distro: 1.9.0
> httpx: 0.28.1
> jsonpatch: 1.33
> orjson: 3.11.9
> packaging: 26.3
> pydantic: 2.13.4
> pyyaml: 6.0.3
> requests: 2.34.2
> requests-toolbelt: 1.0.0
> sniffio: 1.3.1
> tenacity: 9.1.4
> typing-extensions: 4.16.0
> uuid-utils: 0.17.0
> websockets: 15.0.1
> xxhash: 3.8.1
> zstandard: 0.25.0

Taranum01:
Fixes #8606

When `JsonPlusSerializer.dumps_typed` can't encode a value with the default msgpack hook, the previous error message identified only the offending type, not where it sat in the graph state. For deeply nested state (dicts of Pydantic models, lists of dataclasses, etc.) users had to manually walk the state object to find the culprit — and the failure surfaces only at checkpoint time, well after the node that produced the bad value has finished logging.

After the new failure-only walk (`_find_unsupported_path`):
- Top-level value: `TypeError: ... (at $)`
- Dict key: `TypeError: ... (at $\"foo\"`)`
- List index: `TypeError: ... (at $[1])`
- Nested structures: `TypeError: ... (at $\"outer\"\.level1[1]\"level2\"\"bad\"`)`
- Pydantic v2 fields: `TypeError: ... (at $.bad)`

The walk dives into Pydantic v2 `model_fields`, dataclass fields, dicts, and sequences. It returns an empty string when no specific path can be pinned down (e.g. the unsupported value lives behind a non-iterable opaque container), in which case we re-raise the original `ormsgpack.MsgpackEncodeError` so callers can still see the bare type.

Adds six tests in `libs/checkpoint/tests/test_jsonplus.py`:
- top-level path,
- dict key path,
- list index path,
- deeply nested mixed structure,
- Pydantic v2 field path,
- primitives (negative case).

All 105 existing tests in the file still pass.

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -268,6 +268,14 @@ class JsonPlusSerializer(SerializerProtocol):
             except ormsgpack.MsgpackEncodeError as exc:
                 if self.pickle_fallback:
                     return "pickle", pickle.dumps(obj)
+                # Re-raise with a JSONPath-style location so the user can
+                # find the offending value in deeply nested state without
+                # manual inspection. Regression for #8606.
+                path = _find_unsupported_path(obj)
+                if path:
+                    raise TypeError(
+                        f"{exc} (at {path})"
+                    ) from exc
                 raise exc
 
     def loads_typed(self, data: tuple[str, bytes]) -> Any:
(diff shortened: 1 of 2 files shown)


Taranum01:
Reopening. The check-issue-link CI requires the PR author to be assigned to the linked issue; that requires collaborator permissions. Please assign @Taranum01 to #8606 if the fix is acceptable, or merge via a fork.

PR #8631, opened 2026-08-16
fix(checkpoint): widen Store.put()/aput() value type to Mapping[str, Any]

### LINKED ISSUE #8616: TypedDict values are rejected by type checkers for Store.put()/aput() despite being valid at runtime
### Checked other resources

### Related Issues / PRs

_No response_

```python
from typing import TypedDict

class Prefs(TypedDict):
    theme: str

store = InMemoryStore()
value: Prefs = {"theme": "dark"}
store.put(("users", "123"), "prefs", value)
```

```shell
repro.py:9: error: Argument 3 to "put" of "BaseStore" has incompatible type "Prefs"; expected "dict[str, Any]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

### Description

* I'm trying to use the `langgraph` library's `Store.put()`/`aput()` to store a `TypedDict` value (a well-typed, string-keyed mapping matching the store's own requirements).
* I expect the call to type-check cleanly with mypy, since a `TypedDict` is exactly the kind of typed, JSON-serializable payload the store expects.
* Instead, mypy rejects it with the error shown below, because `value` is typed as `dict[str, Any]`, which requires full mutability (arbitrary keys can be added/removed/overwritten) that a `TypedDict` doesn't support. This breaks type-checking in CI for any caller passing a `TypedDict` to `store.put()`/`aput()`, despite the code working correctly at runtime.

Proposed fix: widen the `value` parameter on `BaseStore.put()`/`aput()` (and `PutOp.value`) from `dict[str, Any]` to `collections.abc.Mapping[str, Any]`. This is a compatible widening (plain dicts still satisfy `Mapping[str, Any]`) so it doesn't change behavior or break existing callers.

### System Info

langgraph-checkpoint: 4.2.0
Python 3.14.5, Darwin

MrinalCom:
## Summary

Fixes #8616.

`BaseStore.put()`/`aput()` and `PutOp.value` are typed as `dict[str, Any]`, which rejects a well-formed `TypedDict` under mypy -- a `TypedDict` is not a `dict[str, Any]` (it lacks arbitrary key add/remove) -- even though the call works fine at runtime:

```python
class Prefs(TypedDict):
    theme: str

value: Prefs = {"theme": "dark"}
store.put(("users", "123"), "prefs", value)
# mypy: error: Argument 3 to "put" of "BaseStore" has incompatible type "Prefs"; expected "dict[str, Any]"
```

This widens `value` to `Mapping[str, Any]`, which both `dict` and `TypedDict` satisfy, per the fix proposed in the issue.

## Why this is safe

I traced every place `PutOp.value` / `Store.put value` flows to before touching anything, rather than widening the signature blind:

- `InMemoryStore._apply_put_ops` (`store/memory/__init__.py`) constructs `Item(value=op.value, ...)` directly. `Item.value` stays `dict[str, Any]` (out of scope -- `Item` is a *returned* value, not a caller input, so the original issue doesn't apply there). This one call site needed `cast(dict, op.value)` to keep type-checking clean.
- The Postgres and SQLite store backends (`checkpoint-postgres`, `checkpoint-sqlite`) don't override `put`/`aput` at all -- they only implement `batch`/`abatch`, and both already do `cast(dict, op.value)` at their own serialization call sites (`Jsonb(cast(dict, op.value))`, `orjson.dumps(cast(dict, op.value))`). Both needed **no changes**.
- `batch.py` (the batching wrapper) only forwards `PutOp` through, no direct field access.

So the fix is exactly two files: the type signatures in `store/base/__init__.py`, and one `cast` in `store/memory/__init__.py` to match the pattern the other two backends already use.

## Verification

- Added `test_put_accepts_typed_dict` to `tests/test_store.py`, exercising the issue's exact repro against both `put`/`get` and `aput`/`aget`.
- Confirmed the fix against **mypy** specifically, since that's what the issue was reported against (this repo's own CI uses `ty`, which doesn't flag this particular variance -- so a `ty`-only check would have missed it):
  - Before fix: `error: Argument 3 to "put" of "BaseStore" has incompatible type "Prefs"; expected "dict[str, Any]"  [arg-type]` (byte-for-byte the error from the issue)
  - After fix: `Success: no issues found in 1 source file`
- `uv run ty check langgraph tests`: all checks pass (both before and after -- confirms no regression from `ty`'s perspective either)
- `uv run pytest tests/`: **156 passed, 17 skipped** (17 skips are pre-existing, environment-gated -- e.g. real Redis)
- `uv run ruff check` / `ruff format --diff` / `ruff check --select I` / `codespell`: all clean

## Test plan

- [x] New regression test added and passing (`test_put_accepts_typed_dict`)
- [x] Full `libs/checkpoint` test suite passes with no regressions (156 passed, 17 pre-existing skips)
- [x] Reproduced the exact reported mypy error pre-fix, confirmed resolved post-fix
- [x] `ty check`, `ruff check`, `ruff format`, import sort, and `codespell` all clean

### DIFF
--- a/libs/checkpoint/langgraph/store/base/__init__.py
+++ b/libs/checkpoint/langgraph/store/base/__init__.py
@@ -12,7 +12,7 @@ Core types:
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from collections.abc import Iterable
+from collections.abc import Iterable, Mapping
 from datetime import datetime
 from typing import (
     Any,
@@ -473,7 +473,7 @@ class PutOp(NamedTuple):
         the full path would effectively be `"documents/user123/report1"`
     """
 
-    value: dict[str, Any] | None
+    value: Mapping[str, Any] | None
     """The data to store, or `None` to mark the item for deletion.
 
     The value must be a dictionary with string keys and JSON-serializable values.
@@ -857,7 +857,7 @@ class BaseStore(ABC):
         self,
         namespace: tuple[str, ...],
         key: str,
-        value: dict[str, Any],
+        value: Mapping[str, Any],
         index: Literal[False] | list[str] | None = None,
         *,
         ttl: float | None | NotProvided = NOT_PROVIDED,
@@ -1110,7 +1110,7 @@ class BaseStore(ABC):
         self,
         namespace: tuple[str, ...],
         key: str,
-        value: dict[str, Any],
+        value: Mapping[str, Any],
         index: Literal[False] | list[str] | None = None,
         *,
         ttl: float | None | NotProvided = NOT_PROVIDED,
--- a/libs/checkpoint/langgraph/store/memory/__init__.py
+++ b/libs/checkpoint/langgraph/store/memory/__init__.py
@@ -109,7 +109,7 @@ from collections import defaultdict
 from collections.abc import Iterable
 from datetime import datetime, timezone
 from importlib import util
-from typing import Any
+from typing import Any, cast
 
 from langchain_core.embeddings import Embeddings
 
@@ -408,7 +408,7 @@ class InMemoryStore(BaseStore):
                 self._vectors[namespace].pop(key, None)
             else:
                 self._data[namespace][key] = Item(
-                    value=op.value,
+                    value=cast(dict, op.value),
                     key=key,
                     namespace=namespace,
                     created_at=datetime.now(timezone.utc),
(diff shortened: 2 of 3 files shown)


MrinalCom:
Withdrawing this -- didn't see that @navarra-lisandro (the issue author) already had #8617 open with the same fix, submitted a few days before this one. Sorry for the duplicate noise; deferring to their PR. No need to assign me to #8616, this can stay closed.

PR #8692, opened 2026-08-23
fix(checkpoint): serialize Fraction/complex and preserve dict subclasses

### LINKED ISSUE #8184: Checkpoint serialization downcasts dict subclasses (defaultdict/Counter/OrderedDict) to plain dict, losing default_factory

### LINKED ISSUE #8185: Checkpoint serialization rejects fractions.Fraction and complex (TypeError) though Decimal is supported

Shivani767:
## Summary
- **#8185**: Add `fractions.Fraction` and `complex` handlers to `JsonPlusSerializer._msgpack_default`, using positional constructor args for lossless round-trip (same ext pattern as other stdlib numerics). Register both in `SAFE_MSGPACK_TYPES` for strict mode.
- **#8184**: Enable `OPT_PASSTHROUGH_SUBCLASS` so `defaultdict` / `Counter` / `OrderedDict` reach the encoder instead of being silently downcast to plain `dict`. Each gets an explicit ext handler; `defaultdict` also preserves encodable `default_factory` values (builtins). Unencodable factories warn and fall back to plain dict. Native-base subclass fallbacks (e.g. numpy scalars → `.item()`) prevent regressions from passthrough.

## Test plan
- [x] `pytest libs/checkpoint/tests/test_jsonplus.py` — 106/106
- [x] `make test` in `libs/checkpoint` — 163 passed
- [x] `make format` / `make lint` in `libs/checkpoint`
- [x] Fraction / complex round-trip + strict-mode tests (#8185)
- [x] defaultdict / Counter / OrderedDict type-preservation tests (#8184)
- [x] numpy scalar in dict round-trip (passthrough fallback)
- [x] unencodable defaultdict factory warns and downcasts

Fixes #8184
Fixes #8185

Made with [Cursor](https://cursor.com)

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/_msgpack.py
@@ -30,10 +30,15 @@ SAFE_MSGPACK_TYPES: frozenset[tuple[str, ...]] = frozenset(
         ("uuid", "UUID"),
         # numeric
         ("decimal", "Decimal"),
+        ("fractions", "Fraction"),
+        ("builtins", "complex"),
         # collections
         ("builtins", "set"),
         ("builtins", "frozenset"),
         ("collections", "deque"),
+        ("collections", "Counter"),
+        ("collections", "OrderedDict"),
+        ("collections", "defaultdict"),
         # ip addresses
         ("ipaddress", "IPv4Address"),
         ("ipaddress", "IPv4Interface"),
--- a/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
+++ b/libs/checkpoint/langgraph/checkpoint/serde/jsonplus.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import copy
 import dataclasses
 import decimal
+import fractions
 import importlib
 import json
 import logging
@@ -10,7 +11,7 @@ import pathlib
 import pickle
 import re
 import sys
-from collections import deque
+from collections import Counter, OrderedDict, defaultdict, deque
 from collections.abc import Callable, Iterable, Sequence
 from datetime import date, datetime, time, timedelta, timezone
 from enum import Enum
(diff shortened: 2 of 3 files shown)


Shivani767:
Requested assignment on #8184 and #8185 — PR should reopen automatically once assigned.

Issue #8753, opened 2026-08-29
Cache namespace collapses to "__dynamic__" for non-function callables, so different graphs sharing a cache return each other's results


vasilisnasopoulos:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangGraph rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangGraph (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Related Issues / PRs

#8009 is a collision on the argument side of the key (`default_cache_key`). This one is on the namespace side and is independent of it.

### Reproduction Steps / Example Code (Python)

```python
from typing import TypedDict
from langgraph.graph import StateGraph, START, END
from langgraph.types import CachePolicy
from langgraph.cache.memory import InMemoryCache

class S(TypedDict):
    out: str

shared = InMemoryCache()

class HandlerA:
    def __call__(self, s: S) -> S: return {"out": "from graph A"}

class HandlerB:
    def __call__(self, s: S) -> S: return {"out": "from graph B"}

def build(handler):
    g = StateGraph(S)
    g.add_node("process", handler, cache_policy=CachePolicy())
    g.add_edge(START, "process")
    g.add_edge("process", END)
    return g.compile(cache=shared)

A, B = build(HandlerA()), build(HandlerB())

print(A.invoke({"out": "x"}))   # {'out': 'from graph A'}
print(B.invoke({"out": "x"}))   # {'out': 'from graph A'}   <-- B's node never ran
```

### Error Message and Stack Trace (if applicable)

No error. The second graph returns the first graph's value.

### Description

Two separate graphs, each with a node of the same name, sharing one cache: the second graph silently receives the first graph's result. Its node body never executes.

**Where it comes from.** In `libs/langgraph/langgraph/pregel/_algo.py`:

```python
cache_key = CacheKey(
    (CACHE_NS_WRITES, (identifier(proc) or "__dynamic__"), name),
    xxh3_128_hexdigest(...),
    cache_policy.ttl,
)
```

[...]

PR #8760, opened 2026-08-30
test: add cache callable-identity regression matrix

### LINKED ISSUE #8761: Add tests-only regression matrix for cache callable-identity collisions
## Related bug

This is a **coverage-only tracking issue** for #8753, not a duplicate bug report and not a production-fix proposal.

## Proposed scope

Add a tests-only regression matrix at the public `StateGraph.add_node()` boundary covering:

- distinct top-level named functions as the non-colliding sync/async control;
- callable instances;
- lambdas;
- `functools.partial` bindings;
- closures with different captured values;
- bound methods from different instances;
- `RunnableCallable` wrappers with different bound kwargs;
- distinct named `async def` nodes.

Known collision rows use `xfail(strict=True)`: the suite remains green before a source fix, but every repaired row becomes a fatal `XPASS` until its marker is removed and it becomes an ordinary permanent regression.

## Boundary

- Tests only; no production code changes.
- Does not select namespace isolation versus conservative cache opt-out.
- Kept separate from @Sehlani042's source-change scope in #8753.
- Closing this tracking issue must not close the underlying bug #8753.

## Existing draft

Draft PR #8760 contains one added test file and one commit. Repository automation closed it because external PRs require a closing link to an issue assigned to the PR author.

Validated against exact `main@11ee185999b86bfea2d8c0e69cef9a5e37acf686`:

```text
Python 3.11: 2 passed, 13 xfailed
Python 3.12: 2 passed, 13 xfailed
Ruff check: pass
Ruff format --check: pass
Unexpected passes: 0
Errors: 0
```

Could a maintainer approve this bounded tests-only scope and assign it to @safal207? Once assigned, the repository workflow can reopen #8760 automatically.

safal207:
Fixes #8761

Related bug: #8753

## Summary

Adds a **tests-only** regression matrix for #8753 at the public `StateGraph.add_node()` boundary.

The matrix covers:

- distinct top-level named functions as the non-colliding sync/async control;
- callable instances;
- lambdas;
- `functools.partial` bindings;
- closures with different captured values;
- bound methods from different instances;
- `RunnableCallable` wrappers with different bound kwargs;
- distinct named `async def` nodes.

Known collision rows use `xfail(strict=True)`. This keeps the suite green before a source fix lands, while making an unexpected `XPASS` fatal so each repaired row must be converted into an ordinary permanent regression.

## Scope boundary

- No production code is changed.
- This PR does not choose between namespace isolation and conservative cache opt-out.
- It is intentionally separate from @Sehlani042's source-change scope discussed in #8753.
- Merging this PR closes only the coverage tracking issue #8761, not the underlying production bug #8753.

## Validation

Validated against exact `main@11ee185999b86bfea2d8c0e69cef9a5e37acf686`:

```text
Python 3.11: 2 passed, 13 xfailed
Python 3.12: 2 passed, 13 xfailed
Ruff check: pass
Ruff format --check: pass
Unexpected passes: 0
Errors: 0
```

The two passing rows are the named-function controls; the 13 strict xfails are the currently reproduced callable-identity collisions.

### DIFF
--- a/libs/langgraph/tests/test_cache_callable_identity.py
+++ b/libs/langgraph/tests/test_cache_callable_identity.py
@@ -1,0 +1,192 @@
+"""Regression matrix for cache callable-identity collisions.
+
+These tests cover langchain-ai/langgraph#8753 without selecting a production
+policy. Known collision rows are strict xfails so the suite stays green before
+a source fix lands; once a fix makes a row pass, XPASS is intentionally fatal
+until the marker is removed and the row becomes an ordinary regression.
+"""
+
+from __future__ import annotations
+
+import functools
+from collections.abc import Callable
+from typing import Any
+
+import pytest
+from langgraph.cache.memory import InMemoryCache
+from typing_extensions import TypedDict
+
+from langgraph._internal._runnable import RunnableCallable
+from langgraph.graph import END, START, StateGraph
+from langgraph.types import CachePolicy
+
+
+class State(TypedDict):
+    out: str
+
+
+NodePairFactory = Callable[[], tuple[Any, Any]]
+
+KNOWN_CACHE_IDENTITY_COLLISION = pytest.mark.xfail(
+    strict=True,
+    reason="Known cache callable-identity collision tracked by #8753",
+)
+
+
+def _named_a(_: State) -> State:
+    return {"out": "A"}
+
+
+def _named_b(_: State) -> State:
+    return {"out": "B"}
+
+
+async def _async_named_a(_: State) -> State:
+    return {"out": "A"}
+
+
+async def _async_named_b(_: State) -> State:
+    return {"out": "B"}
+
+
+def _tagged(_: State, *, tag: str) -> State:
+    return {"out": tag}
+
+
+async def _atagged(_: State, *, tag: str) -> State:
+    return {"out": tag}
+
+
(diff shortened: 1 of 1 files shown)


vasilisnasopoulos:
@cwlbraa @jkennedyvz — a small procedural ask rather than a technical one, since the bot points at maintainers here.

This PR was auto-closed for the missing assignment, not for anything about its content. It is tests only: one added file, no production changes, and it deliberately does not choose between the two fixes under discussion in #8753.

The reason it can land ahead of any fix is the way it is marked. The known-colliding rows are `xfail(strict=True)`, so the suite is green today and every row that a later fix repairs turns into a fatal `XPASS` until someone removes its marker. Validated on `main@11ee185`: 2 passed, 13 xfailed on both 3.11 and 3.12 — the two passing rows are the named-function controls, which are what make the other thirteen mean anything.

The bot's own note says reopening or dropping the `missing-issue-link` label is enough. Failing that, assigning @safal207 to #8761 reopens it automatically.

I filed the underlying bug (#8753) and have no stake in this PR beyond wanting the matrix to exist: @safal207 measured most of these rows independently on current `main`, and @Sehlani042 is holding the source change separately so the two do not collide. Whichever policy you eventually prefer, the coverage is the part that makes it checkable.

PR #8819, opened 2026-09-04
feat: Send to same subgraph in parallel with persistent memory

nfcampos:
https://claude.ai/code/artifact/befd728a-68be-470a-9bce-7efa2e25fd7f

langchain/deepagents should then be modified to pass this new send arg in tool calls

### DIFF
--- a/libs/langgraph/langgraph/pregel/_algo.py
+++ b/libs/langgraph/langgraph/pregel/_algo.py
@@ -1341,7 +1341,7 @@ def _scratchpad(
         resume=task_resume_write,
         get_null_resume=get_null_resume,
         # subgraph
-        subgraph_counter=LazyAtomicCounter(),
+        subgraph_counter=LazyAtomicCounters(),
     )
 
 
@@ -1437,6 +1437,29 @@ class LazyAtomicCounter:
                 if self._counter is None:
                     self._counter = itertools.count(0).__next__
         return self._counter()
+
+
+class LazyAtomicCounters:
+    """Lazily created per-key counters; each key counts from 0.
+
+    Used to number repeated subgraph invocations under the same base
+    namespace within one task."""
+
+    __slots__ = ("_counters",)
+
+    _counters: dict[str, Callable[[], int]] | None
+
+    def __init__(self) -> None:
+        self._counters = None
+
+    def __call__(self, key: str) -> int:
+        with LAZY_ATOMIC_COUNTER_LOCK:
+            if self._counters is None:
+                self._counters = {}
+            counter = self._counters.get(key)
+            if counter is None:
+                counter = self._counters[key] = itertools.count(0).__next__
+            return counter()
 
 
 def sanitize_untracked_values_in_send(
--- a/libs/langgraph/langgraph/pregel/_algo.py
+++ b/libs/langgraph/langgraph/pregel/_algo.py
@@ -45,6 +45,7 @@ from langgraph._internal._constants import (
     CONFIG_KEY_RUNTIME,
     CONFIG_KEY_SCRATCHPAD,
     CONFIG_KEY_SEND,
+    CONFIG_KEY_SUBGRAPH_KEY,
     CONFIG_KEY_TASK_ID,
     CONFIG_KEY_THREAD_ID,
     ERROR,
@@ -68,7 +69,7 @@ from langgraph.channels.base import BaseChannel
 from langgraph.channels.topic import Topic
 from langgraph.channels.untracked_value import UntrackedValue
 from langgraph.constants import TAG_HIDDEN
-from langgraph.errors import NodeError
+from langgraph.errors import InvalidUpdateError, NodeError
 from langgraph.managed.base import ManagedValueMapping
 from langgraph.pregel._call import get_runnable_for_task, identifier
 from langgraph.pregel._io import read_channels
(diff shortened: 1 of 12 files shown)


libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -441,7 +442,20 @@ def prepare_next_tasks(
...
+        seen_keys: set[tuple[str, str]] = set()
+        for idx, packet in enumerate(tasks_channel.get()):
+            if for_execution and isinstance(packet, Send) and packet.key is not None:
+                # Keyed packets derive their task id from the key, so two of
+                # them for one node in one step would be one task. Refuse the
+                # step instead of silently dropping one (or, if they invoke a
+                # subgraph, running one memory concurrently). This runs
+                # wherever tasks are prepared, so it holds across processes.
+                if (packet.node, packet.key) in seen_keys:
+                    raise InvalidUpdateError(
```
eliornl, line 454: i tried this with a checkpointer. two Sends with the same key, the step is already saved, then this raises. after that everything on the thread raises the same error: invoke(None), invoke with new input, get_state, and update_state with any as_node. i compared with two nodes writing one LastValue in a step: there get_state also raises, but update_state gets you out and invoke(None) works after it. here there is no way out through the public api.

the check itself is right. a suggestion: do it in apply_writes where the Sends land in TASKS, same place the LastValue error comes from. i tried that locally, update_state recovers and the tests in this file still pass.

libs/langgraph/tests/test_subgraph_key.py
```diff
@@ -0,0 +1,647 @@
...
+        config = {"configurable": {"thread_id": str(uuid4())}}
+        # two threads, same first checkpoint id is impossible, so compare within one step instead
+        graph.invoke(
+            {"messages": [ai(*((("record", {"name": n}, f"id_{n}")) for n in order))]},
+            config,
+        )
+        tasks = graph.get_state_history(config)
+        step0 = next(s for s in tasks if s.metadata.get("step") == 0)
+        ids = {t.name: t.id for t in step0.tasks}
+        assert set(ids) == {"tools"} or len(step0.tasks) == 2
```
eliornl, line 558: `set(ids) == {"tools"}` is always true here, both tasks are named tools. i removed `key=` from make_send_graph and this test still passes, so it does not check what the name says. the checkpoint id is part of the task id, so comparing across the two threads cannot work. maybe assert against the id built with the same inputs, or drop the test?

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -871,6 +900,37 @@ def _first(
             )
         )
 
+        if is_resuming and self.is_nested:
```
eliornl, line 903: this also changes existing checkpointer=True subgraphs, not only keyed ones. i checked, test_stateful_subgraph_new_turn_inside_resumed_task fails on main, the subgraph returns stale state after a resume. good fix, worth a line in the description since people with stateful subgraphs today get different behavior.

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -323,21 +328,45 @@ def __init__(
...
+            # (True).
+            if not isinstance(subgraph_key, str) or not subgraph_key:
+                raise ValueError(f"'{CONFIG_KEY_SUBGRAPH_KEY}' must be a non-empty str")
+            if NS_SEP in subgraph_key:
+                raise ValueError(
+                    f"'{NS_SEP}' is a reserved character and is not allowed in "
+                    f"'{CONFIG_KEY_SUBGRAPH_KEY}'"
                 )
+            ns = NS_SEP.join(
+                (recast_checkpoint_ns(ns, keep_keys=True), f"{NS_END}{subgraph_key}")
```
eliornl, line 352: i ran stream_events(version="v3") on a keyed subgraph. the lifecycle started event comes out as graph_name=None and trigger_call_id='call_a': `_parse_ns_segment` in stream/transformers.py splits the last segment on ':' and reads the key as a task id. the unkeyed version of the same graph gives graph_name='call'. also `run.subgraphs` yields no handle for the keyed instance, since its namespace is two segments deep and the transformer only tracks direct children. the Platform API falls back to the same split when upstream sends no name, so it would show an empty name there too. probably a small change in the transformer to recognise a `:key` segment, plus the paired platform change like the deepagents one.

libs/langgraph/langgraph/pregel/_algo.py
```diff
@@ -982,18 +996,21 @@ def prepare_push_task_send(
...
         )
+        send_key: str | None = getattr(packet, "key", None)
         task_id = task_id_func(
             checkpoint_id_bytes,
             checkpoint_ns,
             str(step),
             packet.node,
             PUSH,
-            str(idx),
+            send_key if send_key is not None else str(idx),
```
eliornl, line 1013: a keyed Send with key "0" and an unkeyed Send at position 0 to the same node get the same task id. i tried it, only one of the two ran, no error. the duplicate check above does not see it since it only tracks keyed packets. probably nobody keys with digits, but the drop is silent. a suggestion: give keyed ids a different shape, e.g. `f"key:{send_key}"`. i checked locally that both tasks run with that.

libs/langgraph/langgraph/pregel/_algo.py
eliornl: resolved in bf8559ba, the check moved to `apply_writes` and the test now shows `update_state` recovering the thread.

libs/langgraph/tests/test_subgraph_key.py
eliornl: resolved in bf8559ba, the test now computes the expected ids from the same inputs the engine uses, and i checked it fails 2/2 without `key=`.

libs/langgraph/langgraph/pregel/_loop.py
eliornl: resolved in bf8559ba for the library side, lifecycle events name the frame and carry `subgraph_key`, and `run.subgraphs` yields the handle. the paired Platform change is still separate.

libs/langgraph/langgraph/pregel/_algo.py
eliornl: resolved in bf8559ba, keyed ids are prefixed with `:` and i checked that key "0" and position 0 both run.

PR #8838, opened 2026-09-07
feat: Queue/steer graphs while running 

nfcampos:
https://claude.ai/code/artifact/101b8a3e-ede7-4e77-9c21-482a928d7f98

### DIFF
--- a/libs/checkpoint/langgraph/checkpoint/base/__init__.py
+++ b/libs/checkpoint/langgraph/checkpoint/base/__init__.py
@@ -39,13 +39,14 @@ logger = logging.getLogger(__name__)
 class CheckpointMetadata(TypedDict, total=False):
     """Metadata associated with a checkpoint."""
 
-    source: Literal["input", "loop", "update", "fork"]
+    source: Literal["input", "loop", "update", "fork", "queue"]
     """The source of the checkpoint.
 
     - `"input"`: The checkpoint was created from an input to invoke/stream/batch.
     - `"loop"`: The checkpoint was created from inside the pregel loop.
     - `"update"`: The checkpoint was created from a manual state update.
     - `"fork"`: The checkpoint was created as a copy of another checkpoint.
+    - `"queue"`: The checkpoint is a queued state update, see `Pregel.queue_state`.
     """
     step: int
     """The step number of the checkpoint.
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -118,6 +118,18 @@ from langgraph.pregel._io import (
     read_channels,
 )
 from langgraph.pregel._messages import ensure_message_ids
+from langgraph.pregel._queue import (
+    CHECKPOINT_META_QUEUE_CONSUMED,
+    QueueApplyError,
+    QueueItem,
+    aack,
+    aack_after,
+    ack,
+    ack_after,
+    alist_pending,
+    list_pending,
+    queue_item_writes,
+)
 from langgraph.pregel._read import PregelNode
 from langgraph.pregel._utils import get_new_channel_versions, is_xxh3_128_hexdigest
 from langgraph.pregel.debug import (
@@ -134,6 +146,7 @@ from langgraph.types import (
     Durability,
     Interrupt,
     PregelExecutableTask,
+    PregelTask,
     RetryPolicy,
     Send,
     StreamMode,
@@ -264,6 +277,11 @@ class PregelLoop:
     ]
     control: RunControl | None
     tasks: dict[str, PregelExecutableTask]
+    # queued state updates, see `Pregel.queue_state`
+    _queue_read_fut: Any  # in-flight read of the pending items, if any
+    _queue_consumed: list[QueueItem]  # applied, awaiting the next checkpoint save
+    _queue_stale_ids: set[str]  # listed as consumed by the checkpoint resumed from
+    _queue_applied_ids: set[str]  # applied in this run; a read may still list them
     output: None | dict[str, Any] | Any = None
     updated_channels: set[str] | None = None
     _graph_lifecycle_events: deque[GraphLifecycleEvent]
@@ -372,6 +390,10 @@ class PregelLoop:
         self.prev_checkpoint_config = None
         runtime = self.config[CONF].get(CONFIG_KEY_RUNTIME)
         self.control = runtime.control if isinstance(runtime, Runtime) else None
+        self._queue_read_fut = None
+        self._queue_consumed = []
+        self._queue_stale_ids = set()
+        self._queue_applied_ids = set()
 
     def _push_graph_lifecycle_event(
         self,
@@ -716,6 +738,8 @@ class PregelLoop:
         self.is_replaying = False
         # save checkpoint
         self._put_checkpoint({"source": "loop"})
+        # read the queue in parallel with the write, per durability mode
+        self._schedule_queue_read()
         # after execution, check if we should interrupt
         if self.interrupt_after and should_interrupt(
             self.checkpoint, self.interrupt_after, self.tasks.values()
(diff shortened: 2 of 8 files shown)


libs/langgraph/langgraph/pregel/main.py
```diff
@@ -2512,6 +2534,154 @@ async def aperform_superstep(
...
+        checkpointer: BaseCheckpointSaver | None = ensure_config(config)[CONF].get(
+            CONFIG_KEY_CHECKPOINTER, self.checkpointer
+        )
+        if isinstance(checkpointer, BaseCheckpointSaver):
+            checkpointer = self._apply_checkpointer_allowlist(checkpointer)
+        if not checkpointer:
+            raise ValueError("No checkpointer set")
+        if (
+            checkpoint_ns := config[CONF].get(CONFIG_KEY_CHECKPOINT_NS, "")
+        ) and CONFIG_KEY_CHECKPOINTER not in config[CONF]:
```
eliornl, line 2638: found why `test_subgraph_is_addressed_by_namespace` is flaky. in `queue_state`, the check for "should i recurse into the child graph" looks at whether the checkpointer key is missing from the config. but that key can already be there, it is when you take the config from `get_state(subgraphs=True)` while the child is running. when that happens we skip the recursion and validate `steer="c"` against the parent instead of the child, and the parent has no `c`. that's the `Node 'c' not found`. whether the key is there depends on timing, so the test passes sometimes and fails sometimes.

i reproduced it locally (about 1 in 3 runs fail) and confirmed from the traceback that `self` was the parent graph with the checkpointer already in the config. i also wrote a version of the test that adds the checkpointer key on purpose, and it fails every time, so it's easy to reproduce.

a suggestion: use a dedicated flag for "already dispatched to the child" on the recursive call instead of relying on the checkpointer key. `get_state` can get away with that check because it only reads by namespace, but here we also check `steer` against `self.nodes`, so it needs to actually reach the child. same thing in the async version. with that change the test passed 20/20 for me, including a 2 level nested case.

libs/langgraph/langgraph/pregel/_loop.py
```diff
@@ -716,6 +738,8 @@ def after_tick(self) -> None:
         self.is_replaying = False
         # save checkpoint
         self._put_checkpoint({"source": "loop"})
+        # read the queue in parallel with the write, per durability mode
+        self._schedule_queue_read()
```
eliornl, line 742: a question about cost. as far as i can see, every run with a checkpointer now does a `list()` on the queue namespace at every superstep in sync mode, even if `queue_state` is never used.

i checked how much it costs. on postgres it hits the primary key, so an empty queue is basically free, and the read runs in parallel with the checkpoint write, so it only adds time when the read is slower than the write. so probably fine, and worth a line in the description since it's the first thing people will ask.

one idea, not sure it's worth it: a compile-time opt-in on the graph, so only graphs that want steering pay the read. the downside is a graph that didn't opt in can't be steered, so maybe you prefer it always on. up to you.

libs/langgraph/tests/test_pregel.py
```diff
@@ -5696,7 +5696,16 @@ def graph(state: dict) -> dict:
                     "step": 0,
                 },
                 "next": [],
-                "parent_config": None,
+                # under durability="exit" the final checkpoint is written at
+                # the boundary where the run finishes, so the event reports
+                # its real parent, the checkpoint the run resumed from
+                "parent_config": {
```
eliornl, line 5702: small note on this. the expected `parent_config` changed from `None` to a real parent. i checked when this happens: only for exit-mode runs that resume from an existing checkpoint (like after an interrupt). fresh threads behave the same as before. so this looks like a fix, not a regression, since before the last checkpoint had no parent at all.

but the Platform API exposes this as `parent_checkpoint_id`, so it does change what users see. i'd suggest mentioning it in the description instead of leaving it only in the test diff.

libs/langgraph/langgraph/pregel/main.py
```diff
@@ -2512,6 +2534,154 @@ async def aperform_superstep(
...
+        self,
+        config: RunnableConfig,
+        values: dict[str, Any] | Any,
+        *,
+        steer: str | None = None,
+    ) -> str:
+        """Queue a state update for the thread's run to apply at a superstep
+        boundary. Safe to call whether or not a run is in flight.
+
+        The update is durable when this returns and is applied exactly once,
```
eliornl, line 2608: small docstring thing on "applied exactly once". that holds under the same one-run-per-thread assumption the rest of the engine has (Platform enforces it). since this is a new public API people will read on its own, maybe worth saying "exactly once per run" or noting the assumption, so nobody reads it as a cross-process guarantee. not a bug, just precision.

libs/checkpoint/langgraph/checkpoint/base/__init__.py
```diff
@@ -39,13 +39,14 @@
 class CheckpointMetadata(TypedDict, total=False):
     """Metadata associated with a checkpoint."""
 
-    source: Literal["input", "loop", "update", "fork"]
+    source: Literal["input", "loop", "update", "fork", "queue"]
```
eliornl, line 42: did you keep the new metadata keys (`steer`, `consumed`, `accepted_at`, `queue_consumed`) out of `CheckpointMetadata` on purpose, to keep them internal? if so makes sense, maybe a short comment. if not, adding them (it's `total=False`) would let us drop the `type: ignore`s.

libs/langgraph/langgraph/pregel/main.py
eliornl: resolved in bb5f4540, the recursive call now uses a dedicated dispatch flag. i re-ran the test 15/15 green.

libs/langgraph/langgraph/pregel/main.py
eliornl: resolved in bb5f4540, the docstring now says exactly once per run and names the one-run-per-thread assumption.

libs/checkpoint/langgraph/checkpoint/base/__init__.py
eliornl: resolved in bb5f4540, the keys live on a private TypedDict and the per-checkpoint list is `__queue_consumed`, which `get_checkpoint_metadata` never copies from config.

libs/langgraph/langgraph/pregel/_loop.py
eliornl: covered in the description now, under "How the loop reads the queue": one read per superstep under sync, one in flight under async, primary key hit on postgres, only adds latency when the read is slower than the checkpoint write.

libs/langgraph/tests/test_pregel.py
eliornl: covered in the description now, under "Behavior changes and compatibility": the final streamed checkpoint event reports its real parent on a resumed exit-mode thread, stored history unchanged.

Nakagawa-master (commented):
The durable queue shape is useful, but I think it is missing one time-validity boundary for **steering / operator intent**.

A queued item is validated against the thread's current state when accepted, then may remain pending until a later boundary — and an idle thread explicitly consumes nothing until a future run. At apply time the reducer is checked again, which catches structural/state incompatibility, but it does not answer a different question:

\`\`\`text
was valid when queued
!=
is still the current intended update when it finally applies
\`\`\`

For example:

\`\`\`text
t0: operator queues steer/update U for node c
t1: run stays idle for hours
t2: operator changes their mind / a newer control-plane decision supersedes U
t3: next run starts
t4: U is still accepted from the durable queue and applies exactly once
\`\`\`

Nothing has to violate a reducer for U to be stale. The same applies mid-run if a queued control survives longer than the assumption under which it was accepted.

The returned update id is a good primitive, but I don't see a corresponding cancellation/revocation or conditional-apply mechanism in this PR.

A bounded API could support one or more of:

- \`cancel_queued_state(config, update_id)\` / async equivalent;
- optional \`expires_at\` / TTL for control-plane steering;
- optional compare-and-set such as \`expected_checkpoint_id\` / state version, so the item applies only if the thread is still descended from the state against which it was accepted;
- a documented supersession key for "latest operator intent wins" use cases.

I would not make every queued data update expire by default. The key is to let callers distinguish durable data from time-sensitive steering/authority.

A regression shape:

\`\`\`text
queue U against checkpoint A
→ U is visible in StateSnapshot.queued
→ revoke U (or advance beyond its expected state)
→ later run reaches the matching node
→ U produces zero state effect
→ queue history records revoked/expired rather than applied
\`\`\`

This is separate from the PR's existing consume-time reducer revalidation: a stale instruction can remain perfectly reducer-valid.

Public source relation: this framing is informed by Nakagawa Master's current-authority / temporal-validity work:
https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/CURRENT_AUTHORITY_REUSE_KIT.md

The concrete finding here is scoped to \`queue_state\`'s durable accept-now/apply-later contract; it does not imply ordinary durable data should require human reapproval.

PR #8868, opened 2026-09-09
feat(langgraph): restrict stream_mode="messages" to specific state keys (#6798)

### LINKED ISSUE #6798: messages stream mode: allow configuring which state key(s) to stream
### Privileged issue

- [x] I am a LangGraph maintainer, or was asked directly by a LangGraph maintainer to create an issue here.

## Problem

`stream_mode="messages"` streams **all** `BaseMessage` instances from **all** state fields. There's no way to restrict it to a specific key.

This is a problem when state has multiple message lists (e.g. user-facing `messages` + internal `agent_messages`). Internal reasoning gets leaked to the frontend with no server-side way to prevent it.

## Reproduction

```python
from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
from typing import Annotated, Sequence
import operator

class State(BaseModel):
    messages: Annotated[Sequence[BaseMessage], operator.add]
    agent_messages: Annotated[Sequence[BaseMessage], operator.add]

def public_node(state):
    return {"messages": [AIMessage(content="Public response")]}

def internal_node(state):
    return {"agent_messages": [AIMessage(content="SECRET internal reasoning")]}

graph = StateGraph(State)
graph.add_node("public", public_node)
graph.add_node("internal", internal_node)
graph.set_entry_point("public")
graph.add_edge("public", "internal")
graph.add_edge("internal", END)
app = graph.compile()

for msg, meta in app.stream(
    {"messages": [HumanMessage(content="hi")], "agent_messages": []},
    stream_mode="messages",
):
    print(msg.content)
# Output includes BOTH "Public response" AND "SECRET internal reasoning"
```

## Root Cause

`StreamMessagesHandler._find_and_emit_messages()` in `langgraph/pregel/_messages.py` extracts all top-level state field values and emits every `BaseMessage` found. No `messages_key` parameter exists to scope it.

https://github.com/langchain-ai/langgraph/blob/630bd9ab953da68e9a00c46e8245176a76c697f7/libs/langgraph/langgraph/pregel/_messages.py#L12

## Proposal

Allow specifying which state key(s) participate in `messages` streaming — similar to how `ToolNode` already accepts `messages_key`:

```python
app.stream(input, stream_mode="messages", messages_key="messages")
# or at compile time
graph.compile(messages_key="messages")
```

## Workarounds

- Tag internal messages with `is_hidden` metadata + filter via ASGI middleware (works but fragile)
- Client-side filtering by node name (data still reaches the frontend)

AaronG18:
## Summary
Closes #6798.

`stream_mode="messages"` currently streams **every** `BaseMessage` found in the state, so an internal reasoning channel (e.g. `agent_messages`) is leaked to the client with no server-side way to prevent it.

This adds a `configurable["__pregel_stream_messages_keys"]` option. When set, only the listed state keys are scanned for messages to stream. When unset, behaviour is exactly as before (covered by a backwards-compat test).

## How it works
- `_state_values(obj, keys=None)` now filters top-level fields when `keys` is given (dict / BaseModel / dataclass).
- `StreamMessagesHandler` / `StreamMessagesHandlerV2` accept `state_keys` and use it when scanning node output for messages.
- Both `Pregel.stream()` and `Pregel.astream()` read `CONFIG_KEY_STREAM_MESSAGES_KEYS` from the run config and pass it to the handler.

Server-side usage (so the leak is preventable without touching client code):
```python
graph.stream(
    input,
    stream_mode="messages",
    config={"configurable": {"__pregel_stream_messages_keys": ["messages"]}},
)
```

`astream()` honours the same option:
```python
async for chunk in graph.astream(
    input,
    stream_mode="messages",
    config={"configurable": {"__pregel_stream_messages_keys": ["messages"]}},
):
    ...
```

## Tests
`tests/test_messages_stream_keys.py` — 7 cases covering both the sync and async paths:
- default streams every message field (sync + async)
- restricted to `["messages"]` — internal `agent_messages` not streamed (sync + async)
- can select the other key instead (sync)
- unknown key streams nothing (sync + async)

The restricted / unknown-key async cases fail without the `astream()` wiring, so both paths are locked in.

## Note on #6878
@mrutunjay-kinagi's PR #6878 implemented a similar idea but has been untouched since 2026-03-24 and currently reports `mergeable_state=dirty`. This PR takes a smaller, config-driven approach with no public signature change. Happy to rework it onto #6878's `messages_key=` kwarg, or to close mine, if that API shape is preferred.

### DIFF
--- a/libs/langgraph/langgraph/_internal/_constants.py
+++ b/libs/langgraph/langgraph/_internal/_constants.py
@@ -72,6 +72,9 @@ CONFIG_KEY_RUNTIME = sys.intern("__pregel_runtime")
 CONFIG_KEY_RESUME_MAP = sys.intern("__pregel_resume_map")
 # holds a mapping of task ns -> resume value for resuming tasks
 CONFIG_KEY_STREAM_MESSAGES_V2 = sys.intern("__pregel_stream_messages_v2")
+# Restrict `stream_mode="messages"` to these state keys, so internal message
+# fields are not streamed to the client (#6798).
+CONFIG_KEY_STREAM_MESSAGES_KEYS = sys.intern("__pregel_stream_messages_keys")
 # when True, attach StreamMessagesHandlerV2 so content-block (v2) events
 # flow through stream_mode="messages"; set by StreamingHandler only.
 CONFIG_KEY_NODE_ERROR = sys.intern("__pregel_node_error")
--- a/libs/langgraph/langgraph/pregel/_messages.py
+++ b/libs/langgraph/langgraph/pregel/_messages.py
@@ -35,14 +35,27 @@ T = TypeVar("T")
 Meta = tuple[tuple[str, ...], dict[str, Any]]
 
 
-def _state_values(obj: Any) -> Sequence[Any]:
-    """Extract top-level field values from a state object (dict, BaseModel, or dataclass)."""
+def _state_values(obj: Any, keys: Sequence[str] | None = None) -> Sequence[Any]:
+    """Extract top-level field values from a state object (dict, BaseModel, or dataclass).
+
+    If `keys` is provided, only those fields are returned. This lets
+    `stream_mode="messages"` be restricted to specific state keys, so internal
+    message fields are not leaked to the client (#6798).
+    """
     if isinstance(obj, dict):
-        return list(obj.values())
+        if keys is None:
+            return list(obj.values())
+        return [obj[k] for k in keys if k in obj]
     elif isinstance(obj, BaseModel):
-        return [getattr(obj, k) for k in type(obj).model_fields]
+        field_names = set(type(obj).model_fields)
+        if keys is None:
+            return [getattr(obj, k) for k in type(obj).model_fields]
+        return [getattr(obj, k) for k in keys if k in field_names]
     elif is_dataclass(obj) and not isinstance(obj, type):
-        return [getattr(obj, f.name) for f in fields(obj)]
+        names = {f.name for f in fields(obj)}
+        if keys is None:
+            return [getattr(obj, f.name) for f in fields(obj)]
+        return [getattr(obj, k) for k in keys if k in names]
     return ()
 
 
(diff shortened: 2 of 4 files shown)


AaronG18:
Hi @niilooy - I re-QA'd my own change here and found a real bug, so I fixed it and force-pushed. Sorry for the churn on an open PR.

**The bug:** the key restriction was only wired into `Pregel.stream()`. `Pregel.astream()` never received `state_keys`, so `stream_mode="messages"` silently ignored `__pregel_stream_messages_keys` on the async path and kept streaming internal message fields - i.e. exactly the leak #6798 is about. My original tests only covered the sync path, which is why it slipped through.

**Fixed:** the restriction is now applied in both `stream()` and `astream()`, with coverage for both (7 cases: default behaviour unchanged, restricted key, alternate key, unknown key). I confirmed the new async tests fail without the fix.

I also cleaned up the branch - the previous version carried unrelated edits (private-channel filtering, input validators, bound-context fallback). Those are gone, so this PR is now only the #6798 change:

- 1 commit, 4 files, +125/-7
- `configurable["__pregel_stream_messages_keys"]` selects which state keys are scanned for messages; when unset the behaviour is unchanged
- applied to both `stream()` and `astream()`

One piece of context, since there is an earlier PR for this issue: #6878 (by @mrutunjay-kinagi) currently reports `mergeable_state=dirty` and has not been updated since 2026-03-24. I am not attached to my API shape - happy to rework this onto #6878's public `messages_key=` kwarg, or to close mine if @mrutunjay-kinagi wants to pick theirs back up.

PS: the `require-issue-link` bot closed this PR right after my force-push, because I am not currently assigned to #6798. Whenever convenient, could you assign the issue to me? The bot reopens the PR automatically once that is done. Always happy to reshape the API to whatever you prefer.

AaronG18:
Hi @niilooy - thanks for assigning me to #6798. The PR is ready for review whenever convenient: a single commit, +125/-7 across 4 files, based on current main, and the restriction is wired into both stream() and astream().

If you would rather go with #6878's public `messages_key=` kwarg path instead, I'm happy to rebase #8868 onto it, or close mine so @mrutunjay-kinagi can pick his back up. No rush either way.

PR #8879, opened 2026-09-09
refactor(pregel): replace object identity check with explicit exiting…

### LINKED ISSUE #8880: Refactor: replace implicit identity check with explicit `exiting` parameter in `PregelLoop._put_checkpoint`
### Checked other resources

### Related Issues / PRs

#8879

### Description

In `langgraph/pregel/_loop.py`, `_put_checkpoint` currently relies on an object identity check `metadata is self.checkpoint_metadata` to gate exit-only behaviour (skipping count bumps and retaining metadata).

As noted by the existing code comment and `TODO` on line 1091:
> `# TODO: replace with an explicit exiting: bool = False parameter.`

Using `is` for control flow is fragile. Replacing this with an explicit keyword-only `exiting: bool = False` parameter makes the loop exit semantics explicit and deterministic without changing caller behavior.

### Proposed Solution

- Add `*, exiting: bool = False` to `PregelLoop._put_checkpoint`.
- Pass `exiting=True` at the exit teardown call site in `_suppress_interrupt`.
- Keep default `exiting=False` for all other call sites.

```python
def _put_checkpoint(
    self, metadata: CheckpointMetadata, *, exiting: bool = False
) -> None:
    if exiting and self.checkpoint["id"] == self.checkpoint_id_saved:
        # checkpoint already saved
        return

BenjaminVijayaraj5102004:
Fixes #8880

### Summary
Refactors `PregelLoop._put_checkpoint` to accept an explicit keyword-only `exiting: bool = False` parameter instead of relying on the implicit object identity heuristic `metadata is self.checkpoint_metadata`.

### Details
- Removes the fragile `exiting = metadata is self.checkpoint_metadata` check and its associated `TODO` comment in `langgraph/pregel/_loop.py`.
- Updates the exit teardown call site in `_suppress_interrupt` to pass `exiting=True`.
- Preserves backwards compatibility across other call sites (`source: input | loop | fork`), which continue defaulting to `exiting=False`.

### Testing
- Ran checkpoint unit tests (`pytest -k "test_checkpoint" -m "not postgres and not redis"`).
- Verified formatting and linting pass via `ruff check` and `ruff format`.

### DIFF
--- a/libs/langgraph/langgraph/pregel/_loop.py
+++ b/libs/langgraph/langgraph/pregel/_loop.py
@@ -1078,18 +1078,9 @@ class PregelLoop:
             self._push_graph_lifecycle_event("resume")
         return updated_channels
 
-    def _put_checkpoint(self, metadata: CheckpointMetadata) -> None:
-        # `is` (object identity) — not `==`. Three of four call sites pass a
-        # fresh dict ({"source":"input"|"loop"|"fork"}); only
-        # `_suppress_interrupt`(will rename to _on_loop_exit soon)
-        # at exit reuses the existing `self.checkpoint_metadata` instance. So
-        # `metadata is self.checkpoint_metadata` is True only on the exit call,
-        # which is what we use to gate exit-only behaviour (skip count-bump,
-        # don't replace metadata). Could be replaced by an explicit
-        # `exiting: bool = False` parameter; left as-is to match the existing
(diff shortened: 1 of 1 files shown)


BenjaminVijayaraj5102004:
wants to merge 1 commit into langchain-ai:main

PR #8886, opened 2026-09-10
feat(langgraph): add response_schema to interrupt()

eliornl:
Adds an optional keyword-only `response_schema` to `interrupt()`, carried on `Interrupt.response_schema` as JSON Schema so clients (LangGraph Studio first) can render a typed form for the resume value instead of a free-form JSON box. Default `None`, so nothing changes for existing graphs.

- Accepts a JSON Schema dict (passed through, not validated) or a Pydantic model / `TypedDict` / dataclass, converted with `TypeAdapter.json_schema()`. For those, the resume value is validated and the validated object is what `interrupt()` returns; an invalid value raises from the node before it is committed, so the next resume works.
- Additive on the `stream_mode="debug"` payload: interrupt dicts gain `response_schema` (it is a dataclass field, serialized with `asdict`).
- `libs/sdk-py`: mirrors the field on the `Interrupt` TypedDict as `NotRequired`; the server omits the key when no schema was given.

JS counterpart: langchain-ai/langgraphjs#2824.

Verified: new tests parametrized across every checkpointer backend (schema kinds, coercion, invalid resume) plus the full `libs/langgraph` suite and downstream `prebuilt`, `cli`, `sdk-py`; round-tripped end to end through the agent server with this branch installed.

### DIFF
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -14,6 +14,7 @@ from typing import (
     NamedTuple,
     TypeVar,
     final,
+    overload,
 )
 from warnings import warn
 
@@ -37,6 +38,7 @@ from langgraph.warnings import LangGraphDeprecatedSinceV10, LangGraphDeprecatedS
 # when used in standalone type aliases.
 StateT = TypeVar("StateT")
 OutputT = TypeVar("OutputT")
+ResponseT = TypeVar("ResponseT")
 
 if TYPE_CHECKING:
     from langgraph.pregel.protocol import PregelProtocol
@@ -859,6 +861,14 @@ class Command(Generic[N], ToolOutputMixin):
             return []
 
     PARENT: ClassVar[Literal["__parent__"]] = "__parent__"
+
+
+@overload
+def interrupt(value: Any, *, response_schema: type[ResponseT]) -> ResponseT: ...
+
+
+@overload
+def interrupt(value: Any, *, response_schema: dict[str, Any] | None = None) -> Any: ...
 
 
 def interrupt(
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -982,9 +982,10 @@ def interrupt(
     # find previous resume values
     if scratchpad.resume:
         if idx < len(scratchpad.resume):
-            conf[CONFIG_KEY_SEND]([(RESUME, scratchpad.resume)])
             v = scratchpad.resume[idx]
-            return adapter.validate_python(v) if adapter else v
+            validated = adapter.validate_python(v) if adapter else v
+            conf[CONFIG_KEY_SEND]([(RESUME, scratchpad.resume)])
+            return validated
     # find current resume value
     v = scratchpad.get_null_resume(True)
     if v is not None:
--- a/libs/langgraph/langgraph/types.py
+++ b/libs/langgraph/langgraph/types.py
@@ -20,6 +20,7 @@ from warnings import warn
 from langchain_core.messages import AnyMessage
 from langchain_core.runnables import Runnable, RunnableConfig
 from langgraph.checkpoint.base import BaseCheckpointSaver, CheckpointMetadata
+from pydantic import TypeAdapter
 from typing_extensions import NotRequired, TypeAliasType, TypedDict, Unpack, deprecated
 from xxhash import xxh3_128_hexdigest
 
@@ -596,13 +597,19 @@ class Interrupt:
     id: str
     """The ID of the interrupt. Can be used to resume the interrupt directly."""
 
+    response_schema: dict[str, Any] | None = None
+    """JSON Schema for the value expected when resuming this interrupt, if the graph provided one."""
+
     def __init__(
         self,
         value: Any,
         id: str = _DEFAULT_INTERRUPT_ID,
+        *,
+        response_schema: dict[str, Any] | None = None,
         **deprecated_kwargs: Unpack[DeprecatedKwargs],
     ) -> None:
         self.value = value
+        self.response_schema = response_schema
 
         if (
             (ns := deprecated_kwargs.get("ns", MISSING)) is not MISSING
@@ -614,8 +621,14 @@ class Interrupt:
             self.id = id
 
     @classmethod
-    def from_ns(cls, value: Any, ns: str) -> Interrupt:
-        return cls(value=value, id=xxh3_128_hexdigest(ns.encode()))
+    def from_ns(
+        cls, value: Any, ns: str, *, response_schema: dict[str, Any] | None = None
+    ) -> Interrupt:
+        return cls(
+            value=value,
+            id=xxh3_128_hexdigest(ns.encode()),
+            response_schema=response_schema,
+        )
 
     @property
     @deprecated("`interrupt_id` is deprecated. Use `id` instead.", category=None)
@@ -848,7 +861,9 @@ class Command(Generic[N], ToolOutputMixin):
     PARENT: ClassVar[Literal["__parent__"]] = "__parent__"
 
 
-def interrupt(value: Any) -> Any:
+def interrupt(
+    value: Any, *, response_schema: dict[str, Any] | type[Any] | None = None
+) -> Any:
     """Interrupt the graph with a resumable exception from within a node.
 
     The `interrupt` function enables human-in-the-loop workflows by pausing graph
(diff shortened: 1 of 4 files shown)


libs/langgraph/langgraph/types.py
```diff
@@ -596,13 +597,19 @@ class Interrupt:
     id: str
     """The ID of the interrupt. Can be used to resume the interrupt directly."""
 
+    response_schema: dict[str, Any] | None = None
```
sydney-runkle, line 602: shouldn't interrupt be generic on this type so we could theoretically type it properly?

libs/langgraph/tests/test_interruption.py
```diff
@@ -90,3 +94,101 @@ async def noop(_state):
...
+
+
+@pytest.mark.parametrize(
+    ("response_schema", "expected_schema", "expected_answer"),
+    [
+        (None, None, {"approved": True, "extra": 1}),
+        (RAW_SCHEMA, RAW_SCHEMA, {"approved": True, "extra": 1}),
+        (Decision, Decision.model_json_schema(), Decision(approved=True)),
+        (
+            DecisionDict,
```
sydney-runkle, line 124: should we also try dataclass

eliornl: added.

libs/langgraph/langgraph/types.py
eliornl: done via overloads on interrupt(): response_schema=Approval now returns Approval, dict or none returns Any. left the Interrupt dataclass untyped since it's the wire payload.

Nakagawa-master:
One resume-contract edge case looks worth pinning before this becomes the typed HITL boundary: **the schema surfaced in the pending `Interrupt` can diverge from the schema that validates the resume.**

`interrupt()` constructs the `TypeAdapter` from the *current re-execution* of the node. A paused task can outlive state/config changes (and, with a persistent checkpointer, a code deployment). If `response_schema` is selected dynamically, the person/client may have rendered schema A, while resume re-runs the node under schema B and validates the answer against B.

A minimal regression shape is:

```python
class A(BaseModel):
    approved: bool

class B(BaseModel):
    approved: bool
    reason: str

current_schema = A

def node(state):
    return {"answer": interrupt("approve?", response_schema=current_schema)}

# first invoke -> pending Interrupt advertises A
# mutate current_schema = B (standing in for state/config/code change)
# resume with {"approved": True}, which is valid for the form actually shown
```

Today that resume is evaluated using B, not the schema carried by the pending interrupt. The inverse (B -> A) is more concerning because a value that would not satisfy what was originally shown can become acceptable after the pause.

I think the invariant should be:

```text
schema shown for interrupt I
== schema contract consumed by resume for interrupt I
```

At minimum, persist/bind a canonical schema fingerprint to the pending interrupt/resume slot and reject a stale resume when the current node produces a different fingerprint, forcing the new interrupt/schema to be surfaced. That avoids needing to persist a `TypeAdapter` across deployments while preventing silent reinterpretation.

Regression cases I'd add:
1. unchanged schema across pause/resume -> current behavior;
2. schema narrows while paused -> old valid answer is not silently validated under the new contract;
3. schema broadens while paused -> old pending interrupt cannot gain new accepted values;
4. ID-mapped resume has the same schema-binding rule as positional/global resume.

This is separate from the already-fixed invalid-resume persistence issue: that one binds *value validity*; this one binds the *validation contract itself* to the interrupt the client actually saw.

libs/langgraph/langgraph/types.py
sydney-runkle: shouldn't this be generic :)

libs/langgraph/langgraph/types.py
```diff
@@ -36,6 +38,7 @@
 # when used in standalone type aliases.
 StateT = TypeVar("StateT")
 OutputT = TypeVar("OutputT")
+ResponseT = TypeVar("ResponseT")
```
sydney-runkle: does this need a bound?

libs/langgraph/langgraph/types.py
```diff
@@ -848,7 +863,17 @@ def _update_as_tuples(self) -> Sequence[tuple[str, Any]]:
...
+@overload
+def interrupt(value: Any, *, response_schema: type[ResponseT]) -> ResponseT: ...
+
+
+@overload
+def interrupt(value: Any, *, response_schema: dict[str, Any] | None = None) -> Any: ...
+
+
+def interrupt(
+    value: Any, *, response_schema: dict[str, Any] | type[Any] | None = None
```
sydney-runkle: confused about type[Any] here

VANDRANKI (commented):
Community review, not a merge-gate approval.

I traced `interrupt()` in `libs/langgraph/langgraph/types.py` end to end for the new `response_schema` parameter. Key design point I checked carefully: `scratchpad.resume.append(v)` always appends the raw, unvalidated resume value, never the validated Pydantic/dataclass instance. Validation (`adapter.validate_python(v)`) happens fresh on every access, including on replay through the `idx < len(scratchpad.resume)` branch. That means the checkpoint only ever persists plain JSON-serializable data, never an arbitrary Pydantic object, which avoids a serialization problem this change could easily have introduced.

I also checked the ordering in the "current resume value" branch:
```
v = scratchpad.get_null_resume(True)
if v is not None:
    validated = adapter.validate_python(v) if adapter else v
    scratchpad.resume.append(v)
    ...
    return validated
```
Validation happens before the append. So `test_interrupt_response_schema_rejects_invalid_resume` (resume with `{"approved": "nope"}` against `response_schema=Decision`, expecting `pydantic.ValidationError`) should leave `scratchpad.resume` untouched when it raises, meaning a second, valid resume attempt at the same interrupt is treated as a fresh attempt rather than being blocked by a half-committed bad value. The test does exactly this: it expects the `ValidationError` on the bad resume, then resumes again with `{"approved": False}` and expects success. That matches what the code does.

The three type-adapter branches (raw JSON-schema dict used as-is with no validation, vs. a Pydantic model/TypedDict/dataclass converted via `TypeAdapter(response_schema).json_schema()` and validated on resume) are each covered by a parametrized test case, including a case with an extra `"extra": 1` key in the resume payload to confirm the raw-dict/no-schema path passes it through untouched while the typed paths drop it via Pydantic's default lenient validation. I did not verify pydantic's TypedDict/dataclass extra-field-dropping behavior from first principles, I'm relying on it matching Pydantic v2's documented default (extra keys ignored unless a model sets `extra="forbid"`), which is a reasonable thing to trust here rather than re-derive.

The `Interrupt` and `langgraph_sdk.schema.Interrupt` (TypedDict) additions are both additive with a default of `None` / `NotRequired`, so this shouldn't break existing callers building an `Interrupt` positionally, since `response_schema` is keyword-only in `__init__` and `from_ns`. I did not fully verify `Interrupt`'s equality/serialization machinery beyond what the tests exercise (`Interrupt(...) == Interrupt(...)` comparisons in the new tests pass structurally as written, so at minimum equality includes the new field).

Overall this reads as a well-scoped feature with good test coverage across the schema-type matrix and the validation-failure-then-retry path, which is the trickiest part to get right. I didn't find a correctness problem tracing it.

libs/langgraph/langgraph/types.py
```diff
@@ -948,27 +971,35 @@ def node(state: State):
...
     # track interrupt index
     scratchpad = conf[CONFIG_KEY_SCRATCHPAD]
     idx = scratchpad.interrupt_counter()
     # find previous resume values
     if scratchpad.resume:
         if idx < len(scratchpad.resume):
             conf[CONFIG_KEY_SEND]([(RESUME, scratchpad.resume)])
-            return scratchpad.resume[idx]
+            v = scratchpad.resume[idx]
+            return adapter.validate_python(v) if adapter else v
```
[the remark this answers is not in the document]
eliornl: fixed: the id-mapped branch now validates before the resume is sent to the checkpoint, so a bad value is never persisted. test_interrupt_response_schema_rejects_invalid_resume covers both plain and id mapped resume with a bad value followed by a corrected one.

libs/langgraph/langgraph/types.py
eliornl: done, Interrupt[ResponseT] with Any default.

libs/langgraph/langgraph/types.py
eliornl: no. pydantic, TypedDict and dataclass share no base, so a bound would exclude one of them. TypeAdapter is the real check.

libs/langgraph/langgraph/types.py
eliornl: done, plain type.

libs/langgraph/langgraph/types.py
```diff
@@ -596,13 +605,19 @@ class Interrupt:
...
 
+    response_schema: dict[str, Any] | None = None
+    """JSON Schema for the value expected when resuming this interrupt, if the graph provided one."""
+
     def __init__(
         self,
         value: Any,
         id: str = _DEFAULT_INTERRUPT_ID,
+        *,
+        response_schema: dict[str, Any] | None = None,
```
sydney-runkle: this should be ResponseT :) can you make sure this is working end to end and type checking correctly please

eliornl: good catch, ResponseT wasn't used anywhere. fixed.

sydney-runkle (approved):

PR #8910, opened 2026-09-13
fix(langgraph): internal frames leak into user tracebacks after internal module renames

CodeAlex52:
# fix: internal frames leak into user tracebacks after internal module renames

## Problem

Exceptions raised in user code (nodes/edges) are supposed to be re-raised with
LangGraph's internal plumbing frames stripped from the head of the traceback, so
the first frames a user sees belong to their own code. That is the feature added
in #4054 (`bfd271e00`, "Remove internal frames from stack traces", 2025-03-27).

Since `8c4e698c5` (#5252, "langgraph[change]: solidify public/private
differentiations", merged 2025-07-02) that filtering no longer removes anything:
the exclusion list in `langgraph/pregel/_runner.py` still matches the pre-rename
file paths.

```python
EXCLUDED_FRAME_FNAMES = (
    "langgraph/pregel/retry.py",      # now _retry.py
    "langgraph/pregel/runner.py",     # now _runner.py
    "langgraph/pregel/executor.py",   # now _executor.py
    "langgraph/utils/runnable.py",    # now _internal/_runnable.py (old path is a 3-line shim)
    "langchain_core/runnables/config.py",
    "concurrent/futures/thread.py",
    "concurrent/futures/_base.py",
)
```

None of the four LangGraph entries matches `co_filename` anymore
(`endswith("langgraph/pregel/runner.py")` is false for
`.../langgraph/pregel/_runner.py`). Worse, because the trim loop

```python
while tb.tb_next is not None and any(tb.tb_frame.f_code.co_filename.endswith(name) for name in ...):
```

walks a *contiguous run from the head* and stops at the first non-matching
frame, the very first stale entry makes the loop exit immediately: the three
still-valid non-LangGraph entries (`langchain_core`, `concurrent.futures`) can
never be reached either. The feature degrades to a complete no-op - silently,
since no test covered it and type checkers cannot see path strings.

## Impact (measured on `main` @ `e539ac122`, langgraph 1.2.11, CPython 3.13.5)

A user node raising `ValueError`, frames visible in the user's `except` block:

| scenario | frames seen (broken) | frames seen (fixed) | extra frames |
|---|---|---|---|
| `graph.invoke()`, single node | 8 | 4 | **+4** |
| `graph.ainvoke()`, single node | 11 | 4 | **+7** |
| `graph.invoke()`, 2 parallel branches | 13 | 4 | **+9** |
| `graph.ainvoke()`, 2 parallel branches | 12 | 4 | **+8** |

The leaked frames include `langgraph/pregel/_runner.py::tick`,
`_runner.py::_panic_or_proceed`, `_retry.py::run_with_retry`,
`_executor.py::done`, `_internal/_runnable.py::invoke/ainvoke` (x2),
`langchain_core/runnables/config.py` (x2) and `concurrent/futures/thread.py`.
After the fix the only LangGraph frames left are `main.py::invoke/stream`, which
were never part of the exclusion list by design.

## Fix

Match on the frame's module name instead of its file path. Module names are
stable across file renames and install layouts (source checkout, wheel,
editable install, namespace packages), and the frame globals always carry the
real module name:

```python
EXCLUDED_FRAME_MODULES = frozenset((
    "langgraph.pregel._retry",
    "langgraph.pregel._runner",
    "langgraph.pregel._executor",
    "langgraph._internal._runnable",
    "langchain_core.runnables.config",
    "concurrent.futures.thread",
    "concurrent.futures._base",
))

def _is_excluded_frame(frame: FrameType) -> bool:
    return frame.f_globals.get("__name__") in EXCLUDED_FRAME_MODULES

def _strip_excluded_frames(exc: BaseException) -> None:
    if tb := exc.__traceback__:
        while tb.tb_next is not None and _is_excluded_frame(tb.tb_frame):
            tb = tb.tb_next
        exc.__traceback__ = tb
```

The four duplicated inline loops are replaced by `_strip_excluded_frames(exc)`.
A naive string update of the four paths produces byte-for-byte identical
trimming (verified: 4/4/4/4 frames in the four scenarios above), but would break
silently again on the next rename; matching modules is rename- and
layout-proof.

## Tests

Two new tests assert that (a) the user's node frame is still present and (b) no
`_runner.py` / `_retry.py` / `_executor.py` / `_internal/_runnable.py` frame is
anywhere in the user-facing traceback:

- `tests/test_pregel.py::test_error_traceback_excludes_internal_frames` (sync)
- `tests/test_pregel_async.py::test_error_traceback_excludes_internal_frames_async` (async)

Against the unpatched source both fail (leaking `_runner.py:248 in tick`,
`_retry.py:645 in run_with_retry`, `_runnable.py:713/447 in invoke`, and the
async equivalents); with the patch applied both pass. The existing
exception/error-handler test selection in `test_pregel.py` + `test_pregel_async.py`
(23 tests: `imp_exception`, `overwrite_parallel*`, `*error_handler*`,
`node_cancellation*`) still passes: 25 passed in total including the 2 new tests.

Commands:

```bash
python -m pytest tests/test_pregel.py::test_error_traceback_excludes_internal_frames \
                 tests/test_pregel_async.py::test_error_traceback_excludes_internal_frames_async -q
```

## Compatibility

- No public API change. `EXCLUDED_FRAME_FNAMES` lived in the private module
  `langgraph/pregel/_runner.py` and was referenced only by the four call sites
  in that file (checked with grep over the whole repo); it is renamed to
  `EXCLUDED_FRAME_MODULES`.
- No user frames are removed: the trim only ever drops a contiguous run of
  frames at the *head* of the traceback, and only while they belong to internal
  modules; the user's node/edge frame terminates the run. This is asserted by
  the new tests.
- Fixes the same bug on Windows as a side effect (path matching with `/`
  separators never matched `\`-separated `co_filename`).

## Links

- Feature: #4054 "Remove internal frames from stack traces" (merged 2025-03-27)
- Regression introduced by: #5252 internal renames (merged 2025-07-02)
- Related user report about the `langgraph.utils.runnable` shim: #5647
- No upstream issue tracking this regression was found (searched issues for
  "internal frames", "traceback", "langgraph.utils.runnable"), so this PR does
  not claim a `fixes #` reference.


CodeAlex52:
Opened #8911 per the contribution policy (bug report with environment, reproduction, frame-count deltas and root cause).

The patch and its two regression tests in this PR are unchanged and still apply cleanly to `e539ac122`: 2 failed before the fix, 2 passed after, 25 related tests still pass, ruff clean.

I'll add `Fixes #8911` here as soon as a maintainer approves and assigns the issue.

PR #8917, opened 2026-09-14
fix(checkpoint): make Redis test skip detection fast

### LINKED ISSUE #5029: Add windows build to CI

HUAHAODIA:
## Summary

- The autouse `setup` fixture in `libs/checkpoint/tests/test_redis_cache.py` probed Redis with a **default** redis-py 8.x client, which retries a refused connection ten times with backoff. On Windows, where a refused loopback connection takes ~2s per address family (`::1`, then `127.0.0.1`), deciding to skip cost **~49s per test**, and the function-scoped fixture made all 16 tests pay it — the suite times out instead of skipping (as diagnosed by @sridarsh7858 in https://github.com/langchain-ai/langgraph/issues/5029).
- This PR probes Redis **once at module level** with a single attempt (`Retry(NoBackoff(), 0)`) and a 1s `socket_connect_timeout`, and also catches `redis.TimeoutError` — which is what a connect timeout raises, unlike the `ConnectionError` the retry loop eventually surfaced. The same fast check replaces the two inline `try: client.ping() except: skip` probes in the async tests.
- When Redis **is** available the test behavior is unchanged: the same client is constructed per test and `flushdb()` cleanup is preserved.

## Verification

Measured on Windows 11, redis-py 8.1.0, no Redis running:

| | before | after |
|---|---|---|
| one skip decision | 48.56s | ~1s (probe, once) |
| `pytest tests/test_redis_cache.py` | times out (>12min, killed) | **16 skipped in 2.25s** |

Also verified on unmodified main that the 13 `test_encrypted.py` failures on Windows are pre-existing and unrelated to this change (looks like a separate Windows blocker for #5029 — happy to dig into that next).

Fixes #5029 (unblocks the Windows run for the Redis cache file; the Redis skip timeout was diagnosed in that issue thread)

From ZCode


HUAHAODIA:
The PR description now links the change with `Fixes #5029` (and the commit message does too). Could a maintainer reopen this PR (or drop the `missing-issue-link` label)?

Context: this fixes the Redis skip-detection timeout diagnosed in the #5029 thread — measured 48.56s per skip decision on Windows 11 / redis-py 8.1.0 before, 16 skipped in 2.25s after. Happy to split this into a separate issue if maintainers prefer a narrower link than #5029.

PR #8943, opened 2026-09-16
feat(langgraph): add `available_tools` to ToolCallRequest

ccurme:
`ToolCallRequest` exposes `.tool`, but currently it's not easy to override that tool (e.g., to re-direct to another tool). Here we add `available_tools: list[BaseTool]` to `ToolCallRequest` so that `(a)wrap_tool_call` hooks can easily update the value of `.tool`.

Used in https://github.com/langchain-ai/langchain/pull/40463.


libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -135,18 +135,25 @@ class ToolCallRequest:
 
     Attributes:
         tool_call: Tool call dict with name, args, and id from model output.
+
```
eyurtsev, line 138:
```suggestion
```

libs/prebuilt/langgraph/prebuilt/tool_node.py
```diff
@@ -1184,6 +1218,7 @@ async def _arun_one(
             tool=tool,
             state=tool_runtime.state,
             runtime=tool_runtime,
+            available_tools=list(self.tools_by_name.values()),
```
eyurtsev, line 1221: could we memoize this in the init?

PR #8953, opened 2026-09-17
fix(checkpoint): fail closed on unencrypted checkpoint rows in EncryptedSerializer

### LINKED ISSUE #8938: EncryptedSerializer silently accepts unauthenticated plaintext rows, voiding its at-rest integrity guarantee
# EncryptedSerializer silently accepts unauthenticated plaintext rows, voiding its at-rest integrity

**Severity: medium.** Structural invalidation of a shipped guarantee: checkpoint encryption is confidentiality-only and not even that per row, because any store-level writer can substitute plaintext forgeries that are deserialized with no decryption and no tag verification.

**Affected versions:** verified on langgraph-checkpoint 4.2.0 / langgraph 1.2.11 at commit 230927fb3a9ac9b2893a30322b4dfea7cdea9a8f (main head at time of report). `EncryptedSerializer` was introduced by PR #3852 (merged 2025-03-14), so the behavior has shipped since the first version that included it.

## Summary

`EncryptedSerializer.loads_typed` treats any stored row whose type string contains no "+" as plaintext and passes it straight to the inner serializer without decryption or tag verification:

```python
def loads_typed(self, data: tuple[str, bytes]) -> Any:
    enc_cipher, ciphertext = data
    # unencrypted data
    if "+" not in enc_cipher:
        return self.serde.loads_typed(data)
```

(libs/checkpoint/langgraph/checkpoint/serde/encrypted.py:26-30)

The AEAD tag checked in the decrypt path (`encrypt_and_digest` / `decrypt_and_verify` in the pycryptodome cipher, encrypted.py:87-105) is the only tampering detection this class provides, and an attacker who can write to the store simply never triggers it: they replace an encrypted row with a plaintext row, and the graph loads the forged state. No signature, MAC, or digest protects checkpoint rows in any shipped saver; for the encrypted serializer specifically, the mechanism that was supposed to detect rewrites is bypassed row by row.

We demonstrated this in the hardened posture, that is with `LANGGRAPH_STRICT_MSGPACK=true` AND `EncryptedSerializer` over AES-EAD, so the constructor-revival path documented in CVE-2026-28277 / GHSA-g48c-2wqr-h844 is blocked and only plain data survives deserialization. Even then the forged state is enough to drive the graph's own authorization logic.

## Reproduction (executed at commit 230927fb3a9ac9b2893a30322b4dfea7cdea9a8f)

1. Build a graph with `InMemorySaver(serde=EncryptedSerializer.from_pycryptodome_aes(key=...))` and run it once; stored rows have type `msgpack+aes`.
2. As the store-level writer (the same capability as database access, SQL injection in a co-located app, shared credentials, or a restored backup), replace the head checkpoint row and channel blobs with plaintext: type `msgpack` with no cipher suffix, where the forged checkpoint's `channel_values` contains `approved=True` and matching `channel_versions`.
3. `graph.get_state(config)` returns the forged values: `approved=True`, no decryption error, no tag failure, no warning.
4. The next invoke on the thread routes on the forged channel value through the graph's own conditional edge and executes the privileged node. Observed output:

```
stored row type before attack: 'msgpack+aes'
stored row type after attack: 'msgpack'
state read from forged plaintext row: approved=True
turn-2 result: {'approved': True, 'log': ['turn-2', 'benign-ran', 'PRIVILEGED-NODE-RAN']}
PRIVILEGED-ROUTE-TAKEN-FROM-FORGED-PLAINTEXT-ROW: True
```

## Expected vs actual

Expected: a checkpointer configured with an encryption cipher refuses or at least flags rows that were not written under that cipher, so tampering with stored rows is detectable.

Actual: untagged rows are deserialized as-is, so a store-level writer controls every channel value the graph resumes with, including values that gate authorization routing, tool selection, and human-approval state. Encryption here is confidentiality-only against passive readers, and even that is opt-out per row by the attacker.

## Impact

Deployments that enable checkpoint encryption (its documented purpose is protecting checkpoint contents at rest) do not obtain integrity. The SqliteSaver and PostgresSaver inherit the same `loads_typed` path, so the exposure is not specific to the in-memory saver used in the demonstration.

## Recommended fix

Fail closed: when the serializer was constructed with a cipher, rows whose type tag carries no cipher suffix should raise rather than be trusted, with an explicit `allow_plaintext=True` migration flag for reading legacy rows written before encryption was enabled (and a documented migration helper that re-encrypts them). A ready two-file patch implementing exactly this is attached (PR form), including the strict-mode allowlist wrapper path in `BaseCheckpointSaver.with_allowlist` which reconstructs the serializer and must preserve the new flag.

## Notes

* The upstream test suite for this class (libs/checkpoint/tests/test_encrypted.py, 17 tests) passes both before and after the attached patch, which also shows the suite never exercises plaintext-row loading through a cipher-configured serializer; a regression test for the refusal behavior is included in the PR form materials.
* This is distinct from CVE-2026-28277 / GHSA-g48c-2wqr-h844 (msgpack constructor revival): that issue is about deserializing crafted objects, this one is about accepting unauthenticated rows even with object revival blocked and encryption enabled.

Rook-CodeVolt:
Closes #8938.

## The bug

`EncryptedSerializer.loads_typed` treats any stored row whose type string has no `"+"` in it as plaintext and hands it straight to the inner serializer, no decryption, no AEAD tag check:

```python
if "+" not in enc_cipher:
    return self.serde.loads_typed(data)
```

The AEAD tag is the only tamper-detection this class has, and it's only checked on the path that already believes the row is encrypted. A store-level writer (DB access, SQL injection in a co-located app, restored backup, shared creds) can drop in an unencrypted row and it loads as-is: no error, no warning, and anything the app routes on from that row (approval flags, tool gating, etc.) is now attacker-controlled.

## The fix

Fail closed by default:

- `EncryptedSerializer` and `EncryptedSerializer.from_pycryptodome_aes` take a new `allow_plaintext: bool = False` constructor arg.
- When `allow_plaintext=False` (the default) and a row has no cipher suffix, `loads_typed` now raises `ValueError("Refusing to load unencrypted checkpoint data ...")` instead of silently deserializing it.
- Existing plaintext data can still be read by opting in explicitly (`allow_plaintext=True`), documented as a migration/back-compat escape hatch, not a default.
- `allow_plaintext` is propagated through `with_allowlist()` / the msgpack-allowlist copy path, so opting in doesn't get silently reset by unrelated config changes.
- Fixed a related bug found while testing this: `from_pycryptodome_aes` accepted `allow_plaintext` as a kwarg but never forwarded it to the constructor, so passing it had no effect.

## Tests

Added to `tests/test_encrypted.py`:
- Unencrypted-row load rejected by default / accepted only with explicit opt-in.
- A plaintext-substitution attack simulation against `InMemorySaver`: splices a plaintext blob into `saver.blobs` (where `InMemorySaver` actually stores per-channel values, keyed by `(thread_id, checkpoint_ns, channel, version)` — they aren't embedded in the checkpoint row itself) and asserts it's rejected by default / only readable with `allow_plaintext=True`.
- `with_allowlist()` preserves `allow_plaintext` in both directions (true and false).

```
$ uv run pytest tests/test_encrypted.py -v
============================= 22 passed in 0.07s ==============================
```

Full `libs/checkpoint` suite (unaffected areas, checking for regressions from the base/`__init__.py` and `encrypted.py` changes):

```
$ uv run pytest
================= 163 passed, 17 skipped in 131.05s (0:02:11) ==================
```

`make format` / `make lint` clean.


Rook-CodeVolt:
Yes — that's exactly what #8953 does. `EncryptedSerializer.loads_typed` currently treats any row with no cipher suffix as plaintext and returns it straight from the inner serializer, no error. The fix flips that to fail-closed by default: unencrypted rows raise `ValueError` unless the caller explicitly opts in with `allow_plaintext=True` (kept only as a documented migration/back-compat escape hatch, not a default).

@sattyamjjain's point above is correct and out of scope for #8953 specifically: fail-closed on the "no cipher suffix" branch doesn't add row-identity binding to the AEAD (no AAD over thread/checkpoint_id/channel/version), so a same-key replay of a genuinely-encrypted row from elsewhere still decrypts and verifies. That's a materially larger change (touches `CipherProtocol.encrypt`/`decrypt` and every saver's row-key construction) and deserves its own issue rather than blocking this fix.

The PR itself has been sitting closed since 2026-09-17 purely on the bot's assignment gate (external contributors need to be assigned to the linked issue before a PR against it stays open) — four assignment requests over as many days with no maintainer action yet. Re-flagging here in case a maintainer with triage access sees this thread and can assign so the bot reopens it; happy to rebase/adjust if anything's changed on main since.

PR #8980, opened 2026-09-18
fix(checkpoint): resolve dotted filter keys in InMemoryStore (#7795)

### LINKED ISSUE #7795: InMemoryStore search filters treat dotted keys as literal top-level keys instead of nested paths
### Checked other resources

- [x] I searched existing issues and did not find a matching report for this specific `InMemoryStore` filter behavior.
- [x] I verified the behavior locally against the current `main` branch checkout.

### Related Issues / PRs

_No response_

### Reproduction Steps / Example Code

```python

store = InMemoryStore()
store.put(("docs",), "hyphen", {"user": {"access-level": "nested"}})

print([item.key for item in store.search(("docs",), filter={"user.access-level": "nested"})])
print([item.key for item in store.search(("docs",), filter={"user": {"access-level": "nested"}})])
```

### Actual Behavior

```text
[]
['hyphen']
```

`InMemoryStore` treats `"user.access-level"` as a literal top-level key, so dotted nested filter keys do not work.

### Expected Behavior

The dotted filter should resolve the nested value and return the stored item, matching the behavior already covered in SQLite store tests for dotted nested keys:

```text
['hyphen']
['hyphen']
```

### Why This Looks Like a Cross-Backend Inconsistency

- `InMemoryStore` currently filters via `item.value.get(key)`, which never resolves dotted paths.
- `SqliteStore` already supports dotted nested filter keys such as `filter={"user.access-level": "nested"}`.
- This creates different filter semantics between local/dev memory-backed flows and SQLite-backed flows.

### Minimal Root Cause

`libs/checkpoint/langgraph/store/memory/__init__.py`

`_filter_items()` currently does:

```python
_compare_values(item.value.get(key), filter_value)
```

That bypasses nested path resolution entirely.

### Suggested Fix

Resolve filter keys through nested path traversal before passing the value to `_compare_values()`.

I already validated a minimal fix and regression test on my fork:

- Branch: https://github.com/zichen0116/langgraph/tree/fix/inmemory-dotted-filter-keys
- Commit: https://github.com/zichen0116/langgraph/commit/6c848e8b3

### Test Plan

Local regression test added and verified:

```bash
python -m pytest libs/checkpoint/tests/test_inmemory_store_dotted_filters.py -q
```

### System Info

- OS: Windows
- Python: 3.13.5
- Reproduced from local checkout of `langchain-ai/langgraph` `main` on 2026-05-14

AaronG18:
Fixes #7795.

`InMemoryStore.search` / `asearch` treated dotted filter keys (e.g. `user.access-level`) as literal lookups, while `SqliteStore` already resolves them against nested values via SQLite `json_extract`. This made the two stores behave differently for the same filter.

This change resolves dotted keys against nested values (mirroring `SqliteStore`), and falls back to a literal top-level lookup when the dotted path does not resolve -- so keys that genuinely contain a dot stay matchable.

Tests: 7 new cases covering sync/async, operators (`$gt`/`$lte`/`$ne`), deep paths, unresolvable paths, and literal dotted keys.


AaronG18:
Hi maintainers - I have a ready fix for #7795 (InMemoryStore now resolves dotted filter keys like SqliteStore does, with 7 new tests). Could someone assign this issue to me so the require-issue-link bot stops auto-closing the PR? Happy to adjust the approach if you'd prefer it done another way.

PR #8981, opened 2026-09-18
fix(langgraph): keep StateGraph.compile() from mutating the builder (#8612)

### LINKED ISSUE #8612: StateGraph.compile() fails on the second call with a default error handler
### Checked other resources

- [x] I searched existing issues and pull requests.
- [x] I reproduced the issue on the current `main`.
- [x] I included a self-contained minimal reproduction.

### Related Issues / PRs

Introduced with `StateGraph.set_node_defaults()` in #7747.

#7996 concerns `cache_policy=None` opt-out behavior and does not cover repeated compilation with a default error handler.

### Reproduction Steps / Example Code

```python

from langgraph.graph import START, StateGraph


class State(TypedDict):
    value: str


def fail(state: State) -> State:
    raise RuntimeError("boom")


def recover(state: State, error: NodeError) -> State:
    return {"value": f"handled: {error.error}"}


builder = (
    StateGraph(State)
    .set_node_defaults(error_handler=recover)
    .add_node("fail", fail)
    .add_edge(START, "fail")
)

first = builder.compile()
print(first.invoke({"value": "start"}))

second = builder.compile()
print(second.invoke({"value": "start"}))
```

### Error Message and Stack Trace

```text
{'value': 'handled: boom'}

ValueError: Auto-generated default error handler node
`__default_error_handler__` already exists.
```

### Description

Compiling the same `StateGraph` builder more than once normally works and is exercised by the repository's `test_parent_command.py` and async equivalent.

When `set_node_defaults(error_handler=...)` is configured, the first `compile()` mutates `builder.nodes` by inserting `__default_error_handler__`. The second call then treats that internally generated node as a user collision and raises `ValueError`.

This prevents compiling one builder with different checkpointers or compile-time settings. The equivalent per-node `error_handler=...` path remains repeatable.

Expected behavior: both `compile()` calls succeed and produce independent compiled graphs.

A regression test could compile the same builder twice with a default error handler and invoke both graphs. The implementation should ideally keep generated handler state local to compilation, or otherwise make repeated compilation idempotent without retaining stale defaults.

If repeated compilation should remain supported, I would be happy to prepare a focused fix and regression test. Could a maintainer confirm the direction and assign this issue to me?

### System Info

- LangGraph: `1.2.11`
- Commit: `644815f9e5bc52ad8f7a5227a456227e9c3e639b`
- Python: `3.13.14`
- langchain-core: `1.5.3`
- langgraph-checkpoint: `4.2.0`
- OS: macOS ARM64

AaronG18:
Fixes #8612.

`StateGraph.compile()` applied node defaults (error handler, retry, cache, timeout) by mutating the `StateNodeSpec` objects in place. That leaked the auto-generated default error-handler node into the builder and made compiling the same builder a second time raise `ValueError: ... already exists`.

This change applies defaults to a local copy of the node specs via `dataclasses.replace`, so the builder is left untouched and `compile()` is idempotent.

Tests: `test_set_node_defaults_error_handler_repeated_compile` compiles a builder twice and asserts the builder is unchanged.


AaronG18:
Hi maintainers - ready fix for #8612: StateGraph.compile() now leaves the builder untouched (defaults applied via dataclasses.replace), so compiling the same builder twice no longer raises. Could someone assign this issue to me so the require-issue-link bot keeps the PR open? Open to adjusting the approach.

PR #9014, opened 2026-09-20
fix(sdk): clear SSE reconnect cursor on empty id

### LINKED ISSUE #9009: sdk-py: empty SSE id does not clear the reconnect cursor
### Checked other resources

### Description

After an SSE stream sends `id: old-cursor` followed by an empty `id:`, both Python clients reconnect with `Last-Event-ID: old-cursor`. The empty ID should clear the cursor, so the reconnect should omit that header.

Reproduced on current main `aa742fb31e2827d569b843e3600aeda2e0528e4b` and independently against PyPI `langgraph-sdk==0.4.4` (latest stable when checked). The example uses public `runs.join_stream()` APIs, real HTTPX transport, and a local HTTP server. It deliberately ends the first response early so the SDK follows `Location: /resume` through its normal reconnect path.

Results are identical for sync and async:

| Stream suffix after the first event | Expected reconnect header | Actual |
| --- | --- | --- |
| Comment heartbeat | `old-cursor` | `old-cursor` |
| Empty `id:` without an event | absent | `old-cursor` |
| Empty `id:` with a values event | absent | `old-cursor` |
| New nonempty ID | `new-cursor` | `new-cursor` |
| ID containing NUL | `old-cursor` | `old-cursor` |

The [SSE specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) sets the last event ID buffer to an `id` field's value unless it contains NUL, including an empty value. It only sends `Last-Event-ID` on reconnect when the last event ID string is nonempty. I have not tested whether a deployed LangGraph server emits empty IDs; this report concerns the SDK's handling of that valid SSE input.

### Related Issues / PRs

#6159 introduced reconnect handling; #6581 exposed stream IDs. #7915 concerns multiline data concatenation, a separate decoder behavior.

Install `langgraph-sdk==0.4.4` and run this script. It prints ten results; four have `contract_holds: false` (the two reset cases in each mode). Its final assertion records that current failing baseline.

<details>
<summary>Self-contained sync/async loopback reproduction</summary>

```python
"""Read-only wire-level reproduction using the unmodified public SDK."""

import asyncio
import json
import threading
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

from langgraph_sdk import get_client, get_sync_client


def run_case(mode, case, suffix, expected):
    observed = []
    first = b'id: old-cursor\nevent: values\ndata: {"step": 1}\n\n' + suffix
    final = b'event: end\ndata: null\n\n'

    class Handler(BaseHTTPRequestHandler):
        protocol_version = "HTTP/1.1"

        def do_GET(self):
            observed.append(self.headers.get("Last-Event-ID"))
            initial = len(observed) == 1
            payload = first if initial else final
            self.send_response(200)
            self.send_header("Content-Type", "text/event-stream")
            self.send_header("Content-Length", str(len(payload) + int(initial)))
            self.send_header("Connection", "close")
            if initial:
                self.send_header("Location", "/resume")
            self.end_headers()
            self.wfile.write(payload)
            self.wfile.flush()
            self.close_connection = True

        def log_message(self, *args):
            pass

    server = ThreadingHTTPServer(("127.0.0.1", 0), Handler)
    worker = threading.Thread(target=server.serve_forever, daemon=True)
    worker.start()
    url = f"http://127.0.0.1:{server.server_port}"

    async def async_read():
        async with get_client(url=url, api_key=None) as client:
            return [part async for part in client.runs.join_stream("thread", "run")]

    try:
        if mode == "sync":
            with get_sync_client(url=url, api_key=None, timeout=3) as client:
                parts = list(client.runs.join_stream("thread", "run"))
        else:
            parts = asyncio.run(async_read())
    finally:
        server.shutdown()
        server.server_close()
        worker.join()

    assert len(observed) == 2, observed
    assert observed[0] is None, observed
    result = {
        "mode": mode,
        "case": case,
        "expected_reconnect_id": expected,
        "actual_reconnect_id": observed[1],
        "contract_holds": observed[1] == expected,
        "events": [part.event for part in parts],
    }
    print(json.dumps(result))
    return result


cases = [
    ("no_reset", b": heartbeat\n\n", "old-cursor"),
    ("empty_id_only", b"id:\n\n", None),
    ("empty_id_with_event", b'id:\nevent: values\ndata: {"step": 2}\n\n', None),
    ("new_nonempty_id", b'id: new-cursor\nevent: values\ndata: {"step": 2}\n\n', "new-cursor"),
    ("nul_id_ignored", b"id: ignored\x00id\n\n", "old-cursor"),
]
results = [run_case(mode, *case) for mode in ("sync", "async") for case in cases]
assert sum(not result["contract_holds"] for result in results) == 4, results
```

</details>

No user-visible exception: the SDK reconnects successfully, but the loopback server receives the stale cursor.

### Proposed Scope

`SSEDecoder.last_event_id` currently collapses the explicit empty reset to `None`; the sync and async HTTP loops then preserve the previous cursor. I propose distinguishing an unseen ID from an explicit empty ID, propagating resets to both reconnect loops, and omitting a stale header after reset. Regression coverage should include both modes, ID-only control blocks, nonempty/NUL controls, and a caller-supplied initial `last_event_id`. ID-only control blocks should remain hidden from application events.

Could a maintainer confirm this scope and assign this issue to `1052326311`? I will prepare the focused sdk-py fix and sync/async regressions once approved.

### System Info

Current-main environment, from `python -m langchain_core.sys_info`:

```text
OS: Darwin (arm64), Darwin Kernel Version 25.6.0
Python: 3.13.14
langchain_core: 1.6.1
langsmith: 0.8.18
langchain_protocol: 0.0.19
langgraph_sdk: 0.4.4
httpx: 0.28.1
orjson: 3.12.0
jsonpatch: 1.33
packaging: 26.0
pydantic: 2.13.5
pyyaml: 6.0.3
requests: 2.33.0
requests-toolbelt: 1.0.0
tenacity: 9.1.4
typing-extensions: 4.16.0
uuid-utils: 0.14.1
websockets: 16.1.1
xxhash: 3.8.1
zstandard: 0.25.0
```

daleselaji-dev:
## Problem
When an SSE stream sends an existing event ID followed by an empty `id:` field, the Python SDK keeps the old cursor on reconnect. Both sync and async clients then send a stale `Last-Event-ID` header and can resume from the wrong event position.

## Root Cause
`SSEDecoder.last_event_id` collapsed an empty event ID into `None`, which was indistinguishable from no ID having been received. The HTTP reconnect loops therefore retained the prior cursor. They also treated an empty cursor as a header value instead of omitting `Last-Event-ID`.

## Solution
Track whether an ID field has been seen separately from its value, preserve an empty ID as an explicit cursor reset, and omit the reconnect header when the current cursor is empty.

## Changes
- Preserve the SSE distinction between no ID and an explicitly empty ID.
- Apply the cursor-reset behavior to sync and async HTTP reconnects.
- Add sync and async regression tests that verify an empty `id:` removes the stale reconnect header.

## Testing
- Baseline focused test: `uv run --project . --no-default-groups --with pytest --with pytest-asyncio --with pytest-mock --with httpx --with syrupy pytest tests/test_client_stream.py -q` — 14 passed.
- Final focused test: same command — 16 passed.
- Full SDK test suite with the repository's non-integration selection — 512 passed, 55 deselected.
- Ruff format check — passed.
- Ruff check — passed.
- `git diff --check` — passed.
- `ty check langgraph_sdk` was attempted but remains unverified because this Windows environment could not resolve the repository's optional `starlette.authentication` and `langgraph.store.base` imports.

## Compatibility/Risk
This follows the SSE specification: an empty `id:` resets the last-event-id, while events without an `id` field retain the existing cursor. Non-empty event IDs and explicit caller-provided reconnect behavior remain unchanged. No external service or credentials are required by the tests.

## Notes for Reviewer
The decoder now distinguishes “no ID seen” from “empty ID seen” without changing the public `StreamPart.id` shape. The sync and async HTTP loops use the same truthy-header rule already used by the higher-level thread/run clients, so a cleared cursor does not produce an empty `Last-Event-ID` header.

## Linked Issue
Fixes #9009


1052326311:
I tested head `5e9c58c19322c20d147f9c1862a903195fecfb6d` against a real loopback HTTP server through `runs.join_stream`. The empty-ID-with-event case is fixed, but two cases remain in both sync and async:

- After `id: old-cursor`, an `id:\n\n` control block still reconnects with `Last-Event-ID: old-cursor`. The blank-block early return in `SSEDecoder.decode` means the outer reconnect cursor never receives the reset.
- With `join_stream(last_event_id="initial-cursor")`, an empty ID **with** an event reconnects with `initial-cursor`. Skipping the new header assignment retains the initial header copied from `reconnect_headers`.

The expected reconnect header is absent in both cases. Nonempty replacement IDs, no-reset streams, and NUL-containing IDs passed as controls. The transport deliberately disconnects after the response payload and supplies `Location: /resume`, so these results exercise actual reconnect requests, not just decoder state. No deployed LangGraph server behavior is assumed.

<details>
<summary>Runnable loopback reproduction (SDK dependencies installed; run against this PR checkout)</summary>

```sh
PYTHONPATH=libs/sdk-py python /path/to/repro.py
```

```python
import asyncio
import json
import threading
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

from langgraph_sdk import get_client, get_sync_client


def run_case(mode, case, suffix, expected, initial_cursor=None):
    observed = []
    first = b'id: old-cursor\nevent: values\ndata: {"step": 1}\n\n' + suffix
    final = b'event: end\ndata: null\n\n'

    class Handler(BaseHTTPRequestHandler):
        protocol_version = "HTTP/1.1"

        def do_GET(self):
            observed.append(self.headers.get("Last-Event-ID"))
            initial = len(observed) == 1
            payload = first if initial else final
            self.send_response(200)
            self.send_header("Content-Type", "text/event-stream")
            self.send_header("Content-Length", str(len(payload) + int(initial)))
            self.send_header("Connection", "close")
            if initial:
                self.send_header("Location", "/resume")
            self.end_headers()
            self.wfile.write(payload)
            self.wfile.flush()
            self.close_connection = True

        def log_message(self, *args):
            pass

    server = ThreadingHTTPServer(("127.0.0.1", 0), Handler)
    worker = threading.Thread(target=server.serve_forever, daemon=True)
    worker.start()
    url = f"http://127.0.0.1:{server.server_port}"

    async def async_read():
        async with get_client(url=url, api_key=None) as client:
            return [part async for part in client.runs.join_stream("thread", "run", last_event_id=initial_cursor)]

    try:
        if mode == "sync":
            with get_sync_client(url=url, api_key=None, timeout=3) as client:
                parts = list(client.runs.join_stream("thread", "run", last_event_id=initial_cursor))
        else:
            parts = asyncio.run(async_read())
    finally:
        server.shutdown()
        server.server_close()
        worker.join()

    assert len(observed) == 2, observed
    assert observed[0] == initial_cursor, observed
    result = {
        "mode": mode,
        "initial_cursor": initial_cursor,
        "case": case,
        "expected_reconnect_id": expected,
        "actual_reconnect_id": observed[1],
        "contract_holds": observed[1] == expected,
        "events": [part.event for part in parts],
    }
    print(json.dumps(result))
    return result


for mode in ("sync", "async"):
    run_case(mode, "id-only reset", b"id:\n\n", None)
    run_case(mode, "reset with initial cursor", b'id:\nevent: values\ndata: {"step": 2}\n\n', None, "initial-cursor")
    run_case(mode, "nonempty ID control", b'id: new-cursor\nevent: values\ndata: {"step": 2}\n\n', "new-cursor")
```

The two reset cases print `"contract_holds": false` for each mode; the nonempty-ID control prints `true`.
</details>

These look like useful regression cases for this fix; sharing the reproduction here to support the existing implementation.

